PR #42, opened 2026-07-22
Self-host fold as a recursive helper, not a privileged loop

kawasima:
## What

`fold` is now an ordinary recursive helper in `souther.list`, not a privileged loop node. `foldFrom(step, seed, xs, i)` walks the list with `List.get`, threads the accumulator, and takes its step as a first-class closure; `fold(step, seed, xs)` desugars to it. The backend turns the tail self-call into a loop, so it runs in constant stack — as `fold` is written in Elm or F#.

The privileged path is removed: the `Core.Fold` node, the `List.fold` built-in, `BodyGen.fold`/`resolveAccType`, and the name-based recognition are gone. The one loop the backend emits is the general tail-call optimization, keyed on the shape of a self-tail-recursive call, not on the name `fold`.

## How

- **TCO** for self-tail-recursive helpers, through both `if` and `match` arms.
- A **recursive helper may take a function parameter**, passed as an `Fn` value and applied inside the method. A closure passed to a recursive helper **must be pure** (it may not call an injected behavior) — an effect cannot enter the pure method through it.
- **Generic recursive helpers**: type variables and empty-collection bottoms are erased to `Object` in codegen.
- A referenced **prelude recursive helper is injected** into the module and emitted on its `$Fns` class, its qualified name mangled to a JVM method.
- **Call-site type resolution** (unify/substitute), empty-seed **bottom refinement**, and **case-seeded accumulator widening** to its sum — in both the checker and the backend.

## Surface change

This opens **user-written recursive higher-order helpers** (`let count(n, f: (Int) -> Int) = ...`), under the same purity rule the standard library's `fold` obeys. This reverses the "a block is not a value / a recursive helper cannot take a function parameter" stance for users, keeping the purity restriction.

## Docs

New **ADR-0051**; `specification.adoc`, ADR-0028/0021/0041/0042/0048, and the ADR index updated to describe `fold` as a recursive helper.

## Verification

Reactor green: compiler **592**, LSP **29**, examples — `mvn test` BUILD SUCCESS.


kawasima:
Addressed the review (7d01c6f). Per-comment:

1. **`match` binding leaks into later arms (BodyGen)** — fixed. A case binding is now saved and restored around each arm, in both value-position `match` and tail-position `emitTailMatch`. Regression test: `CompileOrPatternTest.aCaseBindingDoesNotLeakIntoALaterArmThatReusesTheName`.

2. **Same leak in `emitTailMatch`** — fixed by the same save/restore.

3. **Purity bypassable via `List.fold` sugar** — resolved by removing the closure-purity check entirely, not by widening it. A closure that calls an injected behavior, passed to a recursive helper, is the intended requirement float-out: the closure is behavior code and the caller declares `requires` (undeclared is already E1602). The spec's recursive-helper purity rule is about the helper's **own body** (§fn-declaration), which the existing `rejectInjectedCalls` still enforces. Spec §fn-declaration now states a recursive helper may apply an effectful function argument. Tests: `aClosurePassedToARecursiveHelperMayCallAnInjectedBehavior` (accepted) and `aClosureCallingAnInjectedBehaviorStillNeedsTheBehaviorToDeclareRequires` (E1602).

4. **`list.sou` header still says "privileged loop primitive"** — fixed.

5. **`foldFrom` via `List.get` is O(n·log n)** — accepted as a tradeoff. On the persistent vector `get` is O(log₃₂ n) — about four levels for realistic sizes — so it is a small constant factor, acceptable for a spec DSL over domain-sized collections, and it is what keeps `fold` an ordinary recursive helper rather than a privileged loop. Recorded as known in ADR-0051/ADR-0042; revisit against a profile.

6. **Stale `CompileFoldSelfHostTest` Javadoc** — fixed.

Also from a separate review pass: fixed accumulator over-widening (a case-seeded fold that stays in its case was typed at the whole sum and wrongly rejected), made `enclosingSum` deterministic, and scanned example rows for prelude-recursive-helper injection.

Follow-up: totality via structural-recursion checking is tracked in #43 (out of scope here).

Reactor green: compiler 595, LSP 29, examples.

kawasima:
Fixed the blocking multi-module bug (bab0690): `injectRecursivePrelude` now runs on both `compileModules` paths, not only single-module `compile`. Before this, any `fold` / fold-derived combinator in a multi-file build crashed the compiler with an NPE in `recursiveHelperConstructs` (`List.foldFrom` referenced but its lowered body null). Regression test: `CompileFoldSelfHostTest.foldCompilesAndRunsInAMultiModuleBuild`. Reactor green: compiler 596.

PR #45, opened 2026-07-23
Closed newtype arithmetic and invariant-discharge checking

kawasima:
## What

Two related additions that let a domain express and check value constraints statically without giving up expressiveness.

### Closed newtype arithmetic
`+`/`-` over a single-value numeric newtype now yield that newtype (`金額 - 金額 : 金額`) instead of being a type error. The operator opens each operand to its base, computes there, and re-wraps the result, re-checking the newtype's invariant at construction — so a subtraction that goes out of range aborts like any other construction. `*`/`/` (dimension-changing) and two different newtypes stay rejected. Re-wrapping is a computation, not a fresh mint, so it needs no `constructs` (a `NewData` literal such as `金額(10m)` still does).

### Intraprocedural invariant-discharge check
A new pass verifies that a construction's declared invariant is actually established where it is built. It seeds a numeric abstract domain from the input newtypes' invariants (sound by closed construction), refines it along each `require`/`if` guard, and at every construction asks whether the guards discharge the invariant:

- **proven to violate on a reachable path** → compile error `E2010` (the path-sensitive generalization of the constant `金額(-5)` check);
- **left unproven** → warning `E2011`, a possible abort, with a hint to reify the relation into a type invariant;
- **not expressible in the domain** → opaque, no diagnostic, the run-time check stays.

This is contract-verification (proof obligation per site + run-time fallback), in the lineage of SPARK/Dafny — not refinement typing, so types stay unchanged and the check never becomes solver-dependent. A cross-behavior relation is closed by reifying it as a type invariant (which the seeding then picks up), not by inferring a hidden precondition, keeping with declare-don't-infer (ADR-0017).

The domain is interval + difference-bound (no external SMT). The flagged fragment is tied to the decision procedure, so every flagged construction has a guard the domain can verify.

## Tests
- `CompileClosedNewtypeArithmeticTest` (6): add/subtract keep the newtype, literal operand, invariant abort on the re-wrap, `金額 - 数量` and `*` rejected.
- `CompileInvariantDischargeTest` (5): definite violation errors; a sum of non-negatives discharges; an unguarded subtraction warns; a **reified** relation discharges; a **`require` guard** discharges.

Full suite green across all modules (671).

## Deferred
- a per-construction `unchecked` opt-out (needs an expression-prefix in the CST; non-blocking since the possible-violation is a non-fatal warning);
- eliding the run-time invariant check on a discharged construction (a performance optimization).


kawasima:
Addressed all 5 review comments in 7835357:

1. **NumericDomain rounding (soundness)** — single-atom bound derivation now rounds conservatively (CEILING for an upper bound, FLOOR for a lower one) instead of HALF_EVEN, so a derived bound is never tighter than the true one. Prevents a possible false E2010.
2. **E2011 message references `unchecked`** — removed the `unchecked` suggestion (EN/JA); that opt-out is deferred, so it pointed at nonexistent syntax. Keeps the `require`/reify guidance.
3. Same as 2 (JA message).
4. **`compileWithWarnings` hardcodes `Main`** — added a `compileWithWarnings(source, defaultModuleName)` overload for the CLI's filename-stem naming.
5. **LSP severity** — `fromDiagnostic` now maps `Severity.WARNING` to the LSP warning severity instead of always error, so warnings render as warnings in editors. (JSON renderer already emitted `severity`.)

Full suite green across all modules (671).

PR #47, opened 2026-07-23
Expressiveness refinements: Map.upsert, scalar newtype arithmetic, sum == case

kawasima:
Three independent day-to-day frictions, each an extension within existing architectural decisions — **no new ADRs** (spec updated instead).

## 1. `Map.upsert(key, default, f, m)`
Insert-if-absent / modify-if-present in one call, as a self-host `let` over `Map.get` + `Map.insert`. The absent branch takes a plain `default` (not a `Maybe`), so `Option` stays internal — filling the slot Elm's `Dict.update` needs but ADR-0011 forbids. Typical use: `upsert(k, 1, n -> n + 1, m)`.

## 2. Scalar `*` / `/` on a numeric newtype
`Money * 2`, `2 * Money`, `Money / 2` stay in the newtype — the dimension is unchanged, so it rides the closed `+`/`-` infrastructure (unwrap → compute → re-wrap with the invariant check). `newtype * newtype` (a dimension change / units) stays rejected. The invariant-discharge check linearizes a constant multiply; divide and a variable product are left opaque (run-time check stands).

## 3. `==` / `/=` between a sum and one of its cases
`Role == Junior` now type-checks: a case value is a value of its sum (case→sum is transparent everywhere else in the language), so this is a sum-vs-sum comparison by case. No new keyword — reuses the existing equality path (`Objects.equals`), no codegen change. Matches Elm's `x == Idle` on a no-argument variant and F#'s equality on a nullary case. Unrelated types (`Money == Qty`) still don't compare.

## Tests
- `CompileMapLibTest`: upsert present/absent.
- `CompileClosedNewtypeArithmeticTest`: `n * 2`, `2 * n`, truncating `n / 2`, `newtype * newtype` rejected.
- `CompileSumCaseEqualityTest` (new): sum == case true/false, `/=`, unrelated newtypes rejected.

Spec: §stdlib-map, §newtype-arithmetic, §equality (English examples). Full suite green across all modules (678).


kawasima:
Addressed both review comments in 9091ffd:

1. **`scalar / newtype` wrongly typed as the newtype** — `scalarNewtypeArith` now takes the operator. Division is not commutative, so only `newtype / scalar` scales (`Money / 2`); `scalar / newtype` (`2 / Money`) is an inverse (a dimension change) and is rejected. Multiplication stays commutative. Test: `2 / n` rejected.
2. **sum↔case equality leaking through collection covariance** — the exemption is now checked on the top-level case sets (`leafCases`) directly, not via `assignable` (which recurses into `List`/`Map`/`Option`/tuple covariance). So `List<Junior> == List<Role>` no longer compares; only the direct scalar sum↔case relationship does. Test added.

Full suite green (680).

PR #82, opened 2026-07-25
Close the gaps examples ran into

kawasima:
Writing examples against the current language turned up eight places where a rule could not be written, or could not be examined once written. This closes them.

## Examples can carry the values they need

- **Written temporals.** `Date("2026-07-01")` / `DateTime("2026-07-01T09:00")` take a written string, parsed and checked where they stand — in a body or in a fixture. Before this, a behavior taking a `Date` could not have a single `example`.
- **`Map` fixtures.** A map field is written as its list of `(key, value)` pairs, the form `Map.fromList` takes (Elm's `Dict.fromList`, F#'s `Map.ofList`), and the same list literal a `Set` field already took. The declared type decides what a list means, so a `List` field still reads as a list.
- **E1905 shows values.** A failing row reported `found: Out` / `expected: Out` — the arm name on both sides. Both are now encoded through the derived encoder and shown as values.

## The language reaches what the standard library returns

- **Tuple annotations.** `let entry (e: (String, Int))` parses. The spec already allowed a tuple type in a helper signature; the parser read the leading paren as a function type's parameter list. The two are told apart by whether `->` follows the closing paren. Destructuring is unchanged — `let (k, v) = e`, no second spelling added.
- **`Map.fold` with an empty seed.** `Map.fold(step, [], m)` failed to type while `List.fold(step, [], xs)` passed: the pre-inline step check judged a step whose accumulator was still the seed's bottom. It now leaves that to the inlined check, which sees the pushed-down type.
- **Nested spread.** `Address { ...c.address, city = city }` — the ordinary "change one field of a sub-record" edit, which needed a `let` binding first.
- **`List.take` / `List.drop`.** Elm's, derived from `fold`, so a top-N is one call.

## Diagnostics stop leaking internals

- A hint with no arguments of its own takes the message's. Seven hints (E1002/E1005/E1006/E1305/E1602/E1603/E1606) rendered `{0}` / `{1}` verbatim.
- An anonymous union prints as `` `Int | NotANumber` ``, not `Union[members=[Int, NotANumber]]`.
- A lambda of the wrong arity is reported against the parameter it fills (`` `f` of `let map` ``), not against `$0_f`.
- A record named inside `Some(...)` points at `| Some { field }` — the form that already works, so no second spelling was added.

## Examples follow

`tagging` gains examples for every behavior (a set, a map, an optional assignee) plus `topLabels`, a ranking over `Map.toList` using the tuple annotation, `List.sortBy` and `List.take`; its header comment claiming sets and maps cannot appear in fixtures was stale. `inventory` gains `checkExpiry`, a best-before rule fixed by examples written with dates.

Spec updated for the fixture notation, written temporals, E1905, and `take`/`drop`.

## Verification

`mvn clean install` at the root: 792 compiler tests, 44 LSP tests, green. `examples/`: `mvn clean test` green.

### DIFF
--- a/examples/ordering/src/main/souther/inventory.sou
+++ b/examples/ordering/src/main/souther/inventory.sou
@@ -152,3 +152,36 @@ example putAway
     | "a passed item is put away to a shelf" :
         (InspectionPassed { code = Barcode("9784873115658") }, Location("A-03-12"))
             -> PutAwayRecord { code = Barcode("9784873115658"), location = Location("A-03-12") }
+
+// === checkExpiry: a lot past its best-before date is held back ===
+// The warehouse stores dated lots. Whether a lot may ship today is a pure calendar computation —
+// `Date.daysBetween` over the lot's best-before date — so it stays in the domain rather than becoming
+// an injected behavior; only reading *today* needs the outside world, and it arrives as an input.
+// A date is written `Date("2026-07-25")`, in a body or in a fixture, so the rule is fixed by its
+// examples like any other.
+data Lot =
+    { sku: Sku
+    , bestBefore: Date
+    }
+data Fresh = { sku: Sku }
+data Expired = { daysPast: Int }
+
+behavior checkExpiry : (lot: Lot, today: Date) -> Fresh | Expired
+    constructs Fresh, Expired
+
+let checkExpiry (lot, today) = {
+    let past = Date.daysBetween(lot.bestBefore, today)
+    require past <= 0 else Expired { daysPast = past }
+    Fresh { sku = lot.sku }
+}
+
+example checkExpiry
+    | "a lot still within date ships" :
+        (Lot { sku = Sku("milk"), bestBefore = Date("2026-08-01") }, Date("2026-07-25"))
+            -> Fresh { sku = Sku("milk") }
+    | "a lot past its best-before is held back" :
+        (Lot { sku = Sku("milk"), bestBefore = Date("2026-07-20") }, Date("2026-07-25"))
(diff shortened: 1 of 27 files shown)


kawasima:
## Added: a helper's declared return type now survives inlining

Dogfooding `topLabels` turned up a ninth hole, of the same family, so it is in this PR too.

A helper's declared return type is a declaration into its body (spec §fn-declaration), but inlining dropped it. At a call site that expects nothing concrete — a generic parameter such as `Map.toList`'s — a `fold` over an empty seed inside the body had nothing to fix its accumulator, and the caller had to re-name the result:

```
let counts: Map<String, Int> = labelCounts(board.issues)   // only to make the next line compile
Map.toList(counts) |> ...
```

The inlined body now flows through a binding carrying that declaration, so the push-down applies at the call site and `Map.toList(labelCounts(board.issues))` types on its own. Only a collection-bearing, non-generic return is carried: a scalar has nothing to fix (and leaving those bodies bare keeps a constant-foldable expression plain for the compile-time invariant check), and a generic return says nothing at a call site.

Carrying it surfaced a backend gap worth naming separately: `List.find` / `List.sortBy` / `Option.map` re-type their collection argument to learn the element type, and did so without the recursive helpers' signatures in scope — so an argument holding the `foldFrom` a fold expands to failed to resolve (`List.foldFrom is not a standard-library function`, from codegen). They now use the same helper-aware environment the other re-typing sites use. That one was reachable before this PR, by writing `List.sortBy(f, <expression containing a fold>)`.

`mvn clean install`: 796 compiler tests, 44 LSP, green. `examples/`: green.

PR #90, opened 2026-07-26
Split the type checker, and take the repeated work out of the compile

kawasima:
`TypeChecker.java` had reached 4490 lines and about 150 static methods. Nothing about the checks required them to sit together — the type-level operations, the declaration checks, the helper typing, the pipeline routing, and the expression elaboration each answer a different question and share only the symbol table — so each is now its own class.

## The split

`TypeChecker` keeps what is actually its own job, running the phases and recovering past an error to report the next one, in **392 lines**:

| | lines | |
|---|---|---|
| `Elaborator` | 674 | the `elaborate` switch, let annotations, function values |
| `CallElaborator` | 527 | calls, builtin argument checks, type-variable binding |
| `DataChecker` | 559 | data / sum / decoder / encoder, constant constructions |
| `HelperTyping` | 581 | helper checking and parameter inference |
| `TypeOps` | 507 | `resolveType` / `assignable` / `unify` / `substitute` |
| `SpecChecker` | 431 | a behavior against its implementation, injection, stages |
| `TypeChecker` | 392 | the check phases and error recovery |
| `MatchElaborator` | 298 | the three forms of `match` and their exhaustiveness |
| `PipelineSigs` | 259 | behavior signatures and how `>->` routes cases |
| `BinaryElaborator` | 246 | binary operators and the newtype transparency rules |
| `BottomInfer` | 226 | the bottom an empty collection carries |

This is a move, not a rewrite: every method keeps its body, its signature, and its diagnostics. The one deletion is `sameKind`, which nothing called. `Sig` and `ReqSig` become records of their own, and callers point at the new home rather than going through a delegating `TypeChecker`, so there is one place to look for each operation.

`CheckContext` then collapses the `data`/`symbols`/`reqs` triple, which is fixed for as long as one definition is being checked, into one parameter: `elaborate` goes from six parameters to four. The context is built once per definition, not per expression, so nothing is allocated on the walk.

## The performance work

Profiled first (JFR over the examples) rather than guessed at, which was worth doing: the hypothesis going in was that `Fn.apply(Object[])` allocating an array per element was the hot path, and it turned out to be 0.4% — escape analysis removes it. What actually cost something:

**Generated code.** Every fold-derived combinator grows its result with `acc ++ [f(x)]`, so that expression runs once per element, and it compiled to an `ArrayList`, an `add`, a `List.copyOf`, and a `Lists.concat` that iterates the one-element list and throws it away. The backend now emits `Lists.append(acc, x)`. Over 20 000 elements with no warm-up, map+fold goes 5.1 ms → 3.6 ms and filter+fold 1.9 ms → 1.5 ms. **Once C2 has compiled the loop the two are level** — this helps the interpreted and freshly-compiled paths, which is what a compile evaluating its `example`s and a one-shot CLI run actually hit.

**The compiler.** `TypeChecker.symbols(module)` was walking every definition three times per single-module compile; each module's visible defs and imported signatures were resolved once per pass; `ClassDesc.of` re-validated the same class names at every emission site; and `HelperInliner`'s cycle search re-scanned a helper body once per path that reached it rather than once. A few percent in total, mostly inside the benchmark's noise — the honest summary is that no single hot spot exists, the time is spread across parsing, inlining, and ClassFile emission.

## Verification

`mvn clean test` green after each extraction and each optimization — **853 tests** (852 before, plus the one added below). A compile benchmark over the examples (six module groups, median of 25 rounds) went 44.1 ms → ~42 ms, i.e. no regression from the split.

## Review

A code review found that hoisting the per-module import resolution ahead of the checking pass changed *which* error a failing multi-module compile reports: a third module's unresolvable pipeline, reached through a second module's import, would surface instead of a type error in the first. Fixed in the last commit, with a test that fails on the hoisted version. Half of that finding did not hold — `visibleDefs` is already resolved for every module in pass 1.5 — and the commit message says so.


kawasima:
Confirmed and fixed in 71a4ce7.

It reproduces: an implemented `behavior now : () -> D` used as anything but the first stage dies with `IndexOutOfBoundsException: Index 0 out of bounds for length 0`. As the comment says, that is not a `CompileException`, so the recovery boundary in `checkCollecting` does not catch it and the author gets a stack trace instead of a line number.

Two notes on the surrounding shape:

- **The rule was already stated.** `SpecChecker.checkStagesAreSingleInput` rejects arity 0 as readily as arity 2. It just does not get there first: signatures are built before it runs, and are also built for an imported module that this compile never checks. So the arity is now confirmed where `route` consumes it, raising the same `check.pipe.multiinput` diagnostic — a duplicate from the two phases collapses in `deduped`.
- **Pre-existing, not introduced here.** `git show develop:...TypeChecker.java` has the same `Type in = g.in()` in `route` and the same unguarded `ins.get(0)` in `Sig.in()`; this PR moved both unchanged. Fixing it here anyway since the review surfaced it against the new file.

Test added (`CompileZeroArgStageTest`), 854 tests green.

Issue #95, opened 2026-07-26
Cross-module composition: the sealed-permits premise does not hold on the classpath


kawasima:
## Summary

Cross-module composition is closed off on the grounds that the JVM cannot do it. Measured on JDK 25, it can: the same-package requirement for a sealed type's permitted subclasses is a **javac source rule**, not a load-time rule, and Souther emits classfiles directly. A generated result union permitting an imported module's case class loads from the classpath and still gives Java consumers an exhaustive `switch`.

The decision to keep behaviors module-local may still be the right one, but the premise it currently rests on does not hold and should not be the reason.

## The recorded premise

`Backend.java:154-158`:

```java
// A generated result union is a sealed interface, so all its permitted case classes must be
// in this module's package (JVM: an unnamed-module sealed type permits only same-package
// subclasses). A `>->` whose departed case comes from an imported behavior would need that
// upstream case to join this union, which it cannot. Reject it with a clear message rather
// than emit permits pointing at a case class that is never generated here (E1606).
```

## What was measured

Two packages, `pa.Res` sealed permitting `pa.Ok` and `pb.Shortage`:

```java
// pa/Res.java
package pa;
public sealed interface Res permits pa.Ok, pb.Shortage {}

// pb/Shortage.java
package pb;
public record Shortage(String code) implements pa.Res {}
```

1. `javac` **with** a `module-info.java` (named module): compiles.
2. `javac` **without** it (unnamed module): rejected —
   `名前のないモジュールのクラスResは別のパッケージのシール・クラスを拡張できません`.
3. Take the classfiles from (1), delete `module-info.class`, put them on the **classpath**, and load:
   ```
   loaded ok: Shortage[code=x] / sealed=true
   ```
   The JVM does not reject the cross-package permits at link time.
4. Compile a Java consumer against those classpath classfiles with an exhaustive `switch` and no `default`:
   ```java
   static String f(pa.Res r) {
     return switch (r) {
       case pa.Ok o -> "ok " + o.n();
       case pb.Shortage s -> "short " + s.code();
     };
   }
   ```
   Compiles and runs (`short x`).

So javac enforces the package rule when it *declares* a sealed type in source, and accepts one read from a classfile. Souther never goes through javac to emit the union.

## What this currently blocks

All three of these are rejected today, and all three are ordinary shapes when a domain is split into bounded contexts:

- An imported case in a local behavior's output union — E1606:
  ```sou
  import probe.inventory ( 商品コード, 在庫不足 )
[...]

Issue #57, opened 2026-07-23
Injected behaviors are limited to a single argument (extra params silently dropped)


kawasima:
## Summary

An **injected behavior** (a `behavior` with no `let`, implemented from Java) can only take a single input. Declaring more than one parameter is **silently accepted** and all parameters after the first are **dropped** — surfacing either as a confusing arity error at the call site, or as a silent correctness hole.

This is inconsistent with a **fn behavior**, which can take multiple arguments (`会員へ通知する : (会員, 件名) -> …`). The asymmetry has no principled justification visible to the author.

## Measured

```
module p
data A = { x: Int }
data B = { y: Int }
data R = { z: Int }
behavior send : (a: A, b: B) -> R
    constructs R
behavior use : (a: A, b: B) -> R
    requires send
let use (a, b, send) = send(a, b)
```

- `send(a, b)` (the natural way) → `send expects 1 argument(s), got 2` — a late, unexplained error.
- The same `send` called with **one** argument (`let use (a, send) = send(a)`) → **COMPILES OK**: the declared 2nd input `b: B` is silently dropped, never reaches the Java implementation.

## Root cause

The single-input assumption is baked into three layers:

1. **Checker** — `TypeChecker.ReqSig` holds a single `param` (`spec.params().get(0)`); the call check hard-codes `arity(call, 1)`.
2. **Codegen (base)** — `Backend.generateRequiredBase` takes a single `inputRef` and unconditionally `implements Behavior<In, Out>` (a unary `apply`). For a multi-param spec `inputRef` is `null`, so the base is a raw `Behavior` with a single-arg `apply` — the extra params vanish.
3. **Codegen (requires side)** — the `$Impl` stores each required behavior as a `CD_Behavior` field, and every required-call site emits `invokeinterface Behavior.apply(Object)Object` with at most one boxed argument (BodyGen ~275/980/1243).

## Why this is the wrong shape

The unary `Behavior<I, O>` is itself **correct and deliberate** — it is the *composition* contract for `>->` (a pipeline stage takes one value → one value), documented on the runtime interface. The mistake is treating "injected behavior" as a synonym for "unary `Behavior<I,O>`".

[...]

PR #103, opened 2026-07-26
Let an injected behavior be named where an implemented one can

### LINKED ISSUE #96: Diagnostics: an imported behavior in a pipeline or requires is reported as a missing name or an arbitrary Java call
## Summary

Using an imported behavior in a `>->` pipeline or as a `requires` dependency is rejected with messages that describe a different problem. Both read as "you got the name wrong" or "you forgot to declare it", when the real answer is that imported behaviors cannot be used in either position at all.

## Reproduction

`inventory.sou`:

```sou
module probe.inventory
exposing (商品コード, 引当済み, 在庫不足, 引き当てる)
data 商品コード = String
data 引当済み = { 商品: 商品コード, 数量: Int }
data 在庫不足 = { 商品: 商品コード }
behavior 引き当てる : (商品: 商品コード, 数量: Int) -> 引当済み | 在庫不足
```

As a pipeline stage:

```sou
import probe.inventory ( 商品コード, 引当済み, 在庫不足, 引き当てる )
behavior 引当して出荷する = 引き当てる >-> 出荷指示を出す
```

```
-- 名前が見つかりません
このパイプラインに `引き当てる` という名前の behavior が見つかりません。
```

The name is imported and in scope; it is not missing.

As a `requires` dependency:

```sou
behavior 出荷する : (商品: 商品コード, 数量: Int) -> 出荷指示
    requires 引き当てる
let 出荷する (商品, 数量, 引き当てる) = ...
```

```
-- 任意の Java 呼び出し  E1401
`引き当てる` は behavior でも組み込みでもありません。
ヒント: `let` を持たない behavior を宣言し Java 側で実装してください。任意の JVM 呼び出しは許可されていません。
```

`引き当てる` *is* a behavior, declared with no `let` and implemented on the Java side — exactly what the hint asks for. The message accuses the author of an arbitrary JVM call.

## Where it comes from

The pipeline stage resolver reports through `check.unknown.behavior.msg`, a generic name-not-found message, without distinguishing "no such name" from "the name resolves to an imported behavior". The `requires` path falls through to the E1401 arbitrary-call diagnostic for the same reason.

## Suggested direction

Give each position its own diagnostic that names the actual rule — an imported behavior cannot be a `>->` stage / cannot be a `requires` dependency — and say what to do instead (consume it at the Java boundary, or re-declare it locally). This is worth fixing regardless of how the cross-module composition question is decided; if that restriction is lifted, these messages go away with it.

## Related

- #95 (cross-module composition: the sealed-permits premise)

## Environment

develop @ 996acb3, 0.1.0-SNAPSHOT, JDK 25.

kawasima:
Fixes the two diagnostics of #96, by removing the situations that produced them.

## What was measured

The issue reads as "an imported behavior cannot be used in either position". Half of that is wrong, and the half that is right has a wider cause.

| position | case | before |
|---|---|---|
| `>->` stage | imported, injected, one input | compiled |
| `>->` stage | imported, injected, several inputs | "I cannot find a behavior named …" |
| `>->` stage | **local**, injected, several inputs | same message — no import involved |
| `requires` | imported behavior, any arity | E1401 arbitrary Java call |
| `requires` | local implemented behavior | `check.call.nonrequired` (fine) |
| `requires` | a name that does not exist | E1401 (fine) |

`PipelineSigs.signatures` registered a behavior only when a `let` implemented it or it took exactly one input, so an injection target with none or several dropped out of the map and was reported as a name that was never declared. `TypeChecker` built the required signatures from `module.behaviors()` alone, so a `requires` never reached an imported injection target — while the same behavior named as a `>->` stage was injected and bound through `importedInjected`.

## What changed

- Every behavior's declared signature is registered, implemented or injected. The arity rules are enforced where the composition is written, not by leaving the name unresolvable.
- A multi-input injected behavior can be a pipeline's first stage, which receives the pipeline's own arguments. It is called on its base class with the typed `apply`, the arguments cast from the erased one. A stage after `>->` still takes one input.
- A `() -> R` injection target keeps the unary `Behavior` shape and is called with the value its implementation ignores — the same call a body already makes for a zero-input dependency.
- The required signatures are seeded with the injection targets a module imports, so `requires` reaches one across a module boundary.
- **E1607** checks the `requires` clause where it is written, telling a name that carries an implementation apart from one that is no behavior in scope (with a did-you-mean).
- `specification.adoc`: the first-stage rule, the `requires` section, and the E1607 entry.

## A load failure this also fixes

The same omission meant a multi-input or zero-input injected behavior with a union output got no `<name>Result` interface, although its generated `apply` returns one:

```
2 inputs: [… probe.inventory.引き当てる]              ← no 引き当てるResult
1 input:  [probe.inv1.引き当てるResult, … 引き当てる]
```

A Java implementation of that behavior failed with `NoClassDefFoundError`, after a compile that reported success. `Backend.java:245` — "take its arity from the imported signature so the unary-vs-multi dispatch treats a cross-module multi-input dependency the same as a local one (issue #57)" — was also inert for the same reason, its `importedSigs` lookup always null. Both work now.

## Where the issue's reproductions land

Both keep an imported case in the consuming behavior's own output union, so both now report **E1606**, the rule they actually hit. E1606's hint still explains itself as a JVM limit, which #95 measured and disputes; that wording is left to #95.

## Verification

compiler 917 / LSP 46 / runtime 25 green from a clean build; `examples` green. Twelve new tests, including binding and running a multi-input injected first stage, a `requires` dependency imported across modules, and the generated `<name>Result`.

Closes #96


kawasima:
Fixed in 9ceab55 — thanks, this was a real hole.

Verified it first: `requires` naming a `>->` composition rendered *"Compose `chain` with `>->`, or remove its `let` so it becomes an injection target"*, advice for a `let` that does not exist. The same wording was wrong for an imported behavior, whose implementation is not this module's to change.

- Hint now leads with the advice that holds for all three cases: `Remove {1} from requires and compose it with >->; only a behavior with no implementation of its own is injected.`
- Message says "has an implementation of its own" rather than "carries an implementation", so it reads true for a composition as well.
- Regression test asserts the rendered hint names `requires` and never mentions `let`.

PR #129, opened 2026-07-26
Claim a tail slot instead of copying the tail on every append

kawasima:
## Why

I measured where a generated program allocates, and the cost turned out to sit in one place.

| path | HotSpot 21 (C2) | GraalVM 25 |
| --- | --- | --- |
| `acc ++ [x]` over 1000 elements — what `List.map` emits | 144 B/element | 112 B/element |
| accumulating into an `ArrayList`, converting once at the end | 43.9 B/element | 43.8 B/element |
| an unboxed `long[]` | 8.0 B/element | 8.0 B/element |
| a fold that only reduces (`Int` or `Decimal` sum) | — | ~0.1 B/element |

A fold that only reduces costs nothing: escape analysis erases the per-element `Option.Some` from `List.get`, the `Object[]` from `Fn.apply`, and `BigDecimal`'s intermediates. A fold that *builds* a list does not, because the list escapes. Of its 144 bytes, 16 are the boxed `Long`, 40 the new `PersistentVector`, ~4.5 the trie spine on spill — and about 82 are the tail, which `append` reallocated and copied on every single element.

ADR-0042 deferred a transient accumulator and said to reopen it "with a workload, not on principle". This is that workload.

## What changed

A tail array is now `WIDTH` (32) wide with only `cnt - tailoff()` slots in use, and successive appends write into the same array instead of copying it. Slot `j` belongs to whoever moves `claimed` from `j` to `j + 1`, so **each slot is written at most once, ever**. A version holding `m` elements reads only `[0, m)`, and `claimed == m` when it was constructed, so those slots were all written before it existed. A second append off that same version loses the claim and copies `[0, m)` into a fresh array, so branching keeps working.

**The claim is confined to one thread rather than made atomic.** Every slot a version can read is then written by the thread that froze that version, before the freeze, so the JLS 17.5 guarantee for objects reachable from a final field still covers the whole tail — a vector handed to another thread reads correctly even when it was published through a data race, exactly as when every append allocated a fresh array. A compare-and-set would let two threads write one array and lose that guarantee formally, in a way no test would show and no bug report could be traced back to; it would also put a lock-prefixed instruction on the path whose whole purpose is to be cheap.

Nothing in the compiler changed. The claim is a property of the runtime structure, not of the shape of a step, so it applies to every `++` — the fourteen distinct combinator shapes in `souther/list.sou`, the ones threading an accumulator inside a tuple, and user code alike. That is also why ADR-0042's objections do not reach it: they are objections to a *compiler-side* transient. ADR-0051 is untouched — nothing keys on the name `fold`.

Separately, a class with no state is now generated once and handed out rather than allocated per use: a unit data, a lambda that captures nothing, a decoder/encoder implementation, and `Option.none()`.

## Results

| path | HotSpot 21 (C2) | GraalVM 25 |
| --- | --- | --- |
| `List.map` building a 1000-element list, through the behavior | 160 → 106 B/element | 112 → 66 B/element |
| `Lists.append`, which is what `acc ++ [x]` emits | 144 → 66 B/element | 112 → 38 B/element |
| `PersistentVector.append` alone, unboxed | 128 → 50 B/element | 128 → 50 B/element |
| a behavior returning a unit data | — | 16 → 0 B/call |

`append`'s constant factor also goes from O(32) to O(1) — an average 16-element `arraycopy` and 80 bytes of zero-fill per element are gone — so the wall-clock gain should exceed the allocation gain.

On C2 this **does not reach** the 43.9 bytes the `ArrayList` route costs. The floor there is the 40-byte `PersistentVector` allocated per append, which cannot be removed here: the previous version's `cnt` is its identity, so a new object is mandatory. Reaching it needs a real transient accumulator with a linearity guarantee — still what ADR-0042 deferred, and still deferred. On GraalVM it goes below 43.9 because once the tail copy is gone the remaining vector no longer escapes far enough to defeat that compiler's escape analysis; that figure should not be read as the general case.

## Costs accepted

Both are recorded in ADR-0060.

Appending repeatedly to the same retained *small* list allocates more than before — `base ++ [x]` with a three-element `base` copied 32 bytes each time and now copies 176. The number of copies is unchanged; each is bigger, and the crossover is around six or seven elements. If it matters, the copy path can size its array at `min(WIDTH, nextPow2(n+1))` and treat `a.length` as the capacity; spill only happens at `n == WIDTH == a.length`, so the sealing rule survives that unchanged.

A retained early version whose descendants grew keeps a 32-wide array alive, pinning up to 31 elements that are not logically part of it — at most 128 bytes and 31 references per retained vector, the `String.substring` retention class of problem. Keeping the bulk `Builder` at exact size keeps it out of the decode path, where it would be worst.

## Verification

`mvn clean test`: 1192 green (1183 before, 9 new). `examples/`: green, including the Spring Boot + H2 `ordering` case that starts a context, connects, and checks the first stage's INSERT rolls back when the second returns `OutOfStock`.

The correctness tests are the point: `everyIntermediateVersionStaysIntact`, `branchingOffEveryPrefixKeepsAllThreeCorrect`, `branchingAfterSpillKeepsTheRetainedVersion`, and `concurrentAppendsOffOneVectorAllSucceed` (eight threads racing for one tail slot). Removing the `claimed == used` test makes `branchingOffEveryPrefix...` fail on version 1 immediately, which is what a claim test is for.

`PersistentVectorAllocationTest` asserts the budget rather than reporting it — allocation is counted exactly by the JVM, so unlike a timing benchmark it can carry a real assertion and stay enabled. Its bound clears the cost on both object layouts (50 B/element with compressed oops, 73 without; verified with `-XX:-UseCompressedOops`).

## Not in scope

Found while investigating, left for separate work: `raoh` 0.6.0 is class-file 69 (Java 25), so generated output does not in fact run on Java 21 as `specification.adoc` §Target JDK, ADR-0022 and the non-functional requirements state; an `invariant` written on a unit data is silently dropped with no diagnostic; `-XX:+UseCompactObjectHeaders` (16 → 8 B/call for a unit data) and a CLI AOT cache (startup 0.36 → 0.26 s) are worth documenting; and `PersistentHashMap`/`PersistentHashSet` cannot take this trick, since CHAMP's `assoc` rewrites rather than appends.


kawasima:
Thanks — all three land. Pushed c95a67a.

**The JLS 17.5 wording.** The comment on `Tail.a` said only "final so the guarantee reaches the elements", and read on its own that is the overstatement you describe. Final-ness is necessary and not sufficient: the guarantee covers writes performed before the freeze of the object being read through, and a shared array outlives many freezes.

What the code actually relies on is the one-owner rule, and I had left it implicit. Without it the failing case is exactly the one you point at — T1 writes `a[0]` and freezes `V1`; T2 obtains `V1`, writes `a[1]` and freezes `V2`; a reader of `V2` needs `a[0]`, but T1's write is ordered before T2's freeze only if `V1` reached T2 with a happens-before edge, and through a race there is none. That case is why the claim is thread-confined rather than a CAS, and it is now written down instead of assumed. With one owner per array, every element a version can read was written by the very thread that froze that version, so program order supplies the ordering and JLS 17.5 covers the whole array.

The read direction needed saying too: a non-owner that loses the claim copies the prefix, and it reads that prefix through the *previous* version's freeze, performed by the owner after writing it. The comment on `Tail.a` now points at `owner` and notes that changing either half without the other breaks the argument.

I have not changed the design — the property holds as implemented — but the comment now states the condition that makes it hold rather than leaning on `final` alone.

**The common pool.** Right, and it weakens the test rather than breaking it: the latch is counted down from the main thread after submission, so queued tasks are released and it cannot hang, but with parallelism below eight the appenders run in batches and never contend for the slot the test exists to contend for. It now uses its own fixed-size pool, shut down in a `finally`. Blocking common-pool threads on a latch was poor citizenship regardless.

**The unconditional print.** Removed. The assertion message already carries the figure when it matters, and the class javadoc and ADR-0060 record it when it does not.

PR #130, opened 2026-07-26
Build a CHAMP trie in one pass where linear use is a fact

kawasima:
## Why

ADR-0060 took the fold path's cost out of `PersistentVector` with a claimed tail. The natural next question is whether `PersistentHashMap` and `PersistentHashSet` can have the same. Measured first, per ADR-0042's "with a workload, not on principle" — building a 1000-entry map, `ThreadMXBean.getThreadAllocatedBytes` after warm-up, GraalVM 25.0.3 (arm64):

| path | before |
| --- | --- |
| `PersistentHashMap.from` — every decoded `Map` field, `Map.fromList` | 406 B/entry |
| `PersistentHashSet.from` — every decoded `Set` field, `Set.map`/`filter`, the set algebra | 415 B/entry |
| `assoc` one entry at a time — `groupBy`, `indexBy`, `Map.map`, `distinct` | 357 B/entry |

`from` costing *more* than the element-wise loop it exists to replace is the tell: it had no bulk path at all, and simply looped `assoc` over an intermediate `HashMap`.

## The claimed tail does not transfer

Worth stating plainly, because it is the reason this PR looks different from #129.

A vector's in-place write goes to a slot **no existing version reads** — a version holding `m` elements reads `[0, m)` and nothing beyond, so the write is invisible to everything but the version returned. A CHAMP insert has no such region: it rewrites the `dataMap`/`nodeMap` bitmaps that older versions read to interpret the very same array. Any in-place write there is observable by every version sharing that node.

So ownership cannot be established after the fact — not by a claim, not by a generation counter, not by a thread check. The question is not *who may write* but *who can see*, and the answer is everyone holding an older map. No runtime mechanism changes that.

## What this does instead

Confines the mutation to where linear use is structural rather than hoped for: a package-private, single-use, from-empty **builder** whose trie never leaves the method building it. That is the same confinement `PersistentVector.Builder` already had, and the same reasoning ADR-0042 used for why that one was safe while a general transient was not.

It backs `PersistentHashMap.from`, `Maps.fromList`, `Maps.mapKeys`, `PersistentHashSet.from`, and `union`/`intersect`/`difference` — which covers every decoded `Map` or `Set` field at the boundary, `Map.fromList`/`Set.fromList`, and `Set.map`/`Set.filter` (written over `toList`/`fromList`, so they take the bulk path).

**Ownership is a flag threaded down the call, not a mark on each node.** Starting from the shared empty node and only ever inserting, every node the builder can reach below the root is one it built — nothing adopts a node from elsewhere. The empty node is the one it does not own and is never written, since a write needs an entry or a child to replace and it has neither.

The marked-node version was written first and measured: a reference on every node of every map cost the persistent `assoc` path 6% to speed up the bulk path. That is the wrong trade — `assoc` is the common path — so it was replaced. Node fields stay final, so a map is still safely published through them (ADR-0060); what the flag permits is writing an *element* of an array nothing else can reach.

## Results

| path | before | after |
| --- | --- | --- |
| `PersistentHashMap.from` | 406 B/entry | 167 B/entry |
| `PersistentHashSet.from` | 415 B/entry | 164 B/entry |
| `assoc` one entry at a time | 357 B/entry | 357 B/entry |

Bulk construction about 2.4× leaner; the persistent path deliberately untouched, and `PersistentHashMapAllocationTest` asserts the first row so it stays that way.

## What is not covered

`List.groupBy`, `List.indexBy`, `List.distinct`, `Map.map`, `Map.filter`, `Map.union`, `Map.intersect` and `Set.partition` call `Map.insert`/`Set.insert` per element from generated code, where the accumulator is a genuine persistent value the caller may branch on. Those stay at 357 B/entry. Making them cheaper needs the accumulator to be known linear, which is the compiler-side analysis ADR-0042 deferred and still defers — this PR does not change that verdict, it explains why the runtime cannot substitute for it.

## Verification

`mvn clean test`: 1204 green (1197 before, 7 new). `examples/`: green, including the Spring Boot + H2 `ordering` case.

The correctness tests are the point. `builderProducesTheSameMapAsRepeatedAssoc` checks contents, size and the deterministic iteration order the boundary encoding depends on, across trie depths and with duplicate keys; `builderHandlesCollisionsLikeRepeatedAssoc` drives the `HashCollisionNode` path the builder does not own; `aBuiltMapIsPersistentAfterwards` and `aBulkBuiltSetIsPersistentAfterwards` are the ones that fail if ownership ever leaks past `build`. Making `assoc` pass the ownership flag fails nine tests, two of which predate this change.

The allocation budget clears both object layouts (verified with `-XX:-UseCompressedOops`).


kawasima:
Correct, and the wording understated the builder rather than overstating it. Pushed 4bdf981.

"Only the leaf, whose array changes length, still allocates" describes inserting a *new* key and then quietly implies the same for every put. An overwrite takes `copyAndSetValue` with the flag set, writes the slot in place and returns the node — so nothing allocates on that path, and the parent has nothing to rewrite either.

The accurate statement is that what remains to allocate is what changes an array's length or a bitmap: adding a key to a node, and splitting an inline pair into a sub-node when two keys collide at that level. Both javadocs now say that, and the one on `copyAndInsertValue` no longer calls itself "once per entry" when it is once per key actually added.

Measured it rather than reasoning about it, through `Maps.fromList` (which takes a list and so can carry a repeated key, unlike a `Map` source):

```
  fromList: 1000 distinct keys            107.91 bytes/entry
  fromList: 1 key, 1000 overwrites          0.15 bytes/entry
```

1204 still green.

Issue #150, opened 2026-07-27
The generated `@NullMarked` reaches almost no module, and writing it by hand does not stop the warning


kawasima:
## What happens

Since #142 each generated package carries JSpecify's `@NullMarked` on a `package-info` class emitted
beside the module. When the module's own package already declares a `package-info.java`, that
declaration wins, the generated one is skipped, and a warning is printed instead.

That is 7 of the 8 modules in `souther-examples`: account, businesstrip, contact, invoicing,
issuetracker, joboffer, sharedmoney. Only ordering escapes, because its hand-written Java sits in
`app.ordering` rather than in the module's package.

```
$ mvn -o clean test -pl contact
[WARNING] souther: `example.contact.package-info` is declared by this compilation, so the generated
`@NullMarked` was not written; add it there to keep the module's types non-null for Kotlin
[INFO] BUILD SUCCESS
```

The `package-info.java` in the module's own package is not an accident of those examples — it is the
one Java source that makes javac run a processing round at all (#137). So the layout Souther's own
examples show is the layout where the nullness marking does not land.

## The warning does not stop when you do what it says

`SoutherProcessor` warns from the `FilerException` handler without looking at what the declared
`package-info` says. `examples/issuetracker` now carries `@NullMarked` on its own `package-info.java`
— the generated types really are non-null to Kotlin, verified below — and the same warning still
prints on every build. A warning that survives being fixed is one nobody reads.

## Following the advice needs the jar the spec says nobody needs

> The annotation is read by name, so the JSpecify jar need not be on the consumer's classpath.
> — specification.adoc, Nullness

True of the generated `package-info`, and false the moment a consumer is told to write the annotation
themselves: `examples/issuetracker` had to add `southerCompileOnly("org.jspecify:jspecify:1.0.0")` to
compile its own `package-info.java`.

It cost a second thing in that build. javac emits a `package-info.class` only when the package
declaration carries an annotation, so adding `@NullMarked` produced that class from two compilations
at once — the source set that runs the processor and `main`, which share `src/main/java` — and the
jar task refused the duplicate entry. The example had to take `src/main/java` out of `main` entirely.

## Option: mark the types, not the package

`@NullMarked`'s targets are `MODULE, PACKAGE, TYPE, METHOD, CONSTRUCTOR`, and kotlinc reads it on a
class:

```java
package probe;
import org.jspecify.annotations.NullMarked;
@NullMarked
public class Probe { public static String take(String s) { return s; } }
```
```kotlin
[...]

PR #151, opened 2026-07-27
Mark the classes, not the package

kawasima:
Fixes what issue #150 reports: the generated `@NullMarked` reaches almost no module, and writing it by hand does not stop the warning.

## The marking rides on each class

A module's package is where a consumer's own `package-info.java` sits — in a project whose only Java is that file, by construction. The generated `package-info` lost to that declaration, the marking was skipped, and the warning saying so kept printing after the consumer did what it asked, since it never read what the declared package already carried.

Every generated class carries the annotation instead. They all go through `Descriptors.build`, so it is one place, and it covers the `$`-named classes (`Foo$Dec`, `Foo$Enc`) that an enclosing-element rule would not. With nothing generated under that name any more, the `package-info` generation goes and the processor's `FilerException` handler goes with it — a collision there is now as much a build failure as any other.

Nobody is asked to write the annotation, so nobody needs the JSpecify jar on any classpath.

## `__construct` hands back a typed `Result`

It had no `Signature`, so `Result<T, E>` was raw — a platform type on the Kotlin side, in the one method a Java binding reaches for. It is public for an exposed data and is the path another module's `constructs` takes, so it gets the signature it returns: `Result<T, String>`, the failure side being the invariant message the method builds.

`$Dec.decode`'s raw `net.unit8.raoh.Result` is left alone: consumers reach it through the generic `decoder()`, so nothing sees it.

## Verification

Generated classes run through kotlinc 2.2.21, with the type forced out by mismatching it:

```
Probe.kt:4:18: error: initializer type mismatch: expected 'Int', actual 'String'.
    val x: Int = r.note()

Probe2.kt:5:18: error: initializer type mismatch: expected 'Int', actual 'Result<Receipt, String>'.
    val x: Int = Receipt.__construct(a, "note")
```

`String`, not `String!`; `Result<Receipt, String>`, not `Result<*, *>!`. No `package-info.class` among the output.

`mvn -o clean test`: compiler 1115, LSP 81, formatter 37, all green.

`BOUNDARY_VERSION` stays at 2 — no descriptor and no visibility changes, so a jar built before this still links.

### DIFF
--- a/souther-compiler/src/main/java/souther/compiler/apt/SoutherProcessor.java
+++ b/souther-compiler/src/main/java/souther/compiler/apt/SoutherProcessor.java
@@ -9,7 +9,6 @@ import souther.compiler.meta.ModulePath;
 
 import javax.annotation.processing.AbstractProcessor;
 import javax.annotation.processing.Filer;
-import javax.annotation.processing.FilerException;
 import javax.annotation.processing.RoundEnvironment;
 import javax.annotation.processing.SupportedAnnotationTypes;
 import javax.lang.model.SourceVersion;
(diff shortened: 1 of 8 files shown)


kawasima:
Taken the double compile — the no-package-info case now takes the `keySet` once and uses it for both the predicate and the message.

Not taken: the two comments asking for "contains `@NullMarked`" rather than the whole runtime-visible list. What a generated class states to a consumer's compiler is the boundary, so the set of runtime-visible annotations on it is the assertion, not incidental strictness. The module's own declarations ride on `$Module` as runtime-*invisible* precisely so consumers do not read them; a second runtime-visible annotation arriving on every generated class is a decision to take deliberately, and failing here is how it gets taken. Added the reason as a comment beside the assertion so it does not read as accidental.

Issue #178, opened 2026-07-28
A helper's body-determined parameter type does not reach its expansion


kawasima:
Issue #176 gives a helper a type of its own: its parameter types come from its body rather than from its call sites. That type stops at the helper. Where the helper is expanded, the binding the call becomes carries only the type that was *written*, so a body-determined type is dropped and the argument's own type flows into the re-checked body.

## What happens

```
data A = { x: Int }
data B = { y: Int }
data U = A | B
data W = { u: U }
data X = Int

behavior f : (a: A) -> X constructs X, W

let describe (v) = {
    let w = W { u = v }
    match v with
        | A as a -> a.x
        | B as b -> b.y
}
let f (a) = X(describe(a))
```

`v` is determined as `U` — it is the value of `W`'s `u` field — so `match v` is a match on a sum and the helper checks out standalone. At the call site it is rejected:

```
13|     match v with
        ^^^^^
match needs a sum-typed value, but A was passed
```

Writing the type that the body already determined — `let describe (v: U)` — compiles. Nothing else differs.

## Where it comes from

`HelperInliner` rewrites `describe(a)` to `let $k_v = a in <body>` and carries the parameter's *written* type onto that binding, for the reason its own comment gives: so a value known to be a sum is not narrowed to the argument's specific case when the body is re-checked inline. A body-determined parameter has nothing written, so the binding carries no type and the narrowing happens.

This is not a regression. Before #176 the parameter was typed *from* the call site, so it was `A` there too and the same program failed in the same place. What changed is that the helper now has a settled type — and the expansion does not use it.

## What a fix has to solve

The determined type is a resolved type; `Ast.FnParam` and `Ast.LetIn.declaredType` hold surface syntax. So the type cannot simply be written back.

1. Carry the surface type node alongside the determined type where the determining position has one — a construction's field, a callee's declared parameter, a recursive helper's parameter — and write it onto the helper's parameter. Positions that yield only a resolved type (an operator's other operand, an intrinsic's parameter) stay unfixed, so the rule would hold in some places and not others.
2. Resolve the general case with a type-to-surface writer. That is resolution run backwards, and it is better written after #177 step 2 (names are resolved once, into the tree), where the tree already carries resolved names.

Found while reviewing the #176 branch. See ADR-0066.

kawasima:
Fixed by #181 (merged to develop).

Issue #177, opened 2026-07-28
One question, several implementations: nothing in the compiler answers a name, a type, or a diagnostic once


kawasima:
Two test classes in this repository describe the same defect pattern, each about a different pair of implementations, each written after the defect it describes.

`CompilePathAgreementTest`:

> One module compiles the same way whether it is compiled alone or linked with others. The two are separate sequences over the same stages … so a feature can work on one and not the other. That is what happened to the recursive prelude, which the linking path did not inject, and every module that folded failed there with an NPE while the single-module tests stayed green. **Nothing compared them.** Each path had its own tests over its own inputs.

`DiagnosticPathAgreementTest`:

> A compile reports through `CompileException`, which the CLI and the annotation processor render; an editor reaches the same source through `Analyzer` … The two are separate implementations over one compile, and every diagnostic defect this project has filed — #93, #94, #98 — was one path behaving differently from another. **Nothing compared them.**

The response so far has been to add an agreement test per pair, after the pair has drifted. That covers the inputs someone thought of, for the pairs someone noticed.

## The same shape, further down

The compiler has no single answer to "what does this name denote".

- `Ast.TypeRef` carries a `String`, sum cases carry `String`, `constructs` carries `List<String>`. Names reach the checker as written, never resolved.
- Resolution happens where each check needs it: 60 calls to `Symbols.resolve` / `resolveCase` / `resolveIn` / `declaration` across the checker, the deriver, codegen and the example runner.
- A name written inside another module's declaration has to be resolved there, so `Symbols` carries `resolveIn(inModule, written)` and 32 places thread a `home` argument to say where the text came from. That argument exists only because resolution happens late.
- The grammar open-codes "a name may be qualified" six times in `CstParser`: `qualifiedName`, `stage`, the type-reference head, `caseNameTail`, and the two lookahead predicates `atLetPattern` / `atCtorPattern`. (Three further dotted loops — `fieldAccessChain`, the spread's field path, and the `.field` getter — read a field-access chain, which is a different production and does not collapse into this one.) `AstBuilder` reassembles the dots through two functions rather than one, because the parser wraps a module name in a node and leaves a case name as a run of tokens.

Because each site decides for itself, a capability reaches the positions someone wired it to. PR #175 fixed three instances at once: a field read that followed a name but not a call; a `constructs` clause that took a bare name while `requires` took a qualified one; and E1002 / E1006 / E1305 comparing spellings rather than types. #101, #113, #124, #132 and #154 are the same class.

[...]

PR #179, opened 2026-07-28
Type a helper's parameters from its body, not from its call sites

### LINKED ISSUE #176: A helper `let` has no type of its own: its parameters are typed from its call sites
A helper `let` has no type of its own: its parameter types are read from its call sites, so what a helper means depends on who calls it. That is the one property that has to change before a definition can be asked about on its own.

## What the compiler does today

```
let double (x) = x * 2
```

```
-- IMPLEMENTATION MISMATCH ---------------------
6| let double (x) = x * 2
               ^
helper `let double` parameter `x` cannot be inferred from its call sites; annotate it with its type (spec 13.1)
```

That is with no call site. Add one — `double(i.n)` — and it compiles. The body says `x * 2`, which fixes `x` to `Int` on its own, and the compiler never looks.

The call site is not a fallback, it is the source:

| written | result |
| --- | --- |
| `let double (x) = x * 2`, no call | rejected — "cannot be inferred from its call sites" |
| `let double (x) = x * 2`, one call | compiles |
| `let firstOf (xs) = List.length(xs)`, called with `[ i.n ]` | compiles; the element type comes from the call |
| `let sizeOf (xs) = List.length(xs)`, called at `List<Int>` and `List<String>` | rejected — "used at conflicting types; a helper is monomorphic" |

`HelperTyping.checkHelpers` builds `CallScope`s over the whole module for this (`inferenceScopes`), so a helper's type is a function of every other definition in the module.

## Where it comes from

A helper is not a function, it is a macro. `HelperInliner`'s own description: a helper fn "is not lowered to a class of its own. Instead every call `h(a, b)` is rewritten to `let $k_p1 = a in let $k_p2 = b in <body>`… Because the body is spliced into the caller, the caller's construction-permission check, `requires` inference, and codegen all see the helper's constructions and injected calls directly."

Splicing before typing is what forces the rest:

- A lambda passed to a helper is inlined under a synthetic `$k_p` name, so `HelperInliner` keeps a `lambdaOrigins` side table purely to send diagnostics back to where the lambda was written.
- A recursive helper cannot be spliced, so it needs a separate path — `Compiler.injectRecursivePrelude` emits it as a method instead.
- #166 (`List.filterMap` cannot be given a step) is this: the helper has no signature to write the step's type into, so the lambda is typed with whatever `expected` happens to flow in at the site.
- #159 (a behavior cannot be called from a `let`) is the seam between the two things the language calls a function: a behavior is a compiled unit typed by its declaration, a helper is spliced text typed by its callers.
- A definition cannot be type-checked on its own, so there is nothing to key an incremental compile on (#35).

## What Elm and F# do

Both were measured rather than recalled.

**F# 10.0.302.** Annotations are optional; the type comes from the body, processed in declaration order; call sites are not consulted.

```fsharp
let lengthOf x = x.Length            // error FS0072
type R = { Length: int }
let used = lengthOf { Length = 3 }   // three lines below, and it does not help
```

The message is explicit — "a reference to an object of indeterminate type based on information **prior to this program point**". Where the body does determine the type there is no annotation: `let double x = x * 2` compiles alone, and `let firstOf xs = List.head xs` generalises to `'a list -> 'a` and is used at `int` and `string` in one file.

**Elm 0.19.2.** Annotations are optional; the type comes from the body; it never gets stuck, because it generalises — records included.

```elm
lengthOf x = x.length
-- inferred: { b | length : a } -> a
```

One definition called at `Int` and at `String` in the same module compiles.

So neither reference language reads a definition's type from its callers. Both read it from the body and fill what is left by generalising. Souther has no user generics, so it cannot generalise — and call-site inference is what was put in its place.

## Proposal

Take F#'s rule: infer a helper's parameter types from its body, and require an annotation exactly where the body does not determine them. Without generalisation, the annotation is required in more places than in F# — that is the price of the no-generics decision, and it is already being paid (the "a helper is monomorphic" error above).

It is not only a tightening. `let double (x) = x * 2` with no call site is an error today and would compile.

Spec 13.1 currently sanctions the call-site rule:

> A tuple or record pattern names no type, so a helper whose parameter type cannot be inferred from its call sites still annotates it (`let 二倍 ({ 数量 }: 明細)`).

"cannot be inferred from its call sites" becomes "is not determined by its body".

## Migration cost

Across the bundled prelude and every module in souther-lang/examples there are 240 helper `let`s (a `let` with no behavior of the same name, taking at least one parameter). 235 annotate their parameters. The other 5 name the type with a constructor pattern — `let openAmount (Amount(v))`, `let worstDiscountOf (QuoteLines(lines))`, `let 高額か (予定費用(明細), 基準金額: 金額)`, `let 先方負担を含むか (予定費用(明細))`, `let 立替の合計 (実費用(明細))` — which spec 13.1 already accepts as saying the type.

Nothing in the corpus depends on call-site inference. The change rewrites no existing source.

## Scope

`HelperTyping.checkHelpers` loses `CallScope` / `inferenceScopes`; the parameter type comes from the body or from an annotation, and the "cannot be inferred from its call sites" diagnostic is replaced by one that says what the body left open. Inlining stays, as an expansion of a helper that has already been typed.

This is the first step of #177 — a helper that has a type of its own is what makes a per-definition question possible at all.

kawasima:
Closes #176.

A helper `let` had no type of its own: `HelperTyping` built a call scope for every top-level fn in the module and read an un-annotated parameter from the arguments passed there. `let double (x) = x * 2` was rejected until something called it, the same body meant `Int` or `String` depending on the caller, and no definition could be type-checked alone — which is what #177 step 1 is about.

## The rule

A parameter takes its type from the helper's own body: the first position that names a type, reading the whole body — an operand against a value of known type, an argument of a call whose parameter type is declared, a field of a construction, an `if` condition, the value of a binding (including the binding a helper call inlines to). A parameter the body leaves open is annotated, and so is a function-typed one.

Elm 0.19.2 and F# 10.0.302 were measured rather than recalled. Both take the type from the body and fill what is left by generalising; Souther has no user generics, so call-site inference had stood in for the generalisation. Unlike F#, the whole body is read: a determining position further down the body — or a helper declared further down the file — still settles the parameter, so no annotation is required merely because of where something was written.

`BodyTyping` answers with a type or with nothing and never reports a type error. Every type judgment stays in `Elaborator`, run on the completed environment, so a body that disagrees with the type is reported there, at the position of the disagreement.

## The report

```
-- HELPER DECLARATION --------------------------demo.sou:8:9

8| let total (line) = line.qty * line.price
              ^^^^

8| let total (line) = line.qty * line.price
                      ^^^^
this use does not name a type

Helper `let total` parameter `line` is not determined by its body; annotate it with its type.
```

`check.helper.conflict` is removed: with no call sites read, there is nothing to conflict. `check.helper.fnparam` now fires on the application in the body rather than on what a caller passed.

## Migration cost: 0 lines

The bundled prelude and every module of souther-lang/examples build unchanged (`mvn clean test` there, working tree clean). Of the 240 helper `let`s in that corpus, 235 annotate their parameters and 5 name the type with a constructor pattern.

## Testing

- `souther`: 1189 tests, LSP 81 — green
- `souther-examples`: BUILD SUCCESS, no source changes
- New `CompileHelperBodyTypingTest` (19): one per determining position, the tightening (a call site no longer types a parameter), the open-use report, and the three shadowing/inlining defects found in review — each verified to fail with its fix removed

## Documentation

Spec `[#fn-declaration]` and `[#binding-patterns]`; ADR-0066 supersedes ADR-0050.

## Known gap

What a helper settles about itself does not yet reach its expansion: the binding a call inlines to carries only a *written* parameter type, so a body-determined sum narrows to the argument's case there. Not a regression (call-site inference produced the same narrow type), and it needs a resolved type to be written onto a surface tree — filed as #178.


kawasima:
Good catch — measured and it reproduces:

```
let applyAll (g) = List.map(g, [1, 2])
```

`List.map` applies `g` inside the expansion, not where `g` is written, so scanning the un-expanded body missed it and the report was `check.helper.infer` ("not determined by its body") where `check.helper.fnparam` is what the user needs.

Fixed in abf61dc: the applied-parameter check now reads the expanded body — the same tree the typing rule reads — and the message says the parameter *is used as a function* rather than that it is *applied*, so it fits both shapes (`g(v)` written directly, and `g` handed to a combinator). Covered by a new test asserting the message key for the `List.map` shape, verified to fail before the change. Spec `[#fn-declaration]` and ADR-0066 updated to match.

Issue #35, opened 2026-07-22
Incremental compilation (on demand)


kawasima:
Split out of #31 (future-work prioritization). **Lowest priority / on demand.**

## Goal

Avoid a full recompile of all `.sou` when only some changed, especially when `SoutherProcessor` runs as an annotation processor during a host Maven/Gradle build.

## Current state / readiness

Compilation is fully from-scratch each invocation. `Compiler.compile` / `Compiler.compileModules` (`Compiler.java:44-63`, `:134-202`) parse, derive, desugar, type-check, and generate over **all** sources every call, into fresh maps discarded on return. No content digest, no mtime check, no persisted intermediates, no dirty/stale/invalidate concept. The only "cache"/"memo" (`Backend.java:709`) is intra-call and discarded.

`SoutherProcessor` (`apt/SoutherProcessor.java`) re-reads every `.sou` by filesystem walk (`:62,:85-97`) on a single `done`-guarded round, ignores `RoundEnvironment` for source discovery, and calls `createClassFile` with **no** originating elements (`:71`) — so even Gradle's built-in incremental APT can't skip unchanged rounds.

**Foundation that exists:** a per-module import dependency graph (`imports()` / `byName` / `detectCycles`, `Compiler.java:157-166`) with cycle detection, and naturally per-module output (`Backend.generate` is called once per module and merged, `:182-190`).

**Missing (net-new):** the whole caching/invalidation/persistence layer — content digest or mtime, persisted prior outputs, a reverse-dependency (dependents) map for dirty-set propagation, and Filer originating-element wiring.

## Why deferred

Souther targets small-to-medium domain models and the full build/test already runs in seconds, so the practical payoff is low until a real large-corpus build-time complaint appears.

## Cheap partial win (independent of the caching layer)

Wire Filer originating elements in `SoutherProcessor.createClassFile(...)` (and, where feasible, discover sources via the round environment) so Gradle's incremental APT can skip unchanged rounds. This is small and can land without building any persistence/invalidation layer. Consider doing just this if/when build time becomes a mild annoyance, and reserving the full incremental compiler for a demonstrated need.

## Acceptance

Given an unchanged module set, a rebuild reuses prior outputs and does measurably less work (skipped modules logged); changing one module recompiles it and only its dependents, resolved through the existing import graph.

kawasima:
Done by #182 (develop a4657ec).

[...]

PR #182, opened 2026-07-28
Make the compiler a set of questions it can be asked

kawasima:
Issue #177 step 4, first part. The compiler stops being a pipeline and becomes a set of memoised
questions; the three sequences that used to run over the same stages become three short functions
that differ only in what they decide.

`Compiler.java` goes from 1379 lines to 293 and no longer names a single pass. The hand-rolled
pipelines are gone:

```
Deriver.derive / Lower.run / NewtypeDesugar.rewrite call sites:  7 -> 0
Resolve.module call sites:                                       5 -> 1   (the prelude bootstrap)
```

Measured migration cost: zero lines. souther-lang/examples builds unchanged.

## What is here

**A query engine** (`souther.compiler.query`). A key is one question and the code that answers it.
The store answers it once, keeps the answer, and records what answering it read.

**Nothing throws to report.** An answer is a value or nothing, plus what the attempt found. That is
what let `compiling`, `linking` and `diagnoseModules` collapse: the difference between them was
whether the first error was raised or collected, and that is now the caller's line of code rather
than its own copy of the compile.

**The order the passes ran in is not written anywhere.** Settling a helper's parameter type before
its call is expanded (#178), and inlining a spread's invariant before the importer reads it, were
kept by putting the passes in the right sequence. They are edges now, and reading an answer is what
makes it happen first.

**An edit is absorbed rather than recompiled.** Each answer knows the revision it was made at and
the revision it last came out different at, and an answer equal to the one it replaces leaves
everything that read it alone. Editing any source changes that source's parse, and the layout of the
workspace is built from every parse — but the layout comes out the same, so nothing below it is
looked at. This is #35, and there is no caching layer: the dependencies were already being recorded
to find a question that depends on itself. The language server keeps its compilation between
keystrokes.

## Two divergences found on the way

These are the pattern #177 is about — one question, several implementations — and both were live.

**Cycles were detected twice, by different rules.** The batch walk followed qualified type
references as well as imports; the editor's walk followed only imports. A cycle written without an
import line was invisible in an editor.

**A failing example merged in from an `examples for` file named no source at all**, because the
merge lost which file the row came from. Each row now carries its origin, so the file that has the
row is the file quoted. `MultiFileDiagnosticOriginTest` pinned the old limitation and now pins the
answer.

## Navigation

Go-to-definition, find-references and rename each decided what the cursor was on by matching the
identifier's text. Two modules may declare the same name, and a qualified reference reaches one of
them without an import line, so the text cannot tell them apart:

```souther
module up    exposing ( Amount )    data Amount = Int
module here  exposing ( Amount )    data Amount = String
                                    data Box = { far: up.Amount, near: Amount }
```

Renaming `here`'s `Amount` rewrote the tail of `up.Amount` — silently changing what that reference
means — and go-to-definition on `up.Amount` stopped at `here`'s. Both are covered by
`NavigationResolvesNamesTest`, which fails on develop.

The resolve pass has answered this since #180; it threw the answers away after rewriting the tree.
It keeps them now. Which module and which name is the compiler's answer; where in the file the name
is written stays the syntax tree's, which is what knows about characters.

A name in the value namespace — a behavior, a `let` — is still matched by spelling, because it is
not resolved yet. So is anything in a file that does not compile.

## Not here

Per-definition keys, and error recovery in the type checker. The second is what "the LSP's second
engine goes away" actually needs: without it an editor still cannot ask about half-typed source,
whatever the compiler is made of. Both are separate work.

#35 is finished by this. A develop-targeted PR does not close an issue automatically, so it needs
closing by hand after merge.

## Verification

```
mvn -o clean install -DskipTests && mvn -o test
  souther-syntax    37 tests
  souther-compiler  1222 tests
  souther-lsp       83 tests

mvn -o -f ../souther-examples/pom.xml clean test    BUILD SUCCESS, unchanged sources
```

Every commit compiles on its own.

### DIFF
--- a/souther-compiler/src/main/java/souther/compiler/Compiler.java
+++ b/souther-compiler/src/main/java/souther/compiler/Compiler.java
@@ -272,31 +272,7 @@ public final class Compiler {
     public static Map<String, List<Diagnostic>> diagnoseModules(Map<String, String> sourcesById,
                                                                 Set<String> brokenModuleNames,
                                                                 ModulePath path) {
-        Compilation compilation =
-                Compilation.ofDocuments(sourcesById, brokenModuleNames, path);
-        compilation.answerEverything();
-
-        Map<String, List<Diagnostic>> result = new LinkedHashMap<>();
-        for (String id : sourcesById.keySet()) {
-            result.put(id, new ArrayList<>());
-        }
-        // Every report lands on the source it is about — an error in an imported module on that
-        // module's document, an `examples for` file's failing row on that file — which is what the
-        // reports carry with them. A report about something the caller did not hand over (a module
-        // read off the path) has nowhere to go and is left out.
-        for (Db.Found found : compilation.db().allReports()) {
-            String id = found.sourceId() != null ? found.sourceId()
-                    : compilation.sourceIdOf(found.module());
-            List<Diagnostic> on = id == null ? null : result.get(id);
-            if (on != null) {
-                on.add(found.report().diagnostic());
-            }
-        }
-        Map<String, List<Diagnostic>> published = new LinkedHashMap<>();
-        for (Map.Entry<String, List<Diagnostic>> e : result.entrySet()) {
-            published.put(e.getKey(), List.copyOf(e.getValue()));
-        }
-        return published;
+        return Compilation.ofDocuments(sourcesById, brokenModuleNames, path).diagnostics();
     }
     /** The module name from a source's {@code module <name>} header, for identifying a source that will
      * not parse (so its importers can be skipped, and the LSP can map a broken file to its module).
--- a/souther-compiler/src/main/java/souther/compiler/Compiler.java
+++ b/souther-compiler/src/main/java/souther/compiler/Compiler.java
@@ -1,38 +1,18 @@
 package souther.compiler;
 
-import souther.compiler.types.TypeName;
-import souther.compiler.check.Symbols;
 import souther.compiler.diag.CompileException;
-import souther.compiler.ast.Ast;
 import souther.compiler.diag.Diagnostic;
-import souther.compiler.diag.Region;
-import souther.compiler.diag.SourcePos;
-import souther.compiler.check.Exposing;
-import souther.compiler.check.HelperInliner;
-import souther.compiler.check.Lower;
-import souther.compiler.check.DataChecker;
-import souther.compiler.check.NewtypeDesugar;
-import souther.compiler.check.Resolve;
-import souther.compiler.check.PipelineSigs;
-import souther.compiler.check.Sig;
-import souther.compiler.check.TypeChecker;
-import souther.compiler.codegen.Backend;
-import souther.compiler.meta.ModuleMetadata;
 import souther.compiler.meta.ModulePath;
-import souther.compiler.meta.PublishedModule;
-import souther.compiler.frontend.CstFrontend;
-import souther.compiler.derive.Deriver;
+import souther.compiler.query.Compilation;
+import souther.compiler.query.Db;
+import souther.compiler.query.Report;
+import souther.compiler.query.Output;
 
 import java.io.IOException;
 import java.nio.file.Files;
 import java.nio.file.Path;
-import java.util.ArrayDeque;
 import java.util.ArrayList;
-import java.util.Deque;
-import java.util.HashMap;
-import java.util.HashSet;
 import java.util.LinkedHashMap;
-import java.util.LinkedHashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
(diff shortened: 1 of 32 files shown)


kawasima:
Review findings addressed in 98f4a7a. Every one was real, and each fix has a test that fails
without it. Seven of the nine were regressions against develop; two were in code this PR added.

| Finding | Verified as | Fix |
| --- | --- | --- |
| Behavior-reference cycle crashes the editor | `IllegalStateException` out of `diagnoseModules` | Every module in a cycle is stopped, not just the one the back edge was found at |
| E1501 never reported in an editor | `{a.sou=[], b.sou=[]}` | `Cycles` / `Shadows` are per-module keys and report through the answer |
| Duplicate diagnostics | 3 for one unknown import, 4 for one reserved name | A report belongs to the key that found it; absent answers stop forwarding |
| Parse errors dropped | `{bad.sou=[]}` | `Parsed` names its source, so there is a document to publish on |
| Rename edits the wrong module's `exposing` | reproduced on this PR's own `up`/`here` fixture | Binding sites come from the resolved target, as the references already did |
| Stale reports from unreachable queries | old E1905 on a deleted row | `allReports` reads only what this revision asked |
| `throughCycle` leaks on an exception | key never memoised again (3 computes) | Cleared in the `finally`, with the rest of the state |

Two things worth calling out beyond the list.

**The incremental claim in the description was not true.** Writing the test for it — "changing only a
body does not reach the modules that import it" — made it fail. `Registry.exposedBy` read the whole
module to get its `exposing` line, so adding a behavior to one module re-ran codegen in every module
that imports a type from it. The `exposing` list is its own question now and the test passes. This
was the boundary the whole claim rests on and nothing was holding it down.

**Where value equality carries the cutoff is now written down** (`Db`'s javadoc). Two boundaries
carry it — `Front.Layout`, which stops an edit reaching the whole workspace, and
`Names.Declarations`, which stops it at the module boundary — and both are pinned by
`IncrementalCompilationTest`. It does not hold for a module's classes: `Map<String, byte[]>` compares
arrays by identity, and every module's examples read every module's classes. Comparing the bytes
would not help, since after a real edit they differ; narrowing an example to the classes it reaches
would, and that is per-definition work.

On the cycle-recovery breadth (finding 7): the whole in-flight chain is still discarded, not just the
part inside the cycle, because an answer built from a cyclic one is as entry-dependent as the cyclic
one. The cost the finding describes is gone for a different reason — cycles are now stopped before
any question is asked that would answer itself — and `CyclicModulesTest` pins that `Output.All` stays
kept.

`Front.Layout`'s unused report list and `Output.Classes.stamp`'s unreachable failure path are gone.
A closed document is forgotten rather than held for the session.

Verification: 37 / 1234 / 84, and souther-lang/examples still builds unchanged.

kawasima:
Addressed in 2a248fa. Four of the five held; one does not, and the reason is worth recording.

**1 — `forget` did not drop the text. Correct.** `Front.Text` named no source, so the sweep did not
match it. One nuance: the text string itself was released — `forget` sets the input to nothing first
— so what accumulated was one key with an empty answer per URI ever opened, not the text of every
file. It names its source now, and the test asserts `Front.Text` rather than saying "its text" about
`Front.Parsed`.

**3 — revision is not a round of asking. Correct.** `allReports` reads what was asked since the last
input change. Every caller here asks about the whole compilation, so the two cannot come apart; a
caller that asked only about the file in front of the author would need a generation of its own. The
javadoc says that now instead of leaving it to be found.

**4 — the comment overclaimed. Correct, and the comment was the wrong idea.** A dependency that
answered before the throw keeps its answer, and should: a query is a function of its inputs, so an
answer that was reached is as good as any other. The invariant is the narrower one you wrote — the
failing key leaves no frame, no place in the chain, no cycle mark. Comment replaced.

**5 — two freshness rules. Correct, and `reportsUnder` had no callers.** Deleted rather than
reconciled; `Compilation.reportsUnder` with it.

**2 — the module-wide removal is not too broad.** I tested the case you constructed. Two sources
claiming `shop.cart`, close the one the layout named: the survivor becomes the module and is compiled
from its own source, `shop.cart.Extra` and all (`DuplicateModuleForgetTest`).

The reason is that ownership is not ambiguous in the direction that would matter. Every question about
a module reaches its source through `Front.Layout`, which names one source per module however many
claim it. So while a source was the module, every answer about that module was an answer about that
source — there is no `shop.cart`-keyed memo derived from `b.sou` while `a.sou` is the named one. A
source that claimed a name it did not get is passed a null module name and takes nothing else with it.

That said, the code did not say any of this, which is the part of the finding that stands. It says it
now, and the test holds it down.

37 / 1235 / 84, souther-lang/examples unchanged.

Issue #37, opened 2026-07-22
Type checker: recover past the first semantic error (multi-error diagnostics)


kawasima:
## Context

Follow-up to #29 (cross-module LSP). Mature LSPs report *all* semantic errors in a file, not just the first. Souther's `TypeChecker` aborts on the first error by throwing `CompileException`, so both single-file and workspace diagnostics surface only one semantic error per module at a time.

This is not an LSP concern — it is a redesign of the type checker itself, and it also changes CLI and annotation-processor error output. It is therefore split from #29.

## Proposed change

Introduce an error type (a bottom that unifies with anything, suppressing cascades) and let the checker collect diagnostics and continue inference instead of throwing on the first error.

## Composes cleanly with #29

#29 keeps the LSP diagnostics plumbing list-shaped (`Map<module, List<Diagnostic>>`). When the checker recovers, multiple diagnostics flow through with no LSP-side rework.

kawasima:
Done in #40 (merged to develop, released on main in v0.1.0-rc2). Auto-close didn't fire because the PR merged into develop, not the default branch main.

kawasima:
The other half of this landed in #183 (develop c392506). #40 delivered the collecting; the error type
proposed here did not exist until now, and `TypeChecker.checkCollecting` still carried a comment saying
so.

`Type.Erroneous` is what a name the compiler cannot resolve denotes. It absorbs — every comparison with
it holds, every join with it yields it — so one unknown name is reported where it was written rather
than again at each position the value it produced flowed into. `Type.Nothing` is not this: that is a
real inference result, the element type of `[]`, and must not silence anything.

What it changed beyond multi-error reporting: resolution, imports and declarations stopped ending the
module, so every unknown name in a file is reported at once, an import line that names nothing leaves
the other lines doing their job, and a name declared twice keeps the first. Before this, one mistake
anywhere in a file sent every editor navigation request back to matching the spelling — which silently
does the wrong thing, since renaming a type then rewrote the tail of every qualified reference to
another module's type of the same name.

Two things worth recording for anyone reading this issue later.

Absorbing is only right for a comparison, which can answer "no disagreement". A check that wants a
particular shape — Int or Decimal to add, two lists to append — has no answer to give, and letting one
through puts the error type's own rendering in the message. Those definitions are abandoned instead, at
the recovery boundary #40 built, so every other definition in the module is still checked.

Whether a module may be emitted is a separate question from whether it may be checked, and conflating
[...]

PR #183, opened 2026-07-28
Give a name that denotes nothing a type, so the compiler can carry on

kawasima:
The other half of #37. That issue proposed an error type *and* collecting past the first error; #40
landed the collecting. Without the error type, one unknown name still ended the module — so an editor
could not say what any name in the file meant, which is when an author most wants to ask.

`Type.Erroneous` is what a name the compiler cannot resolve denotes. It absorbs: every comparison
with it holds and every join with it yields it, so the one mistake is reported where it was written
and not again at each position the value it produced flowed into. `Type.Nothing` is not this — that is
a real inference result, the element type of `[]`, and must not silence anything.

## What stops ending the module

| While editing | Before | Now |
| --- | --- | --- |
| one unknown type name | module ends; navigation falls back to matching spelling | the name is reported, the rest resolves |
| two unknown type names | only the first is reported | both |
| one import line that names nothing | the whole scope goes | that line is reported, the others still bring in what they bring in |
| a name declared twice | module ends | the duplicate is reported, the first is still a declaration |

Falling back to spelling is worse than not answering, because it silently does the wrong thing:
renaming a type rewrote the tail of every qualified reference to another module's type of the same
name. `NavigationSurvivesAMistakeTest` and `NavigationSurvivesABadImportTest` fail on develop.

When `import m.nope ( X )` cannot be followed, `X` goes into scope denoting nothing. A use of it takes
the error type and says nothing more, so the author is sent to the import line rather than to each
field that wanted it — one diagnostic for one mistake, the same rule the error type follows a level
down.

## The judgment worth reviewing

Suppressing the consequences of an unknown name check by check does not end: it cannot be constructed,
the behavior does not construct what it declared, the body does not return what it promised. Each is a
condition to add and a place to forget.

So `Unanswerable` says the thing itself — the compiler does not know what this definition means, so
the questions about what it means have no answers. It carries no diagnostic, because the name was
reported where it was written. The unit is abandoned at the recovery boundary the checker already has,
which means **every other definition in the module is still checked**: an author fixing one name is
still told about the rest of the file (`abandoningOneDefinitionStillChecksTheOthers`).

Emitting is gated in two places, at two different moments. Declarations and signatures are checked
against the tree before the module is checked — there is no bytecode for a field whose type nobody
could name. Bodies are caught after, by the abandonment: a unit with no meaning has nothing to emit.
Neither can be forgotten by a pass added later; one puts an error type in the tree and the other
raises `Unanswerable`, and each gate looks for its own.

## Still ends the module

A qualified behavior reference naming no module (`behavior g = nosuch.h >-> f`). Behaviors are the
value namespace, which is not resolved yet — that is the per-definition work left in #177 — so making
this total would produce an "unknown behavior" cascade with nothing to suppress it. Left as it is on
purpose.

## Verification

```
mvn -o clean install -DskipTests && mvn -o test
  souther-syntax    37
  souther-compiler  1240
  souther-lsp       89

mvn -o -f ../souther-examples/pom.xml clean test    BUILD SUCCESS, unchanged sources
```

#37 is already closed; this finishes what it asked for. Worth a note there after merge rather than
reopening.


kawasima:
Both findings held. 876f8bb.

**ChatGPT's P1 — the gate was an early return, and that undoes what the error type is for.** Confirmed
with the source in the finding: `A.value: Nowhere` and an unrelated `g` returning the wrong type gave
one diagnostic, not two. An absorbing type exists so the check can carry on; putting the gate before
the check meant a hole in one declaration silenced every other definition in the file. The structure is
now the one you wrote — the tree is asked after the check, and only about whether the module can be
emitted. `anUnknownTypeInADeclarationStillLetsTheOtherDefinitionsBeChecked` covers it, and your point
about the existing test is right: it put the unknown name in a body, so it never reached the early
return.

Your caveat about absorption holding when an error type flows from a signature into body checking: it
does. `g`'s mismatch is reported and nothing about `A` is, which is what running the check on that tree
produces.

**Copilot's — `Available().value() == null` conflates unknown with unusable.** Confirmed: two sources,
one taking a reserved module name and one importing it, gave the importer
`check.import.unknownmodule` for a module it can see. It asks
`registry.moduleNames().contains(...)` now, so only a module that is genuinely not part of the
compilation is reported as unknown; a module that is here and unusable is reported on its own source
(`ImportOfAnUnusableModuleTest`).

Two things fell out of fixing those, both worth a look.

**A declaration a module may not have is dropped from the tree, not only from the scope.** Keeping the
first of two `data A` left the second in `module.defs()`, so once the module was actually checked the
checker read it and reported `Int` against `String`. `TypeChecker.declared` decides which declarations a
module has and now says it to the tree as well as the scope.

**Whether a module may be emitted was turning into a list of conditions.** It is one named question —
`Names.Sound`, did anything about this module's names come out wrong — plus the tree walk. I tried to
make the first subsume the second and was wrong: an import of a module that is here and unusable leaves
the names it was to bring denoting nothing, and what is wrong was reported on *that* module, so this one
holds a hole having reported nothing. Both are needed and neither implies the other. The reserved-module
test above is what caught that.

37 / 1244 / 89, souther-lang/examples unchanged.

kawasima:
The `/code-review` findings as well, in fcbac49. All seven reproduced; 1 and 2 were already handled by
876f8bb (the review read 19fb15e), 3–7 were not.

**3 — an operator reported the error type as a type.** `a.n + a.n` where `a.n` has no type gave
`operand of arithmetic must be Int or Decimal, got ?`, and `"x" ++ a.n` gave `[String, ?]`. Exactly as
the finding says: `?` is the rendering whose own comment says it should not reach a message.

Absorbing is for a comparison, which can answer "no disagreement". An operator wants a particular
shape and there is no shape to give it, so patching operator by operator would be the forgettable
pattern. The definition is abandoned instead — the same rule a construction of a nameless type already
followed — in one place for every operator.

**4 — a module built against a rejected one was emitted.** Reproduced: `m.dup` reports its duplicate
and emits nothing, while `m.a` emitted `m.a.B`, `m.a.B$Dec`… naming a class nothing would produce.

Soundness is transitive now. And an import that could not be followed *at all* counts the same way,
whether or not anything was reported here — which also closes something a walk of the declarations
would never find: a hole in an annotation on a local, inside a body. Your note that
`holdsAnErroneousType` never visits `Ast.LetIn.declaredType` was right, and this is why that walk is
not where the answer belongs.

**5 — a working import lost to a broken one above it.** Reproduced, including the wrong message: `X`
"conflicts with a local definition" when there is none, because `nameless` never recorded anything in
`from`. A name a failed line only stood in for is not a claim on it, so an import that can do the job
takes it.

**6 — `Resolve.module` had stopped guarding the prelude.** Right, and the comment in `Prelude.load`
still described the guard. It raises again; the recovering form is `resolving`.

**7 — three entry points ignored abandonment.** None of them had a caller: `checkOrThrow`, `check` and
`checkAndElaborate` are deleted rather than corrected, with the `collect` overload that dropped what it
caught. `checkReporting` is the one entry point.

Also useful: your note that findings held identically on `a4657ec` for the `check.codec.nodecoder` case
and `data S = A | Nowhere` saved me chasing two things this PR did not cause.

37 / 1247 / 89, souther-lang/examples unchanged.

PR #186, opened 2026-07-28
A definition is its own question in the body layer

kawasima:
Step 4 of #177, continued: the last of the per-definition keys. Declarations became their own
question in #185; bodies are the other half.

## What it was

Lowering read a module and expanded its bodies as one pass over one tree, and the type check read a
module and answered for all of it. So every body in a file depended on every other one: editing a
behavior's body re-expanded and re-checked all of them, and an editor asking what one definition
means had to have the whole module checked to find out.

## What it is

A body is its own question, at both stages.

- `Bodies.LoweredBody(module, fn)` — one body expanded and desugared. What it reads is the fn it is
  and the helpers of the module.
- `Bodies.CheckedBehavior(module, behavior)` — one body checked against the behavior it implements,
  as the Core the backend emits.

The shared context each of them reads is a value, not an object: `Helpers`, `RecursiveHelpers`,
`ReqSigs`, `RecursiveHelperSigs`, `RecursiveHelperConstructs`, `BehaviorNames`. That is what makes
them shareable — a `HelperInliner` has no equality, so a key answering with one would make every
reader run again whatever changed. `HelperInliner.forHelpers` builds one from the helpers alone.

What is left at the module is its own check — declarations, helpers, `exposing`, compositions —
and the decision it and the bodies come to together: whether there is a module to emit.

`IncrementalCompilationTest` pins both: editing one behavior's body leaves the expansion and the
check of the one beside it as the same answer.

## What changed for an author

- The module's own check is asked before any body, so a module told about its own mistakes is told
  about them first. Two reports move: a `let` that implements a composition, and a composition's
  stages and declared output, now come before a body error rather than after.
- A module whose own check stopped is not asked about its bodies. A structural check builds what a
  later phase reads, so a body checked without it reports being unable to see what has already been
  reported missing.
- One problem is one diagnostic, whichever questions found it. A helper is checked on its own and
  again in each body it is expanded into, and both look at the same line; the check used to drop the
  second because both were inside one answer. Now they are two answers, so the reading of them is
  what recognises them as one.
- A module with two helpers to complain about complains about the earlier one first: the helpers are
  read in the order they are written rather than in hash order.

## Left

Error recovery in the checker was done in #183; per-definition keys are this. Step 4 has nothing
further outstanding.

### DIFF
--- a/souther-compiler/src/main/java/souther/compiler/check/HelperInliner.java
+++ b/souther-compiler/src/main/java/souther/compiler/check/HelperInliner.java
@@ -10,6 +10,7 @@ import souther.compiler.diag.SourcePos;
 import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.HashSet;
+import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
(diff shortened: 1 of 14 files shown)


kawasima:
All four are addressed in c3579cb.

**Problem identity (and the NUL).** The concatenated key is gone. `Diagnostic.identity()` is now the
whole diagnostic — code, message key, arguments, literal message, severity, region, secondary labels,
notes, diff, suggestion — with the arrays it carries compared by what is in them, which is the one
thing a record cannot do for an array component. `Report.problem()` returns that, and `Db.allReports`
keys on a record, so there is no delimiter to collide with. `TypeChecker.deduped` asks the same
question rather than keeping a second, weaker one of its own.

The delimiter was worse than ambiguous: it was two literal NUL bytes in the source. Confirmed and
gone; the branch has none.

**The one-sided test.** Both directions are pinned now, for the expansion and for the check: the
edited body is worked out again, the ones beside it are not. Worth reporting that this found a
defect in the test I had written while addressing it — the fixture's edit matched nothing, so the
text did not change at all, and the one-sided assertion passed anyway.

The shared-helper row of your table is now a test too, but only the half that holds: editing
`doubled` reaches `twice`, which calls it. It reaches `thrice`, which does not, as well — a body
reads the module's helpers, not the helpers it calls. That is real remaining coarseness and the test
says so rather than leaving it to be discovered. Narrowing it is per-helper work (a key per helper,
plus the call closure of a body) and belongs in its own change, not this one.

**`SettledFn`.** Kept as a projection, with the distinction written down: the module is settled
together — one helper's settled type can settle the next one's — and this reads one fn out of the
result. What a reader depends on is what decides how far an edit travels, not what the work was, so
settling per definition can go behind this later without any reader noticing.

**`ReqSigs` answering `Map.of()`.** Pinned as a contract, in `ASignatureThatDoesNotBuildTest`: a
signature that does not build is reported once, no body reports being unable to see what is already
known missing, and nothing is emitted. What holds the second one is the fail-fast interlock — a
`requires` naming something that is not an injection target stops the module check, and a module
whose own check stopped is not asked about its bodies. Checked against develop as well: same single
diagnostic there, so the recovery did not widen what is reported.

Reactor green throughout: compiler 1277, lsp 93, syntax 37.

Issue #187, opened 2026-07-28
A name in `exposing` may rest on one that is not, and a field read fails with IllegalAccessError instead of a diagnostic


kawasima:
A name in `exposing` may rest on one that is not, and the reader gets bad bytecode rather than a diagnostic.

`exposing` gates naming. It does not gate what an exposed name's *signature* rests on, so a module can
publish a behavior whose input and output types it keeps to itself, or a data whose field types it
keeps to itself. The generated class is public and names them anyway.

## Repro: a behavior

```
module up exposing ( x )

data A = { n: Int }
data B = { m: Int }

behavior x : (a: A) -> B
    constructs B
let x (a) = B { m = a.n * 2 }
```

This compiles. What comes out is a public interface over two package-private classes:

```
public interface up.X extends souther.runtime.Behavior<up.A, up.B>
final class up.A extends java.lang.Record        // package-private
final class up.B extends java.lang.Record        // package-private
```

Nobody outside `up` can call it. From Souther, naming the type is where it is reported:

```
-- MODULE ERROR -------------------------------down.sou:7:22
7| behavior toOut : (b: up.B) -> Out
                        ^^^^
`B` is declared in `up` but not exposed by it.
```

From Java, the same, at the same place:

```
error: up の B は public ではありません。パッケージ外からはアクセスできません
```

Both are reported against the reader, for something the writer decided.

## Repro: a data field, which is not reported at all

```
module up exposing ( Order )

data UserId = String
data Order = { by: UserId }
```

```
module down exposing ( Out, read )

import up ( Order )

data Out = { s: String }

behavior read : (o: Order) -> Out
    constructs Out
let read (o) = Out { s = o.by.value }
```

Reading the field needs no name for its type, so nothing is reported. Both modules compile, and
`down.Read$Impl` is emitted calling a package-private class in another package:

```
10: invokevirtual #18    // Method up/Order.by:()Lup/UserId;
13: invokevirtual #24    // Method up/UserId.value:()Ljava/lang/String;
```

```
java.lang.IllegalAccessError: failed to access class up.UserId from class down.Read$Impl
	at down.Read$Impl.apply(down.sou:9)
```

This is #124's family — compiles, then fails at run time on a class the reader was never told it
could not touch.

## Prior art

F# reports it where it is declared, not where it is read (`dotnet 10.0.302`):

```fsharp
module Up

type private A = { n: int }
type private B = { m: int }

let x (a: A) : B = { m = a.n * 2 }
```

```
[...]

PR #189, opened 2026-07-28
Call a behavior that requires nothing, wherever it was declared

### LINKED ISSUE #159: A behavior cannot be called from a `let`, in its own module or an imported one
A `let` may call only the behaviors its behavior `requires`. Two consequences show up together in a model of any size, and they are the same question.

## In its own module

```
behavior a : (i: In) -> Out constructs Out
let a (i) = Out { m = i.n }
behavior b : (i: In) -> Out constructs Out
let b (i) = a(i)
```

```
-- IMPLEMENTATION MISMATCH ---------------------
7| let b (i) = a(i)
               ^
Only a required behavior can be called from a body; compose others with `>->`.
```

So anything two behaviors share becomes a helper, and a behavior that is *also* a building block needs a helper twin. The `pipeline` module carries three: `probabilityFor`/`probabilityOf`, `categoryFor`/`categoryOf`, `roleFor`/`approvalNeeded`.

## From another module

```
module down3
import up ( rateOf )
let use (i) = {
    let got = rateOf(In { n = i.n })
    ...
}
```

```
-- ARBITRARY JAVA CALL  E1401 ---------------
7|     let got = rateOf(In { n = i.n })
                 ^^^^^^
`rateOf` is not a behavior or a builtin.
Hint: Declare a behavior without a `let` and implement it from Java; arbitrary JVM calls are not allowed.
```

The hint proposes re-declaring it as injected and binding the upstream implementation at the boundary, which trades the compile-time link for the single source. Neither route keeps both: `forecasting` copied `pipeline`'s ten-arm stage-probability mapping, and nothing but a comment keeps the two in step — a stage added to `pipeline` breaks three matches there and silently leaves `forecasting` weighing it at whatever its own copy says.

## Proposal to discuss

Let a `let` call a behavior that **has a `let` body and no `requires`** — a pure behavior, which is a named function with an example attached. The compiler already computes exactly that predicate: it is the condition `ExampleVerifier` uses to decide whether a behavior can be run as an example (E1902). Extending it across a module boundary would cover the `forecasting` case, since the callee has an implementation the importer can see.

What would need deciding: whether a `requires`-carrying behavior stays uncallable. I think yes — that is what `>->` is for, and calling one from a body would hide a dependency the signature is supposed to state (ADR-0016, ADR-0017).

Found by the `crm` example in souther-lang/examples (finding F17 and F23 in its README), and verified against `develop` with the repro above.

kawasima:
Closes #159.

`exposing` decided what another module could see, and a second set of rules decided what a visible
name could be used for — rules that were not the same on both sides of a module boundary. Elm, F# and
Haskell differ in nearly every surface detail of their module systems and agree that an export list
controls visibility, and that what a name may be used for follows from what the name is, identically
inside and outside the declaring module. None of the three can express "exported but not applicable".

**What a behavior may be called for is decided by its requirement set**, not by whether an
implementation is written, and the answer does not depend on which module is asking.

| requirement set | how it is reached |
| --- | --- |
| empty | called by name, here or imported. Nothing injected, nothing in `requires` |
| not empty | named in `requires` and called through it, whether or not it has a `let` |

## What was in the way

Nothing decided it. The body-call rule entered as one of four bullets in the commit that made
*required* behaviors callable (`f0fd3ff`, "guards: only required behaviors may be called from a
body"), and the cross-module half was never a rule at all — the name was simply absent from what a
body could resolve, so it was reported as an arbitrary JVM call (E1401).

The specification said the opposite in two places and contradicted itself in a third.

| | |
| --- | --- |
| `[#fn-rules]` | "A behavior of another module may be imported and called" |
| `[#sequential-composition]` | "call it inline in the implementation" when a multi-argument behavior cannot be a stage |
| `[#published-modules]` | "a `let` may not call an imported behavior" |

The first two describe what this PR does. The third is removed.

## What it costs to write

A behavior stopped being shareable the moment any logic was written in Souther for it: `requires`
refused an implemented name (E1607), and a body could not call one. So a rule the specification names
had to exist twice — once as the behavior, once as a helper twin that could actually be called — and
the twin could not cross a module boundary either. A rule with `requires` of its own is now depended
on the same way a Java implementation is, and the boundary composes:

```java
Place.bind(FetchUser.bind(new JdbcLoadRow()))
```

E1607 separates its three cases, because the fix differs: a behavior that requires nothing is called
instead, a `>->` composition cannot be rested on because its requirements are not written, and a name
that is no behavior has to be declared or imported.

## Two things it found

**A `requires` cycle compiled.** `p requires q`, `q requires p`, both implemented — two classes
emitted, neither of which can be built, and nothing said so. `Backend.resolveDeps` only ever recursed
through pipeline stages, so it never saw one. A call cycle would not terminate. Both are E1608 now,
walked as one graph over calls, `requires` and stages.

**`exposing` naming a helper said it was "not a data or behavior of this module"** — of a name defined
a few lines below. It is the first thing an Elm habit writes, and the message sent the author looking
for a typo they had not made.

## Not here

An exposed name may still rest on an unexposed one. Implemented and measured: **12 fixtures in this
repository fail**, and it runs into E1305, which already lets an unexposed *unit* case be constructed.
The rule is right; it carries a migration this change does not, and which positions it reaches has to
be decided first. Taken back out and recorded on #187.

## Verification

`Bodies.CalleeSigs` reads `Shapes.Prepared` and `Imported` — the callee's declaration, never its body:

```
editingACalledBehaviorsBodyDoesNotRecheckTheCaller
  assertNotSame  twice     the edit is `twice`'s, so it is checked again
  assertSame     viaTwice  checked against `behavior twice`, which says what it said
```

Both sides, because one `assertSame` alone passes on an edit that changed nothing.

Nothing is added to inference: a behavior's signature is written (ADR-0017), so both sides of a call
are declared, and typing one is strictly easier than typing a call to a helper (ADR-0066).

A cross-module call that compiles is not evidence that it links, so it was run:

```
down.use(A{n=21}) = Out[o=42]
order.place       = Placed[by=UserId[value=u-1], name=Ada]
```

```
mvn -o clean install -DskipTests && mvn -o test    BUILD SUCCESS
mvn -o -f ../souther-examples/pom.xml clean test   BUILD SUCCESS, unchanged sources
```

Measured migration cost: zero lines, as with every other step of #177.

ADR-0068 records the decision, with a table of where Souther sits against the three and which of its
differences have a reason of their own.


kawasima:
Pushed 2c23b41.

## P1 does not reproduce

The repro in the review does not reach E1608, because it does not compile:

```
-- ARBITRARY JAVA CALL  E1401 -----------------hl.sou:9:24
9| let helper (a: A): B = loop(a)
                          ^^^^
```

A helper `let` cannot call a behavior at all, cyclic or not — `HelperTyping` builds its
`CheckContext` without the callee map. So the edge the graph is missing cannot be written, and the
`$Impl`-per-call recursion the review describes is unreachable.

The path that *does* exist is a helper calling a behavior with a requirement of its own. There E1602
forces the call into the caller's `requires`, so it appears as a `requires` edge. Both cases the
review asked for are caught today, and are tests now:

```
behavior -> helper -> same behavior      `behavior loop` reaches itself: loop -> loop.
behavior A -> helper -> B -> A           `behavior one` reaches itself: one -> two -> one.
```

## What it did find

The reason it does not reproduce is itself a defect: **the specification said the opposite.** "A
`let` body may call a behavior" reads on a helper, which is a `let`. Narrowed to a behavior's `let`,
with the reason — a helper is expanded into what calls it and travels with an invariant that names it
(`invariantHelpers`), so it names only what travels with it. That is what keeps the invariant rule a
single question about the requirement set; allowing it would add a second question about where a name
came from, which is the shape #183 took out.

And the diagnostic was the same lie the `exposing` message told about a helper — `twice` "is not a
behavior or a builtin", of a behavior declared a few lines above:

```
`twice` is a behavior, and it cannot be called from here.
Hint: A helper `let` does not reach a behavior, and a `>->` composition is composed with
rather than called. Call `twice` from a behavior's own `let` and pass the result in.
```

It could not say that at first. `Resolve` counted a behavior's `let` among the module's helpers while
`HelperInliner.helpersOf` did not, so a name reaching an implemented behavior denoted a helper in one
place and a behavior in another. Both ask `helpersOf` now.

## Copilot

Both correct, both fixed.

**E1607 and an imported composition** — confirmed: `requires chain` on an imported `>->` said "not a
behavior in scope", the one thing it was not. It reads the resolved name now rather than scanning
this module's behaviors.

```
`behavior use` declares `requires chain`, but `chain` is a `>->` composition.
Hint: Remove `chain` from `requires`. A composition's requirements are decided by its stages…
```

**ADR index order** — back to ascending.

## Verification

```
mvn -o clean install -DskipTests && mvn -o test    BUILD SUCCESS
mvn -o -f ../souther-examples/pom.xml clean test   BUILD SUCCESS, unchanged sources
```

Three tests added for the rule: a helper cannot call a behavior, a behavior cannot reach itself
through a helper, a composition cannot be called from a body.

Issue #197, opened 2026-07-29
A module cannot publish a helper


kawasima:
ADR-0074 published a module's *values*. A helper — a `let` with parameters — stays module-local, and the ADR records that publishing one is a separate question and not decided there. This is that question.

## What it would give

A rule written once and shared. Today a module that wants another's pure calculation either restates it or declares it as a behavior, and a behavior is a business operation — ADR-0005 keeps the behavior list one-to-one with the spec DSL's, so turning every shared calculation into one puts things in that list which the specification does not name.

F# and Elm publish functions without ceremony. Souther's own direction, decided while ADR-0072 was written, is that the implementation side follows them where nothing else decides otherwise.

## What it brings with it

**Construction authority.** A non-recursive helper is expanded at the call site, so no method survives (`Backend`, "the only non-behavior fns left are recursive helpers"). Only a recursive helper becomes one, on the module's `$Fns` class, which is `ACC_FINAL | ACC_SUPER` — package-private, and commented "not exposed". Making it public lets Java call a data-constructing helper directly. The value still runs its invariant (ADR-0002), so "no unvalidated value can be built" holds; what is lost is ADR-0002's other half, that each origination is stated in a signature. ADR-0059's `__construct` is not the same shape — that opens the construction of one type, a helper opens a calculation.

Recommended treatment: **re-emit an imported recursive helper into the importing module's own `$Fns`**. Its body is already in the jar slices, and a mutually-recursive group is copied as a group. Then `$Fns` stays package-private everywhere, no Java-reachable constructing entry appears, cross-module calls are intra-package `invokestatic`, and the existing check applies unchanged — the copy lives in the importer's package, so the importer's behavior writes `constructs up.T`, which ADR-0059 allows. The cost is duplication per importer, which non-recursive helpers already have through expansion.

Rejected alternatives: making `$Fns` public (opens the Java entry); nestmates (a nest host must be shared, so a different-package importer cannot join); publishing only helpers that construct nothing (decides publishability from the body, which a signature does not show); a Java-boundary wrapper (one more path).

**The body becomes a contract.** Expanding an imported non-recursive helper makes the jar's source slice the linkage target. Souther already does this for a data's invariant, but an executable body is a heavier promise than a declaration: a compiler version difference could change what a library means during the *consumer's* compile. Version agreement has to be decided.

[...]

Issue #163, opened 2026-07-27
A fixture cannot be named, so every example row restates its whole input


kawasima:
`fixture` is not a top-level form, and an example row cannot refer to a value by name:

```
fixture base = In { n = 1 }
```

```
-- SYNTAX ERROR ---------------------------------
6| fixture base = In { n = 1 }
   ^^^^^^^
A top-level definition starts with `data`, `behavior`, or `let`.
```

The compiler's own wording elsewhere is "`quoteAt10` is not a value a fixture can name".

## Cost in the model

Four rows checking a four-band approval matrix are four twenty-line quotes differing in one number. `crm.examples.sou` exists as a separate file only because a `QualifiedLead` fixture is twelve fields. This is the single biggest length reduction available to the CRM example.

## Proposal to discuss

A named fixture plus a spread over one:

```
fixture acme = QualifiedLead { … }
example convertLead
    | "an unconfirmed budget blocks" : (QualifiedLead { ...acme, budgetConfirmed = false }) -> …
```

## A second, related problem

Writing an example row means importing every type that appears anywhere inside the input value, even when the behavior reads none of them. `nextStepFor` takes an `OpenOpportunity` and reads only its stage, but its rows need the ten newtypes that appear inside a `NegotiationReview`; ten of `activity.sou`'s imports exist for fixtures alone, and the import list stops describing what the module does. A named fixture reduces the repetition but not the imports, so whatever shape fixtures take should be decided with this in view.

Found by the `crm` example in souther-lang/examples (finding F20 and F21 in its README), and verified against `develop` with the repro above.

kawasima:
Closed by PR #194 (`guard` / `depends on` renaming is unrelated — ignore), #195 and #196 on `develop`.

**A `let` with no parameter list defines a value** (ADR-0072). An `example` row names it, and spreads it to vary a field; what a row may name is a property of the value graph — a literal, a construction, a spread, a `fromList` over one, or a name of another such value.

**A module publishes its values** (ADR-0074), so a fixture lives with the type it is an instance of and a reader names one thing instead of everything inside it.

Measured on `crm`, moving the `NegotiationReview` fixture to `example.pipeline`:

| | imported names | `activity.sou` |
|---|---|---|
| before | 24 (crm 6 + pipeline 18) | 587 lines |
| after | 15 (crm 4 + pipeline 11) | 543 lines |

[...]

PR #199, opened 2026-07-29
Publish a helper, so a rule is written once

kawasima:
Issue #197. ADR-0075.

`exposing` may name a helper. A module that wanted another's pure calculation restated it or declared it a behavior; declaring it a behavior puts something in the behavior list the specification does not name (ADR-0005), and restating it is the duplication #163 removed for values and left in place for rules.

## What crosses

A published helper crosses **closed**, as a published value does, and is expanded at the reader's call sites. The `>->` restriction stays: a helper does not appear in the spec DSL.

Closing cannot remove a **recursive** helper — it is a method, so the call stays a call. It travels with the body that calls it, under the name of the module that declares it, and so does everything it reaches in turn. A mutually-recursive group arrives whole; one the reader never imported arrives because the body it did import calls it. The importer emits them on its own `$Fns`, the path a recursive prelude helper already takes, so that class stays package-private and no Java-reachable constructing entry appears.

The declaring module is the helper's **identity**, not where its method goes. `ValueName.Helper` already carried the pair, but an imported value resolved to the *reading* module's name, so the declaring one was lost; it is kept now, and `Prepared` writes the spelling to match once — after which everything downstream reads the identity by reading the name. Verified in the emitted bytecode: `top.$Fns` carries `low$summed`, and a reader taking two same-named recursive helpers carries `up$a$step` and `up$b$step`.

## Construction authority

What a published body builds is not the reader's to declare: publishing states that origination, and the reader may have no name for the type. A construction records the module that carried it, so this is **preservation rather than a grant** — a type of some *third* module the body happened to build stays the reader's to declare, which it can, because the module declaring it exposed it. Declaring a carried construction anyway is not called building nothing (E1006 is satisfied by it), so no existing published-value user has to remove a `constructs` entry.

This also closes a hole the published-value feature already had: a value building a hidden type on the way compiled where it was published and was unusable everywhere, because the reader was asked for a `constructs` it could not write.

## Exposed surface

A published helper's argument and return types join it — a reader has to write the arguments. ADR-0074's refusal of a published value that reaches a recursive helper is **withdrawn**: it refused what this carries, and it was not a corner (a value computed with a fold could not be published). The surface rule keeps its reach through such a helper by reading the return type it is required to declare.

## Version

`BOUNDARY_VERSION` 2 → 3. A carried body is source the importing compiler reads back, so what the front end makes of it is part of what a jar promises. `PublishedModule` already refuses a jar that disagrees; the scope it is refusing under is now stated, and tested on a module carrying a published helper.

## Verification

Full reactor `BUILD SUCCESS` — souther-compiler 1404 tests, souther-lsp 95, rest green. `souther-lang/examples` builds clean against this compiler.

Out of scope, noted while testing: an `example` row and a `with` value reject a helper call (`E1903 a fixture cannot call it`), for a local helper as much as an imported one. Pre-existing, and worth its own issue now that a rule can be shared.


kawasima:
Pushed 3439fb8. Fact-checked each point; two held, one did not reproduce as written but pointed at a real defect from another angle.

## 1. carried construction lost through a recursive helper — confirmed, fixed

Reproduced exactly as described:

```
E1006: Behavior `bill` declares `constructs Amount` but never builds Amount
```

for a reader declaring a construction carried by a published recursive helper. `recursiveHelperConstructs` now carries `DataChecker.Constructs` rather than a bare map, and the transitive fixpoint merges the two kinds separately (`Constructs.absorb`). Test: `declaringAConstructionCarriedByARecursiveHelperIsNotOverDeclaration`.

## 2. surface checked constructions, not the type — the point holds, the examples do not

Both examples given fail for other reasons: `let published (n: Int) = usesHidden` does not compile (`unknown identifier usesHidden` — a bare helper name is not a value), and the `Box`/`peek` shape is refused earlier by the field rule (`Box` rests on `Hidden`).

The underlying claim — return type ≠ constructed type — is right, and it is reachable:

```souther
module pricing exposing ( published )
data Hidden = Int
let published (n: Int) : List<Hidden> = []      -- compiled
```

A hidden **unit data** went out the same way (`let published (n: Int) = Marker`), and so did the value form (`let published : List<Hidden> = []`), which is the pre-existing shape this inherited.

The surface check now asks the **type** for values and helpers alike, settled by the standalone helper check and recorded as `Elaborated.definitionTypes`. This also removed the special case that read a recursive helper's `declaredReturn` and walked `calledNames` — the type is the type however the body was written, so `constructedTypes`/`calledNames` are gone from `SpecChecker`. Three tests added, one per shape.

I did not split the check into two systems by definition kind: with the type as the question, values and helpers ask the same thing, and `params().isEmpty()` still decides only what each *becomes* in the reader.

## Copilot's two comments — confirmed, fixed

Both are about unit data, and both held. A unit-data construction is an `Ast.Var`, so `publishedBy` on `Ast.NewData` did not reach it, and the branch re-resolved by spelling. `ValueName.OfType` carries the mark now (2 construction sites), and the branch reads `v.denotes()`. `BodyGen` had the same spelling dependency and would have loaded the reader's own class for a same-spelled unit — it reads the resolved type now. Three tests: carried, same-spelled-in-reader, and the reader's own unit still being its own to declare.

## Verification

Full reactor `BUILD SUCCESS` — souther-compiler 1411, souther-lsp 95. `souther-lang/examples` clean.

Issue #198, opened 2026-07-29
A function value flows only where the syntax puts it


kawasima:
ADR-0025 decided that functions are first-class values. The position that takes a function is still a *spelling* slot rather than a value slot, so the decision does not reach as far as it says.

```
List.all(positive, xs)             -- passes
List.all(List.get(0, fs), xs)      -- "pass a named function or a lambda"
let fs = [positive, positive]      -- "I cannot find a value named `positive` in scope"
let p: (Int) -> Bool = positive    -- syntax error: expected `=`, found `->`
```

The root is `[#delimiters]`: a function type `(A) -> B` may be written only in a `let` argument position. A name is a value where its type can be written, and this is the only place it can. Two of the four lines above are refused for a reason the language states; the other two are refused because the type has nowhere to go.

## What is decided already, and is not in question

A function may not be a `data` field or a behavior's input or output. That is ADR-0004: a function has no external representation, so it cannot cross a decoder or encoder. This issue does not touch it.

The `.field` getter is not a concern here: it desugars to `(x) -> x.field`, a closed lambda that captures no receiver, so opening the type positions does not change how it is compiled.

## The measurement to make first

Partial application was refused on measured grounds — 215 lambdas in the corpus, one place where it would have helped, and the argument to fix was the second one so left-to-right currying could not reach it. The same bar applies here, and has not been met. Before designing:

- how many places in the real corpus become shorter, and what they look like;
- generated classes added, classfile size, run-time allocation;
- the size of the compiler change;
- the size of excluding function types from the boundary *recursively* (below).

## The design point, for when the measurement supports it

The boundary check cannot ask "is a function type written here". It has to be a type predicate — *boundary-representable* — because a function can hide inside a type that is representable at the surface:

```
List<(Int) -> Bool>
Map<String, (Int) -> Bool>
```

and inside anything that stands for one. That is the general form of ADR-0004, and it is what the ADR should say once this is settled.

Follows ADR-0025, ADR-0004, ADR-0072.

kawasima:
Landed on `develop` in #201 (ADR-0076).

A function type is an ordinary type term, writable wherever a type is, and what refuses one is what the position requires of a type — an external representation at a boundary, equality under `==` and for a `Set`'s element or a `Map`'s key, an ordering for `sort` and a `sortBy` key. The three are separate predicates sharing one walk today.

[...]

PR #201, opened 2026-07-29
A function type is a type, and a position asks what it requires

kawasima:
Issue #198. ADR-0025 decided that functions are first-class values; the grammar admitted a function
type in one position — a helper `let`'s parameter — and a name is a value where its type can be
written, so that one position decided where a function could go. A function value could be bound and
applied but not handed to `List.map`, a helper's name could be called but not named, and a binding
could not say what function it held.

A function type is now an ordinary type term, writable wherever a type is. What refuses one is what
the position requires of a type, asked of the type rather than of the syntax that spelled it.

## The measurement first, because it does not support the change

Issue #198 asked for it, and it comes out against the feature on convenience grounds. In
souther-examples — 24 modules, 7007 lines, 289 helpers, 101 lambdas, 76 higher-order calls — no
author writes a function-typed parameter at all; every one of the 23 in the tree is in the standard
library's own `.sou`. No corpus line gets shorter. Partial application was declined on a stronger
showing (one place in 215 lambdas).

The reason to do it is not convenience. ADR-0025 is decided, and the implementation accepted or
refused the same type by the syntax that produced it. ADR-0076 records the measurement alongside the
decision so nobody later reads a corpus need into it.

## What is here

- **The grammar.** `type ::= fnType | unionType`, with the result a whole type — so `->` is
  right-associative, a function may take a function and return one, and `(A) -> B | C` still reads as
  `(A) -> (B | C)`. `(A, B) -> C` takes two parameters and `((A, B)) -> C` one tuple, mirroring
  ADR-0036 for a lambda's parameter list. A parenthesised single type is grouping.
- **`Ast.TypeTerm`**, a sealed sum of a reference and a function type. Chosen over another field on
  `TypeRef` deliberately: that would have left 44 readers compiling unchanged and silently passing a
  function type through. Sealed put all 44 in front of javac.
- **Three capability predicates.** `isBoundaryRepresentable` (a data field, a newtype base, a
  behavior's input and output), `supportsEquality` (`==` / `/=`, a `Set`'s element, a `Map`'s key),
  `supportsOrdering` (`sort`, a `sortBy` key's result). Written as three questions sharing one walk,
  not as one predicate — that a function is the only type failing all three is a fact about the types
  there are now.
- **A name in a value position is the function it names**, written out as the lambda that applies it.
  A recursive helper expands the same way.
- **A binding may state its type**, and that is what settles the parameters of the lambda it binds.

## Two things the design got wrong, found by measuring

**The bug was not where the design said.** The design blamed `HelperInliner`. It is
`Elaborator.collectApplications`, which types every application's arguments in the enclosing scope —
so an application inside a lambda, which is where a combinator's own expansion puts it, throws on a
name bound in that lambda. That is why passing a function to `List.map` reported an internal
`$b2_x`, while `twice(f, v)` and `List.find(f, xs)` were fine. Such an application is now neither a
constraint nor an error, and the fourth line of the issue turned out to be the answer to its second
and third: written down, the type needs no reading off.

**Three holes only measurement found.**

- A behavior's input and output had no check at all. A data field and a newtype base were already
  refused, at any depth, by the codec they derive; the behavior's own signature was not asked.
- The `Map` key restriction is a *boundary* rule, not a global one ("Inside a behavior body a Map may
  be keyed by any value"). A map keyed by a function inside a body was writable.
- A `Set` the checker infers escapes a check on written types. `List.distinct` grows one to remember
  what it has seen, and after expansion it stands in the caller's body with its element type decided.
  The question is now asked of what elaboration decided.

## Elm and F#

Nothing here reaches past either. Both write function types in every type position, a record's field
included. Elm refuses them at a port, by a recursive predicate over the type — the boundary rule
above. Souther's port line falls at the data field instead because a `data` derives a codec by being
declared and an Elm record does not. On equality this follows F#, which refuses the comparison, over
Elm, which compiles it and fails at run time. Currying and user generics stay out, so this is
narrower than either.

## Testing

1426 + 37 + 95 green. souther-examples builds clean against this compiler. The escaping cases are run
rather than only compiled — a function chosen at runtime and mapped over a list, and a function taken
out of a list and applied, both assert on values. The formatter is held to a nested function type,
since a type it cannot see is a type it drops silently.

## What is not done

- **A `$Fn` class per value reference.** Three references to one helper in a list literal generate
  three classes at 639 bytes each. A non-capturing eta-expansion of the same helper could share one.
  Measured, not optimised; it was outside the review's acceptance conditions.
- **The evaluate-once invariant is unverified.** `List.map(selectFunction(flag), xs)` must evaluate
  its function expression once, and the implementation binds it to a local so that it does — but a
  pure language offers nothing to observe the difference with. Asserting it means reading bytecode.
- **`BOUNDARY_VERSION` 3 → 4 asks for a rebuild**, not only of the artifacts using the new form. The
  check is an exact match and this does not change that.
- **The issue's own second line does not typecheck as written.** `List.all(List.get(0, fs), xs)` — 
  `List.get` returns an `Option`, so the example needs a `match`. The issue body should be corrected;
  it has not been touched here.

Follows ADR-0025, ADR-0004, ADR-0072. Issue #198 stays open until this merges (a PR to `develop`
does not close it).


kawasima:
Both fixed in 83ea622. Fact-checking them first turned up one correction and one narrowing.

## P1 — real, but the example given does not reproduce

`f(unknownName)` still reports `unknownName` correctly: an unknown value name is reported by the
resolve pass, which runs before elaboration, so it never reaches the catch. The masking is real for
anything only elaboration finds. With a control:

| written | before |
| --- | --- |
| `i.noSuchField` (no function value) | `I cannot read a field noSuchField on this value.` |
| `f(i.noSuchField)` | `Cannot infer the type of the function f…` |
| `f(List.length(v))`, `v: Int` | `Cannot infer the type of the function f…` |

The criticism behind it stands as written: the code branched on an exception having happened rather
than on which one, so any semantic error added later would have gone the same way.

Fixed by the third option. What puts an application out of reach is known without trying — a binder
below the inference point. Lambda parameters, a `let`, a `match` arm's binding are tracked while
walking, and an application reaching one of them is skipped; every other application is typed, so
what is wrong inside one is reported as itself. Both rows above now report the real mistake, and an
application inside a lambda that reads only outer names is still a constraint — that case has a test
that runs the values, since skipping it would have been a quiet regression.

## P2 — real, and reachable in a narrower range than described

| written | before |
| --- | --- |
| `let p: Option<(Int) -> Bool> = None` | compiles |
| `let p: ((Int) -> Bool)? = None` | `parse.optional.function` |
| `data R = { f: ((Int) -> Bool)? }` | the boundary refuses it (the check never fired here) |
| `let p: Int? = None` in a user module | `?` is for a data field or the core |

So in a user module the annotation position is closed to `?` for every type, and the check was only
putting a wrong reason in front of the right one. The genuine asymmetry — one type accepted or
refused by which of its two spellings was written — is reachable inside the shipped core. The
conclusion holds either way, and the refusal is gone: `optional` takes a term, `T?` is `Option<T>`
for whatever T is, and where `?` may be written is a rule of its own and is what reports.

## Unrelated, found while checking P2

`let p: Option<Int> = None` compiles in a user module, while `let p: Int? = None` there is refused
because "a user model never names an optional elsewhere" (ADR-0011). The two disagree. Not from this
branch — `Option` was nameable in an annotation before it — so it is left alone here and worth its
own issue.

1430 + 37 + 95 green; souther-examples still builds clean.

kawasima:
Confirmed and fixed in 0c7301f.

The report is right, and it reproduces — though not through the path I first tried. A lambda in an
argument position is expanded by the inliner before inference runs, and expansion α-renames its
parameters, so two source lambdas both spelling `x` no longer collide by the time `reaches` sees
them. What does reach it is a binder the expansion leaves alone. A `match` arm is one:

```souther
let y = 10
let r = f(match o.m with | Some y -> y | None -> 0)   -- cannot infer the type of `f`
let r = f(match o.m with | Some z -> z | None -> 0)   -- compiles
```

Two programs differing by the name of a bound variable, and only one of them compiles. That is the
defect exactly as described.

Fixed by subtracting binders on the way down, per the first of the two options: a lambda's
parameters, a `let`, a `match` arm's binding, and the binder an attempted construction opens each
take their own name off the set for what they cover. The regression test compiles both spellings of
the pair above and holds them to the same output, so the property under test is that the name of a
bound variable decides nothing — not that one particular program happens to compile.

The failure mode was conservative in both directions: it over-skipped, so the result was
`check.fn.noinfer` naming the annotation, never a wrong type or a wrong acceptance. That is why it
survived the suite and the corpus, and it is also why the test had to be a pair rather than a single
program.

`freeVariables(expr)` as a named function would say the same thing more plainly, and it is the better
shape if a second reader ever needs it. There is one reader, and it wants a predicate rather than a
set, so the subtraction is written into the walk for now.

CI on head `0c7301f`: `build` passing —
https://github.com/souther-lang/souther/actions/runs/30467082742/job/90627512791

1431 + 37 + 95 green; souther-examples still builds clean.

kawasima:
Fixed in 6324bc3, non-blocking or not — the point is the one this branch is about, so it should not
ship inconsistent with itself.

Reachable through both paths, confirmed before changing anything:

```souther
Map.size(Map.singleton(p, 1))              -- a Map key, reported as a Set element
Map.size(List.groupBy((i) -> p, xs))       -- same
```

`uncomparableCollection` now answers with which collection asked, as suggested:

```java
public sealed interface UncomparableIn {
    Type type();
    record SetElement(Type type) implements UncomparableIn {}
    record MapKey(Type type) implements UncomparableIn {}
}
```

so the reader has it to report rather than having to guess. Written collections already said the right
thing — the written path resolves the key and the element separately, each with its own message — so
this is the inferred path catching up to them. Two tests, one per path above.

CI on head `6324bc3`: `build` passing —
https://github.com/souther-lang/souther/actions/runs/30467546701/job/90629129541

That is twice now that the status has not come back through the API; if it stays unreadable from
there, `gh pr checks 201` reads it directly.

1433 + 37 + 95 green; souther-examples still builds clean.

PR #203, opened 2026-07-29
An example row may state the rule it is about

### LINKED ISSUE #200: An example cannot state a row with a helper
An `example` row states its input, its expected result, and the value it fakes a dependency with. A *value* may be named in all three (ADR-0072, ADR-0074). A *helper* may not be called in any of them — and now that a module publishes its helpers (#197, ADR-0075), stating a row with the rule it is about is the natural thing to want and the one thing a row cannot do.

None of this is about publishing. Every case below is a module's own helper.

## What happens

**In the input.** Refused, clearly:

```souther
let taxed (a: Amount) = Amount(a.value * 11 / 10)

example bill
  | (taxed(Amount(100))) -> Receipt { total = Amount(121) }
```

```
E1903: could not build input #1 of `bill`: `taxed` is not a newtype; a fixture cannot call it
```

`ExampleVerifier.newtypeInner` builds a fixture out of literals, records, collections and newtype constructors. A call to anything else is refused there.

**In the expected result.** Not refused — answered wrongly:

```souther
example bill
  | (Amount(100)) -> Receipt { total = taxed(Amount(100)) }
```

```
E1905: example does not hold: expected Receipt but was Receipt { total = 110 }
```

The row is correct: `taxed(Amount(100))` is 110 and the behavior produces 110. The expected side is not evaluated, so the row fails, and the message reports a mismatch against an expected value it never built — `Receipt` with no fields. Of the three this is the worst: the other two say no, and this one says something untrue.

**In a `with` value.** The clause stops being recognised at all:

```souther
example bill
  | (Amount(1)) with quoted = taxed(Amount(100)) -> Receipt { total = Amount(110) }
```

```
E1908: `bill` depends on `quoted`, but no fake is given for it.
```

For contrast, a value works in all three positions:

```souther
let standard = Amount(100)

example bill
  | (standard) -> Receipt { total = Amount(100) }      -- compiles
```

## Why it matters more now

A published helper is the shared statement of a rule. An example of a behavior that applies that rule has to restate the rule's result as a literal, which is the duplication #163 and #197 removed everywhere else: the row says `Amount(121)` where it means "the taxed price of 100", and nothing connects the two, so a change to `taxed` leaves a row that is wrong and still passes for the wrong reason.

A row naming a *value* was worth having for the same reason (#163). A row calling a helper is the other half of it.

## What it would take

A helper is expanded at its call site, so a row's expressions could be expanded the same way before the fixture is built — the helper's body is already reachable there, and a published one arrives closed. Then all three positions are ordinary expressions again. The questions to settle:

- Whether the expansion happens before `ExampleVerifier` reads the row, or whether the verifier learns to apply a helper. The first keeps the fixture builder's rule ("literals, records, collections, newtype constructors") unchanged, which is what makes a record serve as both a value and a fixture.
- A recursive helper is not expanded, so a row calling one has to run a method rather than fold an expression. An example is already evaluated within a time budget, so this is a question about what the verifier may call, not about termination.
- Whether the expected side rejecting what it cannot evaluate — rather than comparing against a value it never built — is worth fixing on its own. It is a wrong answer today whatever is decided about helpers.

Follows #197, ADR-0075. Related: #163, ADR-0072, ADR-0074.

kawasima:
Closes #200 (the default branch is `main`, so this needs closing by hand after merge).

## What it does

An `example` row may apply a helper — the rule the row is about — wherever a fixture value is written: an input, an expected result, a `with` value, a `fake` table's rows, a record field, a list element, a map key or value, and an argument of another helper application.

```souther
let taxed (a: Amount) = Amount(a.value * 110 / 100)

example bill
  | (taxed(Amount(100))) -> Receipt { total = Amount(110) }
```

The helper's arguments are built as fixtures against its parameter types, it is run as the method its module emits, and the value it returns is re-materialised into the form a fixture is written in. So a row stays a fixture, and one path serves an application at the top of a position and one nested inside a record.

## What the measurements changed

**Issue #200's proposal — expand the helper, then build the fixture — does not work.** Expanding the simplest realistic helper leaves the fixture builder what it has no case for:

```
taxed(Amount(100))
  ->  LetIn($0_a = Amount(100), Amount(FieldAccess($0_a, value) * 11 / 10))
```

`LetIn`, `FieldAccess` and `/` are none of them fixture forms. Expansion is not a way to avoid an evaluator; it is a way to need one. ADR-0032 already answered this for a constant construction's invariant: run the bytecode rather than fold it a second time.

**Re-materialising through the derived encoder does not work either.** A `Date`'s neutral form is the parsed temporal while its encoder writes ISO text, so a newtype over a date came back as text its own decoder refuses. It is read field by field through the declared types instead — which is what a written record fixture is built as, so one decoder reads both.

**No transitive closure is needed for what gets emitted.** A non-recursive helper is emitted from its inlined body, so what it reaches is already expanded into it, and a recursive helper it reaches is a method already. A value the row *names* is another matter: its body is read the way the row's text is, so the helpers it applies are followed as far as the values name each other.

**The owner's `$Fns` cannot be called across modules.** It is package-private in every module (which is how ADR-0075 keeps publishing a helper from opening a Java-reachable construction entry), and the declaring module emits no method for a non-recursive helper of its own. The reader emits it, as ADR-0075 already decided for a recursive one.

## Two wrong answers fixed first (`cca0c86`)

Independent of the feature:

- A row's expected value was never built before the comparison, so a row that held was reported as a mismatch against a value nothing had built: `expected Receipt but was Receipt { total = 110 }` for a correct row. It is built before the behavior runs, and a fixture that cannot be built is reported as one (E1903, its own message).
- A `with` value that could not be built was reported as `depends on ... but no fake is given for it`, which the row plainly gives. That gets its own E1908 message at the value's own position.

## Also

- The whole row is evaluated under one time budget — fixtures, helpers, the behavior, the comparison — so a row cannot buy more time by applying more helpers. E1910 names the helper it was in.
- A helper with no method (a standard-library intrinsic, or one whose body produces a function) is refused by that reason rather than by the general "is not a newtype".
- A value whose body applies a helper is a fixture, so ADR-0072's refusal of one "whose body computes" is withdrawn.
- The prelude-recursive-helper walk now reads a row's `with` values and `fake` tables, which it did not.

## Review round: three real bugs, all of them as reported (`1c256f1`, `5144daa`)

**A worker that will not stop held shared state.** Cancelling a row's evaluation is a request, not a stop — a pure Souther computation reaches no interrupt point — so everything the timed-out worker can still write to was at risk. `checkRowNow` added to the caller's diagnostics list from the worker thread. A row's evaluation now has its own list and hands it back only when it finishes; a late result is dropped, and only the waiting thread touches the aggregate. The evaluation also has its own `ExampleVerifier`, so the state a fixture builds up belongs to the row. `RowEvaluation` names that unit.

**`applying` raced.** It is now an `AtomicReference` per row, and it is read *before* the cancel — the order was wrong, since an interrupt can let the worker's `finally` clear it first. A regression test puts a non-terminating row before a mismatching one and pins that both diagnostics appear, in row order, each with its own reason. What remains is that a computation loop cannot be stopped at all; the deterministic step budget is still the fix for that.

**An interrupt was reported as a failing example.** No result was produced and no comparison made, so calling it a mismatch put a diagnostic on a model that may be correct. The interrupt flag is restored and the evaluation is abandoned.

**Responsibilities split.** `HelperInvoker` owns the `$Fns` lookup, the reflection, the classification of a failure inside a helper, and which helper is running. `NeutralForm` owns what the neutral form of a value is. `FixtureException` and `NonTerminationException` become their own files, being the vocabulary three classes share. `ExampleVerifier` keeps what an example means, and is 1392 lines from 1700 — smaller than before this branch.

**One deviation from the review.** Extracting only the re-materialiser was declined: a row reaches the neutral form two ways (the author's text, and a value a helper returned) while the rules that decide the form itself — which discriminator a case carries, when a unit case is a bare name, which written list is a map's entries — are the same rules either way. Putting them on one side and calling them from the other splits one rule across two places, so they live together and re-materialising is that class's entry point.

## Verification

`mvn clean install` green: 1459 compiler tests, 37 fmt, 95 LSP. `souther-examples` builds and its 19 tests pass against this compiler.

ADR-0077 records the five points fixed before implementation — transitive reachability, the row-wide budget, cross-module ownership, re-materialisation failure, the intrinsic refusal — plus what the feature does not check: a row that states the rule under test on both sides asserts nothing, and no check will say so.


kawasima:
Review addressed: the four required points and both recommended ones (`1c256f1`, `5144daa`).

## 1. A worker that outlives the timeout held shared state — as reported

The core of it is what the review says: cancelling is a request, not a stop, since a pure computation reaches no interrupt point. So everything the surviving worker can still write to is at risk. `checkRowNow(out)` added to the caller's diagnostics list from the worker thread.

A row's evaluation now has its own list and hands it back only when it finishes; a late result is dropped. Only the waiting thread touches the aggregate. The evaluation also has its own `ExampleVerifier`, so which values a fixture is expanding and which helper it is inside belong to the row. `RowEvaluation` names that unit in the type.

## 2. `applying` raced — as reported

It is an `AtomicReference` per row now, read **before** the cancel. The order was wrong, exactly as the review says: an interrupt can let the worker's `finally` clear it before the waiting thread reads it.

A regression test was added (`aRowThatDoesNotFinishLeavesTheRowsAfterItAlone`): a non-terminating row followed by a mismatching one, pinning that both diagnostics appear, in row order, each with its own reason.

What remains is that a computation loop cannot be stopped at all. The deterministic step budget is the real fix and is unchanged here.

## 3. `InterruptedException` classified as a mismatch — as reported

No result was produced and no comparison made, so the classification was wrong. The interrupt flag is restored and the evaluation is abandoned with a `CancellationException`. No diagnostic lands on a model that may be correct.

## 4. Splitting the responsibilities

`HelperInvoker` — the `$Fns` lookup, the method name, the reflection, what a failure inside a helper means, and which helper is running. The fixture side hands over a name and its arguments and reads back a value or the reason the row can state nothing.

`NeutralForm` — what the neutral form of a value is. **Extracting only the re-materialiser was declined**, because a row reaches that form two ways (the author's written text, and a value a helper returned) while the rules that decide the form itself — which discriminator a case of a sum carries, when a unit case travels as a bare name, which written list is a map's entries — are the same rules either way. Putting them on one side and calling them from the other splits one rule across two places. Re-materialising is that class's entry point (`NeutralForm.of`).

`FixtureException` and `NonTerminationException` are their own files now, being the vocabulary three classes share rather than one class's internals.

`ExampleVerifier` keeps what an example means: the target, the row's checks, the fakes, the comparison, the diagnostics. 1392 lines from 1700, smaller than before issue #200 was implemented. The two items the review marked as follow-ups (the whole fixture builder, and the fake/behavior execution) are not in this branch.

## Verification

`mvn clean install` green: 1459 compiler tests, 37 fmt, 95 LSP. souther-examples builds and its 19 tests pass against this compiler.

Issue #202, opened 2026-07-29
`Option<T>` cannot be written as a type, except that it can


kawasima:
ADR-0011 decided that `Option<T>` cannot be written as a type. The rule is enforced on the spellings
`T?` and `Some(...)`, and not on the type name itself, so a model reaches the same type by naming it.

```
let p: Int? = None                      -- refused: `?` is for a data field, or the core
let p: Option<Int> = None               -- passes
let p = None                            -- refused: E1303, an optional is made at a `?` field
```

The third line is the one that shows this is not only a second spelling. `None` on its own is refused,
and it passes on the second line because the written `Option` annotation gave it the expected type the
guard allows. Naming the type reaches the construction rule too.

## Where it is writable today

Measured on `develop` at 6e1a18c. Every one of these compiles in an ordinary module:

```
let p: Option<Int> = None               -- a local binding's annotation
data R = { f: Option<Int> }             -- a data field, beside the `Int?` that means it
behavior run : (o: Option<Int>) -> R    -- a behavior's input
behavior run : (r: R) -> Option<Int>    -- a behavior's output
let unwrap (o: Option<Int>) = ...       -- a helper's parameter
let firstOf (xs: List<Int>): Option<Int> = List.get(0, xs)   -- a helper's declared return
```

The fourth is the one ADR-0011 names outright: "None of the three may appear in a behavior output."
Its reason is that `>->` would have to decide whether it consumes `Some` / `None` or treats
`Option<会員>` as one case, and saying it in business vocabulary — `-> 会員 | 会員なし` — removes the
question. Composing such a behavior does fail today, with E1701 "these behaviors cannot be composed",
which blames the pair rather than the output that should not have been written. The author is told the
two do not fit, not that the absence belongs in the model's own sum.

## What is decided already, and is not in question

ADR-0011. `Option` is not a surface type; an optional is made in one place, a construction giving a
`?` field its value; a model reads an optional everywhere else and passes it on. This issue does not
reopen any of that. It is about the rule reaching the type rather than two of its spellings.

## The mechanism

`TypeOps.denoted` answers the built-in type names, and `Option` sits in that switch beside `List`,
`Set` and `Map`, which are surface types. Nothing there asks which module is being compiled. The
`?` form is guarded instead, in `AstBuilder.optional`, which refuses it outside a data field unless
the module is in the reserved `souther` namespace — so the guard is on the mark, not on what the mark
desugars to.

[...]

Issue #166, opened 2026-07-27
`List.filterMap` cannot be given a step, because its step type cannot be written


kawasima:
`filterMap` is declared in `souther/list.sou` as taking `(f: ('a) -> 'b?, xs: List<'a>)`. A model cannot supply that `f`, because it cannot state the return type:

```
let pick (n: Int): Int? = ...
```

```
-- SYNTAX ERROR --------------------------------
4| let pick (n: Int): Int? = ...
                      ^
An optional type `T?` may be written on a data field, or in the core (the reserved `souther`
namespace); a user model never names an optional elsewhere.
```

So `filterMap` is present in the stdlib and unreachable from a model. The rule that blocks it is ADR-0011 (Option is not a surface type), which is a decision worth keeping — but it leaves a published stdlib function with no callers a model can write.

## What the model writes instead

The projection answers a list of nought or one and `concatMap` does the work, which is `filterMap` open-coded at every call.

## Proposal

Let `filterMap`'s step answer a 0-or-1 list as well as an optional. That keeps ADR-0011 intact and makes the function reachable. Alternatively, remove it from the stdlib — either way it should not stay in a state where the only callers it can have are core ones.

Found by the `crm` example in souther-lang/examples (finding F8 in its README), and verified against `develop` with the repro above.

kawasima:
Fixed by #172. The premise turned out to be wrong —  is reachable, including with a named step, as long as the model does not write the type. What was missing was the diagnostics saying so, which is what changed; the language is unchanged.

kawasima:
Correction to the comment above, which lost its code spans: the function that is reachable is `List.filterMap`, and the way to name a step is a helper that leaves its result type off, so `filterMap(assigneeOf, issues)` compiles and runs. Writing `Name?` is what the rule refuses.

PR #204, opened 2026-07-29
Put the Option rule on the type, not on two of its spellings

kawasima:
Found beside issue #200. Needs closing by hand after merge (the default branch is `main`).

## The re-examination moved the answer

Issue #202 proposes guarding the name as well. Measured, **that costs more than the one corpus line**. Five of the six positions the name reached are the reading ADR-0011 already permits, written instead of inferred — and **two of those cannot be written any other way, because the annotation is mandatory there**:

```souther
partial let firstOver (xs: List<Int>) : Option<Int> = ...          -- a recursive helper declares its return
let firstKept (step: (Int) -> Option<Int>, xs: List<Int>) = ...    -- a function parameter is annotated
```

The second is how a user writes a `List.filterMap`-shaped combinator. Refusing the name leaves no way to say what its step answers. The corpus has no helper of that shape, which is why the measurement did not show it.

## The leak was the reasoning, not the name

`None` was accepted **wherever the expected type was an optional**, on the grounds ADR-0011 states itself: an expected optional reaches an expression only from a field's own type, since nowhere else in a model can `T?` be written. Naming the type is another way to ask for one, so the premise stopped holding and `let p: Option<Int> = None` made an absence outside a construction.

A second leak had the same root and showed up **as one helper with two answers**: `| None -> None` inside a helper declaring an `Option<T>` return was accepted, while the same helper inlined as a `List.filterMap` step was refused. Which answer you got depended on where the helper was used.

## What is decided

**Making an optional is the `?` field being given its value, and nothing else.** The permission travels with the check rather than with the expected type, so a field's branches still have it (`if g then v else None` compiles) and an annotation never does. A helper answers the optional it read, which is what ADR-0011 says a model does everywhere but the one place:

```souther
partial let firstOver (n: Int, xs: List<Int>) : Option<Int> = {
    let head = List.get(0, xs)
    match head with
        | Some v -> if v > n then head else firstOver(n, List.drop(1, xs))
        | None -> head                  -- the absence it read, passed on
}
```

That keeps the property issue #166 rests on — a step for `List.filterMap` has to answer an optional it read, so absence the model owns stays a case of the model's own sum — and keeps it in every position rather than in the one the call site happened to check.

**A behavior does not answer an optional.** Read off the resolved output type, so `-> Option<Int>` and `-> Int?` are refused as the same thing, and the report names the rule instead of blaming the pair that could not compose (E1701).

**The `>->` question ADR-0011 declined to answer has been answered all along.** Measured: an optional output composes when the next stage takes the same optional, and it travels as one case. The prohibition stands on the other two grounds — the business vocabulary a reader of the answer matches on, and keeping the runtime's `Option` out of the Java-facing signature an exposed behavior generates (ADR-0008; `JvmTypes:185` maps an optional to `souther.runtime.Option`).

`T?` stays the field mark. Each spelling has a job: `?` marks where an optional is made, the name says what the type is.

## Verification

`mvn clean install` green: 1474 compiler tests, 15 of them new. **Migration cost across souther-examples: none** — the one `Option<...>` written there reads an optional and passes it on, and every `None` is a `?` field's value or an example fixture. Its 19 tests pass against this compiler.


kawasima:
Both points addressed (`2522897`).

## [Blocker] The permission does not reach into arguments or lambdas

Measured, and it does not reproduce — including the sharpest form of it, an argument whose parameter type says `Option<Int>` outright:

| written inside `Note { value = … }` where `value: Int?` | result |
| --- | --- |
| `Option.withDefault(0, None)` | E1303 |
| `List.get(0, List.filterMap(x -> None, [ 1, 2, 3 ]))` | E1303 |
| `List.length(List.map(x -> None, [ 1, 2, 3 ]))` | E1303 |
| `{ let p = None  p }` | E1303 |
| `orZero(None)`, where `let orZero (o: Option<Int>) : Int` | E1303 |
| `orZero(None, 1)`, where `orZero` is recursive (so the call survives inlining) | E1303 |
| `if n > 0 then n else None` | compiles, as it must |
| `match k with \| Big -> n \| Small -> None` | compiles, as it must |

The mechanism: `None` is accepted only when the permission **and** an expected optional are both there, and the second one is what stops at those boundaries. `CallElaborator.CallArgs` elaborates every argument with `Elaborator.elaborate(args.get(i), env, ctx)` — no expected type; arguments are typed bottom-up and then checked against the parameter type. A lambda body is the same, which is ADR-0011's own consequence ("a lambda handed to a stdlib combinator is typed with no expected type at all"). So the flag alone grants nothing, and the leak needs a position that has both — which is the field's value and its branches, where the propagation into `if`/`match` is deliberate.

That said, the structural objection stands on its own: the property was resting on a second mechanism in another file, and nothing here said so. Two changes:

- The permission is **dropped at both boundaries** — `CallElaborator.CallArgs` and `Elaborator.elaborateFunctionValue` — with the reason written down. Redundant with the expected type today, and stated so the rule holds where it is written.
- The seven shapes above are a test (`theFieldsPermissionDoesNotReachWhatItsValueCalls`, plus the two that must keep passing). If arguments ever gain expected types, that test is what notices.

## The ADR's title contradicted its content

Fixed by splitting rather than retitling. The Option decision is now **ADR-0078**, "An optional is made in one place and named anywhere it is read", with the measurements and the reasoning. ADR-0011 keeps its `Never` and `Unit` decisions, its status line says `Option` is revised by 0078, and the index row says so too. Renaming 0011 would have edited a dated record of what was decided then; the repository's pattern is a later ADR revising an earlier one (0075 revises 0074, 0076 amends 0025).

## Verification

`mvn clean install` green: 1475 compiler tests, 37 fmt, 95 LSP. souther-examples builds and its 19 tests pass.

Issue #211, opened 2026-07-30
The expected side of an example row cannot name a value directly, only spread one


kawasima:
ADR-0072 lets an `example` row name a value in its input, its expected result, or a \`with\` fake, and
spread one to vary a field. But a bare name on the expected side is read as a case of the behavior's
output union, not as a reference to a value — so naming one there, with no spread at all, is refused.

## Repro

```souther
module f26
data In = { n: Int }
data Out = { m: Int }
behavior run : (i: In) -> Out
    constructs Out
let run (i) = Out { m = i.n }

let priced = Out { m = 1 }

example run
    | (In { n = 1 }) -> priced
```

```
-- EXAMPLE  E1904 ------------------------------
11|     | (In { n = 1 }) -> priced
          ^
`priced` is not one of the result cases of `run`.
Hint: Expected one of: Out
```

A construction that spreads `priced` (`Out { ...priced }`) is accepted, and is usually the better reading
anyway when a field is being varied — but where a row's whole expectation is a fixture already named with
nothing to vary, it still has to be written as a one-field spread of itself just to reference it.

## Cost in the model

`pipeline` (souther-lang/examples) names its deal fixtures and spreads them to vary the field a
transition's row is about; where a row's expected side is the fixture unchanged, it still has to write
`-> Qualification { ...acmeQualified }` rather than `-> acmeQualified`, for a construction that varies
nothing.

## Proposal to discuss

Read a bare name on the expected side the same way one is already read in the input and `with` positions:
first as a value in scope, and only as a case name of the output union when it resolves to neither.

Found by the `crm` example in souther-lang/examples (finding F26 in its README), verified against
`develop` with the repro above.

kawasima:
Fixed by #213, along with #206: which case a row's expectation asserts is read through what it names, so a bare name on the expected side stands for the value it denotes, and only a name denoting a case asserts a case and nothing under it.

One difference from the proposal here. It reads a bare name as a value first and as a case name only when it resolves to neither; #213 asks the case first. The two answer the same thing — a spelling that is both a `data` and a `let` is already `check.dup.valuename`, so no name reaches both branches — and asking the case first leaves what a case name means untouched.

Two things fell out of it. A named expectation standing for a literal (`let answer = 42`) works, and rendered as `answer(42)` before it was reached at all. And a failing named expectation shows the value it stands for (`Out { m = 1 }`) rather than the name.

[...]

Issue #206, opened 2026-07-30
A fixture value built by a spread takes the spread source's case


kawasima:
Found by the `crm` example in souther-lang/examples while taking up finding F20 (a fixture can be named now, ADR-0072/0074). Verified against `develop` at 47bcd03.

A value whose body spreads another value carries **the spread source's case** into an `example` fixture, not its own. Where the position is a union, the row runs against the wrong case and the example still evaluates — so it reports a mismatch that names the right expectation and the wrong actual, or passes silently when both cases answer alike.

```souther
module probe.spread

data Common = { id: String }
data Open   = { ...Common }
data Closed = { ...Common, closedOn: Date }
data Deal   = Open | Closed
data Answer = { label: String }

behavior labelOf : (deal: Deal) -> Answer
    constructs Answer

let labelOf (deal) =
    match deal with
        | Open   -> Answer { label = "open" }
        | Closed -> Answer { label = "closed" }

let flatClosed   = Closed { id = "d-1", closedOn = Date("2026-07-30") }
let openDeal     = Open { id = "d-1" }
let spreadClosed = Closed { ...openDeal, closedOn = Date("2026-07-30") }

example labelOf
    | "a flat fixture keeps its case"               : (flatClosed)   -> Answer { label = "closed" }
    | "a fixture built by a spread keeps its case"  : (spreadClosed) -> Answer { label = "closed" }
    | "the row writing the case around it"          : (Closed { ...spreadClosed }) -> Answer { label = "closed" }
```

The first row holds. The second and third do not:

```
-- EXAMPLE  E1905 --------------------------spread.sou:25:72
    | "a fixture built by a spread keeps its case" : (spreadClosed) -> Answer { label = "closed" }
It is:
    Answer { label = "open" }
But it needs to be:
    Answer { label = "closed" }
```

Three things this says:

- The value is a `Closed` construction, and it reached the behavior as an `Open`.
- Writing the case around it in the row (`Closed { ...spreadClosed }`) does not fix it — the inner value's tag wins over the enclosing construction's own case.
- Nothing is reported about the fixture. The mismatch is the only signal, so two cases whose rule agrees would leave a green example asserting the wrong value.

## The same root, refusing rather than lying

A value built by a spread cannot be named by another module's row at all:

```
`nextStepFor` の入力 #1 を構築できませんでした: an example fixture must be a literal or a construction
```

and where the union's decoder needs a discriminator it is missing rather than wrong:

```
[...]

PR #213, opened 2026-07-30
A fixture keeps the case it constructs

kawasima:
Closes the three faces of #206, and #211 with them — a name on the expected side is read the same way, so the same change answers both.

## What was wrong

Three faces, one root each.

**An input ran against the wrong case, silently.** `ExampleVerifier.record()` copied every entry of the spread source's neutral form into the enclosing construction. That form carries the source's own discriminator, and the tag written afterwards is a `putIfAbsent`, so the construction's own case never landed. Where the position is a union, the tag is what decides which case the behavior sees — so a row reported a mismatch naming the right expectation and the wrong actual, or passed where both cases answer alike. Where the source's case is no case of that union at all, the row was refused instead (`type: must be one of [...]`).

**An imported value built by a spread could not be named.** A spread holds a name rather than an expression, so closing a published body binds the value ahead of the construction. `raw()` had no case for that binding and said the fixture must be a literal or a construction.

**A `fake` row and a `with` value could not name a value.** `fixtureType` read the case off the spelling of the expression, and a value's name is not a case name. It does not show at an input, whose declared type is always a named sum with a decoder that dispatches on the tag; it shows where the type is a union written inline (`-> Account | NotFound`), which has no decoder, so the case has to come from the value. The expected side read a name as the case it asserts for the same reason.

## What changed

- A spread copies fields, and only the ones the construction has — which is what the language copies (`Narrow { ...wide }` drops a field `Narrow` does not declare). The discriminator is not a field.
- A fixture reads a binding the way it reads a value, so a closed published body that spreads one holds.
- `constructedCase` reads the case a fixture stands for, following a name to the value's body. `fixtureType`, the arm check (E1904), the built expectation and the rendering of a failure all read it.
- The expected side tells a name denoting a case from a name denoting a value: the first asserts the case and nothing under it, the second asserts the value (#211). A named expectation standing for a literal now renders as the literal rather than as `answer(42)`.

  #211 proposes reading a bare name as a value first and as a case name only when it is neither. The order is the other way round here — a case first — because the two answer the same thing: a spelling that is both a `data` and a `let` is already `check.dup.valuename`, so no name reaches both branches. Taking the case first leaves what a case name means untouched.
- `check.fake.unbuildable` replaces the three sites that reported an unbuildable fake as a missing one. `lookUp depends on lookUp` and "no fake was supplied" where one was written are both gone.
- Two sentences in the specification: what a spread copies and where the case comes from, and that E1908 covers a `with` value and a `fake` row alike. No ADR — the rules were already stated (§example-evaluable already says a value whose body is a spread is fixture-evaluable); the implementation did not hold them.

## Verification

13 new tests. Swapping `develop`'s `ExampleVerifier` back in fails 6 of the 7 spread tests, including the refusing face from the issue's second comment. The seventh documents the language rule the field filter mirrors, and passes either way. Full suite green (1507 + 95 + 37), `mvn clean install` clean.

## Not in this change

- souther-lang/examples `main` is already red against `develop` (`forecasting.sou:175`, E1401 — a helper calling a behavior), so the crm fixtures could not be measured end to end here. Unrelated to this change: `develop`'s compiler fails at the same position.
- A helper applied at the top of an expected result compares its neutral form against the live value, so a row that is right is reported as a mismatch (`-> taxed(Amount(100))` gives `expected 110 but was Amount(110)`). ADR-0077 says a row may apply a helper wherever a fixture value is written, so this is a gap of its own, one line from what this change touches. Filed as #214.

### DIFF
--- a/souther-compiler/src/main/java/souther/compiler/ExampleVerifier.java
+++ b/souther-compiler/src/main/java/souther/compiler/ExampleVerifier.java
@@ -780,8 +780,16 @@ public final class ExampleVerifier {
         return constructedCase(e, new LinkedHashSet<>());
     }
 
-    /** As above; {@code followed} are the names already followed, so a value defined in terms of itself
-     * stops here and is reported as the cycle it is where the fixture is built. */
+    /**
+     * As above; {@code followed} are the names already followed, so a value defined in terms of itself
+     * stops here and is reported as the cycle it is where the fixture is built.
+     *
+     * <p>The binding a closed body carries is not read here, and does not need to be: a value is
+     * substituted where it is named, so the only name closing leaves standing in a fixture is the one a
+     * spread holds — a spread cannot hold an expression — and that name is read where the spread is
+     * copied. A closed body therefore ends in the construction, whether it was published itself or
+     * named by another value that was.
+     */
     private TypeName constructedCase(Ast.Expr e, Set<String> followed) {
         return switch (e) {
             case Ast.NewData nd -> nd.typeName().denotes();
--- a/souther-compiler/src/main/java/souther/compiler/ExampleVerifier.java
+++ b/souther-compiler/src/main/java/souther/compiler/ExampleVerifier.java
@@ -26,6 +26,7 @@ import java.util.ArrayDeque;
 import java.util.ArrayList;
 import java.util.Deque;
 import java.util.LinkedHashMap;
+import java.util.LinkedHashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
@@ -356,9 +357,10 @@ public final class ExampleVerifier {
                 return;
             }
         }
-        // validate the expected arm/value against the output cases before running
+        // validate the expected arm/value against the output cases before running. Which case the row
+        // asserts is read through what it names, so a row may name a value where it may name a case.
         String expectedArm = expectedArm(row.expected());
-        TypeName named = expectedArm == null ? null : symbols.resolveCase(expectedArm);
+        TypeName named = constructedCase(row.expected());
         // a name nothing here denotes is not an arm of the target either, and says so the same way
         if (expectedArm != null && !outCases.isEmpty()
                 && (named == null || !outCases.contains(named))) {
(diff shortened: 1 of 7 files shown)


kawasima:
## [P1] does not reproduce, and the reason is now in the code

The scenario is a good question and I ran it rather than reasoning about it. An alias of a value built by a spread is closed to this — dumped from `Bodies.Helpers` while the row was being read:

```
deals.named -> LetIn[name=$s1_openDeal, value=NewData[Open …],
                     body=NewData[Closed, spreads=[$s1_openDeal]]]
```

A value is *substituted where it is named*, so `let named = spreadClosed` does not become `… in spreadClosed`; it becomes the same construction with its own fresh binding around it (`$s0`, `$s1`, `$s2` for the value and two names for it — a binding per published value, not a shared one). The tail of a closed body is therefore always the construction, so `constructedCase(let.body())` reaches it, and the trailing `Var` that would need the environment is not a shape the inliner can produce. The only name closing leaves standing in a fixture is the one a spread holds — a spread cannot hold an expression — and that one is read where the spread is copied, through `valueBody`, which does consult `bindings`.

Checked at every position that reads it:

- an input typed by an imported union — `(named)`, `(namedTwice)`, and an imported spread-of-a-spread;
- an expected result, which is where the case a name stands for decides what is compared rather than a decoder dispatching on a tag — `-> spreadClosed`, `-> named`, `-> namedTwice`.

All hold. Added as `anImportedNameForAValueBuiltByASpreadKeepsItsCaseHoweverManyNamesItPassedThrough`, and the reason is stated on the `LetIn` case so the next reader is not left to ask it (4dd9b78).

One correction to the repro sketched in the review: an inline union in this module cannot list an imported case at all (E1606 — a case class carries its unions from its own module's generation), so an imported name never reaches a `fake` output or a `with` value typed by an inline union. Where those are typed by an imported *named* sum, its decoder is there and the fallback is exact. The reachable position for an imported name is the expected result, which is what the new rows cover.

I did not add the environment push. It would be a branch no input can reach, and if a later change did make a trailing `Var` reachable, `constructedCase` returning null falls back to the declared type — a row that fails, not one that passes with the wrong case. Happy to pass an explicit environment instead if you would rather have the two resolvers agree by construction than by a stated rule.

Issue #207, opened 2026-07-30
A cross-module `>->` cannot rename an imported departed case into one this module declares


kawasima:
An aggregate that spans two contexts — one behavior in module `A`, the next in module `B` that cannot
import `A` back — cannot be written as one composition, even though `>->` is exactly the compile-time
link the rest of the language uses for this.

## Repro

```souther
module up exposing ( Out, ADeparture, a )
data ADeparture
data Out = { n: Int }
behavior a : (i: Int) -> Out | ADeparture
    constructs Out, ADeparture
let a (i) = if i > 0 then Out { n = i } else ADeparture
```

```souther
module down
import up ( Out, ADeparture, a )
data BDeparture
data Out2 = { m: Int }
behavior b : (o: Out) -> Out2 | BDeparture
    constructs Out2, BDeparture
let b (o) = if o.n > 0 then Out2 { m = o.n } else BDeparture

behavior ab = a >-> b
```

```
-- IMPORTED CASE IN AN OUTPUT UNION  E1606 ------
9| behavior ab = a >-> b
   ^
The output union of `ab` includes `ADeparture`, a case declared in another module.
Hint: A case class carries the unions it belongs to, decided when its own module is generated, so an
imported case cannot join a union declared here; consume `ADeparture` at the boundary, or re-express it
as a case of this module.
```

The hint's second half — "re-express it as a case of this module" — is exactly what cannot be written:
there is no syntax naming what a local case an imported departure lands on, only the refusal of the union
as-is.

The reverse direction is closed too: writing the composition from `up`'s side instead would need `up` to
import `down`, which is refused as a cyclic module dependency (E1501) since `down` already imports `up`.

## Cost in the model

Lead conversion (`crm` in souther-lang/examples) produces an Account, a Contact **and** an Opportunity
together; a conversion that made two of the three is a data-integrity bug. The Opportunity is built in
`example.pipeline`, which imports `crm`, so `crm` cannot import it back (E1501); and
`crm.convertLead >-> pipeline.openFromLead`, written in `pipeline`, would put `crm`'s `ConversionBlocked`
into a union `pipeline` declares (E1606, as above). So the two calls are joined in Java instead of one
composition — the one place in this domain where an aggregate crosses a context boundary is exactly where
the compile-time link `>->` gives everywhere else is missing.

## Proposal to discuss

A translating composition, naming the local case an imported departure lands on:

```
[...]

PR #221, opened 2026-07-30
An imported departure is answered by a case this module declares

kawasima:
`>->` cannot cross a module boundary when a stage departs by a case the importing module did not declare: the case class settles its unions when its own module is generated, so it cannot join one declared here (E1606, ADR-0057). That decision stands. What was missing is that the diagnostic never said what to write instead — "re-express it as a case of this module" names no shape — and the `crm` example read it as a language limit, joining the two contexts in Java and leaving the aggregate rule stated nowhere.

The shape was writable all along. An operation that spans two modules is one behavior with a body: it calls the stages, opens the imported result with `match`, and answers the imported departure with a case it declares itself. The hint now names that shape, and the specification shows it.

Two tests are added. One pins the shape the hint recommends, with the imported mainline carried on to a behavior of this module that has a requirement of its own — so it is received rather than built, which is the one thing a call does differently from a stage.

## The bug that shape found

Calling an imported behavior of two or more arguments emitted a typed `apply` whose return type named the *reading* module: `inv.AllocateFor.apply` called for a `ship.AllocateForResult`, a class nothing writes. It compiled and threw `NoSuchMethodError` on the first call.

`<名>Result` is a name the backend makes up, so `cd(String)` finds nothing declaring it and falls back to the module being generated — right for a behavior declared here, wrong for an imported one. It now resolves through the declaring module, the way the behavior's own class and `$Impl` already do. One argument was unaffected: that call goes through the erased `Behavior.apply(Object)Object`, which names no union at all.

Refs #207.


kawasima:
Both test points addressed in 9a26459, and the second one turned out to widen the fix's stated scope.

**1. The operation now runs.** `anOperationSpanningTwoModulesIsOneBehavior` binds a counting stand-in for `Instruct` and applies the composed behavior. `assertSame` on the object that stand-in returned is what pins "received rather than built" — a self-built stage could not produce that instance — and the departure case asserts the count stays at 1, so a departed value does not go on to the next stage.

**2. Zero arguments was a real, separate case, and it was broken.** Measured on `develop`:

```
up/Mint.apply:()Ldown/MintResult;
```

It reaches the typed `apply` by the other road. A behavior of no arguments cannot have a body at all — a `let` with no parameter list is a value (ADR-0072) — so it is always *injected* and reached through `depends on`, i.e. `requiredCall`/`requiredApplyDesc` rather than `behaviorCall`. Both funnel through `typedApplyDesc`, so the fix already covered it; there was no test saying so, and the PR text wrongly scoped the defect to two or more arguments.

`anImportedBehaviorOfNoArgumentsIsCalledOnItsOwnResultUnion` covers it. It reads the call descriptor off the emitted bytes instead of running the call: an injected behavior's base is an abstract class with a protected constructor, so a stand-in cannot be made by reflection alone. It also asserts the named class is one the compilation actually wrote, which is the property the bug violated.

Both new assertions were confirmed non-vacuous by reverting `refTypeOrNull` and watching them fail:

```
anImportedBehaviorOfNoArgumentsIsCalledOnItsOwnResultUnion:210
  expected: <()Lup/MintResult;> but was: <()Ldown/MintResult;>
anImportedBehaviorOfTwoArgumentsIsCalledOnItsOwnResultUnion:259
  NoSuchMethod 'ship.AllocateForResult inv.AllocateFor.apply(inv.Sku, java.lang.Long)'
```

**3. Specification wording** rewritten as suggested — it now says what each side supplies rather than negating "requires nothing".

Full suite green.

Issue #225, opened 2026-07-30
The checker does not know which stdlib operation preserves which property


kawasima:
Sub-issue of #222. This is the table that decides how much of a model the language tracks.

With #223 and #224, a term is an atom and a predicate is a fact, but both are keyed by the expression as
written. `List.length(List.map(f, xs))` and `List.length(xs)` are still unrelated, so the ordinary shape
— guard the input, construct from a mapped list — does not discharge.

## Scope

Normalization rules on the canonical key, stated per operation and per property, as part of the
specification rather than as an implementation detail:

| operation | length | uniqueness | element bounds |
| --- | --- | --- | --- |
| `List.map` | preserved | not preserved | depends on the closure |
| `List.filter` | bounded above | preserved | preserved |
| `List.reverse` | preserved | preserved | preserved |
| `List.sortBy` | preserved | preserved | preserved |
| `Set.map` | bounded above | — | depends on the closure |

The table above is a sketch to be settled, not the proposal. What each cell must say is precise enough
to normalize a key or to add a bound.

## Done when

- `QuoteLines`'s `nonEmpty` clause discharges at `buildQuote` in souther-lang/examples.
- The rules are in the specification, so an author can predict which constructions are judged safe.

kawasima:
Merged via #231.

Issue #222, opened 2026-07-30
A guard on a collection or string property discharges nothing, and an unguarded construction of one is not reported


kawasima:
The invariant discharge check (`[#invariant-discharge]`, ADR-0056) reasons over numeric relations on the
constructed value. A property of a collection or a string is outside it entirely — not merely unproven,
but unrepresented — so a construction that must abort is accepted in silence, and a guard that states
exactly the invariant discharges nothing.

## Measured on `develop`

| invariant | body | reported |
| --- | --- | --- |
| `data Money = Int invariant value >= 0` | `Money(n)`, no guard | **warning: may violate its invariant** |
| `data Name = String invariant String.length(value) >= 3` | `Name(s)`, no guard | *nothing* |
| `data Lines = List<Int> invariant List.length(value) >= 1` | `Lines(xs)`, no guard | *nothing* |

The last two are the hole. There is no guard at all and the construction will abort at run time, and the
compiler says nothing — while the numeric one, which is the same shape of mistake, is reported.

This is not about multi-clause invariants. A single-clause list invariant has the same lack of cover.

## Not a proof engine

`NumericDomain` is a linear-form domain over **atoms keyed by a string** (`LinearForm.atom(String)`). It
already has everything needed to reason about `List.length(xs)` — what is missing is that nothing ever
turns that expression into an atom. For

```souther
guard List.length(xs) >= 1 else NoItems
Lines(xs)
```

what the checker needs is that both occurrences of `List.length(xs)` name the same atom. That is term
interning over an uninterpreted function application, not a decision procedure.

The work above that is a **table of which stdlib operation preserves which property**, applied as
normalization of the atom's key — `List.length(List.map(f, xs))` is the same atom as `List.length(xs)`;
`List.filter` is not, and contributes bounds instead. Defining that table is the deliverable, and it is
what decides how much of a model the language tracks rather than leaves to a run-time check.

## Scope, in the order it can ship

Each is separately valuable and separately completable:

1. **A pure total stdlib call becomes a term the checker can name** — `List.length`, `String.length`,
   `Set.size`, `Map.size`. Closes the silence in the table above with no preservation rules at all.
2. **A fact domain for predicates that are not comparisons** — `List.allUniqueBy`, `List.all`,
   `List.member`, `Set.member`. A guard assumes the fact; a construction discharges it by the same
   canonical key.
3. **Property-preservation rules per stdlib operation** — `List.map` preserves length, `List.reverse`
[...]

PR #228, opened 2026-07-30
A pure total stdlib size call is a term the invariant checker can name

### LINKED ISSUE #223: A pure total stdlib call is not a term the invariant checker can name
Sub-issue of #222, and the one that closes the silence on its own.

`NumericDomain` reasons over atoms keyed by a string (`LinearForm.atom(String)`). Nothing ever turns a
stdlib call into such an atom, so `List.length(xs)` is invisible to it: a guard stating it establishes
nothing, and a construction needing it is not reported.

```souther
data Lines = List<Int> invariant List.length(value) >= 1

let build (xs) = Lines(xs)          // aborts at run time; no diagnostic today
let build (xs) = {
    guard List.length(xs) >= 1 else NoItems
    Lines(xs)                        // discharged by the guard, but not recognised as such
}
```

## Scope

Give a canonical atom key to a pure, total, statically-known stdlib call whose result is numeric:
`List.length`, `String.length`, `Set.size`, `Map.size`. No preservation rules — `List.length(xs)` and
`List.length(List.map(f, xs))` are two different atoms at this increment.

The invariant's `value` is substituted with the constructed expression before the key is formed, so the
clause and the guard agree on the key exactly when they name the same expression.

## Turning flagging on

A clause of this shape becomes flaggable: an unguarded construction is "possibly violated". The
guardrail in `[#invariant-discharge]` holds, because a guard on the same term now discharges it.

## Done when

- The three rows of #222's table all report.
- A guard restating a length invariant discharges the construction rather than being ignored.
- Nothing outside the listed calls changes behaviour.

kawasima:
Closes #223 (a sub-issue of #222).

## The hole

The invariant-discharge check reasons over atoms keyed by a string, and nothing ever turned a stdlib
call into one. So a property of a collection or a string was not merely unproven but unrepresented:
a construction that must abort was accepted in silence, and a guard stating exactly the invariant
discharged nothing. Measured on `develop`, the three rows of #222's table:

| invariant | body | before | after |
| --- | --- | --- | --- |
| `data Money = Int invariant value >= 0` | `Money(n)`, no guard | E2011 | E2011 |
| `data Name = String invariant String.length(value) >= 3` | `Name(s)`, no guard | *nothing* | E2011 |
| `data Lines = List<Int> invariant List.length(value) >= 1` | `Lines(xs)`, no guard | *nothing* | E2011 |

## What is a term now

`List.length`, `String.length`, `Set.size` and `Map.size` applied to a nameable location — a
parameter, a field chain, a newtype's wrapped value — become an atom keyed by the call written over
that location's path (`List.length(b.items)`). A guard and an invariant clause mean the same term
exactly when they size the same location. The size of a container is never negative, and the
procedure is told that much about every size atom a clause names.

Resolving the invariant's leaf names needed more than it had: the construction site handed over an
affine form per field, which cannot say *what* `value` is, only what it computes to. It now hands
over the value's canonical path as well, which is what lets `List.length(value)` at the clause and
`List.length(xs)` in the guard land on one key.

No preservation rules: `List.length(List.map(f, xs))` names no location, so a clause needing it
stays outside the fragment and its run-time check stands. That is #225.

## A rebinding drops the facts about the name

Found while adding the above, and a hole of its own. A `let`, a match binding, a combinator's
closure parameter and an attempt's binder can each shadow an outer name, and the facts held under
that name were about what it used to denote. Only an exact numeric assignment forgot anything;
`p.a` survived a rebinding of `p`. `NumericDomain.forgetIf` now drops every atom rooted at a name
the binding takes over.

## A guard restating an invariant now discharges it whatever its shape

Turning the flagging on exposed this. The domain derives over intervals and differences, and a
relation of neither shape was dropped on assertion — so `crm`'s `PossibleDuplicate`, whose
invariant is `List.length(accounts) + List.length(contacts) >= 1` and which is guarded by exactly
that, was reported anyway. Nothing the author could write would silence it, which is the one thing
the flagging rule promises. A form the two shapes do not hold is now kept as it was written, and
proves what it states plus any constant no greater than zero.

## Verification

- `mvn -o clean test` green; 12 tests added to `CompileInvariantDischargeTest`.
- Every module of souther-lang/examples compiled before and after. `develop` reports three warnings
  (`Quantity`, `UnitPrice`, `DiscountRate` in `crm`); this branch reports those three plus one:
  `Label` in `issuetracker`, built inside a `Set.map` over a set a `Set.filter` has just made
  non-empty of empties. That is honest — the check does not know `Set.filter` preserves an element
  property, which is #225 — and today it is silenced by attempting the construction rather than by
  a guard, since a `Set.map` closure has nowhere to put an `else`.
- Comparisons were run through `souther compile`, not the Maven build: the annotation-processor
  path reports no warnings at all.
- Specification: `[#invariant-discharge-terms]` states what the procedure can name, and the
  bounded-fragment paragraph now says a restated relation is kept as written.

### DIFF
--- a/souther-compiler/src/main/java/souther/compiler/check/InvariantChecker.java
+++ b/souther-compiler/src/main/java/souther/compiler/check/InvariantChecker.java
@@ -206,7 +206,8 @@ final class InvariantChecker {
                             paths.put(fi.name(), p);
                         }
                     }
-                    check(type, new Bindings(forms::get, paths::get), d, nd.pos(), attempted);
+                    check(type, new Bindings(forms::get, paths::get,
+                            TypeOps.fieldTypes(type, symbols)), d, nd.pos(), attempted);
                 }
             }
             case Ast.Binary bin when isArith(bin.op()) -> {
@@ -215,8 +216,8 @@ final class InvariantChecker {
                     LinearForm value = affineOf(bin, types);
                     if (value != null) {
                         // an arithmetic result is a form, not a location, so it names no path
-                        check(type, new Bindings(name -> "value".equals(name) ? value : null, _ -> null),
-                                d, bin.pos(), attempted);
+                        check(type, new Bindings(name -> "value".equals(name) ? value : null, _ -> null,
+                                TypeOps.fieldTypes(type, symbols)), d, bin.pos(), attempted);
                     }
                 }
             }
@@ -227,7 +228,8 @@ final class InvariantChecker {
     /** How an invariant's leaf names resolve at a construction site: to the affine form of what the
      * field is being given, and to that value's canonical path — so a size call over the field names
      * the same atom the body names when it calls the same function on the same container. */
-    private record Bindings(Function<String, LinearForm> form, Function<String, String> path) {}
+    private record Bindings(Function<String, LinearForm> form, Function<String, String> path,
+                            Map<String, Type> fields) {}
 
     /** Runs the discharge check for a construction of {@code type} whose field values resolve through
      * {@code binds}. A definite violation is an error; an unproven one a warning; a fully-discharged
@@ -358,7 +360,7 @@ final class InvariantChecker {
                     String p = resolvePath.apply(fieldName);
                     return p == null ? null : LinearForm.atom(p);
                 },
-                resolvePath);
+                resolvePath, fields);
         NumericDomain out = d;
         for (Ast.Expr inv : TypeOps.effectiveInvariants(data, symbols)) {
             List<Constraint> cs = invConstraints(inv, binds);
@@ -368,7 +370,11 @@ final class InvariantChecker {
                 }
             }
         }
-        if (!data.newtype()) {
+        if (data.newtype()) {
+            // A newtype's `.value` is the same location as the newtype, so what its base guarantees is
+            // guaranteed of this very atom: `data Outer = Inner` carries Inner's invariant.
+            out = seedAt(path, fields.get("value"), out, depth + 1);
+        } else {
             for (Map.Entry<String, Type> f : fields.entrySet()) {
                 out = seedAt(path + "." + f.getKey(), f.getValue(), out, depth + 1);
             }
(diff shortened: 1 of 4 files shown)


kawasima:
[P1] fixed in `bca15eb`, and it reproduced on all three shapes you named.

## The fix

Not a newtype test added to `invPath` — the two path rules are now one. A clause is a path over the
declaration's own fields, read by the very function a body path is read by (`pathKey`), and only the
field it is rooted at is then replaced by what the construction gives that field:

```java
String local = pathKey(e, binds.fields());          // one rule for what a location is
String root  = local up to the first '.';
String given = binds.path().apply(root);            // and the construction's own value for it
```

`Bindings` therefore carries the declaration's field types, which is your "path と type を組にして保持する"
— read off the declaration rather than threaded from the construction, since the clause is written
against the declaration.

## Your test 3 turned up a second hole

The nested case still warned after the path fix, for a different reason: **seeding never descended
through a newtype's base.** `data Outer = Inner` has no spread and is a newtype, so `seedAt` recursed
into neither, and `Inner`'s invariant never reached a parameter of type `Outer`. Since the two are one
location — which is the whole point above — what the base guarantees is guaranteed of the same atom.
Four lines, same commit.

This one was never about clauses: it also means a numeric `data 管理職 = レベル = Int` carried nothing of
`レベル`'s bound.

## Tests added

1. record field is a newtype, invariant sizes `field.value`, **no guard** — discharged by the input
   type alone (this is the one the seeding hole broke)
2. the same with a guard on `n.value`
3. two newtypes deep, both by guard and by seeding
4. and the negative: a guard on *another* `Name`'s `.value` does not discharge, so collapsing `.value`
   does not collapse two newtypes into one

`mvn -o clean test` green (1569). souther-lang/examples unchanged at four warnings.

The fix is merged forward into #229, #231, #232 and #233 so the stack stays consistent; each has a
merge commit and was re-tested on its own.

Issue #209, opened 2026-07-30
An attempted construction cannot say which of a type's invariant clauses failed


kawasima:
An attempted construction (`if T { … } as v else …`, `guard T(v) as x else …`, ADR-0070) runs the type's
invariant and takes one `else` value when it fails. Where a type carries more than one independently
meaningful rule, the attempt cannot say which one failed, so a behavior that needs to answer differently
per rule writes the rules again as `guard`s ahead of the construction.

This is about **classifying the failure**, not about proving the construction safe. The two are separate
and neither subsumes the other:

- #222 — is this construction safe? How many constructions can be judged so without an attempt.
- this issue — when it is not, which of the type's rules failed.

Discharging every clause statically would not remove the need to answer `DuplicateProduct` rather than
`NoLines`. Classifying failures would not make one more construction provably safe.

## Repro

```souther
module f24
data NoItems
data DuplicateItem
data Lines = List<Int>
    invariant List.length(value) >= 1 && List.allUniqueBy(x -> x, value)

behavior build : (xs: List<Int>) -> Lines | NoItems | DuplicateItem
    constructs Lines, NoItems, DuplicateItem

let build (xs) = {
    guard List.length(xs) >= 1 else NoItems
    guard List.allUniqueBy(x -> x, xs) else DuplicateItem
    Lines(xs)
}
```

Each guard is meaningful on its own — a caller needs "no items" apart from "a duplicate item" — and both
clauses of `Lines`'s own invariant are written a second time to get them.

## Cost in the model

`quoting` (souther-lang/examples) has `QuoteLines`, whose invariant says a quote has at least one line and
no product twice; `buildQuote` departs by `NoLines` for the first clause and `DuplicateProduct` for the
second.

Measured across all seven example models: 20 multi-clause invariants, of which 18 are ranges
(`value >= a && value <= b`) — one rule with one meaningful failure, which this issue does not concern.
Of the two that are genuinely independent rules, `crm`'s `WorkedLeadCommon` is reached only through a
decoder, where a violation is already a decode failure carrying a path. `QuoteLines` is the one site.

## Proposal to discuss

Name the clauses on the type; map them to departures at the attempt. The failure's *name* stays with the
caller, as ADR-0070 decided — what the type gains is an identifier per clause, not a business case.

```souther
data QuoteLines = List<QuoteLine>
    invariant nonEmpty: List.length(value) >= 1
    invariant uniqueProducts: List.allUniqueBy(.product, value)

[...]

Issue #83, opened 2026-07-25
Decoder errors: derive Raoh constraints from invariants instead of one `invariant_violation` code


kawasima:
## Summary

Every invariant violation a derived decoder reports carries the same error code. `CodecGen.emitConstructCall` emits `Result.fail(path, "invariant_violation", "invariant violated on <TypeName>")` for every data and every invariant, with no meta (`CodecGen.java:817-823`; the message text is built in `ValueClassGen.java:353`). Raoh's message layer is keyed by code, so the boundary can replace that one string but cannot tell one invariant from another, and cannot interpolate the bound that was broken.

Souther already knows the shape of the invariant at compile time, and Raoh already has a typed constraint API with a code and meta per constraint. The derivation should map one onto the other, the same way it already maps a data's shape onto Raoh's decoder construction (ADR-0004).

## What the boundary can do today, and what it cannot

Errors that come from Raoh's own combinators — a missing key, a type mismatch — carry Raoh's standard codes and meta, with `customMessage=false`, so `Issues.resolve(MessageResolver)` replaces their text and `resolve(resolver, locale)` localises them. Nothing in Souther gets in the way of that.

Invariant violations are the gap. `Result.fail(path, code, message)` builds the issue through `Issue.of`, so `customMessage` is `false` and a resolver *can* override the text — but a resolver only sees `(code, meta)`. The code is one shared constant and the meta is empty. The only discriminators that survive are the path, and the type name embedded in a message string that the resolver is about to replace. So "must be at least 1 character" for `Label` and "must be positive" for `Amount` cannot both be expressed.

There is also no surface syntax for attaching a code to an `invariant`.

## The vehicle already exists

`Ast.Require(Expr cond, String errorCode, SourcePos pos)` is defined as a decoder statement (`Ast.java:290`), `TypeChecker` checks its condition is `Bool` (`TypeChecker.java:1842`), and `CodecGen.emitRequire` lowers it to a Raoh failure with that code (`CodecGen.java:780`). Nothing constructs a `Require` — neither `AstBuilder` nor `Deriver` — so the path is unreachable. Deriving decoder constraints from invariants looks like the intent this node was cut for.

`emitRequire` as written needs two fixes when it is put to use: it reports at `Path.ROOT` rather than the value's path, and it passes the code as the message.

## The invariants that actually occur

All 18 invariants across `examples/`:

| invariant | occurrences | Raoh constraint | code |
|---|---|---|---|
[...]

PR #235, opened 2026-07-30
An invariant clause may be named, and what fails is one clause

kawasima:
Issue #209. An invariant was one boolean expression and its clauses had no identity, which cost the same thing in two places.

Inside the domain an attempted construction (ADR-0070) ran the whole invariant and took one `else`, so a type whose rules depart differently could not be attempted at all — `crm`'s `buildQuote` restated `QuoteLines`' two rules as guards to answer `NoLines` and `DuplicateProduct` apart. At the boundary a clause no Raoh constraint states fell into a single `refine` over the whole invariant, and a product data — outside that mapping altogether — reported `invariant_violation` with **no metadata**, so a `MessageResolver` could not tell which type had rejected the value.

## What this adds

**A clause may be named.** `invariant nonEmpty = List.length(value) >= 1`. Spelled with `=` because that is how everything defined as something is spelled (ADR-0026). `invariant <expr>` keeps its meaning exactly; no source in the corpus changed to compile. Two clauses that apply to one data cannot share a name, a `...` spread's included clauses among them.

**Clauses are checked in declaration order** and the first that does not hold is the one the failure names — so reordering a published type's clauses is a breaking change, and where several fail only the first is reported.

**An attempt departs per clause.**

```
guard QuoteLines(List.map(r -> toLine(r), lines)) as ls else
    | nonEmpty       -> NoLines
    | uniqueProducts -> DuplicateProduct
```

The arms are total over what can fail: every named clause has an arm, and `| _ -> …` is written exactly when the type has a clause carrying no name. Which arm is taken is the failing clause's to decide, so the order the arms are written in says nothing. `else <value>` keeps its meaning, so a site that does not care which rule broke is undisturbed.

**One failure value, three destinations.** `__construct`'s failure side becomes `souther.runtime.InvariantFailure` (the rejecting type and the clause), read by the attempt's arm, by the decoder's issue metadata (`{type, clause}`), and by the message an abort or a compile-time-rejected constant carries. A clause's name is never the issue's `code`: the code stays Raoh's, so a resolver keyed on that vocabulary keeps working.

**Raoh's collection constraints.** `List.length` bounds become `nonempty`/`minSize`/`maxSize`/`fixedSize`, `List.allUniqueBy(x -> x, value)` becomes `unique`, `Map.size` bounds the record decoder's. A `Set` cannot join them — Souther decodes one as a list and drops the duplicates while mapping it. Recognition reads the DISCHARGE representation, since `List.allUniqueBy` is a self-hosted helper and by emission it is the fold it derives from (#83 never met this: `String.length` is a builtin).

## One thing decided differently from the plan

Interleaving mapped constraints and refines in declaration order is not possible: Raoh's `refine` answers a plain `Decoder` and `ListDecoder.chain` is private, so a typed constraint cannot follow a refine in the chain. Hoisting the mapped clauses in front would mean a value breaking an earlier unmapped clause and a later mapped one is reported as the later one — the boundary and an attempt naming different rules for the same value. So the chain follows declaration order, and **a mapped clause declared behind an unmapped one trades its Raoh code for its place in the order**. Within one clause the conjuncts still map individually, so #83's existing behaviour is unchanged.

## Boundary changes worth knowing

- a product data's invariant failure gains `{type}` (and `{clause}` once named) where it carried nothing;
- a collection rule that was `invariant_violation` becomes Raoh's own code;
- a type mixing a mapped clause behind an unmapped one reports the unmapped one — declaration order, now;
- `__construct`'s published signature is `Result<T, InvariantFailure>` rather than `Result<T, String>`. Java reaches a type through its decoder, so the documented surface is unmoved.

## Verification

`mvn -o clean install` green: 1652 compiler tests + 98 LSP. New: `CompileNamedInvariantClauseTest` (18 — per-clause departure, arm order vs declaration order, spread-carried names, cross-module naming, value and tail position, E2014–E2019), 14 added to `CompileInvariantConstraintTest` (collection codes and metadata, clause in the metadata, the ordering rule both ways, a product data's issue), a formatter round trip, and a constant construction naming the clause it broke.

`souther-examples` builds and passes with `crm`'s `buildQuote` rewritten to the attempt: the last two guards restating a type's own invariant are gone, and finding F24 is retired.

Docs: ADR-0080, and `[#invariant-clause-name]` / `[#invariant-mvp]` / `[#attempt-departures]` / `[#decoder-error]` / `[#invariant-discharge-capability]` in the specification.


kawasima:
Review addressed. Both P1s were real; the P2 does not reproduce.

**`_` as a clause name (P1) — fixed.** Reproduced: `invariant _ = value > 0` was accepted at the declaration, and the arm reading `_` is the wildcard for the clauses that carry none, so the clause was unanswerable and every mapping site reported it missing (E2015). Now refused where the clause is declared, with its own diagnostic (`check.invariant.underscore`) and a hint, in both message catalogues. Test asserts the report and that it lands on the clause line, not at a use.

**The type was a simple name (P1) — fixed.** Reproduced: two modules each declaring `Id` gave `{clause=shaped, type=Id}` for both. `InvariantFailure` now holds the module and the name apart, and the metadata is `{module, type, clause}` — the structured form, `type` staying the name a reader knows so nothing keyed on it breaks. The abort message and the issue's default one name the type as Souther identifies it (`invariant violated on sales.Id: shaped`). Test decodes through both types and asserts the modules differ and the names do not.

Worth recording: `{type}` as a simple name came in with #83 and this PR only spread it wider (to a product data, which carried no metadata at all). Fixing it here rather than separately, since the metadata is what the PR is asking a resolver to switch on.

**Diamond spread (P2) — does not reproduce.** A diamond would reach one declaration twice and count its clause twice, but it cannot arise: a spread flattens the fields it takes in, so the second path brings a field the first already brought and E1004 refuses the declaration before any invariant is read.

```
data Base = { total: Int }
    invariant valid = total >= 1
data Left = { ...Base, l: Int }
data Right = { ...Base, r: Int }
data Combined = { ...Left, ...Right, c: Int }
```

→ `E1004: Field 'total' from '...Right' conflicts with a field of 'Combined'.`

So `checkClauseNames` cannot see the same clause twice today. It is an assumption rather than a proof, so it is pinned by a test (`aDiamondSpreadIsRefusedBeforeItsClauseIsCountedTwice`) — if the field-collision rule ever loosens, that test fails and the clause identity question you describe becomes real. Carrying an owner on `InvariantClause` for a case the language does not admit looked like the wrong trade now.

`mvn -o clean install` green: 1655 compiler tests + 98 LSP. `souther-examples` still green. ADR-0080 and the specification updated for both fixes.

PR #245, opened 2026-07-31
A benchmark is a module, and a fold is the loop it is

kawasima:
A benchmark that carries what it measures, and then the four things it found.

## What is measured

`souther-bench` is a module of its own, holding the sources it times: a CRM of four files, an issue tracker, and a handful of behaviors that differ in what they accumulate. It checks that they still compile before it times anything, and `CorpusTest` compiles them in CI so they cannot rot while nobody is looking. Six measurements — `cold`, `warm`, `phase`, `edit`, `scale`, `run` — and beside them the runtime's own structures, including the same counting written by hand on a `java.util.HashMap`, so a figure is read against what the work costs rather than in the air.

## What it found

**An edit costs what it reaches.** `Output.Examples` asked for every class of every module, so one added definition sent every example row in the workspace round again. It now asks for the modules it imports. A leaf edit re-runs 12 example rows where it re-ran 62, and costs 18ms where it cost 60 to 69.

**Example rows run beside each other.** Each already had a thread, for the timeout, and then waited for the row before it. Running them together needs the store left alone while they run, so a row reads through a thread-safe wrapper over the declarations and `check()` is split into prepare, start and read.

**A fold that only grows a collection builds it, as the loop it is.** `fold` is the one loop the language has, so what a fold costs is what a program's loops cost. `List.map` and `List.filter` wrote a new vector per element; the walk now carries a builder, two walks in a row are joined into one, and the walk itself is emitted where it stands instead of handing a first-class `Fn` to a shared method. A map accumulated the same way carries a transient.

**An upsert descends the trie once.** It read the key and then wrote it, and both descended.

## Numbers

Per element, on the runtime corpus (`souther-bench.jar run`):

| | before | after |
|---|---|---|
| a fold summing | 2.82 | **1.33** |
| a filter then a map | 12.18 | **2.93** |
| a fold counting into a map | 31.33 | **8.74** |

For scale: the same counting written by hand on a `java.util.HashMap` measures 4.07. What is left of the gap is a CHAMP descent counting bitmaps where a HashMap indexes an array.

Compile-side, on the CRM corpus: a warm compile 138.7 to 105.9ms, a leaf edit 59.5–68.6 to 17.7ms.

## What changes besides speed

Joining two walks changes when each step runs, not what either answers: two steps that can both abort now abort on a different element. Which abort comes first is not something a program may rest on (ADR-0029).

One bug fell out on the way, and is its own commit: `Map.get` never boxed its key, so an Int-keyed `Map.upsert` inside a fold emitted bytecode the verifier rejects. Every existing test keyed by String.

## Checks

Full reactor green (1700 + 98 + 40 + 1). The examples repository builds and its tests pass against this compiler.


kawasima:
Thank you — the P1 is right, and chasing it found something worse underneath it. The example-row parallelization is reverted; everything else stands.

## The parallelization deadlocks the compiler

The 6-hour CI cancellation is not a CI problem. The job hung in `CorpusTest`, which does nothing but compile. It reproduces locally under `-XX:ActiveProcessorCount=2`, and a thread dump says why:

```
"souther-example-eval" BLOCKED
    java.lang.ClassLoader.loadClass(ClassLoader.java:549)     <- waiting for another name's lock
    java.lang.ClassLoader.defineClass1(Native Method)         <- while inside defineClass
    souther.compiler.MemoryClassLoader.findClass(39)
    souther.compiler.ExampleVerifier.decoderFor(1525)
```

Two rows defining classes at the same time wait on each other. A generated class names others — an `$Impl` its interface, a decoder its value class, a sealed interface its cases — so a thread inside `defineClass` for one can block on the lock a second thread holds while it is inside `defineClass` for the other. Registering the loader as parallel-capable is what opens that path: per-name locks let two definitions run at once, and the JVM's deadlock avoidance covers the loader's own lock, not a cycle between two definitions on two threads.

It compounds into a hang rather than a failure. A thread blocked inside `defineClass1` is in a native frame and cannot unmount, so it keeps its carrier; with two carriers both are kept and no further row is ever scheduled. A row that has not started has spent none of its budget — by design — so the wait for it never ends. The dump shows two definitions blocked on each other and fourteen rows never scheduled.

## On the P1

Your reading of `awaitRow()` is correct and the fix you describe is the right one. The revert removes the code it applies to: the sequential path is `Future#get(timeout)` from the moment the row starts, so "each row finishes within the budget" holds again and the diagnostic's wording matches what it does.

Re-landing this needs rows that define no classes, which means defining the module's classes on one thread before any row starts. That is worth doing and is not a patch on this change. What it bought was 45.6ms to 37.3ms on the examples phase of a 106ms compile; the narrowing it came with is untouched and is where the edit-latency win actually is (a leaf edit 60-69ms to 18ms), and it is single-threaded.

## The other points

**Builders are now sealed.** Both `PersistentVector.Builder` and `PersistentHashMap.Builder` refuse to be written to after `build()`, and the vector's builder counts modifications as `AbstractList` promises. You put it exactly right: the safety rested on `GrowingFold`'s analysis rather than on the runtime type, and that is the wrong place for it now that a builder is reachable as a `List` and a `Map`. (Copilot raised the same thing.)

**Cancelling the remaining rows** is moot with the revert.

**`Timing.ofPrepared`** (Copilot): the doc claimed a set-up phase that does not exist. Renamed to `ofRounds` and the doc says what it does — a run is handed its round number because an edit must differ from the one before it, and everything the run does is timed.

## Checks

Full reactor green (1700 + 98 + 42 + 1). `CorpusTest` under `-XX:ActiveProcessorCount=2`, twelve runs, no hang — before the revert it hung on the first. CI is running on this head.

The benchmarks themselves are not run in CI and should not be: a shared runner says nothing reliable about how long anything takes. `CorpusTest` only compiles the corpora, which is what caught this.

Issue #242, opened 2026-07-31
A behavior answering "a value, or a reason there is none" cannot name a primitive or an imported type as the value


kawasima:
The commonest shape in a business model is "here is the amount / the date, or here is why there is none".
Its success member is almost always a primitive or a newtype some other module owns — money, a deadline, a
count — and an output union admits neither. So each such behavior declares a wrapper type for the value
alone, and the wrapper is not a domain concept.

## Repro

A primitive:

```souther
module m
data NoAnswer
behavior half : (n: Int) -> Int | NoAnswer
    constructs NoAnswer
let half (n) = if n >= 0 then n / 2 else NoAnswer
```

```
-- BOUNDARY TYPE ---------------------------------m.sou:3:29

3| behavior half : (n: Int) -> Int | NoAnswer
                               ^

Union members must be data types.
```

An imported newtype:

```souther
module up exposing ( Yen )
data Yen = Int
    invariant value >= 0
```

```souther
module down
import up ( Yen )
data NothingOwed
behavior owed : (a: Yen, b: Yen) -> Yen | NothingOwed
    constructs NothingOwed
let owed (a, b) = if a >= b then a - b else NothingOwed
```

```
-- IMPORTED CASE IN AN OUTPUT UNION  E1606 -----down.sou:4:1

4| behavior owed : (a: Yen, b: Yen) -> Yen | NothingOwed
   ^

The output union of `owed` includes `Yen`, a case declared in another module.
Hint: … Give `owed` a body instead of a composition: call the stages, open the imported result with
`match`, and answer `Yen` with a case this module declares.
```

`owed` already *is* a body rather than a composition, so the hint has nothing to offer here. This is
narrower than #207, which is about a `>->` carrying an imported **departed** case; here the imported type
is the mainline answer and the behavior is written out longhand.

## Cost in the model

Two sites in `hr`, one of each kind:

* `example.payroll`'s `computeTaxableAmount` wanted `-> Yen | DeductionsExceedGross`. It declares
  `data TaxableAmount = Yen` and returns that. The behavior also had to be **renamed** from
  `taxableAmount`, since a behavior is capitalised into a class of the same name as its own wrapper —
  so the workaround for one rule forced a worse name for another.
* `example.filing`'s `deadlineFor` wanted `-> Date | NoStatutoryDeadline` (the occasional revision is
  filed "promptly" and no provision names a day). It declares `data FilingDeadline = Date`.

[...]

PR #246, opened 2026-08-01
A union answers with a value, or a reason there is none

kawasima:
Closes the shape issue #242 raised: a behavior answering "the amount, or the reason there is none" could not name the amount. Its value is almost always a primitive or a type another bounded context owns, and neither could be a member of an output union, so every such behavior declared a type whose only job was to hold the value. `hr` has two — `payroll`'s `data TaxableAmount = Yen`, which then travelled downstream into `withholdingTaxOf`'s parameter and forced the behavior beside it to be renamed, and `filing`'s `data FilingDeadline = Date`.

## The two refusals were one

`successType` required every member to be a data type, which stopped the primitive. Behind it, E1606 stopped the imported type. The reason is the same in both: a case class receives the unions it belongs to when *its own* module is generated, and neither `java.lang.Long` nor a class an imported module already emitted can be given an interface from here.

ADR-0057 recorded that reason and listed **wrapper cases** among the alternatives, rejected "for now" and left available *if cross-context unions turn out to be worth that*. This is that measurement, and ADR-0081 takes them.

## Three rules

**`[#union-member]`** — a member is nominal, tells itself apart at run time, and has a `match` type-arm form. That admits the primitives and named data types, a data whether this module declares it or imports it. A collection fails the second: its type argument is erased, so `List<Order>` and `List<Item>` are one runtime type and no arm could choose between them; that there is also no arm form to write is the surface showing the same fact.

**Effective members** — a named sum contributes its leaves, since a leaf is what an arm names and what the `"type"` discriminator carries. Each effective member goes by a name of its own, so `up.Yen | other.Yen` is refused, and so is a local `Yen` beside an imported one, including where one arrives as a leaf.

**`[#bridge-case]`** — a data this module declares carries the result-union interface itself; anything else joins through `<MemberName>Case`, a record this module emits, one per member, implementing every result union of this module the member belongs to. That is the rule a local case class already follows; what differs is not what membership means but which module owns a class it can be put on.

A bridge case has no codec — belonging to a union does not change a member's external representation — and it is a JVM form only: not a Souther type, not written in a `match` arm or an `example` row, not in an external representation.

## Two boundaries

A Souther body holds only Souther values. The union's JVM form exists across `apply` and nowhere else, and the two meet at `ResultBoundary.inject` (a behavior's return) and `.project` (a call). Both walk the union's members, which are known where they are emitted, so neither asks a value whether it happens to be wrapped, and nothing is put into a bridge case twice.

Projection is at the call, not at the `match` arm that reads the result: a body may answer with a call's result without matching it, and the callee module's bridge cases are not members of this module's union — the `ClassCastException` ADR-0057 measured.

## What this does not change

E1606 stays for a declared sum's cases, and its reason moves from the JVM to the model. A named sum introduces a type name; an anonymous union introduces none.

Whole-program membership stays rejected and is not needed: a bridge case is the consumer's own class, so an imported module is not regenerated and resolving an import from a published jar stays available.

## Verification

The Java `switch` is checked by putting values through it, not by javac accepting it — the shape this replaces also compiled, and then threw at the first value.

`CompileOutputUnionMemberTest` (new) walks the member kinds, the three ways two members can come to share a name, one bridge case serving two behaviors, an `example` row, a member's codec, a Java implementation of an injected behavior, and a value answered by one module, opened by a second and answered again by a third — where a missed conversion arrives as the wrong class rather than merely nested. The four tests that expected E1606 now check that the value arrives and runs. `CrossProjectImportTest` covers the classpath-only import.

Two latent things fell out: `subtypeOf` accepted a `String` where `Int | DivisionByZero` was expected, because `namesOf` answered an empty set for a primitive; and an `if` whose branches are the value and the reason is settled by the output written for the behavior (`TypeOps.joinAt`) rather than by unioning two primitives unconditionally, which would have lost the error in `if flag then 1 else 1.5`.

`souther-examples` drops both wrappers on its own branch.


kawasima:
Fixed in 3036a65. Reproduced first: `data YenCase` beside an imported `Yen` member did write both to `down.YenCase`, the data won, `permits` still named that class, and the first value threw `NoSuchMethodError` — from legal source, as reported.

Took the explicit-diagnostic route. A bridge case is a name this module emits, so it is subject to the rule the other emitted names already are: no two of them may be one class. `IntCase`, `StringCase` and `DateCase` are names a model reaches for, so they are not reserved — the collision is decided by what the module holds, and a name nothing collides with stays available. Reported from the same place the behavior/data class collision already is, and before anything is written, so nothing lands under a name about to be taken.

Three ways to collide are covered, not two. Beside the data case and the primitive case, a behavior capitalizes into a class name too (`behavior intCase` against a member `Int`), and two types of one spelling from two modules collide when they are members of *two different unions of this module* — the member-name rule only refuses them within one union. Four regression tests, one per road.

`[#bridge-case]` and ADR-0081 both say this now.

The suffix point is right and is why the rule is stated over emitted names rather than over the suffix: moving to `$Bridge` would relocate the collision rather than remove it, and the ADR keeps `case YenCase(var y)` as what a Java consumer writes.

PR #251, opened 2026-08-01
What a value constructs belongs to its definition

### LINKED ISSUE #240: A limit written as a value costs every rule that reads it a construction authority
A threshold a rule is stated against is naturally a `let` value: it is written once, it carries the
comment saying where the figure comes from, and #197 made it publishable so several modules can read one.
But a value is elaborated where it is named, so *reading* the limit counts as *constructing* its type —
and a behavior that only compares against a limit has to declare `constructs` of it.

## Repro

```souther
module m

data Hours = Decimal invariant value >= 0.0m
data TooShort

let floorHours = Hours(20.0m)

behavior judge : (h: Hours) -> Hours | TooShort
    constructs TooShort
let judge (h) = {
    guard h >= floorHours else TooShort
    h
}
```

```
-- INSUFFICIENT CONSTRUCTION AUTHORITY  E1002 -----m.sou:6:1

6| behavior judge : (h: Hours) -> Hours | TooShort
   ^

Behavior `judge` constructs `Hours` but does not declare `constructs Hours`.
Hint: Add `constructs Hours` to `judge`, or stop constructing Hours here.
```

`judge` originates no `Hours`. It returns the one it was given. The only `Hours` built anywhere on that
path is the limit's own, which the declaration of `floorHours` already accounts for.

## Why it reads wrong

The asymmetry is with a **behavior**, not with reading:

> A behavior's `constructs` is its own and adds nothing to its caller's. A helper `let` is different — it
> is not a specification statement and declares no authority, so its construction set folds into the
> behavior that calls it. (§calling-a-behavior)

So a behavior that calls `judgeInsuredCategory` does not inherit its `GeneralInsured`, while a behavior
that names `floorHours` does inherit its `Hours`. Both are "reach a definition that builds something";
one is charged and one is not.

And the charge lands on the clause whose stated purpose is to tell those two apart:

> What `constructs` handles is *reading the model*: from the declaration, whether a behavior *creates* a
> new value or merely *passes through* an existing one. (§what-constructs-is-for)

`constructs Hours` on `judge` says it creates one. It does not. The clause that exists to make that
readable is made unreadable by the presence of a named limit.

## Cost in the model

`example.employmentinsurance` declares `weeklyHoursFloor = WeeklyScheduledHours(20.0m)` — the
20-hours-a-week coverage threshold of 雇用保険法第6条第1号, with the comment recording that the 2028
reduction to ten hours is legislated and not yet in force, which is the whole reason the figure is a value
rather than a literal in the comparison. `judgeInsuredStatus` therefore carries
`constructs … WeeklyScheduledHours` and a comment explaining that it does not mean what it says.

The alternative is to inline the literal at each comparison, which is what naming the limit was for.

## Proposal to discuss

Two shapes, either would do:

1. A value's construction set stays its declaring module's, as a *published* value's already does across
   modules — "publishing the definition is what states that origination" (§exposed-values). It is odd
   that the same definition costs a reader nothing from another module and a `constructs` entry from its
   own.
2. Failing that, a `constructs` entry that is satisfied by a value reference alone could be dropped from
   the E1006 over-declaration check, so at least the author can omit it.

Found while writing the `hr` example in souther-lang/souther-examples (recorded there as finding F31).

kawasima:
Fixes #240.

A threshold a rule is stated against is written once as a value — the figure gets one place to live
and one comment saying where it comes from. But a value is substituted at each reference, so the
construction in its body ended up standing in the reading behavior as the node that behavior's own
construction would be. `constructs` was counting what the substitution reached rather than what the
model says:

```souther
let floorHours = Hours(20.0m)

behavior judge : (h: Hours) -> Hours | TooShort
    constructs TooShort          -- E1002: `constructs Hours` was required here
let judge (h) = {
    guard h >= floorHours else TooShort
    h
}
```

`judge` originates no `Hours`. It returns the one it was given, and the only one built on that path
is the limit's own. The charge landed on the clause whose stated purpose is to tell those two apart.

## What the six cases measured

The asymmetry was with a behavior, not with reading — and the split was not the one the
specification describes.

| Case | Before | After |
| --- | --- | --- |
| Construct `Hours(20.0m)` in the behavior body | E1002 | E1002 |
| Read a private value of this module | E1002 | accepted |
| Read a value **this module publishes**, from within it | E1002 | accepted |
| Read a published value of another module | accepted | accepted |
| Call a helper that constructs `Hours` | E1002 | E1002 |
| Call a behavior that constructs `Hours` | accepted | accepted |

`[#exposed-values]` explains the cross-module row as publication stating the origination, but the
implementation branched on the module boundary: the same published value cost a reader nothing from
another module and an entry from its own. It now answers the same from either side.

A helper is unchanged and is the other case. Its body is checked as though written inline, which is
what tells a helper from a behavior.

## Provenance instead of the published-body mark

A construction now records where it came from rather than only whether it crossed a boundary, with
the two ways in kept apart because they are not answered alike. A published body hands over what its
module declares, so its mark names the module and is compared against the one declaring the type — a
third module's type stays the reader's to declare. A value hands over what its definition made
whatever module declares the type, so its mark names none.

Reusing the published-body mark was refused. It means crossed-a-boundary and not-originated-here at
once, and only the second is wanted; being compared against the *type's* module it would also still
charge a value of this module that builds an imported type, which is one of the cases the new tests
pin.

The entry stays legal if written. A construction a body was handed is carried, not absent — the
answer one arriving in a published body already gets. Making a value's stricter would leave two
answers to one question, decided opposite ways by which seam the construction came in on; whether
`carried` should be declarable at all is a question about that seam and belongs with it.

## Not an ADR

Reversing this costs a flag and it constrains nothing after it, so it does not meet the bar
`docs/adr/README.md` now states. The rule it changes is in `specification.adoc`
(`[#constructs]`, `[#fn-value-semantics]`, `[#exposed-values]`); the reason is here. The bullet in
ADR-0072 charging the naming behavior described the substitution mechanism rather than a chosen
design, and is withdrawn.

## Verification

`mvn clean install` then `mvn test`: 1757 compiler, 98 language server, 1 benchmark, all green.
`souther-examples` builds and tests green against this compiler.

Nine new cases in `CompileValueConstructionAuthorityTest`. Two existing tests stated the old rule
directly and are inverted rather than deleted — what they cover still happens, it is carried now
instead of originated.

Follow-up in `souther-examples` once this merges: `example.employmentinsurance` drops
`WeeklyScheduledHours` from `judgeInsuredStatus` and the comment explaining that the clause does not
mean what it says (recorded there as finding F31).

### DIFF
--- a/souther-compiler/src/main/java/souther/compiler/ast/Ast.java
+++ b/souther-compiler/src/main/java/souther/compiler/ast/Ast.java
@@ -1,6 +1,7 @@
 package souther.compiler.ast;
 
 import souther.compiler.diag.SourcePos;
+import souther.compiler.types.ConstructionOrigin;
 import souther.compiler.types.Type;
 import souther.compiler.types.TypeName;
 import souther.compiler.types.ValueName;
@@ -690,22 +691,26 @@ public interface Ast {
      * not go back to matching the spelling against the module's own definitions.
      */
     /**
-     * A construction. {@code publishedBy} is the module whose published value or helper carried it
-     * here, and null where the body being read wrote it.
+     * A construction. {@code origin} says where it came from: written here, or carried in by a
+     * published body or by a value this body named.
      *
-     * <p>Expansion makes the two look alike: a construction spliced in from another module's
-     * published body is the same node the reader's own would be, and the permission check reading
-     * that body would ask the reader to declare `constructs` for a type it may have no name for. So
-     * the construction says where it came from. Every rebuild of this node carries it — the component
-     * has no default, which is what stops a pass from quietly dropping it and turning a carried
-     * construction back into the reader's own.
+     * <p>Expansion makes the two look alike: a construction spliced in from another body is the same
+     * node the reader's own would be, and the permission check reading that body would ask the
+     * reader to answer for it. So the construction says where it came from. Every rebuild of this
+     * node carries it — the component has no default, which is what stops a pass from quietly
+     * dropping it and turning a carried construction back into the reader's own.
      */
-    record NewData(Name typeName, List<FieldInit> inits, List<ValueRef> spreads, String publishedBy,
-                   SourcePos pos) implements Expr {
+    record NewData(Name typeName, List<FieldInit> inits, List<ValueRef> spreads,
+                   ConstructionOrigin origin, SourcePos pos) implements Expr {
 
         /** The same construction, carried into a reader by {@code module}'s published body. */
         public NewData publishedBy(String module) {
-            return new NewData(typeName, inits, spreads, module, pos);
+            return new NewData(typeName, inits, spreads, origin.publishedIn(module), pos);
+        }
+
+        /** The same construction, carried into a body by a value that body named. */
+        public NewData carriedByValue() {
+            return new NewData(typeName, inits, spreads, origin.carriedByValue(), pos);
         }
     }
 
@@ -805,7 +810,7 @@ public interface Ast {
                 for (FieldInit i : nd.inits()) {
                     inits.add(new FieldInit(i.name(), f.apply(i.value()), i.pos()));
                 }
-                yield new NewData(nd.typeName(), inits, nd.spreads(), nd.publishedBy(), nd.pos());
+                yield new NewData(nd.typeName(), inits, nd.spreads(), nd.origin(), nd.pos());
             }
             case Match m -> {
                 List<Case> cases = new ArrayList<>();
--- a/souther-compiler/src/main/java/souther/compiler/ast/Ast.java
+++ b/souther-compiler/src/main/java/souther/compiler/ast/Ast.java
@@ -761,15 +761,23 @@ public interface Ast {
      * behavior, a function-typed parameter, or the type a newtype construction wraps — answered once
      * during resolution.
      */
-    record Call(String fn, ValueName denotes, List<Expr> args, SourcePos pos) implements Expr {
+    record Call(String fn, ValueName denotes, List<Expr> args, ConstructionOrigin origin,
+                SourcePos pos) implements Expr {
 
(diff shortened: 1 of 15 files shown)


kawasima:
Fixed the recursive-helper case in 0f7abe1, and pinned the unit-data one.

The review is right about the leak and about where it is. `carriedByValue` marked the nodes the
substitution leaves behind — a construction, and the name of a unit data — and a recursive helper is
lowered to a method rather than expanded, so what it builds stays behind an `Ast.Call` that had
nothing to say where it came from. `DataChecker` then absorbed its `recConstructs` as the reading
behavior's own.

Confirmed before fixing, with the non-recursive helper as the control:

| | before | after |
| --- | --- | --- |
| value reaching a **recursive** helper that builds `Hours` | E1002 | accepted |
| value reaching a **non-recursive** helper that builds `Hours` | accepted | accepted |
| the recursive helper called by the behavior itself | E1002 | E1002 |

`Ast.Call` carries a `ConstructionOrigin` now, and the set it stands for is absorbed as carried when
the call came in on a value. The mark is on the call rather than on the helper, so a behavior that
both names such a value and calls the helper itself answers for the second and not the first — the
third row is a test of its own.

Threading the origin through `Ast.Call` means every rebuild site passes it explicitly. The parser's
two-argument form still exists and means "written here"; there is no defaulted overload for the
rebuild sites, for the reason `NewData` has none — a pass that quietly dropped it would turn a
carried construction back into the reader's own.

On the second point: a unit data reached through a value was already handled, because
`ValueName.OfType` carries the same provenance and `carriedByValue` marks the name. Added
`aUnitDataNamedByAValueIsCarriedToo` to pin it against a regression in that path.

`mvn clean install` then `mvn test`: 1760 compiler, 98 language server, 1 benchmark, all green.
`souther-examples` green.

Issue #239, opened 2026-07-31
A match arm cannot say the combination it covers cannot arise


kawasima:
A statutory table indexed by two banded sums is written as a nested `match`, which is what makes adding a
band a compile error in the places that have to answer for it. Some such tables have blank cells, because
the two axes are not independent — and exhaustiveness asks for an arm for each blank with nothing to write
there that says the combination cannot arise.

## Repro

```souther
module m

data UnderThirty
data ThirtyOrOver
data AgeBand = UnderThirty | ThirtyOrOver

data UnderOneYear
data TwentyYearsOrMore
data ServiceBand = UnderOneYear | TwentyYearsOrMore

data Days = Int invariant value >= 90 && value <= 330

behavior daysFor : (age: AgeBand, service: ServiceBand) -> Days
    constructs Days

let daysFor (age, service) =
    match age with
        | UnderThirty ->
            match service with
                | UnderOneYear      -> Days(90)
                // Twenty insured years cannot have accrued before the age of thirty.
                // There is nothing to write here that says so.
                | TwentyYearsOrMore -> Days(90)
        | ThirtyOrOver ->
            match service with
                | UnderOneYear      -> Days(90)
                | TwentyYearsOrMore -> Days(240)
```

Omitting the arm is `E1201`:

```
-- NON-EXHAUSTIVE MATCH  E1201 ------------------m.sou:17:18

17|         | Old -> match s with
                     ^^^^^

This `match` on data `Service` does not cover every case.
Hint: Add a case for: Long
```

which is the right diagnostic for an arm that was forgotten and the wrong one for an arm that cannot be
reached. There is no `_` wildcard (by design), and no other form that stands for "unreachable".

## Cost in the model

`example.employmentinsurance`'s `prescribedBenefitDays` writes the statutory basic-allowance matrix: five
age bands by five insured-period bands, from 雇用保険法第23条. Two of the twenty-five cells are blank in
the published table:

* under thirty with twenty or more insured years — you cannot have accrued twenty years before thirty;
* an ordinary recipient with under a year — `judgeBenefitEntitlement` has already turned them away, since
  entitlement needs twelve months.

Both arms currently answer with the neighbouring cell and carry a comment saying the combination cannot
arise. That is a number a report will happily print, and a reader has no way to tell it from the
twenty-three that are real. The alternative — refactoring the two axes so the impossible pair is not
representable — would mean a bespoke sum per age band, which trades one honest table for five that have
to be kept in step.

[...]

PR #253, opened 2026-08-01
An arm may say the combination it covers cannot arise

kawasima:
Closes the gap issue #239 reports: a `match` arm had no way to say that the combination it covers does not arise, so the two blank cells of `example.employmentinsurance`'s basic-allowance matrix answered with the neighbouring cell — a number a report prints, indistinguishable from the twenty-three that are real.

## The form

```souther
| TwentyYearsOrMore ->
    unreachable "twenty insured years cannot precede the age of thirty"
```

* Typed `Never`, which fits every position and contributes no case to a join, so the `match` is typed by the arms that answer a value.
* Exhaustiveness counts the arm like any other. No arm may be omitted, and there is still no `_`.
* Reaching it aborts with `UnreachableReached` — separate from `ConstraintViolation`, because what failed is a premise the model wrote down rather than a value against its type's condition. Both reach a boundary as an abort a business failure is distinguished from.
* An `example` row that reaches it is **E1911**, carrying the reason and where it was written.
* The reason is a required string literal, so it is readable without running the model.
* Barred from an `invariant`, which answers whether the value being built holds on every path through itself. An `example` expectation is a value, so E1903 already refuses one there.

## It is a statement, not a proof

The compiler does not decide that the point cannot be reached, and could not: what rules out the pair is a fact about the world (twenty years cannot precede thirty) or a check the caller has already made (`judgeBenefitEntitlement` turned the recipient away), and neither is written in the types the rule sees. The specification and ADR-0083 say this in those words, because a name borrowed from Rust invites the other reading.

The two answers that need no new syntax were considered and are recorded in the ADR. Proving the arm away needs the relation to be readable somewhere and it is written nowhere; making the pair unrepresentable trades one table shaped like the published one for a sum per age band.

## E1307, found while implementing

Fitting every type is not being one. The code around the abort is still emitted and still reads a shape — a long for an `Int`, a reference for a data — so the position has to state what it holds, or a branch beside it has to. `let n = unreachable "..."` and `1 + unreachable "..."` emitted a class file nothing could load; they are now E1307, which is the rule `[]` already follows.

## Verification

`mvn clean install` green across every module, including 13 new cases covering a value-answering arm beside an unreachable one, a primitive position, tail position, an `if` branch, a binding, a total recursive helper, both refusals, and the two E1307 positions.

Not in this change, and recorded as separate work: the static match context in the abort message, a common supertype for the abort exceptions, coverage that tells declared-unreachable from uncovered, and discharging an arm from types and invariants. `souther-examples`' `hr` follows in its own PR.


kawasima:
Fixed in acaa0b7 — the synthesized file name is gone; the message carries the reason with `(line:column)`.

Two things the investigation turned up, both of which argue for dropping the name rather than deriving a better one:

* The compiler is never given a file name at all. `Descriptors.sourceFileName` derives a generated class's `SourceFile` attribute from its module name by the same rule, as a documented policy (ADR-0043), so there was no truer source to reach for.
* `pkg` is the module being emitted, not the module the expression was written in. An `unreachable` arriving with another module's inlined helper was being reported under the reading module's name against that other module's line — wrong in a second way the review's example does not cover.

A run-time reader has the frame's own file and line beside the message, and an E1911 reader has the row's place, so nothing is lost. `CompileUnreachableTest` now pins the message shape: the reason, `(22:21)`, and no `.sou`.

On `1 + unreachable "..."` being E1307 — agreed it is not the only defensible typing. It follows the `[]` rule the language already has, and if the operator-states-its-operand-type reading is the one we want, it is a change to that rule rather than to this form.

Issue #272, opened 2026-08-02
A lambda parameter does not take a function type from the collection it walks, so List.all over a list of functions is refused where List.fold is not


kawasima:
A list of functions folds and does not `all`:

```sou
let fs: List<(Int) -> Int> = [(x) -> x + 1]
List.fold((acc, f) -> f(acc), i.n, fs)             // compiles

let fs: List<(Int) -> Bool> = [(x) -> x > 0]
List.all((f) -> f(i.n), fs)                        // `f` is not a function here, so it cannot be
                                                   // applied to arguments
```

Both combinators are self-hosted helpers over the same element type:

```sou
// souther/list.sou:36
let all (p: ('a) -> Bool, xs: List<'a>) = List.fold((acc, x) -> acc && p(x), true, xs)
```

The element type is written on the binding, so `'a` has a function type to bind to; the lambda's parameter does not receive it. `List.fold`'s step binding is resolved by its own path (`Elaborator.resolveStepBinding`), which is why the two differ.

This is not new — the same program is refused identically on `develop` before #269's branch, measured. It surfaced while checking what ADR-0086 does and does not deliver, and it is the one position where "a function is a value like any other" is not true.

kawasima:
Fixed by #282.

`HelperInliner.rename` decided substitution differently by syntactic category for the same binding: a read of a parameter was rewritten whenever the binding was in the substitution, an application of one only when the parameter's *declared* type was a function. A lambda given to a function-typed parameter has untyped parameters, so its own parameter was bound by a fresh `let` while the application inside kept the lambda's original `BindingId` — `let f = x in f(n)` with two different `f`s. The report was the "no type yet" reading of `check.apply.notfunction`, not the one an author can act on.

A callee is now renamed as the subexpression it is. Applying a name is an `Ast.Var` held in `Apply.function`, so it goes through the same arm a read does and there is no separate rule for the callee position to fall out of step with.

A second diagnostic went with it: a lambda parameter shadowing an imported name was told to qualify a name that was a binding in force, because the untyped call fell through to a name-based stdlib fallback.

Issue #271, opened 2026-08-02
A behavior's name is a value in an argument and not in a binding, so one name still has two answers


kawasima:
A behavior's name handed to a combinator works, and reaches the behavior's class rather than the `let` behind it — the emitted `demo.Go$Impl` references `demo.Twice`:

```sou
behavior twice : (n: Int) -> Int
let twice (n) = n * 2

let go (i) = Out { ys = List.map(twice, i.xs) }        // compiles
```

Bound to a name it does not:

```sou
let go (i) = {
    let f = twice
    Out { ys = List.map(f, i.xs) }                     // `twice` is a behavior, and cannot be held
}                                                      // as a value here
```

One name, two positions, two answers — which is the shape ADR-0086 set out to remove, with one position left.

## Why the binding position is the one missing

A name written where a value goes is η-expanded by `HelperInliner.valueOf` (`check/HelperInliner.java:1255`), which handles a library name and a helper:

```java
private Ast.Expr valueOf(Ast.Var v) {
    if (v.denotes() instanceof ValueName.Stdlib lib) {
        return libraryValueOf(v, lib);
    }
    if (!(v.denotes() instanceof ValueName.Helper)) {
        return v;                                  // a behavior falls out here
    }
```

Expanding needs the callee's arity, and the inliner is built from the helper table (`HelperInliner.forHelpers`) — it does not hold the behavior signatures, so the arity is not in hand there. The argument position works because the combinator's function-typed parameter is elaborated where the signature *is* known.

`Elaborator.notAValue` now names what the denotation is rather than calling the name unknown, so the refusal is legible; it is still a refusal.

## What this asks for

Either the behavior signatures reach the point where a name becomes a function value, or the reification moves to where they already are. Whichever way, the rule is settled by ADR-0086 and by the specification (`[#blocks]`): a behavior's name handed over is the behavior. Only the binding position is unimplemented.

kawasima:
Closed by #286. A behavior's name bound to a `let` is the behavior, as it already was in an argument — one name, one answer, with no position left.

What was in the way was that reifying a name is written four times in the inliner, each getting its arity from a different table, and a behavior was the case none of them covered. They share one `etaExpand` now; `declarationArity` answers for a helper, a library function and a behavior alike.

The remainder of #277 is the representation question — an explicit function reference and the closure conversion `Core.Block`'s javadoc is waiting for — and that issue has been rewritten to hold only it.

Issue #270, opened 2026-08-02
A rounding mode is written at the call and never evaluated, so three Decimal operations are special forms rather than functions


kawasima:
`RoundingMode` is the last argument in Souther that is not a value. It is written at the call and read as the identifier it is written as, never evaluated, so three operations are special forms rather than functions:

```sou
let round (d: Decimal, scale: Int, mode: RoundingMode) : Decimal = intrinsic "decimal.round"
let toInt (d: Decimal, mode: RoundingMode) : Int = intrinsic "decimal.toInt"
let divide (a: Decimal, b: Decimal, scale: Int, mode: RoundingMode) : Decimal | DivisionByZero
```

`RoundingMode` is a type only core may name, `requireRoundingMode` (`check/CallElaborator.java:505`) checks the argument is one of seven identifiers, and the backend reads a JVM constant from it:

```java
// codegen/BodyGen.java:1079
void emitRoundingMode(Core arg) {
    code.getstatic(CD_RoundingMode, ((Core.Builtin) arg).name(), CD_RoundingMode);
}
```

## What that costs, measured

Each of these is refused today, and each is ordinary in every other position:

```sou
let f = Decimal.toInt                                 // `Decimal.toInt` takes a rounding mode, which
                                                      // is written at the call and not passed as a
                                                      // value, so it cannot be handed over by name
Decimal.toInt(d, if up then HALF_UP else HALF_DOWN)   // the rounding mode of `Decimal.toInt` must be
                                                      // one of HALF_UP, HALF_EVEN, ...
let mode = HALF_UP                                    // `HALF_UP` is written at the position that
Decimal.toInt(d, mode)                                // reads it, not evaluated
```

So a model cannot state a rounding policy once and pass it, cannot choose one from a business rule, and cannot hand the operation to a combinator — three restrictions with one cause. It is also the only exception left in ADR-0086's rule that a declared function is a value: every other standard-library function can be handed over by name.

## What it would take

Declare it as ordinary data in a core module:

```sou
data RoundingMode = HALF_UP | HALF_EVEN | HALF_DOWN | UP | DOWN | CEILING | FLOOR
```

1. **A prelude module must be able to declare data.** `Prelude.load` passes `Symbols.none()` and reads only `module.fns()` (`Prelude.java:143`, `:154`), because prelude signatures use primitives and type variables and nothing else.
[...]

PR #273, opened 2026-08-02
What a call applies is an expression, so a function name is a value

### LINKED ISSUE #261: The empty Map is written `Map.empty()`, not `Map.empty` as the spec says, and the rejection calls the qualifier an unknown local
The `Map` section says: "`empty` is the empty map: like the empty list `[]`, its key and value types are fixed by context … so it is a built-in **taking no argument rather than a call**." Written that way it does not compile, and what comes back names the qualifier as an unknown local variable.

## Repro

```
data Out = { m: Map<String, Int> }
data Req = { n: Int }
behavior go : (r: Req) -> Out
    constructs Out
let go (r) = Out { m = Map.empty }
```

```
-- NAMING ERROR ---------------------------------empty.sou:5:24

5| let go (r) = Out { m = Map.empty }
                          ^^^

I cannot find a value named `Map` in scope.
Hint: did you mean `p`?
```

The caret is under `Map`, the message calls it a value that is not in scope, and the hint offers a local binding as the correction. `Map` is a reserved standard-library qualifier and the compiler knows it.

## What is accepted

| Written | Result |
| --- | --- |
| `Map.empty` | NAMING ERROR above |
| `Map.empty()` | compiles |
| `import Map ( empty )` then `empty` | NAMING ERROR |
| `import Map ( empty )` then `empty()` | compiles |

So `empty` is reached as a call after all. Whichever spelling is meant to be the right one, the two do not agree today.

## Cost in the model

The empty map is the seed of every accumulating fold, which is how the `Map` section itself says a count or a grouping is built (`upsert(k, 1, n -> n + 1, m)`). It is among the first things a model reaches for, and the message sends the reader looking for a misspelled parameter.

## Proposal to discuss

Either accept `Map.empty` as the spec writes it, or change the sentence and keep `Map.empty()`. Either way the rejection of the other spelling should say what it is — a standard-library name reached the wrong way — rather than treating `Map` as a local that might be `p`. The same reading applies to any qualified standard-library name a model spells without its argument list.

Found while writing an allocation model against souther-lang/examples. Verified against develop at 248eba1 with the repro above.

kawasima:
Issue #261 reported that the specification writes `Map.empty` and the compiler rejects it, and that the rejection names `Map` as an unknown local and suggests a parameter called `p`. Both symptoms come from one shape:

```java
record Call(String fn, ValueName denotes, List<Expr> args, ConstructionOrigin origin, SourcePos pos)
```

What is applied was not a subexpression. It was a string beside the node, so "what does this name mean" was asked by two ladders that disagreed, and the parser chose between them by looking three tokens ahead for a `(` — which is why `Map.empty` had no path to the library and `Map.empty()` did.

**`Ast.Apply(Expr function, …)` replaces `Ast.Call`.** A name that names a function is a value of function type, and applying something is an operation on the value it evaluates to.

## What this changes for a model

```sou
Out { m = Map.empty }                                    // the issue's repro
List.map(String.trim, xs)                                // a library function handed over by name
let step: (String) -> String = if flag then String.trim else String.uppercase
let steps: List<(String) -> String> = [String.trim, String.uppercase]
adder(1)(x)                                              // postfix, iterated
(if flag then String.trim else String.uppercase)(s)      // applied to any expression
```

Every one of these is refused on `develop`, measured.

`Map.empty()` is now `check.apply.notfunction` — a value applied, which is an ordinary type error and needed no new diagnostic key.

## What was underneath

**One ladder answers a name, wherever it is written.** The position contributes one bit — whether the name is applied — which decides only whether a type written as a value is a unit data's construction or a newtype taking what it wraps.

**An import contributes a table, not a tree rewrite.** `Exposing` rewrote bare imported names by walking the tree with a `default -> e` catch-all: it missed `Ast.Var`, missed `Ast.IfConstructed` entirely, never visited `examples` or `fakes`, and ran before `Resolve` so it knew no bindings — an `import List ( map )` beat a parameter named `map`. The walk is gone. Resolution consults the table last, after the bindings in force and the module's own declarations.

**One spelling reaches the value namespace once**, whichever way it arrives. An import that collides with a declaration is refused at the import line, and a type may not be named like a library qualifier. This had to land before the two ladders could merge: without it, merging would have silently changed what compiling programs meant.

**`Prelude.BUILTINS` is gone** — deleted, not emptied, so there is nowhere for a signature to be written in Java again. Every standard-library function and value states its own signature in a core declaration. What the type language cannot say — an ordering constraint, "this argument is a rounding mode" — is a side condition beside a declared signature, never in place of one. A two-way test holds it: every kernel the backend can emit has a declaration, and every declaration has an emitter.

`Int.divide` and `Decimal.divide` shared one arm of one switch and were told apart only by argument count, so they were interchangeable aliases — `Int.divide(a, b, 2, HALF_UP)` type-checked as the Decimal one. They are separate declarations, and the specification's sentence that `Decimal.divide(a, b)` cannot be written is true.

## Things the work overturned

**An argument list may not cross a line break.** The plan said the opposite: Souther's grammar has no newline token, so reading a next-line `(` as an application looked like the choice that kept postfix application from being an exception. It stopped the standard library loading — `list.sou` has a block whose result is a tuple on the line after a call.

**A behavior's name handed over is the behavior.** The plan had it as a capability leak to be refused. The emitted `demo.Go$Impl` references `demo.Twice`, so `List.map(twice, xs)` already goes through the behavior's class rather than the `let` behind it: nothing leaks, and a Java implementation still replaces what the caller got. The specification says this instead of the opposite. The binding position is not implemented yet (#271).

**The storage limit is not about recursion.** A plain lambda and a library name are refused identically when stored, so the rule is about a function whose type is unknown where it is stored.

## Two defects found in this branch's own work, and fixed

**The formatter reassembled a callee from the identifier tokens under the node**, which only works while a callee is a name: it printed `adder(1)(x)` as `adder(1, x)` and a parenthesised choice as nothing. Nothing in the corpus applies anything but a name, so its suite was green while it was wrong.

**Resolution wrote the library name it answered with over the spelling the source had**, because sixty-odd readers key their tables by the spelling. That put two questions back into one field: a report on an imported `trim` quoted `String.trim` and underlined eleven characters where four were written, and the width is what the editor draws. An application answers both now — `written()` is what a report quotes, `reaches()` is the name of the declaration it reaches. `HelperInliner.expands` already said its table is keyed by the denotation and never by the text; it was being handed the text.

A name that was answered is also no longer reported as unknown. A behavior's name and a rounding mode both resolve and are both refused for a stated reason, and `unknown identifier` sent the reader looking for a spelling mistake.

## Specification and ADR

Spec: the `Map.empty` spelling, postfix application and the line-break rule, named functions as values, the four boundaries that require an external representation, what an import brings in, and the value-namespace rules. Two existing statements were narrower than the implementation and are corrected — "a function type belongs only in a helper's parameter position", which contradicted `[#delimiters]` and `[#fn-rules]`, and the list of what stays integrated in the compiler.

ADR-0086 records the decision and what the work overturned.

Filed rather than fixed here: #270 (a rounding mode is not a value), #271 (a behavior's name in a binding), #272 (a lambda parameter over a collection of functions — pre-existing, measured on `develop`).

Closes #261 after merge.


kawasima:
All four are confirmed by measurement, fixed, and pinned. Sweeping the rest of the diff for the same shape found five more.

**[P1] `rename` rebuilds an application from a name** — confirmed, and worse than reported. It does not lose the callee quietly: the empty spelling reaches the check and throws ```` reached the check unresolved, at 4:43`. The same application written directly in a behavior compiles, so this was exactly the reported split. `rename` now has an arm for a callee that is not a name, renaming it as the expression it is. Pinned in `CompilePostfixApplicationTest`.

**[P1] A helper parameter is not typed through a bare import** — confirmed. `let size (s) = length(s)` under `import String ( length )` reports `parameter `s` is not determined by its body`; written `String.length(s)` it compiles. `calleeParams(call.reaches())`.

**[P2] An invariant's term key** — confirmed, and observable rather than only suspected. The bare form emits **E2011** ("the guards do not establish the invariant, so the check remains at run time") where the qualified form emits nothing, so the invariant was silently demoted to a run-time check. `Prelude.hasQualified(c.reaches())` and `c.reaches() + "("`. Pinned by asserting zero warnings, not by asserting it compiles — compiling was never the difference.

**[P2] An import colliding with a data** — confirmed, and it fails worse than "silently ineffective": the import goes dead and the name can then not be called at all (``map` is a standard-library function and must be called qualified`). `Exposing.validate` reads `module.defs()` and refuses at the import line; an own `let` still shadows.

**The sweep you asked for found five more**, all the same shape — a qualified library name compared against a spelling:

- `InvariantConstraints.sizeBound("List.length", …)` and `("Map.size", …)` — the size bounds an invariant lowers to
- `ConstEval` — its switch read `case "length", "String.length"`, so a bare name was folded as `String.length` whether or not it reached it. That also folded a *user helper* named `length`. The bare arms are gone with the fix.
- `TotalityChecker` (×2) and `HelperTyping` (×3) and `DataChecker` — the recursive-helper call graph, over a table that is keyed qualified for the prelude
- `CallElaborator` / `ExampleVerifier` — the temporal constructors

The rule is now mechanical, as you put it: `written()` is what a report quotes and underlines, `reaches()` is what a declaration table is keyed by.

Both documentation points taken. `Exposing`'s class comment described the rewrite it no longer does — the responsibility inverted and the comment did not. The ADR index summary is now "a behavior's name handed over is the behavior rather than the `let` behind it — binding one to a local `let` is the single position still unimplemented", which is what the ADR body says and what #271 tracks.

Full suite green after the fixes. Nothing was accepted without a probe first: three of the four reproduced exactly as described, and the invariant one needed a sharper measurement than "does it compile" before it showed.

kawasima:
Both documentation blockers are fixed, the small cleanups are in, and the four larger simplifications are filed. Note the head had moved past 6f4cfdc when you reviewed: **your item 5 (Exposing analysing twice) was already done** in 005b429d — one `read()` answers the table and the kept imports together, and `exposedNames` is gone with its last caller.

**[P2] ADR polymorphic alias** — you are right, and it was wrong in a specific way worth recording. I measured before rewriting: `let r = List.reverse` applied to a `List<Int>` and a `List<String>`, and `let e = Map.empty` read at two key and value types, all compile. Nothing was added for that. A name bound to a declaration *is* the declaration, expanded at each use — the same `scopedLambdas` path a helper bound to a name already took — so a polymorphic one instantiates afresh per use. The ADR now says that, and says what let-polymorphism would actually add (a binding that generalises without being expanded).

**[P2] `CompileFunctionBindingTest`** — taken, including the test name. The limit is a function whose type nothing settled, and `aRecursiveHelperIsBoundAndAppliedButIsNotStored` named the two things that are *not* the cause. It is `aFunctionWhoseTypeNothingSettledCannotBeStored`.

**`theAppliedExpressionIsWorkedOutOnceAndFirst`** — the sharpest catch, because the test passes. Arithmetic cannot show which side ran first, so the name claimed what the assertion could not. Split: the runtime test is `anApplicationOfAComputedFunctionWorks`, and `theCalleeExpressionStandsOnce` asks the settled body and counts — the callee stands once, which is the half that can be looked at. I did not invent a proof of ordering; the invariant holds by construction and no test here says so.

**Small cleanups**, all in: dead `primitiveHeaded`; dead `Apply.denoting` (you called it — zero callers, and it cast to `Var` without the type saying so); `libraryValue`'s unused `Scope`; the doubled `libraryValue` call — the arm answers now rather than a guard and the arm both asking; `ORDERED_ELEMENT` is an enum saying whether the ordered element is the list's or the option's; both misplaced javadocs.

**Filed rather than done here**, with the reasoning you gave:

- #274 CST `CALL_EXPR`/`APPLY_EXPR` — and your caveat is the reason it is separate: unifying them makes the qualified form a `FIELD_ACCESS` chain in both cases, so what the parser hands the resolver moves.
- #275 the prelude's three maps, and `hasQualified` answering from `HELPERS || INTRINSICS || SUGARED` rather than the declarations.
- #276 the two copies of the four-step signature application. Recorded that the duplication already cost once — the missing third step is what blocked every kernel taking a function.
- #277 reifying a named function where the signatures already are. #271 closes with it.

On your item 2, the `Apply` name API: I went a different way and should say so rather than let it look overlooked. `namedCallee()` is gone — it had one caller and asked what a pattern match on `function()` answers directly. `written()` and `reaches()` stayed on `Apply`: routing ~160 sites through `namedCallee().map(Var::reaches).orElse(…)` costs more than the empty spelling does, and the emptiness is no longer a convention — `reaches()` is empty exactly when `denotes()` is not a library name, which is what applying something that is not a name leaves. If you would still rather have it on `Var` only, say so and I will move it.

Full suite green.

PR #278, opened 2026-08-02
Refuse a value too deep for the boundary with a path, not Jackson's words

### LINKED ISSUE #262: A JSON nesting-depth rejection reports Jackson's internals instead of a decode failure with a path
Self-referential data is allowed and its codecs follow the self-reference, so a tree arrives from outside as nested JSON. Past a certain depth the input is refused by Jackson, and the refusal is reported in Jackson's words.

## Repro

```
data Node = { n: Int, kids: List<Node> }
data Out = { depth: Int }
let depth (t: Node): Int = 1 + List.fold((acc, k) -> Int.max(acc, depth(k)), 0, t.kids)
behavior go : (t: Node) -> Out
    constructs Out
let go (t) = Out { depth = depth(t) }
```

250 levels of `{"n": …, "kids": [ … ]}` answer `{"depth":250}`. 251 give:

```
--input is not valid JSON: Document nesting depth (501) exceeds the maximum allowed (500, from `StreamReadConstraints.getMaxNestingDepth()`)
```

The ceiling is Jackson's 500, and this shape spends two levels per level of tree, so it lands on 250.

## Cost in the model

Not the limit — 250 levels of a bill of materials or a category tree is deep enough for most domains, and a limit is fine. It is that `StreamReadConstraints.getMaxNestingDepth()` is not part of the vocabulary a model author works in, and the message does not say the thing that would help: which path was too deep, and that the limit counts JSON nesting rather than levels of the data.

Everything else the decoder rejects is reported in the boundary's own terms with a path — `/bins/A0: must be exactly 3 characters`, `/rule/type: must be one of [...]`. This one is not.

## Proposal to discuss

Report a nesting overflow as a decode failure like the others: a Raoh issue with a path and a code, saying the input nests deeper than the boundary accepts. Whether the limit itself should be configurable, or raised now that self-referential data is a supported shape, is a separate question worth settling at the same time.

Found while checking how deep a self-referential data can arrive from outside, against souther-lang/examples. Verified against develop at 248eba1 with the repro above.

kawasima:
Closes #262 once merged (a PR onto `develop` does not fire the keyword; the issue is closed by hand).

Everything else the decoder rejects is named at its JSON pointer — `/bins/A0: must be exactly 3 characters`. A refusal for depth was the one that read `StreamReadConstraints.getMaxNestingDepth()`, which is not a name a model author works in, and said nothing about which branch went deep.

## Three doors, not one

The issue names the reading side. Measuring found two more, and fixing only the first moves the reader to a worse report at the second:

| | before | after |
|---|---|---|
| reading past 500 levels | `Document nesting depth (501) exceeds the maximum allowed (500, from \`StreamReadConstraints.getMaxNestingDepth()\`)` | `--input nests deeper than this boundary accepts: /kids/0/kids/0/kids/0/kids/0/… and below (JSON nesting past 500 levels)` |
| writing past 500 levels | the same words, plus a few hundred `java.util.ImmutableCollections$ListN[0]` of reference chain | `the output nests deeper than this boundary can render: …` |
| encoding deeper still | a bare `StackOverflowError` stack trace | `the output nests too deeply for this boundary to encode.` |

A behavior can build a value deeper than anything it was handed — `List.fold((acc, x) -> Node { kids = [acc] }, …)` over a flat list — so the second and third are reached from input the boundary accepted.

## How the path is got

The exception a depth refusal throws carries neither a location nor the parser that raised it, so the parser and the generator are made in `Runner` rather than left to `readTree(String)` and `writeValueAsString`. Holding them is what gives `streamReadContext().pathAsPointer()`. The generator's context is read before it is closed — closing resets it to the root, where the parser's survives.

The path is cut after eight segments: nesting that deep is nesting that repeats, so the whole pointer runs to hundreds of segments of the same few names, and the head is the part that says which branch went deep.

No depth is claimed, only the limit. Reading and writing both stop at the level that crosses it, so how much deeper the value goes was never counted; reporting the context's own count read `500 levels, the limit is 500` on the writing side.

## The limit itself

It stays at 500 and is not made settable. Raising it was measured: 1000 levels of tree answer, and 3000 die inside the codecs with no diagnostic at all. Five hundred does not guard a real limit, but what sits behind it is reported worse than what sits in front, so raising it is a question about how deep the codecs can recurse, not about this.

## Also

Reading through a parser answers no node at all where reading a string answered the missing node, which a multi-input `--input` then walked into. Empty input now reaches the same refusal as input of the wrong shape.

## Testing

Four tests in `RunnerTest` covering all three doors and the empty input; each also asserts the Jackson class name never reaches the reader. Full suite green (2106).

### DIFF
--- a/souther-compiler/src/main/java/souther/compiler/Runner.java
+++ b/souther-compiler/src/main/java/souther/compiler/Runner.java
@@ -585,6 +585,14 @@ public final class Runner {
             StreamWriteConstraints.defaults().getMaxNestingDepth();
 
     /**
+     * The depth a context stands at when its side has just refused a value for nesting. The two
+     * sides count differently, measured: reading stands on the level it refused to enter, and
+     * writing on the last level it accepted.
+     */
+    private static final int READ_REFUSAL_DEPTH = READ_DEPTH_LIMIT + 1;
+    private static final int WRITE_REFUSAL_DEPTH = WRITE_DEPTH_LIMIT;
+
+    /**
      * Parses {@code --input}.
      *
      * <p>The parser is made here rather than left to {@code readTree(String)} because the exception a
@@ -599,14 +607,17 @@ public final class Runner {
         JsonParser parser = JSON.createParser(input);
         try {
             JsonNode tree = JSON.readTree(parser);
-            parser.close();
             return tree == null ? JSON.missingNode() : tree;
-        } catch (StreamConstraintsException _) {
-            throw tooDeep("run.input.toodeep", "--input nests deeper than this boundary accepts",
-                    parser.streamReadContext(), READ_DEPTH_LIMIT);
         } catch (Exception e) {
+            TokenStreamContext where = parser.streamReadContext();
+            if (refusedForDepth(e, where, READ_REFUSAL_DEPTH)) {
+                throw tooDeep("run.input.toodeep", "--input nests deeper than this boundary accepts",
+                        where, READ_DEPTH_LIMIT);
+            }
             throw fail("run.input.badjson", "--input is not valid JSON: " + e.getMessage(),
                     e.getMessage());
+        } finally {
+            parser.close();
         }
     }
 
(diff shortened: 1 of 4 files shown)


kawasima:
## [P1] `StreamConstraintsException == nesting depth` — fixed in e0472148

You are right, and the measurement confirms it. On Jackson 3.2.1 the reader holds six limits behind the one exception type:

```
nesting=500 docLen=-1 numLen=1000 strLen=100000000 nameLen=50000 tokenCount=-1
```

A 1001-digit number was reported as `--input nests deeper than this boundary accepts: (root) and below (JSON nesting past 500 levels)`. That is a refusal that meant one thing reported as another, which is worse than the leak the PR set out to close.

### The discriminator

Not the message text — that is prose Jackson is free to rewrite, and the number in it is not the number the message needs. It is the context's own depth. Measuring where each side stands when it throws:

| input | exception | context depth |
|---|---|---|
| 501 levels of nesting | `StreamConstraintsException` | **501** |
| 1001-digit number at level 500 | `StreamConstraintsException` | 500 |
| 1001-digit number at level 499 | `StreamConstraintsException` | 499 |
| 50001-char property name | `StreamConstraintsException` | 1 |

A refusal for depth stands one level past the limit; a refusal for any other reason stands at the limit or shallower, because the level that crosses the limit is the deepest one ever reached. So `depth >= limit + 1` is exact rather than heuristic.

The writing side counts differently — it stands on the last level it **accepted**, at 500, not 501 — which the first attempt at this got wrong and the suite caught. The two refusal depths are now named separately with that reason written down.

Everything that is not the depth refusal falls through to `run.input.badjson` / `run.output.json` with exactly the wording it had before this PR. Giving the other five limits words of their own is a different change from #262, and not one I have made here.

### The test

`aLimitOtherThanDepthIsNotReportedAsDepth` writes the over-long number at the **500th** level — the deepest a non-depth refusal can reach, one short of where the depth refusal stands. That is the sharp case: reading the depth as *at* the limit rather than *past* it misclassifies it.

I checked the test actually pins the comparison rather than merely passing: loosening `READ_DEPTH_LIMIT + 1` to `READ_DEPTH_LIMIT` makes it fail with `--input nests deeper than this boundary accepts: /0/0/0/0/0/0/0/0/…`. A first version of the test nested the number 249 tree levels deep, which only reaches depth 499 — it passed under both comparisons and proved nothing. Replaced.

Verified on the built CLI:

```
$ souther run tree.sou --input @251-levels
--input はこの境界が受け取れる深さを超えています: /kids/0/kids/0/kids/0/kids/0/… 以下（JSON の入れ子が 500 レベルを超えました）

$ souther run tree.sou --input @1001-digit-number-at-level-500
--input が正しい JSON ではありません: Number value length (1001) exceeds the maximum allowed (1000, …)

$ souther run tree.sou --input @50001-char-name
--input が正しい JSON ではありません: Name length (50001) exceeds the maximum allowed (50000, …)
```

## `close()` on the exception path — fixed in the same commit

Both are closed in `finally` now. You are right that it is ownership rather than a leak here. The generator's context is read inside the `catch`, before the close resets it to the root — that reset is why it was written this way, and it is now in a comment rather than implied.

## `StackOverflowError` breadth — left as is

Agreed it is the defensible reading, and thank you for saying why. The encoder is structural recursion over a value the type checker already accepted, and Souther has no way to build a cyclic value, so the only depth that reaches it is the value's own.

Full suite green: 2107.

PR #284, opened 2026-08-02
Make a value reference one node, and reach it with the walk

### LINKED ISSUE #267: A spread names a value where the one walk over an expression cannot reach it, so thirteen passes each remember spreads on their own
`Ast.mapChildren` and `Ast.forEachChild` are how a pass reaches what an expression is made of, and every pass that walks an expression delegates to them. A spread is not reachable that way. `NewData.spreads()` holds `Ast.ValueRef` — a name, not an expression — so the walk goes past it, and each pass that cares has to remember on its own.

`HelperInliner` says so where it does:

```java
// check/HelperInliner.java:517
// a spread names a value the way any other position does, and `mapChildren` does not
// reach it — it holds a name rather than an expression
case Ast.NewData nd -> {
    List<Ast.ValueRef> spreads = new ArrayList<>();
    ...
```

Thirteen files reach `spreads()` for themselves, across 27 references: `Resolve`, `NewtypeDesugar`, `HelperInliner`, `DataChecker`, `Exposing`, `InvariantChecker`, `Elaborator`, `CallElaborator`, `ExampleVerifier`, `CodecGen`, `BodyGen`, `GrowingFold`, `Core`.

## What it has cost

Two commits in the last stretch of history exist only because a pass did not:

```
5aa8060 Read a value wherever one is named: a spread, and a cycle
5060888 Carry what a spread resolves to, so a binding still wins
```

#154's NPE was the same thing seen from the other side — `fieldTypes` was the only gate that looked at a declaration's spread, so nothing else did. In `souther-lang/examples`, a fixture built by spreading another turned out to take its case from the one it spread, which is again a pass that read the fields and not the name.

The pattern is that a spread is a name in a position, and the language's other name-bearing positions — a `Var`, a `Call`'s target — are reached by the one walk. Whether a rewrite covers spreads is decided per pass, by whoever wrote it, and the answer is not visible where the walk is.

## Why this is the same defect as #260

#260 was a rule the language states — sameness — decided in more than one place, so the places drifted. This is a question the language asks — where does a name occur — answered in thirteen places, so they drift the same way. In both, adding the next reader is what goes wrong, and nothing says so at the point of writing it.

The difference is that `Ast` already solved the neighbouring version of this. `forEachChild` is exhaustive over sealed `Ast.Expr`, so a node kind added later stops the build. A name-bearing position added later does not.

## Proposal

Give `Ast` the counterpart of `mapChildren` for names: one place that knows every position in an expression where a value is named — a `Var`, a `Call`'s target, a spread — and let the passes that rewrite or collect names delegate to it, as they already delegate the children.

Then a fourteenth pass gets spreads by writing the same delegation the other thirteen do, rather than by knowing that spreads exist.

Two things to settle while doing it:

- Whether the same walk should reach names in a declaration (`Ast.Data`'s own spreads), or whether that stays separate. #154 says a declaration's spread and an expression's spread are not quite the same position — a declaration's may not be qualified — so they may want two walks with one shape rather than one walk.
- Whether `Core` needs the same thing. `GrowingFold` and `Core` are on the list above, so the question does not stop at the surface.

kawasima:
Closes #267.

`Ast.mapChildren` and `Ast.forEachChild` are how a pass reaches what an expression is made of. A spread was in neither, so thirteen files reached into `NewData.spreads()` for themselves, and whether a pass covered spreads was decided by whoever wrote it.

## What was actually wrong

Two things, and the issue names the second one.

`Ast.ValueRef` and `Ast.Var` were the same record under two names — the same spelling, the same `denotes`, the same `local` and `denoting`; `ValueRef`'s own javadoc called itself the counterpart of `Var#local`. Only `Var` was an `Expr`, so only `Var` was reachable, and a spread held the other one.

`mapChildren` and `forEachChild` each enumerated the children of a node in a switch of their own, so a slot had to be added to both.

## What this does

`Ast.Var` is now the one representation of a value reference. A construction's spreads, a spec's `depends on` and a pipeline's stages hold one.

Both walks are derived from a single exhaustive walk over the *slots* a node has, and a slot's kind is what may stand in it:

- an expression slot takes any expression;
- a name slot takes only a name — a construction's spread, and in Core the binding an application invokes;
- a construction slot (Core only) takes only a `NewData` — an attempt tests whether a construction holds, and nothing else may stand there.

`forEachChild` hands every slot over as a child, so a pass that asks what an expression names reaches a spread without knowing that spreads exist. `mapChildren` takes an operator per slot kind, so a rewrite cannot put an expression where the language admits only a name — the constraint is in the type, not in a cast or a runtime check, and the one cast Core had is gone. There is no one-operator form: each caller says what it does with a name.

## The bug it fixes

A published value whose body spreads another of its module's values travelled without it. `ModuleMetadata.reach` matched `Var` and `Apply` and nothing else, so the spread name was never reached, the spread value's text was not carried in the jar, and the importing module failed with ``​`base` is spread but names no binding``. It is fixed here without `reach` being touched — which is the point.

Two more readers stop reaching for spreads by hand, and the inliner's renaming asks one function wherever a name stands.

## What it does not do

The declaration spread (`Data.includes`) names a type, not a value, and stays out of this walk — a different namespace with different resolution and different rules about qualification.

The `spreads().isEmpty() && inits().size() == 1 && inits().get(0).name().equals("value")` triple, repeated in three files, is a separate rule duplicated in three places. Separate issue.

`Ast.Construct.spreads` is deleted in its own commit. Every construction of one passed an empty list and both readers ignored the field.

## The rest of the read-only walks

Twenty-five Ast sites now see a spread. Most ask about calls or binders and are unaffected. Of the five that ask about names, one was the bug above; the other four turn out to be unreachable or unobservable, and each is that way for a reason that lived outside the walk:

- `HelperInliner.references` is only asked about bindings that hold a function, and a spread must name a data value.
- `Elaborator.reaches` runs only for a function that escaped, and the applications it skips come from a combinator's own expansion, where no argument is a spread.
- `HelperParams` would gain only an `openUse`, which the one caller collects into a map it discards.
- `Shapes.leftmost` walks invariant clauses, and an invariant may not construct, so it never meets a construction.

The inliner's renaming had drifted the same way: a spread kept its own position where a name standing alone took the call site's. Unifying it is right, but its effect is narrower than a diagnostic — a spread's position reaches `Core.Read.pos()` and from there the line table of the generated code, so what changes is which line a spread inside an *imported* helper is attributed to. No test here demonstrates it.

## Checks

Full suite green: 1972 compiler, 98 LSP, 1 bench.

`CompileValueSpreadTest.aPublishedValueCarriesTheValueItSpreads` is the regression test for the bug; it fails on `develop` with the diagnostic above.

`ast.EverySlotIsAChildTest` and `core.EverySlotIsAChildTest` pin the walk contracts themselves — which slots are children, and which operator each goes through. Both were checked by mutation: reverting the Ast name slot fails 3 of 4, and reverting either Core slot fails 3 of 5.

The derived `forEachChild` keeps the node it walked when nothing changed, so a read-only walk allocates as little as before. `souther-bench warm` on the crm corpus: `develop` median 121.6 / 121.0 ms, this branch 125.3 / 122.9 ms, with the branch's minimum lower in both rounds.


kawasima:
Third round. The construction slot had stopped `GrowingFold` recursing, and it was a real regression — confirmed by measurement before fixing:

```
BUILDERS on develop: 1
BUILDERS on this branch (before the fix): 0
```

for a `List.map` written in a field of a construction an attempt tests. Silent, because the answer is the same either way and only the emitted code differs.

The cause is the one named in review: a name slot is a leaf, so `s -> s` is complete there, but a construction slot is not, and both of `GrowingFold`'s recursions were left saying nothing at it. Fixed with a `NewData`-answering `mapChildren` over the *same* slot definition the walk over an expression reads — a private `atSlots(NewData, …)` that the big switch's `case NewData` also delegates to, so the construction's slots are still written once and the cast stays gone.

`CompileGrowingFoldTest.aWalkInsideAnAttemptedConstructionIsRewritten` counts the builder in the bytecode. Mutation-checked: putting `c -> c` back fails it.

I checked the rest of the slots for the same shape. A name slot holds a `Read`/`Var`, which is a leaf, so identity there loses nothing; the construction slot is the only non-leaf one and `IfConstructed` is its only holder. On the Ast side there is no construction slot — `Ast.IfConstructed.construct` is an `Expr` and recursion is automatic.

Also fixed: a javadoc I had left floating above `Resolve.name` that belonged to `expr`.

Full suite green: 1973 compiler, 98 LSP, 1 bench. `souther-bench warm` crm median 124.8 ms, min 114.5 ms — unchanged against the `develop` baseline in the description.

On the ADR: that was decided before this round — the repo's bar is reversal cost and binding force, and the judgement was that this applies the rule ADR-0084/0085 already state rather than adding one. I have not reopened it.

PR #286, opened 2026-08-02
Reify a behavior's name where it is bound

### LINKED ISSUE #271: A behavior's name is a value in an argument and not in a binding, so one name still has two answers

kawasima:
A behavior's name handed to a combinator reaches the behavior's class; the same name bound to a `let` first was refused. One name, two positions, two answers — with one position left (#271).

## What was in the way

Reifying a name — turning a named function into the function value it names — is written four times in `HelperInliner`, and each gets its arity from a different place:

| | arity from |
|---|---|
| `valueOf` | the helper's `FnDef` |
| `libraryValueOf` | the library declaration |
| `forwardDependencies` | the callee's declared `Ast.FnType` |
| `desugarNamedBlock` | the helper a `fold` step names |

A behavior is the case none of the four covers, and the inliner is built from the helper table, so the arity to expand to was not in hand there.

## What this does

The four share one `etaExpand` — the block, its parameters and the application it wraps — while keeping their own entry points. A name standing for a declaration and a `depends on` binding forwarded to its callee decide arity for different reasons, and one lookup rule over `ValueName` would have hidden that.

`declarationArity` answers for a helper, a library function and a behavior alike. What reification needs is how many arguments the block takes; what stands behind the name — a Souther body, a shipped kernel, a Java implementation — is not the name's business.

Which behaviors a body may name is a fact about the module, so the query layer works it out. `Bodies.NamedBehaviorArity` reads `CalleeSigs` and drops the types, because the expansion is written before anything is typed; asking its own question keeps a change to a behavior's input types from expanding every body of the module again.

## What is still refused

A behavior with a requirement, and a composition. Both are reached through the clause that names them rather than by standing for a declaration, so there is nothing for a name to be. `CompileImportedNameTest.aNameThatDenotesSomethingUnusableSaysWhatItDenotes` pinned the binding position that this fixes; it now pins a composition, which is still answered and still not a value.

A behavior taking no input is refused too: a block taking no parameter is not a shape this language writes.

## Evidence

`CompileBehaviorByNameTest` — the argument position, the binding position, bound and applied, chosen between at run time, imported, shadowed by a local, and the two refusals. Five of its nine fail on `develop` with ``​`twice` is a behavior, and cannot be held as a value here``. One reads the emitted `demo.Go$Impl` for `demo/Twice`, so the value reaching the behavior rather than the `let` is checked rather than assumed.

Full suite green (1982 tests). `souther-examples` builds and verifies against this.

Closes #271.

### DIFF
--- a/souther-compiler/src/main/java/souther/compiler/check/HelperInliner.java
+++ b/souther-compiler/src/main/java/souther/compiler/check/HelperInliner.java
@@ -1282,9 +1282,9 @@ public final class HelperInliner {
      * <p>Empty covers a declaration taking nothing and every name that stands for no declaration at
      * all: a binding, a construction, a library name the library does not declare, and a behavior
      * this body may not name. Each of those is left as it was written and reported where it is
-     * used. A declaration taking nothing is two things under one answer — a library value, which is
-     * already the value it is, and a behavior taking no input, which has no function value because
-     * a block taking no parameter is not a shape this language writes.
+     * used. A declaration taking nothing has no function value to become rather than one this
+     * declines to make: a {@code let} with no parameter list is a value and is written without
+     * {@code ()}, and there is no block taking no parameter to expand it to.
      */
     private OptionalInt declarationArity(Ast.Var v) {
         int arity = switch (v.denotes()) {
--- a/souther-compiler/src/main/java/souther/compiler/check/HelperInliner.java
+++ b/souther-compiler/src/main/java/souther/compiler/check/HelperInliner.java
@@ -21,7 +21,9 @@ import java.util.LinkedHashMap;
 import java.util.LinkedHashSet;
 import java.util.List;
 import java.util.Map;
+import java.util.OptionalInt;
 import java.util.Set;
+import java.util.function.IntFunction;
 import java.util.function.Predicate;
 
 /**
@@ -54,6 +56,17 @@ public final class HelperInliner {
      * applying the declaration by a spelling and why a lambda naming itself had to be refused.
      */
     private final Map<BindingId, Ast.FnDef> scopedLambdas = new HashMap<>();
+    /**
+     * The behaviors a body expanded here may name, with how many inputs each takes — the module's
+     * own callable ones and the ones it borrows (spec {@code [#calling-a-behavior]}).
+     *
+     * <p>Apart from {@link #helpers} because a behavior is never expanded: what stands behind its
+     * name may be a Java implementation, so a body that reached past it to the {@code let} would be
+     * a second answer to the same name. All this holds is what reifying the name needs, which is its
+     * arity; the query layer works out which behaviors are here, since which of them may be named is
+     * a fact about the module rather than about any one body.
+     */
+    private Map<String, Integer> callableBehaviors = Map.of();
     private final Set<String> recursive = new HashSet<>();   // own helpers on a call cycle (spec 13.1)
     private final Map<String, LambdaOrigin> lambdaOrigins = new HashMap<>();   // $k_p -> where it was written
     private int counter = 0;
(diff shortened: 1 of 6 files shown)


kawasima:
Second round, from review: the standalone helper path.

## The path was real, the hole was a diagnostic

A non-recursive helper has no `LoweredBody`; it is checked standing on its own through `TypeChecker.checkModule` → `HelperInliner.forModule` → `HelperTyping.checkHelpers`, and that inliner was told nothing. Measured before wiring it:

| written in a helper | reported |
|---|---|
| `List.map(twice, xs)` | `E1401: a helper `let` does not reach one` |
| `let f = twice` … `List.map(f, xs)` | `` `twice` is a behavior, and cannot be held as a value here `` |

So the case was refused either way — no program was accepted that should not be — but the same shape gave two answers, one of them naming a rule about bindings that does not exist. With the table wired, both say E1401: the name becomes the function value it names, and a helper may then not apply it.

`BodyForInvariantDischarge` is wired for the same reason. It is the same body as the emitted one, and the two representations should not disagree about what a name means.

`InjectionSigs.arities` is now the one projection from the callable signatures to what reification needs. `checkModule` already receives `calleeSigs`, so it derives the table rather than taking a tenth parameter, and `Bodies.NamedBehaviorArity` — which exists so `LoweredBody` does not depend on types — calls the same method.

## Arity 0 is not a suppressed judgment

Measured rather than argued. The language has no zero-parameter function:

- `let seed () = 7` → "`let seed` takes no parameters, so it is written without `()`"
- `() -> 7` → "expected an expression"

A `let` with no parameter list is a value. There is no block taking no parameter for a nullary declaration to expand to, so `OptionalInt.empty()` is the only representable answer rather than a rule this chooses. The javadoc says that now.

## The identity key found a real bug, and it is not this branch's

`b.name()` was the right thing to check. A module may declare a behavior and import one of the same spelling:

```sou
import up ( twice )                              // up.twice : (n: Int) -> Int
behavior twice : (a: Int, b: Int) -> Int         // demo.twice
let go (i) = Out { n = twice(i.n, 3) }           // compiles
```

`demo.Go$Impl` references `up/Twice`. The call is typed against `demo`'s declaration — `twice(i.n)` is refused with "expects 2 argument(s), got 1" — and emitted against `up`'s. Measured on a clean `develop` worktree at 4b069b0, with the direct call and no binding anywhere, so it predates this branch. Filed as #287.

This branch inherits it in the binding position, which is the point of the branch: `let f = twice` now reaches whatever `twice(...)` reaches. `aBoundNameAndAWrittenNameReachTheSameBehavior` pins that agreement and not either answer, so it stays true when #287 is fixed.

The key stays `b.name()`. The arity table and the call typing are both `CalleeSigs`, which puts the module's own behaviors in first and the imported ones with `putIfAbsent` — the same precedence resolution uses — so the two cannot pick different entries. What can disagree is the emitted target, which is #287 and a different table.

## Also added

`aHelperReachesNoBehaviorInEitherPosition` — the two spellings, asserting the diagnostic codes match rather than quoting one.

Full suite green. `souther-examples` clean verify green against an installed build.

PR #289, opened 2026-08-02
Refuse an import beside a declaration of any kind

### LINKED ISSUE #287: A module declaring a behavior and importing one of the same spelling calls the imported one
A module may declare a behavior and import one of the same name. The call is typed against the module's own declaration and emitted as a call to the imported one.

```sou
module up exposing ( twice )

behavior twice : (n: Int) -> Int
let twice (n) = n * 2
```

```sou
module demo

import up ( twice )

data In = { n: Int }
data Out = { n: Int }

behavior twice : (a: Int, b: Int) -> Int
let twice (a, b) = a * b

behavior go : (i: In) -> Out constructs Out
let go (i) = Out { n = twice(i.n, 3) }        // compiles
```

`demo.Go$Impl` references `up/Twice` and not `demo/Twice`, so a behavior declared with one input is called with two. Writing `twice(i.n)` instead is refused with "twice expects 2 argument(s), got 1", which is the module's own declaration answering — so the two sides of the same call disagree about which `twice` it is.

Measured on `develop` (4b069b0). Not new: the same body compiles the same way before and after #286.

## What the rule is

ADR-0086: "One spelling reaches the value namespace once, whichever way it arrives — the module's own declarations, an import, or a library qualifier ... an import that collides with a declaration is refused at the import line." Two behaviors arriving under one spelling are refused when both come from imports — `Bodies.Imported` reports it, because one behavior name is one injected field. A declaration colliding with an import is not checked anywhere.

So there are two ways to fix it, and they are not the same decision:

1. Refuse at the import line, as the collision of two imports is refused. One spelling, one behavior, and the module says which by not writing the import.
2. Let the declaration win, as it does for a value name, and make the emitted call agree with the typing.

The second keeps a working program working; the first is what the ADR's sentence says. Whichever is chosen, what must not survive is the current state, where the type check reads one declaration and the backend emits the other.

## Where the two answers come from

The call is typed against `Bodies.CalleeSigs`, which is `InjectionSigs.callable(...)`: the module's own behaviors are put in first and the imported ones with `putIfAbsent`, so the declaration wins. What the backend emits reaches the behavior class through a different path, which answers with the import.

## Why this is being reported now

#286 makes a behavior's name in a binding reify like the same name written directly, so `let f = twice` now reaches whatever `twice(...)` reaches. That is the point of it — one name, one answer — and it means the binding position inherits this too. `CompileBehaviorByNameTest.aBoundNameAndAWrittenNameReachTheSameBehavior` pins the agreement rather than either answer, so it stays true when this is fixed.

## References

- ADR-0086 (a function name is a value, and application is an operation on one)
- `Bodies.Imported.collision`, which reports the import-against-import case

kawasima:
A module may declare a behavior and import one of the same spelling. The call is typed against the module's own declaration and emitted against the imported one.

```sou
module demo
import up ( twice )                        // up.twice : (n: Int) -> Int
behavior twice : (a: Int, b: Int) -> Int
let twice (a, b) = a * b
let go (i) = Out { n = twice(i.n, 3) }     // compiles
```

Running it: `NoSuchMethodError: up.Twice.apply(Long, Long)`. `demo.Twice` is emitted and never called. A clean compile producing a broken class.

## The check was written for a data and stopped there

Three ways a name reaches the value namespace, and the collision was checked for one:

| declared beside the import | before |
|---|---|
| a data | refused |
| a `let` | compiled; the declaration won everywhere |
| a behavior | compiled; typed against one, emitted against the other |
| (a library import beside any of them) | compiled; the declaration shadowed the import on purpose |

`Names.valueNamespaceCollisions` is where this belongs — its javadoc already says a data, a `let`, a behavior, a published definition and a library qualifier all reach the namespace, and that one name means one thing there whichever way it arrived. What it compared was `Bodies.publishedNames` against `m.defs()`: published helpers against declared types, so a behavior on either side was invisible to it.

It now compares what each import brings into the namespace — published definitions, the declaring module's behaviors, its types, and for a library line the functions it names — against every name this module declares.

`Exposing` decides the library imports and drops them before `Names` sees the module, so the library half is refused there. That means removing a deliberate shadow: `ownNames.contains(name) -> continue`, commented "the module defines its own `name`; that shadows the import". ADR-0086 says an import that collides with a declaration is refused at the import line, and the type half of the same method already refused; the `let` half was the one that did not.

## What still shadows

A binding. It is written inside a body, and what an import says is what a name means where no binding answers it — ADR-0086's "a parameter named `map` now wins over `import List ( map )`" is unchanged, and `aBindingSpelledLikeAnImportIsStillTheBinding` pins it.

## Evidence

`CompileImportCollisionTest` — the three declaration kinds beside a user-module import, a library import beside a `let`, and the binding that is still not a collision. All four refusals report `check.import.conflict`, which is the key the data case already used; its hint said "One name cannot stand for two types" and now says "two things".

Measured before the fix: the behavior case compiled on `develop` at 4b069b0 and threw `NoSuchMethodError` when run.

`CompileBehaviorByNameTest.aBoundNameAndAWrittenNameReachTheSameBehavior` was written against the collision, since that was the shape where the two tables could come apart. With the collision refused it compares the two spellings against an imported behavior and against a declared one, which is the invariant it was pinning either way.

Full suite green (1990 tests). `souther-examples` clean verify green — no module in the corpus shadowed an import with a declaration of its own.

Closes #287.


kawasima:
Review found three things. All three reproduced, all three fixed in 69722faf.

## The library half was raised, not reported

`Exposing.read` is called from `Front.Exposed.compute` with nothing catching it, so a `CompileException` escaped the question that reads the module. Measured:

```
Compiler.diagnoseModules({
  "c.sou": "module probe.c / import List ( map ) / let map (n: Int) = n + 1",
  "d.sou": "module probe.d / data Broken = { a: NoSuchType }",
})
```

threw instead of answering, so `d.sou`'s own diagnostic was lost with it. In an editor that is one "the compiler could not finish reading this file" across the workspace while the author is part-way through writing the `let`.

It is `Validated.conflicts()` now, and `Front` attaches it to the module it read. The data case was already raised this way, so that one is fixed with it — `stdlib-vs-data` used to throw too.

## The library branch in `importedIntoValues` was dead

`Front` drops the library lines before `valueNamespaceCollisions` is asked, so the branch never ran and the library half was refused by the raise alone. The test passed for the wrong reason. The branch is gone: a library line is read where the table it fills is built, which is `Exposing`, and `aLibraryImportBesideADeclarationIsRefused` now passes because of that report.

## What the source module exposes was not consulted

The published definitions were filtered by `from.exposing()`; the behaviors and types added beside them were not. A name the line asks for and the module does not expose was reported both as `check.import.notexposed` and as `check.import.conflict`, and the second told the author to rename a local definition that is not the problem. Now only what the module exposes counts as brought in.

## Added

- `theCollisionIsReportedWithoutStoppingTheOtherFiles` — the two-file case above, asserting each file keeps its own diagnostics.
- `aNameTheSourceDoesNotExposeIsNotAlsoACollision` — one line, one report.

Full suite green (1989 tests). `souther-examples` clean verify green.

kawasima:
Both fixed in afa229c4.

## [P2] The attached files joined after the imports were read

Confirmed, and the gap is the library half only:

| declared in | beside | before |
|---|---|---|
| the model file | `import List ( map )` | refused |
| an attached file | `import List ( map )` | **compiled** |
| an attached file | `import up ( base )` | refused |

The user-module half was already covered because `Names.valueNamespaceCollisions` reads `Front.Available`, which answers with the assembled module. The library half is read in `Exposing`, which ran on the model file alone — so a `let map` written beside the rows was the one declaration that could still shadow a library import.

The files join first now, as suggested. They write no imports of their own, so what is read is still the model file's lines; what is complete by then is the set of names an import can collide with.

`aValueAnAttachedFileDeclaresCollidesWithALibraryImport` pins it in `CompileImportCollisionTest`, beside the other declaration kinds rather than in `ExampleFileValuesTest` — what it is about is the collision rule, and the attached file is which of the rule's inputs was missing.

## [P3] The javadoc stated the rule this replaces

`Exposing`'s class comment said a binding **or the module's own declaration** wins over an import. A binding does. A declaration is a conflict, and the declaration standing as the name's meaning after the report is what recovery does with a module that will not be emitted — not a shadowing rule. Reworded to say that.

## Test count

1990 now; the PR body said 1984 and is updated.

`souther-examples` clean verify green.

Issue #75, opened 2026-07-25
Compiler crash: a block whose result expression is swallowed by call continuation raises NullPointerException instead of a syntax error


kawasima:
## Summary

A statement that ends in a bare identifier followed by a line starting with `(` parses as a call (the parenthesised line becomes the argument list), so the block is left with no result expression. `AstBuilder` then dereferences a null result node and the compiler dies with a `NullPointerException` instead of reporting a syntax error.

## Reproduction

```sou
module r
data Out = { v: Int }
behavior run : (i: Out) -> Out constructs Out
let run (i) = {
  let a = i.v
  (a)
}
```

```
Exception in thread "main" java.lang.NullPointerException: Cannot invoke "souther.compiler.cst.SyntaxNode.kind()" because "n" is null
	at souther.compiler.frontend.AstBuilder.expr(AstBuilder.java:426)
	at souther.compiler.frontend.AstBuilder.foldStatements(AstBuilder.java:731)
```

`AstBuilder.foldStatements` reaches `expr(result)` with `result == null` — the comment there says "a block always ends in a result expression", which is exactly the assumption the parse above breaks.

## How it is hit in practice

Souther is layout-independent, so `i.v` followed by `(a)` is one call. That is the intended rule, but the shape shows up naturally when a tuple is the result expression of a step block:

```sou
let (_, out) = fold((acc, x) -> {
    let (seen, ys) = acc
    (Set.insert(x, seen), ys ++ [x])
}, (Set.empty(), []), i.xs)
```

Here `acc` swallows the following tuple as an argument list, and the crash is what the author sees. The same code with a statement that does not end in a bare identifier compiles.

## Expected

A block with no result expression is a syntax error with a position and a message — ideally one that names the likely cause (the previous statement's trailing name absorbed the next line as a call).

## Environment

develop @ bbc80d1, 0.1.0-SNAPSHOT (JDK 25). Not new: the same crash reproduces at 0e6a19d.

kawasima:
Fixed by #77, merged into `develop`. The parser now reports `parse.block.noresult` at the closing brace, and the message names the usual cause — a result expression absorbed by the line above, not a missing one. `AstBuilder` keeps its own guard for a build over an already-errored CST.

PR #293, opened 2026-08-02
One node for an application, and a qualified name answered once

### LINKED ISSUE #274: The CST still has two nodes for an application, so a callee is reassembled from tokens where the AST has an expression
The AST is one node — `Ast.Apply(Expr function, …)` — and the CST is still two. `name(args)` and `Mod.name(args)` parse to `CALL_EXPR`, and an argument list after anything else parses to `APPLY_EXPR`, so `identExpr()` still decides "call or variable" by looking three tokens ahead for a `(`, and the formatter still reassembles a name callee from the identifier tokens under the node.

That reassembly is the shape that produced two defects in #273: it printed `adder(1)(x)` as `adder(1, x)`, and it is what `HelperInliner.rename` was doing when a non-name callee reached the check with an empty name and threw.

The form that has one path:

```
name(args)  →  VAR_EXPR(name)  →  postfix  →  APPLY_EXPR(VAR_EXPR(name), args)
```

which deletes `CALL_EXPR`, `AstBuilder.call()`, `Formatter.call()`, the reassembly, and the `nth(3) == LPAREN` lookahead.

## What makes this its own change

A qualified name is the part that does not fall out. `Map.empty` is a namespace member or an ordinary field read depending on whether `Map` is bound, which `Resolve` decides and the parser cannot. Today `CALL_EXPR` flattens `Map.empty(…)` into one name at parse time and `Resolve.memberOfNamespace` folds the no-argument form; unifying the two CST nodes means the qualified form is a `FIELD_ACCESS` chain in both cases and the fold is the only path. That is the right end state — one way to read `Q.m` — but it moves what the parser hands the resolver, so it wants its own change and its own measurement.

kawasima:
Closes #274 once merged (`gh issue close 274` — a PR into `develop` does not close it).

The CST had two nodes for one operation. `name(args)` and `Mod.name(args)` were a call; anything else applied was an application; the parser chose between them by looking three tokens ahead for a `(`. That left the callee of a call as a name reassembled from the identifier tokens under the node — in the AST builder, in the formatter, and in the `|>` desugaring — while the AST has held one node with a callee expression since #273.

`CALL_EXPR` is gone. A name is a variable, an argument list after it is an application, and a qualified callee is the field read it follows.

## What carried the weight

Resolution. `Q.m` was folded into one name only where `Q` was a standard-library qualifier and only where the chain was two segments long; everything else reachable through a qualifier got there because the parser had flattened it. The fold now asks the whole chain, and a chain that answers nothing is taken apart so the part that is a name is asked in turn — `probe.a.defaultAmount.value` folds `probe.a.defaultAmount` and reads `.value` off it.

A report names the member of the namespace where the answer ran out (`probe.a.NoSuch`, `up.defaults`). A chain rooted at a name nothing declares is reported at the root (`unknown`), because nothing in front of the dot is a namespace to have a member of.

## What this fixes

**A module whose own name is dotted, in a value position.** The lookahead read a single dot, so `probe.a.Amount(n)` was already the field-read shape and had no path to the module — while the same spelling worked in a type position. It works in both now.

**A report that quoted a name nobody wrote.** Applying something other than a name binds it first and applies the binding, so `deps.count(x)` reported `` `$fn0` is not a function here `` and underlined four characters from the right place. The callee's spelling now travels with the application: the written form is what a report says, the binding is what the application means.

## What changes for a program

**An argument list must begin on the line its callee ends on.** That was already the rule for an applied expression and is now the only rule: a `(` that opens a line is a parenthesised expression, so a result written under a statement is that result rather than an application of the name above it. No `.sou` in this repository or in `souther-lang/examples` writes an argument list across a line break. The case #75 reported — a block's result absorbed by the line above — cannot arise, and its test now pins the result being the result.

**Editor colours.** The callee of an application is the function, and for a qualified callee that is its last name: `Map.map(f, xs)` colours `map` as a function and `Map` as what it is written as. Before, both were the function.

## What did not fall out

A function cannot be stored in a data field — codec derivation runs on every declared data and refuses it — nor be a newtype's base. So a field read never answers a function, and applying one is never a working program: what this changes there is the report, not what compiles.

Reaching another module's values through a qualifier is still not a thing; only its types are. `up.someHelper` was refused before and is refused now, with the full spelling named.

## Measurement

- Full suite green: 2013 / 101 / 73 / 1.
- `souther-lang/examples` builds and its tests pass.
- Formatter output is byte-identical to `develop` over all 52 `.sou` in both repositories.
- `souther-bench warm phase`, same JDK and settings, is within noise: crm warm median 117.4 ms against 117.7 ms; phase `parse` 4.71 ms against 4.47 ms; `resolve names` 1.10 ms against 1.09 ms.

### DIFF
--- a/souther-compiler/src/main/java/souther/compiler/ast/Ast.java
+++ b/souther-compiler/src/main/java/souther/compiler/ast/Ast.java
@@ -986,8 +986,14 @@ public interface Ast {
      * carries what it denotes — a helper, a library function, an injected behavior, a function-typed
      * binding, or the type a newtype construction wraps — answered once during resolution.
      */
-    record Apply(Expr function, List<Expr> args, ConstructionOrigin origin, SourcePos pos)
-            implements Expr {
+    record Apply(Expr function, List<Expr> args, ConstructionOrigin origin, String appliedAs,
+                 SourcePos pos) implements Expr {
+
+        /** Applying whatever {@code function} is, with nothing standing in for what the source
+         * wrote — every application but one a lowering rewrote. */
+        public Apply(Expr function, List<Expr> args, ConstructionOrigin origin, SourcePos pos) {
+            this(function, args, origin, null, pos);
+        }
 
         /** Applying a name, as the parser read it, before resolution has said what the name denotes. */
         public Apply(String fn, List<Expr> args, SourcePos pos) {
@@ -1013,11 +1019,15 @@ public interface Ast {
          * <p><b>Never a lookup key.</b> An import lets a library name be written without its
          * qualifier, so a table keyed by a declaration's name misses on the spelling — silently,
          * because a miss is what a table keyed by names does with one it has not got. Every
-         * question of the form "which declaration is this" asks {@link #reaches()}, whose answer is
-         * this spelling for every name an import cannot have renamed.
+         * question of the form "which declaration is this" asks {@link #reaches()}.
+         *
+         * <p>{@link #appliedAs} answers where a lowering replaced what was written with a binding
+         * it introduced: applying something other than a name binds it first, and the binding is
+         * named nothing an author could have typed. The two are separate for that reason — this
+         * one is read by reports and nothing else.
          */
         public String written() {
-            return function instanceof Var v ? v.name() : "";
+            return appliedAs != null ? appliedAs : nameApplied();
         }
 
         /**
@@ -1028,9 +1038,19 @@ public interface Ast {
          * bare, so that is what it answers; every other name is filed as it is written. The empty
          * answer is not a convention: it is what applying something that is not a name leaves, and
          * no declaration is named the empty spelling.
+         *
+         * <p>Never {@link #written()}: a spelling a report quotes is not what this application
+         * reaches, and a lowering that put one there would otherwise have every table in the
+         * compiler looked up with it.
          */
         public String reaches() {
-            return denotes() instanceof ValueName.Stdlib lib ? lib.qualified() : written();
+            return denotes() instanceof ValueName.Stdlib lib ? lib.qualified() : nameApplied();
+        }
+
+        /** The name in the callee position, or the empty spelling where what is applied is not a
+         *  name. What both spellings above are built from, and what neither may skip past. */
+        private String nameApplied() {
+            return function instanceof Var v ? v.name() : "";
         }
 
         /** What the name this applies denotes, or null where what it applies is not a name. */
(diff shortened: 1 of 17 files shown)


kawasima:
Both points addressed in `c47fd60e`. One correction on the first.

## The line rule is already in the specification

`specification.adoc:1591`, unchanged on `develop`:

> Application is written after what is applied, and what is applied is any expression: a name (`f(x)`), a parenthesised expression (`(if flag then String.trim else String.uppercase)(s)`), or the result of an application (`adder(1)(x)`). What is applied is evaluated once, and before the arguments. **An argument list MUST begin on the line its callee ends on** — a `(` that opens a line is a parenthesised expression, so a block whose result is a tuple written under a call reads as that block's result and not as an application of the line above.

That is the sentence being asked for, and it landed with #273. What this PR changes is that the implementation now agrees with it: `CALL_EXPR` was the one shape the rule did not hold for, so the same two lines meant one thing with an identifier callee and another with any other. `git diff develop -- specification.adoc` is empty because there is nothing to add.

The tests are a fair ask and are in. The `souther-syntax` module has no test source root — CST tests live in `souther-compiler/src/test/java/souther/compiler/cst/` — so they went beside `CstCallResultFieldAccessTest`, parameterised over the four callee shapes:

```
amountOf              a bare name
up.amountOf           a qualified name
amountOf // a comment a comment between the two
amountOf(i)           the result of an application
```

each followed by `(i)` on the next line, asserting nothing was applied across the break and that the `(i)` is a parenthesised expression.

## `reaches()` — correct, and fixed

The contract was broken. `reaches()` falls back to `written()` for anything that is not a library name, so putting the diagnostic spelling in the callee's `Var` made `d.count` the answer to "which declaration is this". The `instanceof ValueName.Local` guard added to `CallElaborator` was the leak showing, as you read it.

They are separate slots now:

```java
record Apply(Expr function, List<Expr> args, ConstructionOrigin origin, String appliedAs, SourcePos pos)
```

`written()` prefers `appliedAs`; `reaches()` cannot see it and reads the callee position only. The lowering keeps `Var("$fn0", Local(...))` where it was and puts `d.count` in `appliedAs`. The test you asked for:

```java
assertEquals("d.count", lowered.written());
assertEquals("$fn0", lowered.reaches());
```

Two things fell out while doing it. Four passes rebuild an application over rewritten arguments and dropped the new slot — the first run reported `$fn0` again, from the inliner's own re-entry. They go through `Apply.withArgs` now, so a rewrite with no opinion about the slot keeps it instead of each one having to remember. And the standard-library report no longer asks what the applied name denotes: it asks what the application reaches, where a lowered callee reaches its binding and has no dot in it, which is a better question than the one I had it asking.

Full suite green (2018 / 101 / 73 / 1), `souther-lang/examples` builds and its tests pass.

PR #294, opened 2026-08-02
Read a relation over a list at the element a closure is handed

### LINKED ISSUE #288: A relation quantified over a list is not assumed of the element inside a combinator's closure, so neither remedy the warning names is available
A construction the checker cannot prove warns, and the warning names two remedies: guard the relation,
or reify it into an input's invariant so it is assumed at the construction. For a relation that holds
of every element of a list, and a construction inside a combinator's closure, neither one works. The
invariant is stated, the guard is written, and the warning is exactly the same.

## Repro

The relation on the input's own type, which is what the hint asks for:

```souther
data Qty = Int
    invariant value >= 1
data Line = { quantity: Int }
data Cart = { items: List<Line> }
    invariant List.all(i -> i.quantity >= 1, items)

behavior toQty : (cart: Cart) -> List<Qty>
    constructs Qty
let toQty (cart) = List.map(i -> Qty(i.quantity), cart.items)
```

```
warning: Constructing `Qty` here may violate its invariant: the guards do not establish it. Guard it with `guard`, or reify the relation into `Qty`'s (or an input's) invariant so it is assumed here.
```

The same over a list of bare `Int` (`invariant List.all(q -> q >= 1, quantities)`, `List.map(q ->
Qty(q), …)`) warns identically, and so does the unqualified spelling with `import List ( map, all )`.

The other remedy, guarding the same relation immediately before the map:

```souther
let toQty (cart) = {
    guard List.all(i -> i.quantity >= 1, cart.items) else Invalid { why = 1 }
    Qtys { qs = List.map(i -> Qty(i.quantity), cart.items) }
}
```

Same warning.

Two controls say the seeding itself is fine. A scalar invariant on the input, read directly, is
silent:

```souther
data Cart = { quantity: Int }
    invariant quantity >= 1
let toQty (cart) = Qty(cart.quantity)          // no warning
```

And a relation on the element's own type is assumed inside the very same `List.map`:

```souther
data Pos = Int
    invariant value >= 1
data Line = { quantity: Pos }
data Cart = { items: List<Line> }
let toQty (cart) = List.map(i -> Qty(i.quantity.value), cart.items)   // no warning
```

So facts do reach the element binding of a closure. What does not reach it is a relation quantified
over the list.

## Where it comes from

`InvariantChecker.walkCall` binds a combinator closure's element parameter to the list's element type
and seeds it from that type alone:

```java
Known k2 = rebind(k, p);                 // anything known about the name p is dropped
if (elem != null) {
    t2.put(p, elem);
    k2 = seedParam(p, elem, k2);         // the element carries its type's invariant
}
```

The container's own clause takes the other path: `seedAt` runs `obligations` over each invariant
clause of the input's type, and `List.all(i -> i.quantity >= 1, items)` yields at best a `Fact` keyed
on that call as written — true of `items`, related to nothing inside the closure. A `guard` of the
same shape goes through `assumeCond` and lands in the same place. Either way the element parameter
enters the closure knowing only what its own type says, which for a bare `Int` field is nothing.

## Cost in the model

The warning is the whole of what the checker says about an unproven construction, and for this shape
both remedies it names are unavailable. An author who follows the advice writes an invariant that
changes nothing, and is left with a warning that repeats the same advice.

The shape is not unusual. "Every line of this cart has a quantity of at least one" belongs to the
cart, not to the line's field type, and in `souther-lang/examples` the element type is not even the
author's to change: `inventory.allocatable` constructs `Eaches(i.quantity)` from `cart`'s `LineItem`,
whose `quantity` is a bare `Int` published by another module. Stating the rule on `PricedCart` is the
only place it can go, and there it does not discharge.

#226 traced a projection through a `List.map` closure so that uniqueness stated of the source list
survives the map. This is the same seam in the other direction: a property stated of the source list
assumed of the element the closure is handed.

## Proposal to discuss

Where a combinator's list argument is the term a known clause quantifies over, instantiate the
quantified body at the closure's element binding: from `List.all(p, xs)` known — as an input's
invariant clause or as an asserted guard — assume `p` of the element parameter bound to an element of
`xs`. `List.map`, `List.filter` and `List.fold` all already bind that parameter in `walkCall`, so the
binding this needs exists.

If instantiating a quantifier is outside the fragment the checker is meant to hold, then the hint has
to stop offering it: a remedy the checker cannot use for the shape in front of it is worse than no
remedy, because it reads as a claim that the shape is expressible.

Found while reviewing the open findings in `souther-lang/examples` (F28 there). Verified against
develop at 1f78e801 with the repros above.

kawasima:
`List.all(p, xs)` settled a fact keyed on the call, which relates it to nothing inside. A construction inside a `List.map` over `xs` read only what the element's own type says, so neither remedy E2011 names was available for a relation quantified over a list: stating it as the input's invariant changed nothing, and guarding the same `List.all` before the map warned identically.

```souther
data Qty = Int invariant value >= 1
data Line = { quantity: Int }
data Cart = { items: List<Line> } invariant List.all(i -> i.quantity >= 1, items)

behavior toQty : (cart: Cart) -> List<Qty> constructs Qty
let toQty (cart) = List.map(i -> Qty(i.quantity), cart.items)   // warned; now discharged
```

## What changed

A relation is kept as the clause it was written as, together with the rule that names its free leaves, so it can be read again at the element. `walkCall` already binds that element; where the container it walks is one a relation was stated of, the clause is read with the quantifier's own parameter standing for the element. That reading is `obligations` over a `Bindings` — the same reading seeding a type's invariant is, differing in direction and not in what it reads.

- An input type's invariant clause and a `guard` or `if` on the path record the same way.
- The container the closure walks need not be the one stated of. A *permutes* or a *subset* construction holds only elements of what it was built from, so `List.map(f, List.filter(q, xs))` carries what was stated of `xs`. A *maps* or *collapses* one does not.
- A relation the clause in turn states of a container is recorded as one, so a container of containers reaches its innermost element.
- A relation is dropped where a binding takes over a name it reads, the container included: inside `List.map(cart -> ..., cart.groups)` the term `cart.items` is the group's.
- `List.any` records nothing, and neither does a `List.all` the path has denied.

A clause comparing a field chain now bounds a term the numeric domain names, rather than only settling a fact a guard has to restate. The spec has read a chain as a term since it was written; the clause side had not.

## Verified

- `CompileQuantifiedRelationTest` — 14 rows: both repros from the issue, the bare-`Int` list, the unqualified `import List ( all, map )` spelling, a helper expanded into the closure, a selection, a fold, nesting, and the five that must still warn.
- `CompileInvariantDischargeTest` — one row for the chain leaf.
- Full suite green (2033 compiler, 101 lsp), and `souther-lang/examples` builds against it.

Fixes #288

### DIFF
--- a/souther-compiler/src/main/java/souther/compiler/check/InvariantChecker.java
+++ b/souther-compiler/src/main/java/souther/compiler/check/InvariantChecker.java
@@ -163,19 +163,12 @@ public final class InvariantChecker {
             "Set.contains", new Carried(1, Set.of(Shape.PERMUTES)),
             "Map.containsKey", new Carried(1, Set.of(Shape.PERMUTES)));
 
-    /** A stdlib call that states a predicate ({@code predicate}) of every element of a container
-     * ({@code container}). Stating one is stating the predicate of each element, so the elements a
-     * combinator's closure is handed carry it — which is the other direction of {@link #CARRIED},
-     * where a predicate travels from a container to one built from it. {@code List.all} is the only
-     * such call the library has. */
-    private record Quantifier(int predicate, int container) {}
-
-    private static final Map<String, Quantifier> QUANTIFIERS =
-            Map.of("List.all", new Quantifier(0, 1));
-
-    /** The constructions every element of whose result is an element of what it was built from, so a
-     * relation stated of the source is a relation of each element here. */
-    private static final Set<Shape> KEEPS_ELEMENTS = Set.of(Shape.PERMUTES, Shape.SUBSET);
+    /** The calls that state their predicate of <em>every</em> element, so what they say of a
+     * container is what holds of each element a closure is handed. Which argument is the predicate
+     * and which the container is what {@link #COMBINATORS} already answers of any combinator, and how
+     * far the statement travels is what {@link #CARRIED} already answers of any predicate — so a
+     * quantifier is the name and nothing else. {@code List.all} is the only one the library has. */
+    private static final Set<String> QUANTIFIERS = Set.of("List.all");
 
     /** Emptiness, by the size call it means. This is not a rule about what an operation does to a
      * property (spec §invariant-discharge-preservation) but about what a predicate <em>says</em>:
@@ -211,12 +204,14 @@ public final class InvariantChecker {
      * relation as the clause it was written as, together with the rule that names its free leaves, so
      * it can be read again at the element a combinator's closure is handed.
      *
-     * <p>{@code reads} is every term the clause names from outside itself, {@code container}
+     * <p>{@code through} is how far it travels: a construction of one of those shapes holds only
+     * elements of what it was built from, so what was stated of the source still holds of each of
+     * them. {@code reads} is every term the clause names from outside itself, {@code container}
      * included. A binding that takes over one of those names leaves the clause saying something else,
      * so the relation is dropped there.
      */
-    private record Quantified(String container, Ast.Block predicate, Bindings binds,
-                              List<String> reads) {}
+    private record Quantified(String container, Set<Shape> through, Ast.Block predicate,
+                              Bindings binds, List<String> reads) {}
 
     /** What the guards have settled on the current path: numeric relations, predicates known to hold
      * or to fail, and relations known of every element of a container. Threaded functionally through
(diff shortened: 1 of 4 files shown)


kawasima:
P1 を修正しました（96ff3a18）。指摘のとおり、コンテナ引数はクロージャの外側で評価されるので、`rebind` の後に引くのは順序の誤りでした。

`elementRelations` はコンテナが読まれる場所、つまりパラメータが名前を奪う前に引くようにしました。束縛が正当に届くのは述語が要素以外に読む項のほうだけなので、依存を2つに分けています。`reads` は述語の自由項で、コンテナはその横に置いて、この walk が関係を通り過ぎる束縛（`let` など）のときにだけ問われます。

回帰テストを2本入れました。指摘のケースそのものと、フィルタが要ることを示すケースです。

```souther
data Line = { quantity: Int, floor: Int }
data Cart = { items: List<Line>, floor: Int }
    invariant List.all(i -> i.quantity >= floor, items)

let toQty (cart) = List.map(cart -> Qty(cart.quantity - cart.floor + 1), cart.items)
```

節の `floor` はカートのもので、クロージャの中では同じ綴りが要素のものです。順序だけ直してフィルタを外すと、この2つを同一視して discharge します（外して実際に落ちることを確認しました）。指摘のケースのほうも、旧順序では落ちることを確認しています。

本体スイート 2035 + 101 全green、`souther-lang/examples` も通ります。

Issue #307, opened 2026-08-03
The generated example block is not in the same order twice


kawasima:
`souther examples --generate` prints a different block for the same sources on a second run. The rows are the same rows — sorting both outputs makes the diff empty — but the order of behaviors, of the rows under one behavior, and of the `// no row for ...` notes moves.

This defeats what the block is for. A generated block is read against the previous one to see what the last change to the model added, and a reordering makes every line look changed. `PatternValues` is written to answer the same string every time for exactly this reason; the ordering above it undoes that.

## Reproducing

In `souther-lang/examples`:

```
for i in 1 2 3
  java -cp <compiler classes> souther.compiler.Main examples --generate \
      (fd -e sou . crm | rg '/main/' | sort) > run$i.txt
end
diff run1.txt run2.txt | wc -l     # 219
diff run1.txt run3.txt | wc -l     #  44
diff (sort run1.txt) (sort run2.txt) | wc -l   # 0 — same content
```

Narrowed:

- The file list is fixed (`sort`), so it is not the input order.
- `-XX:ActiveProcessorCount=1` does not settle it, so it is not the order rows finish being evaluated in.
- Same content every run, different order every run, single-threaded. That is the signature of iteration over a hash container keyed by something using identity `hashCode`, somewhere between `Adequacy.Generated` and `GeneratedRows.of`.

The layers that were checked and are ordered: `Front.Layout` / `Front.Declared` (`LinkedHashMap` through `Ordered.map`), `Adequacy.Generated`'s `out` (`LinkedHashMap` over `prepared.behaviors()`), `GeneratedRows`' `byBehavior` and `said` (`LinkedHashMap` / `LinkedHashSet`), and `Generator`'s pair and single sets (`TreeSet`). Whatever is unordered is upstream of those.

## What to hold it to

Two runs of one unchanged model produce byte-identical blocks. Worth asserting on a model with several behaviors across more than one source, since a single-behavior model would pass without the ordering being fixed.

kawasima:
Fixed in #310.

PR #308, opened 2026-08-03
A format rule says what a value looks like, so one can be written

kawasima:
A format rule is the commonest thing an identifier says about itself — an office number is `[0-9]{2}-[0-9]{6}`, an invoice is `INV-[0-9]{4}-[0-9]{6}` — and nothing could write one. A record holding such a field could not be composed at all, so every combination it took part in came back as one whose values the model refused, and the generator that names the gaps could fill none of them. Twenty-seven of the format rules in `souther-lang/examples` are of this kind.

This reads the pattern and writes the shortest string it accepts. Also `Date` and `DateTime`, which had no representative either.

## What comes out

The same string every time: the first branch of an alternation, the first character of a class, the fewest repetitions a bound allows. Nothing random. A generated block is read against the previous one to see what changed, and a value that differed between two runs would make every row look changed.

```
[0-9]{2}-[0-9]{6}          -> "00-000000"
0[1-9]|[1-3][0-9]|4[0-7]   -> "01"
006[0-9A-Za-z]{12}([0-9A-Z]{3})?  -> "006000000000000"
[^@\s]+@[^@\s]+\.[^@\s]+   -> "a@a.a"
[ｦ-ﾟァ-ヴー 　]+             -> "ｦ"
[^ -~]+                    -> "あ"
```

## What it does not read, it does not guess at

A backreference, a lookaround, a named group, a unicode property: no value, which is the answer the caller already handles. What protects the constructs nobody thought about is that the result goes back through the pattern before it is offered, so reading one wrongly produces nothing rather than a value that does not match. Verified against possessive quantifiers, whose meaning is genuinely elsewhere.

And a value the pattern accepts is still refused unless a row can carry it. A string literal reads five escapes; a character outside them would go into the row as itself, and a newline would end the line the row is commented out on. A value holding one of those, or longer than an observation reads back, gets the same answer as a pattern nothing can be written for.

## Effect on the corpus

Measured against the same base, over every model in `souther-lang/examples`:

| | before | after |
|---|---|---|
| generated rows | 103 | 516 |
| combinations where every candidate was refused | 96 | 26 |

The report's own measurements are unchanged: this fills rows, it does not move what counts as covered.

## Notes

- `[^\w]` and friends: a member of a class stands for a set, and the two questions a class is asked want different amounts of it. One character is enough to be *in* the class; to be out of a negated one, all of them have to be excluded.
- The block's ordering varies between runs of the same model. That is not from this branch — it reproduces on `develop` — and is filed as #307.

### DIFF
--- a/souther-compiler/src/main/java/souther/compiler/partition/FixtureTemplate.java
+++ b/souther-compiler/src/main/java/souther/compiler/partition/FixtureTemplate.java
@@ -51,9 +51,27 @@ public record FixtureTemplate(String text, Ast.Expr value) {
                 value.signum() < 0 ? new Ast.Neg(literal, NOWHERE) : literal);
     }
 
+    /**
+     * A string, written the way the language reads one back.
+     *
+     * <p>Every escape a string literal has, because a row is offered as text to paste into a model and
+     * has to come back as the value it was made from. A tab or a newline written as itself would end
+     * the line the row is on as well.
+     */
     public static FixtureTemplate string(String value) {
-        return new FixtureTemplate("\"" + value.replace("\\", "\\\\").replace("\"", "\\\"") + "\"",
-                new Ast.StringLit(value, NOWHERE));
+        StringBuilder written = new StringBuilder("\"");
+        for (int i = 0; i < value.length(); i++) {
+            char c = value.charAt(i);
+            switch (c) {
+                case '\\' -> written.append("\\\\");
+                case '"' -> written.append("\\\"");
+                case '\n' -> written.append("\\n");
+                case '\t' -> written.append("\\t");
+                case '\r' -> written.append("\\r");
+                default -> written.append(c);
+            }
+        }
+        return new FixtureTemplate(written.append('"').toString(), new Ast.StringLit(value, NOWHERE));
     }
 
     public static FixtureTemplate bool(boolean value) {
--- a/souther-compiler/src/main/java/souther/compiler/partition/FixtureTemplate.java
+++ b/souther-compiler/src/main/java/souther/compiler/partition/FixtureTemplate.java
@@ -60,6 +60,17 @@ public record FixtureTemplate(String text, Ast.Expr value) {
         return new FixtureTemplate(Boolean.toString(value), new Ast.BoolLit(value, NOWHERE));
     }
 
+    /** A date, written the way a row writes one: the constructor applied to an ISO 8601 string. */
+    public static FixtureTemplate date(String iso) {
+        return new FixtureTemplate("Date(\"" + iso + "\")",
+                new Ast.Apply("Date", List.of(new Ast.StringLit(iso, NOWHERE)), NOWHERE));
+    }
+
+    public static FixtureTemplate dateTime(String iso) {
+        return new FixtureTemplate("DateTime(\"" + iso + "\")",
+                new Ast.Apply("DateTime", List.of(new Ast.StringLit(iso, NOWHERE)), NOWHERE));
+    }
+
     /** The absent optional, which the language names rather than any module. */
     public static FixtureTemplate none() {
         return new FixtureTemplate("None",
--- a/souther-compiler/src/main/java/souther/compiler/partition/Generator.java
+++ b/souther-compiler/src/main/java/souther/compiler/partition/Generator.java
@@ -9,9 +9,12 @@ import souther.compiler.observe.ObservedValue;
 import souther.compiler.types.Type;
 
 import java.math.BigDecimal;
+import java.util.ArrayDeque;
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.Comparator;
 import java.util.LinkedHashMap;
+import java.util.LinkedHashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Optional;
@@ -35,9 +38,10 @@ public final class Generator {
 
     /** How many rows one call will write. Past this the output stops being something a person reads
      * and pastes, and a model that wants more than this has axes it should be measured at fewer of. */
-    /** How many candidates one position is asked for. A rule can only put so many values in front of
-     * a position, and a model that put more there than this is one whose row nobody would read. */
-    private static final int MAX_CANDIDATES = 16;
+    /** How many assignments of values one row is tried at. The choices multiply, so this is a bound on
+     * the search and not on any one position — and reaching it is reported as the search having
+     * stopped, which is a different thing from every assignment having been refused. */
+    private static final int MAX_TUPLES = 256;
 
     static final int MAX_ROWS = 200;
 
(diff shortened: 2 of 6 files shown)


kawasima:
All four reproduced, and the framing in the review — the generator proposes, the constructor disposes — turned out to name the cause exactly. It was not where the rules are read.

## The cause behind 1 and 2

`Generator.compose` ended in `stands.get(0)`. A position that no axis divides took the first value offered and never asked again, so the retry that `build`'s own javadoc describes as "the same index at every position at once" reached only the axis positions. A field carrying a format rule is precisely a position no axis divides — a format has no classes to cover — which is why the candidate machinery that already existed never applied to the values this PR added.

So returning a list from `insideTheFormat` was necessary and useless on its own. The fix is that the attempt index now reaches every position, and how far the attempts go is raised by what composing reports rather than counted from the axes up front. The rows that reach a boundary had the same defect — they composed once and gave up — so both consumers now go through one walk (`attempts`).

With retry actually reaching the position, what a newtype wraps stops being an answer and becomes candidates: the base representative, plus one per readable pattern, deduped. A rule that cannot be read adds nothing and takes nothing away. Verified:

```
(?i)x                     -> V("x")     (was: no value)
[a-z]+ then x+            -> V("x")     (was: no value)
x+ then [a-z]+            -> V("x")
(?=x)x                    -> V("x")     (was: no value)
```

Both orders now give the same row, because each readable rule contributes a candidate and the constructor settles which one stands. Solving the conjunction is not needed for that and is not attempted.

For finding 2, `caseClass` and `representativesOf` now go through the same `insideTheNewtype`. That also fixes a case you did not name: a *numeric* newtype used as a sum case was reading no bound either, for the same reason.

```
Contact = Email | NoContact   -> Email("a@a")
Wallet  = NoAmount | Amount   -> Amount(100)     (invariant value >= 100)
```

## 3 and 4

`writable` checked that a character could be *spelled* and not that the value could be *written*. A lone surrogate passes `>= 0x20`, and UTF-8 encodes it as `?` — so the value reaching the file is not the value generated. It now has to survive `StandardCharsets.UTF_8.newEncoder().canEncode`, which is the property, rather than a hand-rolled pairing check. A whole pair still passes: `😀` comes back as the two chars it is.

`\r` added to `PREFERRED`; `[^\x00-\x0C\x0E-￿]` now yields it.

## Tests

All four you named, under the names you gave them, plus `aRuleLeavingOnlyACarriageReturnGetsOne`. `souther-compiler` is at 2371 green.

## Effect on the corpus

None — 516 generated rows before and after, and no report line changed. The example models happen not to contain a pattern the reader refuses, a newtype with two format rules, or a formatted newtype used as a sum case, so the corpus never exercised any of this. The tests do.

kawasima:
Both reproduced, and the framing is right: the check is the model's own constructor over the whole input, so the thing being tried is the tuple. A position's index means nothing on its own.

## The cause

The single index was tenable while only the axes had candidates — an axis's value is already fixed by the assignment being filled, so nothing was being coupled. Widening the candidates to every position turned that same loop into a walk down the diagonal of the choices. The previous round's fix created the defect it is being reported for.

So the search is now over assignments. The positions a row chooses at are collected once, by the walk the values are composed by — one place decides where the positions are, which is what keeps collecting and composing agreeing — and the assignments over them are tried breadth-first: the one where every position takes its first value, then every assignment one step from one already tried.

Outward from all-firsts rather than an odometer, for two reasons. A position's first value is the one that stands for it, so a row that took a later one everywhere is further from what the model says the row is about. And when the bound cuts the search off, it cuts it off evenly — an odometer truncated at N never moves the positions collected first off their first value at all.

Your model:

```
request.flag=No : Req { left = Left("x"), right = Right("x"), flag = No }
```

and the same with `Right`'s two invariants written in the other order gives the identical row. Three runs of each produce identical output.

## The bound

`MAX_TUPLES = 256`, on the assignments rather than on any one position, which is the unit that actually multiplies. Reaching it now reports `SEARCH_LIMIT`:

```
// no row for `request.flag=Yes` in `take`: the search stopped before reaching it
```

While fixing that I had put the count of assignments tried into `detail`. That was wrong for a reason worth recording: `detail` is what a report is *really about* when it is not the combination — a position nothing can write a value for — and `subject()` prefers it precisely so that one such position is not said a hundred times. A count is not that, and putting one there both printed a nonsense sentence and collapsed unrelated combinations into one. It carries no detail now; the combination is the subject and the reason says the search stopped.

## Effect

`souther-compiler` at 2373 green. Over `souther-lang/examples` the generated rows are unchanged — 516, byte-identical content — which is what it should be: those models' positions do have several candidates each now, and all-firsts still builds, so the walk finds the same first answer the diagonal did.

New tests under the name you gave, plus one for the bound:

```
candidatesAtDifferentPositionsAreNotWalkedInLockstep
whatTheSearchDidNotReachIsNotReportedAsRefused
```

kawasima:
All three reproduced. The first is the one that mattered and the cause was not where I had been looking.

## The check was the only unbounded thing here

`LONGEST` bounds what the reader *writes*. Nothing bounded what the engine *does* with it — and a repetition written over something of more than one length gives the engine somewhere to back up to for every copy of that thing in the string, which multiplies. Measured here before the fix:

```
(?:x?x){30} ->  40 ms
(?:x?x){33} -> 172 ms      and doubling per step
```

Every part of that is a construct the reader understands, so nothing refused it. Reading a valid model was enough to stop the compiler, and the value being checked is one this PR manufactures. A regression, as you say.

So each part of a pattern now reports what checking it will cost alongside what it writes, and an answer that would ask for more backing up than a bound gets the same treatment as a pattern the reader cannot read.

The bound is on **what the answer asks for**, not on the shape that asks. That distinction is load-bearing: a repetition over a variable-length atom is not by itself expensive, because `+` and `*` write as few copies as they can. Refusing the shape would have cost the corpus its domain-name rule —

```
[a-z0-9]([a-z0-9\-]*[a-z0-9])?(\.[a-z0-9]([a-z0-9\-]*[a-z0-9])?)+
```

— which nests them three deep, answers `"a.a"`, and is checked in under a millisecond. Counting what the emitted string asks for keeps it. All 27 format rules in the example models are untouched; the most demanding asks for 5, against a bound of 20.

```
(?:x?x){20}       -> "xxxx…"  (20 chars, 4 ms)
(?:x?x){30}       -> (none)   0 ms
(?:x?x){200}      -> (none)   0 ms
(?:(?:x?x){8}){8} -> (none)   0 ms
```

`StackOverflowError` from the engine is now caught beside the `PatternSyntaxException` that was already there.

## 2

Right, and the decomposition is exact rather than an optimisation: `check.refuse(p, …)` is per parameter and no rule can relate two of them, so the choices under one parameter never needed to multiply with another's. Each parameter is searched on its own now. Your 8+8 case builds:

```
(One { a = VA("x"), b = VB("x"), c = VC("x"), d = VD("x") },
 Two { e = VE("x"), …, flag = Yes })
```

where every field's only valid value is the second it was offered.

## 3

Fixed — `MAX_ROWS` is back under its own javadoc.

## Tests and effect

`aPatternWithNestedVariableRepetitionIsNotEvaluated` under the name you gave (run preemptively, since the failure being guarded against is no answer at all rather than a wrong one), plus `aNestedRepetitionWhoseAnswerIsShortIsStillRead` to hold the bound to costing nothing real, plus `eachParameterIsSearchedOnItsOwn`.

`souther-compiler` at 2376 green. `souther-lang/examples` unchanged: 516 generated rows, byte-identical content.

kawasima:
Reproduced, and you are right about the instrument. I have taken the estimate out entirely rather than keeping it as a heuristic in front of the budget.

```
 8 ->    4 ms       12 ->   56 ms       16 -> 6548 ms
10 ->   17 ms       14 ->  402 ms
```

The estimate counted one way a check runs away — the places a repetition over a variable-length atom puts in the answer — and scored `x*`×16 + `x{16}` at 16 against a bound of 20. Two variable-length parts side by side compete for the same characters and no shape gives that away. Fixing the estimate to catch this shape would have left the next one, which is the pattern you have been pointing at for three rounds now.

So the answer is handed to the engine through a `CharSequence` that counts what is read of it and throws at the character past its budget, with the count shared across `subSequence` and fresh per call (rows are checked on more than one thread). `MOST_READS = 100_000`. Measured:

```
x*×12 + x{12}      -> (none)   12 ms
x*×16 + x{16}      -> (none)   12 ms
x*×32 + x{32}      -> (none)    6 ms
(?:x?x){40}        -> (none)    2 ms
(?:(?:x?x){8}){8}  -> (none)    7 ms
```

Two things worth recording from doing it this way.

**The estimate was refusing patterns that cost nothing.** `(?:(a|aa)x){30}` scored 30 and was refused; the engine checks it in **60 reads**, because the `x` separates the branches and there is nothing to back up over. It is answered again. That is the false-negative cost of any shape-based bound, and it is why the reading is the right thing to count.

**The separation is not close.** The 27 format rules in the example models read between 1 and 15 characters each — the domain-name rule, the most nested of them, reads 5. The budget is four orders of magnitude above the most demanding of them and is reached in well under a millisecond.

Tests: your `adjacentVariableQuantifiersCannotExhaustTheMatcher` (at 12, 16, 20 and 32), the nesting one renamed to `aPatternWhoseAnswerCannotBeCheckedInTimeIsNotOffered` since the mechanism is no longer about nesting, and `aNestedRepetitionWhoseAnswerIsShortIsStillRead` now also pins `(?:(a|aa)x){30}` so the bound cannot quietly go back to refusing it.

`souther-compiler` at 2377 green. `souther-lang/examples` unchanged: 516 rows, byte-identical.

PR #311, opened 2026-08-03
A helper carries the type variables its body leaves open

### LINKED ISSUE #303: A helper whose body fixes only the container has no type to be given, though the same body inside core is monomorphized by expansion
A helper whose body fixes only the container is annotated, and the reason is not that the types are
monomorphic. Core's are not: a core signature's `'a` is implicitly universally quantified, and a
non-recursive core helper carrying one is monomorphized by inline expansion at each call site
(`[#stdlib]`). What closes it for a user helper is a pair of rules that meet: a user module cannot
*write* a type variable (`[#stdlib]`), and a definition's type is not read from its call sites
(ADR-0066). The type that fits the body is therefore one nothing can give it — it cannot be written,
and it will not be taken from the callers.

## Repro

```souther
let count (xs) = List.length(xs)
let firstOf (xs) = List.get(0, xs)
let sizeOf (s) = Set.size(s)
let keysOf (m) = Map.keys(m)
```

```
helper `let count` parameter `xs` is not determined by its body; annotate it with its type (spec 13.1)
```

Each names its container and nothing names the element, so `'a` stays unsolved and the report is the
one for a parameter the body left open (#302 widened what the body settles; these are what it leaves).

## The same body inside core compiles

`List.isEmpty` is this shape, written with the variable core is allowed to write:

```souther
let isEmpty (xs: List<'a>) = List.length(xs) == 0
```

It is not compiled to a polymorphic method. It is expanded at each call site with `'a` resolved to that
call's concrete type (`[#stdlib]`), which is also how a non-recursive *user* helper reaches its callers:
settled types are written back onto the parameters and the body is expanded per call (#178, ADR-0066).
A published helper travels the same way — its body crosses and is expanded at the reader's call sites
(ADR-0075). So the machinery a polymorphic user helper would need is the machinery already running for
every helper; what is missing is permission for a definition to hold a type the surface cannot write.

## What ADR-0010 decides, and what it does not

ADR-0010 limits polymorphism to the stdlib types and states that neither `data` nor `behavior` takes
type parameters. It says nothing about the type inference gives a `let`. So an internally polymorphic
helper is a question ADR-0010 left open rather than a reversal of it, and the user-facing restriction it
protects — a business model written in concrete business vocabulary, with no `'a` in it — is untouched
by the compiler inferring one it never prints in a model.

ADR-0011 was cited as the no-generics decision in ADR-0066 until #302 corrected it. It is about
`Option` / `Never` / `Unit` not being surface-writable and never said this.

## Proposal to discuss

A non-recursive helper may carry the type variables its body leaves open, monomorphized at each
expansion exactly as a core helper is. A recursive helper is lowered to a method and has no expansion to
monomorphize, so it writes its types as it does today.

Three things to decide with it.

A definition would hold a type its author cannot write. Hover and the "not determined" report would
become `(List<'a>) -> Int` — a type that can be shown and not written. That is either an honest reading
of what the definition means or a new kind of thing in the surface, and which one it is decides this.

`exposing` a helper publishes its signature, and a reader writes the arguments (ADR-0075). Expansion
still works across the boundary, so the mechanical answer is yes; what has to be decided is whether the
exposed surface may carry a variable. The narrow answer is that a published helper writes its types,
which keeps `'a` out of every place a reader reads.

The same helper would become usable at two element types in one module — `count([1, 2])` and
`count(["a"])` — which is impossible today and is what Elm's generalisation gives. This is the visible
behavioural change, and it is worth saying whether it is wanted rather than tolerated: a model that
calls one helper at two element types is a model with an abstraction over the element, which is what
ADR-0010's vocabulary argument is against.

## Not in this

A field read off a value nothing names (`let qtys (ls) = List.map((l) -> l.qty, ls)`) stays annotated:
reaching `Line` from `.qty` is a structural question a nominal model does not ask (ADR-0012), and F#
10.0 refuses the same shape with FS0072. A user module writing `'a` itself stays a compile error.

Measured at 2f94987d with the repros above, on top of #302's change (2178 compiler tests and every
module of souther-lang/examples green).

kawasima:
Closes #303.

`let count (xs) = List.length(xs)` did not compile. The body says `xs` is a List and nothing says
what it holds, so `'a` stayed unsolved and the report asked for an annotation. The same body inside
core compiles, and not because its types are monomorphic: a core signature's `'a` is implicitly
universally quantified, and a non-recursive core helper carrying one is monomorphized by inline
expansion at each call site. A non-recursive user helper reaches its callers the same way. The
machinery was already running for it; what was missing was permission.

A non-recursive helper now retains the type variables that remain inside a body-determined outer type
constructor, scoped to the definition and instantiated afresh at each expansion.

```souther
let count (xs) = List.length(xs)
let go (i) = Out { n = count(i.ns) + count([ "a", "b", "c" ]) }   // runs, answers 5
```

## What decides what

**A bare variable determines nothing.** `List<'a>` says the value is a list; `'a` says nothing. The
predicate is stated over what a type is — its outermost layer denotes a type constructor — rather
than over a list of the constructors there are, so a constructor added later means what it already
says of it. `let id (v) = v` is annotated as it was, and a parameter read only through a field still
writes its type.

**A stated answer always wins.** The walk records an open answer and keeps going; `let total (xs) =
List.sum(xs) + 1` is `List<Int>`.

**The variables belong to the helper.** A library signature is resolved once and shared by every call
site, so `List.length` and `Set.size` both hand back the `'a` the library wrote. The instantiation
that never happened is done where a declaration is read, over the variables that declaration wrote
and that call did not solve — a variable that arrived through what the position asked for was minted
already and is carried through. So what links two parameters is a position that reads them together,
not a spelling two libraries share:

```souther
let sizes (xs, ys)    = List.length(xs) + List.length(ys)         // two element types
let mixed (xs, s)     = List.length(xs) + Set.size(s)             // two, though both wrote 'a
let sameHead (xs, ys) = List.get(0, xs) == List.get(0, ys)        // one — `==` reads them together
let sameKeys (m1, m2) = Map.keys(m1) == Map.keys(m2)              // one key type, two value types
```

**The readings are unified, locally and symmetrically.** Variables are settled to types and to each
other, constructors are taken apart, a variable is refused a type it stands inside, and what was
settled is substituted through once every reading is in. Unlike the unification a call is typed with,
none of the readings is the one being checked and nothing settled here reaches past the parameter.

**What the body gets wrong is not decided here.** A body that disagrees with the type it settled is
reported by the standalone check that follows, at the position of the disagreement — which ADR-0066
already says happens to a settled type the rest of the body will not take. `let bad (xs, s: String)
= List.length(xs) + (s * 2)` names the arithmetic rather than blaming `xs`.

## What does not change

A user module still cannot write `'a`. A recursive helper is lowered to a method, has no expansion to
monomorphize, and writes all of its types. `data` and `behavior` still take no type parameters. A
helper is still a declaration expanded at each use rather than a value of a polymorphic type.

Two things the issue raised turned out not to arise on this implementation. Hover renders the
definition's own source line and never a `Type`, so a carried variable reaches no editor surface. A
published helper crosses as the source its author wrote, which carries no annotation at all, so the
reader settles it again by the same rule — no type variable is written into any published text, and
`exposing` needs no rule of its own. Tested across a jar boundary, including that the reader still
may not write `'a` itself.

## Also in here

**`check.helper.infer` splits in two.** What is left annotated is a parameter read only through a
field and one the body says nothing about, refused for different reasons. The first is told that
reaching a type from a field is a question a nominal model does not ask, so no amount of further body
will settle it.

**A minted variable is shown as `_`.** Its name is not one an author wrote or could write.

**A relation a signature makes across its own parameters does not survive the expansion of a
self-hosted callee.** `let has (xs, y) = List.member(y, xs)` annotates `y`, where the intrinsic
`Set.union` of the same shape links its two. Pre-existing — both parameters were annotated before this
— and filed as #313 rather than pinned as a rule, because which side of that line the library put a
function on is not something an author can see.

**An `example` row cannot apply such a helper.** A call settles the element from the argument it is
given; a fixture is built before there is a call to settle it. The report says that rather than that
the type is unsupported. Monomorphizing a fixture from the row's own argument is the honest end
state and is out of scope here.

**A question about a declared type is asked of what it denotes** (first commit). The inliner read the
spelling, and a reference the compiler writes carries no spelling. Not reachable today — only an
author-written return type is asked, and `Resolve` resolves every one — so this is a tidy-up rather
than a fix, made because a settled type has no spelling at all.

## Decisions

New ADR-0092, amending ADR-0066, whose decision named this repro as one that must be annotated.

**What this is.** A helper an author defined is usable at two element types in one module, which is
limited parametric polymorphism over a user-written definition however it is compiled — inline
expansion is an implementation of it, not a reason it is not one.

ADR-0010 is not reopened, but what it decides is stated more narrowly than "no user-defined generics",
because that phrase mixes what an author writes with what the language means. As four things, all
still true: no user-written type parameters, no generic `data`, no generic `behavior`, no polymorphic
values. What is added is the thing it never asked about — that a non-recursive helper is generalised
by the compiler over what its body left open, and instantiated at each expansion. ADR-0010's own
status line and its README row carry the same narrowing.

ADR-0086 is not amended: it declined a binding that generalises *without* being expanded, and recorded
that `let r = List.reverse` used at two element types already compiles.

ADR-0028 confined generics and recursion to core together; ADR-0038 returned the recursion half. This
is the same move on the generics half.

The vocabulary cost is stated in the ADR rather than hidden: `let ids (xs) = List.length(xs)` written
by an author who meant `List<CustomerId>` now compiles, where before the annotation was demanded.

## Verification

2402 compiler tests, all modules, and every module of souther-lang/examples pass, **with no existing
annotation moved**. `souther-bench` warm compile is unchanged within noise (crm 120.4ms vs 122.2ms,
issuetracker 14.7ms vs 15.1ms).


kawasima:
Pushed `d766150f`. All three blockers were reproduced first, and two of them turned out to be one root cause.

## The root cause

`Type.Var` was a name and nothing else, so every question about a variable's *identity* had to be answered some other way — by ignoring it (`CandidateMerge` treating each occurrence as its own wildcard) or by guessing from the spelling (a minted name carries a `.`). `Type.Var` now names the parameter it was minted for, or nothing where the core wrote it, and the rules ask that.

- a variable the core wrote is attached to nothing
- one minted for the parameter being settled is what is being worked out
- one minted for **another** parameter says this one holds whatever that one holds

`Type.show` asks it too rather than looking for a dot.

## 1. `CandidateMerge` — confirmed, fixed

Measured before the change:

```
merge(('a,'a),(Int,String)) = (Int, String)
merge(('a,'a),('b,'c))      = ('b, 'c)
merge(('b,'c),('a,'a))      = ('a, 'a)
```

All three are `null` now. The merge keeps what each side's variables stand for while the two are read, **in both directions** — holding only the left map fixes the first two cases but leaves the third asymmetric, which is what the reversed pair above shows.

`TwoReadingsOfOneValueTest` pins the algebra you asked for: commutativity over a table of pairs, foldability in either association, and a repeated variable holding one type. Nothing outside the call is bound, so it is still not a unification.

## 2. Bare result variables — confirmed as a rule gap, fixed at the rule

`determinesOuterType` now accepts a bare variable **that is attached elsewhere**, and still refuses one the core wrote and one minted for the parameter being settled. So `let id (v) = v` is open however the walk reaches it, and the comment no longer contradicts the code.

Two notes on the reproduction. `sameSum` does not reach the gate — `List.sum` refuses a variable element on its own ("sum needs a list of Int or Decimal"), so it fails earlier. The shape that does reach it is `let has (xs, y) = List.member(y, xs)`, and it is **still refused**, for a different reason that the rule fix cannot reach: `List.member` is self-hosted, so the call expands to a binding per parameter and the one variable its signature wrote is split across two declared types that are read independently. Keying the freshening by the call would need the expansion to be identifiable on the tree; the only thing there is the stamped position, and a nested expansion shares it, so that would claim relations the body did not make.

It is pre-existing — both parameters were annotated before this PR, and now one of them is — and it is pinned as a known limit in `aRelationInsideAnExpansionIsNotCarriedAcrossTheBindingsItBecomes` rather than left to be rediscovered.

## 3. Nested function types — confirmed, the wording was wrong

`let forwarded (fs) = countFns(fs)` with `countFns (fs: List<(Int) -> Int>)` compiles, and compiled before this PR: `settles` refused a function type at the root only, and a concrete `List<(Int) -> Int>` has a `List` at its root. So this is not something the change introduced, and refusing it now would narrow a rule that has nothing to do with type variables.

It is also the right answer. What must be written is a parameter that is **applied**; a collection of functions is a value the expansion carries like any other. The ADR sentence claiming "refused wherever it stands" was the error, and it and the comment now say what the code does and why.

## 4. ADR wording — agreed

You are right that "nothing here is user-defined" mixes syntax with semantics. A helper an author defined is usable at two element types in one module, which is limited parametric polymorphism over a user-written definition however it is compiled.

ADR-0092 now states what ADR-0010 decides as four things — no user-written type parameters, no generic `data`, no generic `behavior`, no polymorphic values — all still true, and says plainly that what is added is the thing it never asked about. ADR-0010's own status line and its README row carry the same narrowing.

## Verification

2389 compiler tests, all modules, and every module of souther-lang/examples pass, still with no existing annotation moved.

kawasima:
Pushed `d9e4ebd8`. All seven were reproduced against `HEAD` first; findings 1–4 were real and are fixed, and 2 turned out not to need `TypeOps.join` touched.

## 1 — regression, mine, fixed

Confirmed exactly as reported. Widening `visitShared`'s expectation put an open answer ahead of an arm that states the whole type.

Both are asked at each level now, in the order that says what the rule means: a type stated outright wherever it comes from, and only then one that says what the value is and leaves what it holds open. `List.length(if true then xs else [1,2,3])` takes the `List<Int>`; `List.length(if b then xs else [])` still takes the List the position asks for.

## 2 — reachable independently, fixed without touching `join`

I reproduced a second case that finding 1's fix did not cover:

```souther
let n (bb: Bool, xs, ys) = List.length(if bb then xs else ys)
```

`ys` took the position's freshly minted variable over the one `xs` already carries, so the two arms genuinely disagreed at the join.

At the open level the **arms are asked before the position**: an arm answering a variable already attached to another parameter says these positions hold one thing, where the position has only just minted one for this parameter and says nothing about the arms. With that, both arms carry one variable and `join` never meets two. So `TypeOps.join` needs no `Var` arm — the imprecision that would have required it is gone rather than absorbed downstream. I'd rather not widen a join every `if` and `match` in the compiler goes through to paper over a walk that can be right.

## 3 — confirmed, fixed

`Map<'a,'a>` × `Map<String,'b>` was `null` both ways; it is `Map<String, String>` both ways. What a variable stood for is read *with* what it stands for now rather than compared to it.

## 4 — confirmed, fixed

`merge(v, List<v>)` returned `List<'a>` — a type with no values, written back onto the parameter. Occurs check added, tested in both directions and through a tuple.

Fixing 3 exposed something the original had hidden: two variables standing for each other sent the merge into `StackOverflowError`. What each side's variables have been read as is now followed to what it comes to before the two are put together, so the recursion only ever descends through constructors. Where both readings leave a position open, one name is taken by name order, so the answer is a function of the two readings and not of which was given first. `TwoReadingsOfOneValueTest` covers the new pairs, including the ones that used to loop.

## 5, 6, 7 — fixed

Both comments claimed things the code does not do, and you are right that 6 reads as if the merge were redundant when it is the reason for it. The plain form of four type-mismatch messages printed the record `toString`; they use `Type.show` now, as the diagnostics beside them already did.

On `Type.show` rendering both a minted variable and `Type.Nothing` as `_`: noted, not changed here. `Nothing` only reaches a message through an unresolved empty collection, which has its own report, and giving minted variables a distinct rendering means inventing surface for something an author cannot write. Worth its own issue if a real message is found where the two are confusable.

## Verification

2391 compiler tests, all modules, and every module of souther-lang/examples pass, still with no existing annotation moved.

kawasima:
Pushed `7a493b01`. The blocker was real; both counterexamples reproduced.

## The fold

You are right that returning only a `Type` cannot express what the fold needs. `merge(a, b)` answering `a` drops that `b` is `a`, so the third reading is read against a record that no longer holds it — the occurs check misses `List<b>`, and `(a, b)` reduced to a type has already lost that its two positions are one thing.

The readings are given to one reader now rather than merged two at a time. `CandidateMerge` holds what each variable stands for for as long as the readings of one parameter are being taken, and `offer` hands each reading to it as the walk finds them — which is what the walk does anyway; there was never a pair to fold, only a sequence I had written as one.

Both sides are followed to what they come to before they are put together, so the occurs check asks about the value rather than about the names in front of it. `a`, `b`, `List<b>` is refused in every order, and `(a,b)`, `(b,a)`, `(Int,a)` is `(Int, Int)` in every order — both are in `whatTwoReadingsSettledIsStillSettledWhenTheThirdArrives`, in the four orderings that distinguish them.

I took the `of(List<Type>)` shape you listed. A persistent constraint environment on `BodyTyping` is the same thing one level out; putting it on the merger keeps the answer to "can these be one value" in the object that answers it, and `forget()` at the start of each parameter's walk is what makes it local. Nothing outside it stands for anything afterwards, so the ADR sentence holds.

Fixing this also exposed that my earlier occurs check was over-strict once both sides are resolved — standing for the very thing it comes to is not holding it — which is why `(a,b)` then `(b,a)` had started returning null. Covered by the same test.

## The source-level tests

Agreed, and they were missing. `CompileHelperCarriesItsVariablesTest` now has both shapes written the way an author writes them:

- `anArmThatStatesTheWholeTypeWinsOverAPositionThatStatesTheContainer` — asserts `xs` walks Ints, not that it merely compiles: it takes `[1, 2]` and refuses `["a"]`.
- `twoArmsHoldingTwoParametersHoldOneThing` — `n(true, [1], [2])` fits and `n(true, [1], ["a"])` does not, so the link is checked and not just the absence of an error.
- `aPositionStatingTheContainerReachesAnArmWhereNoArmStatesMore` — the `else []` case still takes the position's answer, and is called at two element types.

## `List.member`

Filed as #313, and the test's javadoc now says it pins what happens rather than what should. Your framing is the reason: `Set.union` is an intrinsic of the same shape and does link its two parameters, so what decides it is which side of the intrinsic / self-hosted line the library put a function on — not a rule an author can read. The issue records where it goes (the expansion has the information at the point it writes the bindings) and why position cannot be used to group them.

## PR body

Updated: the ADR paragraph now says what the ADR says — limited parametric polymorphism over a user-written definition, with ADR-0010 stated as the four things it decides — and the count is 2395.

## Verification

2395 compiler tests, all modules, and every module of souther-lang/examples pass, still with no existing annotation moved.

kawasima:
Pushed `706605f6`. The blocker was real and your reading of why is the one that fixed it.

## The counterexample

`a`, `(b, b)`, `(b, Int)` answered `(b, Int)` — the record knew `b` is an Int and the answer did not say so. Reproduced before changing anything.

## What was actually wrong

Calling this "not a unification" was hiding the requirement from me, exactly as you say. It holds variables equal, binds them to types, takes constructors apart, refuses a value the chance to hold itself, and substitutes — that is unification, done locally and symmetrically. The thing it is not is `TypeOps.unify`: one direction, and its bindings go on to the rest of the call. Nothing here outlives the parameter. Reading it as "a function that merges two types" is what left the substitution out, and it is what made me answer each counterexample with one more clause instead of finishing the abstraction.

The class says that now, and is named for what it holds rather than for the operation: `Readings`, with `add` and `answer`.

## The structure

Taken as you laid it out. Collecting and solving are one object, so there is no half-built type a caller can hold and no failure state kept in two places — `BodyTyping` has `readings` and nothing else, `offer` is `readings.add(t)`, and `typeOf` ends with `pinned != null ? pinned : readings.answer()`. A new session per parameter rather than `forget()`.

`settledThrough` follows one chain rather than everything seen, and drops each variable on the way back out, so a position beside another holding one variable is not left unsettled. You were right that this alone would have produced `(Int, b)`.

## What the property tests found

Adding them was worth more than the fix. On the first run the order property failed on `(b, b)`, `(String, Int)`, `(b, a)` — a fault none of the examples reached: a variable settled to another is only reached through it, so settling the one in hand and not the one it came to left the second standing for itself. That is fixed by settling both, and the property is what named it.

They run over every type of depth two above two variables, two primitives and the constructors a reading carries, in every order, and check: order invariance, idempotence, that the answer has nothing left to settle (`answer(answer) == answer`), that a value which would have to hold itself is refused in every order, and that a variable standing twice holds one type. All five hold.

## Verification

2401 compiler tests, all modules, and every module of souther-lang/examples pass, still with no existing annotation moved. PR body count updated.

kawasima:
Pushed `e9811bc0`. The blocker was real, reproduced before changing anything, and the standard form you laid out is what fixed it.

## The occurs check

`(a, a)` beside `(List<b>, b)` answered `(List<List<b>>, List<List<b>>)` one way and `null` the other. Your diagnosis is exact: the question was asked about the variable the reading was written with and about what *it* had come to, so when the variable that ended up being settled was on the other side, it was asked about the wrong one.

## What the shape of the fault was

Enumerating four `settleTo` calls was not a fix that happened to be incomplete — it was the fault. Once "which variable is being settled to which type" is not stated anywhere, there is no one thing to ask the occurs question about, and every counterexample needs its own clause.

Taken to `prune → bind → occurs → decompose → substitute` as you wrote it:

- `unify` takes both sides as far as what is settled says, then compares, then binds, then decomposes.
- `bind(variable, type)` is one variable and one type. One occurs check, one `settled.put`, one direction.
- Which of two variables is settled to the other is decided by name in `unify`, where the choice is, rather than inside the bind.
- `answer()` substitutes through at the end, unchanged.

`mine`, `yours`, the four `settleTo` calls and the two-sided occurs condition are all gone.

## The property tests

All three points taken.

**The universe is recursive now.** It is built by putting a constructor around what is already there rather than around leaves, so `(List<b>, b)` — the shape this fault needed — is in it, along with `Set` and a Map with a variable key. Depth two is kept to the two variables so every pair of it can still be taken.

**Groups are exhaustive where it matters.** Every pair of both universes, and every triple of a prefix of each. I left triples bounded rather than taking all ~200k orderings: pairs are where the constructor combinations live and triples are where the alias staleness lives, and the bounded set already covers both. Runtime is 1.2s; say the word if you want the full cross product and I will take the cost.

**Normalization is asked of the session that settled it.** You were right that a fresh session answers yes about a type that kept a variable the first one knew the value of. It now asks the same session twice and then reads the answer back into it.

## Wording

`Readings`, its javadoc, `HelperParams.offer`, ADR-0092 and the PR body all say the same thing now: unification, done locally and symmetrically, distinguished from the unification a call is typed with by direction and by nothing settled reaching past the parameter. You are right that leaving "not a unification" in the ADR is the risk — it is what let me treat the standard requirements as optional twice.

## Verification

2402 compiler tests, all modules, and every module of souther-lang/examples pass, still with no existing annotation moved.

PR #317, opened 2026-08-03
One application of a signature decides its variables once

### LINKED ISSUE #313: A relation a signature makes is lost when a self-hosted helper's call expands, so whether a parameter settles depends on how the library wrote it
`let has (xs, y) = List.member(y, xs)` asks for an annotation on `y`, although the body says what `y`
is: `List.member` declares `(e: 'a, xs: List<'a>)`, one variable written twice, so `y` is the element
of `xs`.

```souther
let has (xs, y) = List.member(y, xs)
// helper `let has` parameter `y` is not determined by its body; annotate it with its type
```

The same shape settles where the callee is an intrinsic. `Set.union (a: Set<'a>, b: Set<'a>)` is one,
and `let both (a, b) = Set.union(a, b)` links its two parameters and refuses a call at two element
types. `List.member` is self-hosted, so what decides this is not the language's rule but which side
of the intrinsic/self-hosted line the library put the function on.

## Where it goes

A call to a self-hosted helper is expanded, and the expansion writes one binding per parameter with
that parameter's declared type on it:

```
let $k_e: 'a = y in let $k_xs: List<'a> = xs in <the body of member>
```

The settling walk reads each binding's declared type on its own and mints the variables that
declaration left open. Two bindings, two declarations, two mints — and the fact that the signature
wrote one variable in both is gone by the time anything reads them. There is nothing on the tree
saying these bindings came from one call: the inliner stamps the call's position over the copy, so
the position is shared by nested expansions too and cannot be used to group them.

Nothing about this is new — before a helper could carry a type variable at all, both parameters of
`has` were annotated, and now one of them is. It is filed because the remaining half is decided by
something an author cannot see.

## What would fix it

The expansion has the information at the point it writes the bindings. Either it renames the callee's
unsolved variables once per call as it expands, so what lands on the bindings is already one variable
per signature variable, or it records which bindings one expansion produced so a reader can group
them. The first keeps the knowledge where it exists.

Pinned meanwhile as a known limit in
`CompileHelperCarriesItsVariablesTest.aRelationInsideAnExpansionIsNotCarriedAcrossTheBindingsItBecomes`.

Found while reviewing #311.

kawasima:
Closes #313.

`let has (xs, y) = List.member(y, xs)` asked for an annotation on `y`, although the body says what `y`
is: `List.member` declares `(e: 'a, xs: List<'a>)` — one variable written twice, so `y` is the element
of `xs`.

`let both (a, b) = Set.union(a, b)` — the same shape — settled. What told them apart was `Set.union`
being an intrinsic and `List.member` being written in Souther: an intrinsic is called, a
Souther-bodied helper is expanded. Which side of that line the library put a function on decided the
surface type inference, and nothing an author can read says where that line is.

## The root cause

**A type variable's identity had a shorter lifetime than the scope over which it means something.**

The same fault in four places, each found by measuring rather than by reading:

| where | what it did |
|---|---|
| an expansion | decided nothing — it split the signature into one declared type per binding and each was read on its own |
| a standing call | decided afresh for **each parameter being settled**, so one application said different things depending on which parameter asked |
| the readings of a parameter | settled that one variable is another, then dropped it before the next parameter was read |
| the answer | was written back before a later parameter settled something about it |

`Set.union` settled anyway, but not because the minting agreed — the second parameter's walk reads the
first one's type back out of the scope and unifies through it. The agreement was accidental.

The fix is the same move each time: give the identity the lifetime of the thing it is about. An
application, and a body.

Audited afterwards: `Type.Var` identity is now created in exactly two places, both application-scoped
(`HelperInliner` per expansion, `Freshening` per call). Everything else only tests whether something
is a variable or erases it; `TypeOps.unify` and `CallElaborator.applySignature` bind into a map built
per call. There is no fifth place.

## What changed

**Every binding an expansion writes belongs to that expansion** (`edd12f56`). Only the copied callee
body's did; the argument bindings, the lambda binder and the return carrier were minted under the
inlining pass's own owner, which is one owner for every expansion in a definition. The `k` those
carried in their names was standing in for the owner they did not have. One minter per expansion —
two minters under one owner would each start their ordinals at zero. Changes no inference result.

**A bare variable says something on grounds other than the parameter itself** (`b7ad2ad4`). "Who
minted it" was a proxy that holds only while every variable is minted for a parameter. The question is
whether the body tied the variable on grounds independent of the parameter being settled, and that is:
it stands in what **another parameter** is known to be. Not "occurs anywhere in scope" — a binding the
body made out of the parameter carries it too, and `let f (v) = let xs = [v] in List.member(v, xs)`
says nothing about `v` that `v` did not say first.

**One application of a signature decides its variables once** (`dccaa2ef`). The expansion instantiates
the callee's signature over its parameters and return together before splitting it. Renaming by name
is safe because every unsolved variable in a helper's declared types is that helper's own: a variable
enters a type only where the core writes one in a signature or where a helper's own settling decides
one, and `Type.Ref` carries no type argument. `Type.Var` stopped naming the parameter it was minted
for — that was the proxy above — and carries whether the compiler made it, which is what `Type.show`
needs to render `_`.

**What a call decides is the call's, and what the readings settle is the body's** (`9f50c7a5`). A
standing call's decision is keyed by the call, so every parameter reaching it reads the same one. What
two readings settle about a variable is kept for the body. And each parameter's type is asked for once
every parameter is in — `let bothContain (xs, ys, y)` learns its two calls' elements are one thing
only while `y` is being read.

## What I did not do, and why

**`walking` is untouched.** The plan had a phase for widening what a closure parameter may be bound
to. Measuring after the above showed `has` settles without it: what was blocking was not constraint
propagation but the same root cause left in two more places.

**Two signature positions leave no carrier behind, and are filed rather than fixed.** An expansion
writes a binding per value argument, and what the signature said about those survives on them. A
declared return (**#318**) and a function parameter the callee never applies (**#320**) leave none.

Both are the same rule failing the same way, not a different rule: the instantiation is distributed
into the fragments an expansion leaves, and those positions leave no fragment. What separates them
from this PR is the frontier, not the principle — closing #318 puts an open type through
`Elaborator`'s annotation path for the first time (measured: 17 failures across 8 classes), and #320
needs a core helper that does not apply its function parameter, which none does.

Neither is reachable today. Every self-hosted helper in core that names a variable in its result
builds that result out of its arguments, and every combinator applies its function parameter, so both
relations are reproduced by the body. `whatASignatureSaysBetweenItsArgumentsAndItsResultReachesTheCaller`
pins one of them.

I first argued that a variable-carrying declared return says nothing worth carrying, on the grounds
that it cannot fix an empty-collection seed. That was wrong: `List<'a>` says the outer is a `List`
*and* that the element is the argument's. It is recorded in #318 for whoever picks it up.

**No ADR.** I wrote one and deleted it. Against `docs/adr/README.md`'s own criterion it fails: five of
its six decision bullets were implementation mechanics — "an implementation caught up with a rule
already stated" — and the rule itself is ordinary polymorphic application semantics, which belongs in
the specification. The one part that reads as a constraint on later passes (a rewriting must preserve
what it rewrites) is not a decision anyone would make the other way. #313 is a defect, not a design
choice. `[#type-variables]` gains the rule instead:

> A single application of a function instantiates each variable its declared signature wrote once, and
> every occurrence of that variable — in its parameter types and in its result — denotes that one
> instantiated type. Two applications of one function instantiate separately. This holds however the
> application is compiled.

ADR-0092's bullet about where variables come from is brought up to date.

## Scope

This fixes the violations a program can reach and separates the ones it cannot. It does not claim the
stated rule is fully implemented: #318 and #320 are the two signature positions that leave no carrier
after an expansion, and closing the class means keeping an application's instantiation as something
the expansion carries rather than something distributed into what it splits into. Worth weighing that
before doing either by hand.

## Tests

Type-variable scope was settled first, because the expected values depend on it: the binder is the
signature, a lambda parameter carries no type, and no core body annotates with a variable — so
shadowing cannot arise, and two unrelated variables in one expansion can only come from two
signatures.

- two applications of one helper decide separately, at two element types
- two tied by one parameter hold one thing, and refuse two
- a signature variable in three positions (`Map.insert`) ties three parameters
- an expansion inside an expansion decides on its own
- a repeated signature variable is one identity whichever way the callee is written — `Set.union`
  (intrinsic) and `List.member` (self-hosted) side by side
- what was built from the parameter is not evidence about it
- which parameter is written first decides nothing
- every binding one expansion writes is that expansion's, two calls are two expansions, a nested one
  is another, and no two bindings of one expansion are the same binding
- a parameter whose readings cannot be one value settles nothing for the parameters after it
- a report about a lambda given to a function parameter names that lambda, where two nested
  combinators call theirs the same

## Verification

2440 compiler tests, all modules, and every module of souther-lang/examples pass, with no existing
annotation moved. `souther-bench` warm, measured against the base on the same machine and read on
`min` because the medians swing 140–180ms: crm 116.1ms vs 115.9ms, issuetracker 14.6ms vs 14.4ms.

### DIFF
--- a/docs/adr/0092-a-helper-carries-the-variables-its-body-leaves-open.md
+++ b/docs/adr/0092-a-helper-carries-the-variables-its-body-leaves-open.md
@@ -90,15 +90,12 @@ each expansion. A bare unconstrained variable is not a body-determined parameter
   the position of the disagreement — which is what ADR-0066 already says happens to a settled type the
   rest of the body will not take. A mistake standing beside an open element is reported as the mistake
   it is: `let bad (xs, s: String) = List.length(xs) + (s * 2)` names the arithmetic, not `xs`.
-- **The variables are the helper's own, and a variable carries where it came from.** A library
-  signature is resolved once and shared by every call site, so two unrelated calls hand back one
-  spelling — `List.length` and `Set.size` both wrote `'a`. Variables are therefore minted where a
-  declaration is read, from the variables that declaration wrote and that call did not solve, and a
-  minted variable names the parameter it was minted for. That origin is what the rules ask, rather
-  than the spelling: a variable the core wrote is attached to nothing, one minted for the parameter
-  being settled is what is being worked out, and one minted for another parameter says this one holds
-  whatever that one holds. What links two parameters is a position that reads them together, not a
-  spelling two libraries share.
+- **The variables are the helper's own.** A library signature is resolved once and shared by every
+  call site, so two unrelated calls hand back one spelling — `List.length` and `Set.size` both wrote
+  `'a`. The instantiation that never happened is done where a callee is applied, and what a variable
+  carries is whether the compiler made it: an inferred one is shown as `_`, since its spelling is
+  internal and says nothing to a reader. What links two parameters is a position that reads them
+  together, not a spelling two libraries share.
 - **Two readings hold what a variable stands for while they are read, in both directions.** A
   variable says one thing everywhere it appears, so `('a, 'a)` read against `(Int, String)` is two
   answers rather than one left open. Holding one direction only would make the answer depend on which
--- a/souther-compiler/src/main/java/souther/compiler/check/Freshening.java
+++ b/souther-compiler/src/main/java/souther/compiler/check/Freshening.java
@@ -1,84 +1,65 @@
 package souther.compiler.check;
 
-import souther.compiler.types.BindingId;
+import souther.compiler.ast.Ast;
 import souther.compiler.types.Type;
 
+import java.util.IdentityHashMap;
 import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
 
 /**
- * The variables one helper parameter carries, minted where a declaration is read.
+ * What each call standing in one body decides for the variables its callee's signature left open.
  *
- * <p>A library signature is resolved once, when the library is loaded, and every call site reads
- * that one value; nothing instantiates it. A call builds its own bindings and substitutes, so a
- * variable the call solves becomes what it was solved to and one it does not solve comes back out as
- * the variable the library wrote. Two unrelated calls therefore hand back one name —
- * {@code List.length} and {@code Set.size} both wrote {@code 'a} — and a parameter that took its
- * answer from each of them would claim the two hold the same thing.
+ * <p>A library signature is resolved once, when the library is loaded, and every call site reads that
+ * one value; nothing instantiates it. A call builds its own bindings and substitutes, so a variable
+ * the call solves becomes what it was solved to and one it does not solve comes back out as the
+ * variable the library wrote. Two unrelated calls therefore hand back one name — {@code List.length}
+ * and {@code Set.size} both wrote {@code 'a} — and a parameter that took its answer from each of them
+ * would claim the two hold the same thing.
  *
- * <p>So the instantiation that never happened is done here, at the one place where the two can be
- * told apart: reading a declaration, with that call's bindings in hand. A variable the declaration
- * wrote and the call did not solve is the declaration's own, and is minted afresh. A variable that
- * arrived through what the position asked for was minted already — by an earlier parameter of this
- * helper, or at an outer call — and is carried through untouched. The question is where the variable
- * came from, not what it is spelled.
+ * <p>So the instantiation that never happened is done here, and it belongs to <em>the call</em>. One
+ * application of a signature decides its variables once, and every parameter of the helper that reads
+ * that call reads the same decision: {@code Set.contains(v, s)} says {@code v} is what {@code s}
+ * holds, whichever of the two the walk is settling. Deciding afresh for each parameter would say that
+ * of neither.
  *
- * <p>One declaration is instantiated once, over everything it declares at that call: a signature
- * that writes {@code 'a} in two of its parameters wrote one variable, and {@code Set.union(a, b)}
- * says the two sets hold the same thing.
- *
- * <p>A minted name is stable for a given source: it names the binding the parameter is
- * and counts the declarations read for it in the order they are read, both of which the same source
- * gives the same way twice. It carries a {@code .}, which no written type variable can — the lexer
- * takes only identifier characters after the apostrophe — so it is never a name an author or the
- * core could have written.
+ * <p>An expansion follows the same rule where it inlines a call, instantiating the callee's signature
+ * once over the bindings its arguments become. Whether a call stands or is expanded is then not
+ * something the answer depends on.
  */
 final class Freshening {
 
-    private String parameter = "";
-    private int declarations;
-
-    /** Starts again for {@code target}: another parameter's variables are not this one's. */
-    void forParameter(BindingId target) {
-        declarations = 0;
-        parameter = target.toString();
-    }
-
-    /** {@code declared} with the variables it wrote replaced by this parameter's own. */
-    Type instantiate(Type declared) {
-        Map<String, Type> rename = renaming(List.of(declared), Map.of());
-        return rename.isEmpty() ? declared : TypeOps.substitute(declared, rename);
-    }
+    /** What each call has decided, by the call it is. Two calls written the same way are two
+     * applications, so they are told apart by which node they are and not by how they read. */
+    private final Map<Ast.Apply, Map<String, Type>> decided = new IdentityHashMap<>();
 
     /**
-     * What to rename the variables {@code declared} wrote and {@code solved} did not solve to.
-     *
-     * <p>Asked of the declaration as it was written, before {@code solved} is substituted into it: a
-     * variable that came in as one of {@code solved}'s answers was minted somewhere already, and
-     * asking after the substitution would see it standing in the declaration and mint it again. The
-     * caller substitutes both at once.
+     * What {@code call} decides for the variables {@code declared} left open — one fresh variable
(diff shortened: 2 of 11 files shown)


kawasima:
Pushed `4966cb47`, `6eb12040`, `23af5a55`. All four reproduced or traced first; three are fixed and one is filed with the measurement that says why.

## The common cause of 1, 3 and 4

Findings 1 and 4 are the same fault, and it is mine from this PR: **identity carried by a spelling.** #303 gave `Type.Var` an origin; `dccaa2ef` reduced it to a boolean, and the synthetic lambda name lost its `k` at the same time. Both then rested on names that are no longer unique. Finding 3 is the other half of widening a lifetime without making the writes atomic.

## 3 — confirmed, fixed

Your test sketch reproduces exactly: after the refused parameter, `a` read as `String`. A constructor is taken apart one position at a time, so what the positions before a disagreement settled is written while it is still open whether the readings agree at all. What was settled before a parameter began is what stands when it settles nothing. Pinned in `aParameterThatSettlesNothingLeavesNothingSettled`.

## 4 — confirmed, fixed

`lambdaOrigins` is keyed by `BindingId` now, looked up from `call.denotes()`, and dropped with `scopedLambdas` when the expansion is done rather than left standing. The field comment no longer describes a name that no longer exists.

## 1 — mechanism confirmed, possibility removed

The mechanism is real: each `BodyTyping` made its own record, and the count a decision's name carries starts over with the reader, so the first call two readers met was named alike — and the two answers are unified into one `bind`.

I could not write a program that shows it. Your `f (seed, xs)` compiles: the closure parameters a combinator hands out are settled by the seed and the collection it walks, so both readers meet concrete types before they meet an open one. Rather than keep hunting for a trigger I removed the possibility — there is one body, so there is one set of decisions, and the reader one step in is given it. That is the honest fix for identity that rests on a count restarting.

## 2 — you are right about the counter-example, and the answer is to narrow the sentence

`let emptyLike (xs: List<'a>) : List<'a> = []` is decisive and my "a type naming a variable fixes nothing" was wrong about it: `List<'a>` says the outer is a List *and* that the element is the argument's.

I implemented it — pass the instantiated return through and drop the guard — and measured: 17 failures across 8 classes (`CompileHelperReturnPushdownTest`, `CompileMapLibTest`, `CompileSetLibTest`, `CompileMapUpsertFoldTest`, `CompileLetAnnotationTest`, `CompileHigherOrderIntrinsicTest`, `CompileMapFixtureTest`, `CompileExampleMismatchTest`). The binding it lands on is `annotated = true`, so it puts an open type through `Elaborator`'s annotation path for the first time. That is a frontier of its own — what an annotation means when it names something the annotation cannot state — and it wants deciding on its own terms.

Filed as **#318**, with the measurement, the shape that would lose the relation, and the note that `keepDeclaredReturn`'s collection-only condition needs settling with it. Nothing reaches the gap today: every self-hosted helper in core naming a variable in its result builds that result from its arguments.

So I took your option 1 and narrowed the sentence rather than leaving spec and implementation apart. It says **result** rather than **return type**, because what the rule is about is what the application answers, not which binding a declaration is carried on — and that statement is true of everything reachable, including through the body.

Function parameters need no distribution: a lambda given to one is registered without a declared type, so no variable of the signature is written anywhere for a reader to find. The relation such a signature states between a function parameter and a value parameter reaches the reader through the value parameter, which is instantiated.

## Verification

2439 compiler tests, all modules, and every module of souther-lang/examples pass, with no existing annotation moved.

kawasima:
Pushed `80f8d6a4`, and the four items are done.

## The function parameter — you are right, and it is filed

My "function parameters need no distribution" covered the gap with the library's implementation habit rather than closing it. `let witness (f: ('a) -> Bool, x: 'a) = true` is the shape: one application decides `'a` from the lambda, and whether the callee's body applies `f` is not something a caller's typing should depend on. Today it does — the lambda disappears with the expansion and `x` keeps a bare fresh variable.

Filed as **#320**, with what it shares with #318 stated plainly: both are the same rule failing the same way, because the instantiation is distributed into the fragments an expansion leaves and these two positions leave none. Not reachable today for the same kind of reason as #318 — every combinator in core applies its function parameter — and I could not construct a failing program without adding a helper to core.

## The framing of #318 — corrected

You are right that "return type versus result" is not the separation. The separation is the frontier: closing #318 puts an open type through `Elaborator`'s annotation path for the first time. The rule covers it either way. The PR body says that now, and #320 says the same about itself.

The spec change stands for its own reason: what the rule is about is what the application answers, and saying "return type" made it read as being about which binding a declaration is carried on.

## The PR body

Rewritten rather than left to the comments. The stale spec quote, the wrong "a type naming a variable fixes nothing" claim and the test count are corrected, and there is a **Scope** section saying what this does not claim:

> This fixes the violations a program can reach and separates the ones it cannot. It does not claim the stated rule is fully implemented […] closing the class means keeping an application's instantiation as something the expansion carries rather than something distributed into what it splits into.

That is your reading of the remaining structure, and I think it is the right one — the two open issues are one decision, not two fixes.

## The lambda origins test

Added. `List.map` and `Map.map` both call their function parameter `f`; the inner one is written with the wrong arity and the report names the inner lambda's line. The stale `$k_p` in the record's comment is gone.

Same caveat as finding 1 last round, stated in the commit rather than hidden: I could not make the old code fail. Registration is outer-then-inner and the arity of an outer call is checked before the inner registers, so the overwrite never lands. The test holds the correct behaviour so a future change cannot quietly lose it.

## Verification

2440 compiler tests, all modules, and every module of souther-lang/examples pass, with no existing annotation moved.

PR #319, opened 2026-08-04
A secondary region names the file it is in, and a report says where it is said

### LINKED ISSUE #314: A secondary region cannot name a file, so one problem written in two of them is two diagnostics
A secondary region says a line and a column, and a line and a column stop saying which file once
a diagnostic is about two of them.

`LabeledRegion` holds a `Region`, which is a pair of `SourcePos`. That was enough while every
secondary was in the same file as the diagnostic it hangs off, which every one of them is today —
`Elaborator` pointing at the two branches of an `if`, `BinaryElaborator` at the two operands,
`Names` at the earlier of two imports. It stops being enough as soon as one problem is about
something written in two files.

## Where it already bites

E1919 (#306) reads a `fake` against an `example` row recording what the same behavior owes. A module's
fakes are what its attached files' rows run against and the other way round, so the two statements
need not be in one source. One disagreement therefore comes out as **two** diagnostics, one at each
statement, each quoting its own file, because a single diagnostic with a cross-file secondary would
have the renderer draw the second caret against the wrong text.

That works, and it is what ADR-0092 records. It leaves two things behind. A reader is told twice
about one problem, so anything counting E1919s counts 2N. And the cross-reference in the message
body is a bare `25:28` — useful when both statements are in one file, useless when they are not,
which is exactly the case the pairing exists for.

## What is in the way

Not the position. `souther.compiler.diag.SourceRef` is a `(sourceId, SourcePos)` already, and
`Compilation.sourceIndexOfId` maps a source id to the index a `Located` carries.

The renderers are. `DiagnosticRenderer.render(Diagnostic, SourceContext, Locale)` is handed exactly
one `SourceContext`, and `HumanRenderer.snippet` quotes lines out of it for the primary region and
for every secondary alike. `renderAll` already takes an `IntFunction<SourceContext>` — the resolver
exists — but applies it once, to `Located.sourceIndex`.

So the change is roughly:

- `LabeledRegion` carries a source alongside its region, absent meaning "the diagnostic's own".
- `Located` can answer for a secondary's source, or the renderer takes the resolver rather than one
  resolved context.
- `HumanRenderer` quotes each secondary against its own file and names the file when it is not the
  diagnostic's own. `JsonRenderer` writes it as a field.
- `Main.report` and `SoutherProcessor` hand over the resolver they already build.

## The editor is the part to be careful with

`LspDiagnostic` is `(range, severity, code, message)`. There is no `relatedInformation`, and
`Analyzer` never reads `Diagnostic.secondary()` at all — secondaries are dropped on the way to the
editor today.

That is why this cannot be done as a pure renderer change and then have E1919 collapse into one
diagnostic. Two diagnostics is what puts a marker in *both* files in the editor. Collapsing to one
with a cross-source secondary would make the warning vanish from the second file until
`LspDiagnostic`, its serializer and `Analyzer` learn related information. Either that lands in the
same change, or the collapse waits for it.

## What this would let E1919 become

One diagnostic, quoted at the recorded row, with a secondary at the stand-in naming its file. The
comparison does not change at all: `Output.Disagreements` already answers a source-independent list,
and `Output.SaidDisagreements` is the only thing that projects one disagreement onto two sources.
Replacing that projection is the whole of it.

## Not this issue

Which of the two statements the model should be held to. ADR-0092 leaves that open on purpose, and
nothing about how a diagnostic quotes two files bears on it.

kawasima:
Closes #314.

## What was in the way

A position in a diagnostic is a line and a column, and which file that is came from
whoever held the diagnostic — `Db.Found.sourceId` inside the query layer, `Located`
on the way out. So every region in one diagnostic had to be in one file, and a
problem written in two of them could only be said as two diagnostics, one at each
end. E1919 is that shape, and its cross-reference was a bare `25:28`: useful when
both statements are in one file, useless in the case the pairing exists for.

Two things were in the way, not one. **A secondary region could not name a source**,
and **a report could not be said at more than one**. Only the first is about
quoting. The second is why the count doubled, and it doubled because neither
statement is the one in the wrong, so the pair has no single file to belong to.
Fixing only the first would have left `2N`; collapsing without the second would have
forced a ranking ADR-0093 refuses to make, and taken the marker out of one of the
two files in the editor.

## What changed

**A region says which file it is in.** `LabeledRegion` carries a source, null meaning
the diagnostic's own, read through `sourceIdOr` so the null never reaches a resolver.
`HumanRenderer` quotes each region from its own file and names it when it is not the
diagnostic's; `JsonRenderer` writes it as `file`, the same field the top level uses —
the compiler's source id is a list position on the command line and a document URI in
an editor, so it names nothing a tool could hold on to.

**One projection, one fan-out.** `DiagnosticView.of(d, primarySourceId,
publishedSourceId)` answers which place is the anchor and which are elsewhere. On the
row's file the anchor is the primary region and the stand-in is elsewhere; on the
stand-in's file the two change places. Both readers call it. `allReports`, `warnings`
and `firstError` stay one entry per problem, so there is no deduplication anywhere;
`Compilation.diagnostics()` is the only place one report becomes several entries.

**A report says where it is said.** `Report.Delivery` carries the primary source and
the others it is also said at, for the reports whose key cannot answer — E1919 is
asked of the module, so its key names no source, and the recorded row may be in an
attached file. `Db.Found` derives rather than stores, so the declaration and the
resolution cannot come apart.

**One identifier.** An index is a position in the list a build passed, and a workspace
has no such list — `Compilation.ofDocuments` never fills the index map, so a secondary
naming an index could never resolve in an editor. `Located`, `CompileException` and
the renderers are keyed on the source id now, and the id a plain list of sources gets
is written once in `Compilation.idOfSourceIndex` rather than assumed at both ends.

**E1919** is one warning, quoted at the recorded row, pointing at the stand-in in
whatever file that is, and said at both. Four message keys became two. Which of the
two carries the caret is where a reader starts and not a ranking.

**The editor.** `LspDiagnostic` gains `relatedInformation` and `Analyzer` reads
`Diagnostic.secondary()`, which it never did — the nine existing secondaries were
dropped on the way there. Both files keep a marker, on the statement each of them
wrote, and each links to the other.

## What this looks like

```
-- EXAMPLE (WARNING)  E1919 -----------------clash.sou:10:48

10|     | "m-1 is a member" : (MemberId("m-1")) -> Found { id = MemberId("m-1") }
                                                   ^

clash.examples.sou:4:28
4|     | (MemberId("m-1")) -> Missing { why = "no such member" }
                              ^
the `fake findMember` row

This row and the `fake findMember` row state different answers for the same input.
Hint: The row says Found { id = "m-1" }; the fake says Missing { why = "no such member" }.
```

## Worth a look in review

Secondaries in the diagnostic's own file also become `relatedInformation`, so nine
existing diagnostics gain linked locations in the editor. Restricting it to
cross-file would mean the editor shows less than the terminal for a reason a reader
cannot see, but it is a visible change to diagnostics this issue is not about.

`Analyzer.diagnostics(String text)`, the single-document path, emits no linked
locations: a link is a URI and a document compiled from its text alone has none. The
server itself uses the workspace path.

## Not an ADR

ADR-0093 already recorded that if secondary regions gained source identity the two
projections would become one diagnostic with a cross-source secondary and the
comparison would not change. This is the implementation catching up with a rule
already stated, which `docs/adr/README.md` excludes. Its Consequences and the README
row are corrected and point here; the Decision is left as the record it was.

## Verification

`mvn clean install` then `mvn test`: 2463 + 113 + 73 + 1 pass, 13 of them new.
`souther-examples` builds clean against this compiler.


kawasima:
Pushed 21bfeb03 with the first three points fixed. The fourth is a policy question I've left for you.

## 1 — regression, fixed

Confirmed before fixing: `Compilation.ofSource(...).diagnostics()` returned `{"0": []}` with one report in `allReports()`.

`ofSource` said "a compile of one source names none" by clearing the index map, and that map *was* the exposure. Moving to ids put the rule in `sourceIdOf(Found)`, which the bucket routing also went through. Split as you suggested:

- `primarySourceIdOf(Found)` — private, always answers. What `diagnostics()` files by and what the ordering reads.
- `sourceIdOf(Found)` — public, null when the compile names none. What `warnings()` and `firstError()` put on a `Located`.

One difference from your sketch: inside `diagnostics()` the `Located` carries the **resolved** id, not null. The bucket key and `Located.primarySourceId()` have to agree or the projection has nothing to anchor on, and a map keyed by source id is already naming sources — suppressing the id inside it while keying by it outside would be incoherent. So the assertion is `assertEquals("0", found.get("0").get(0).primarySourceId())`, and the "names none" rule is asserted on `sourceIdOf(Found)` instead. `WhereADiagnosticIsSaidTest` covers both.

## 2 — fixed structurally rather than by a check

You're right that the type didn't guarantee it. Rather than validate the list, I removed it: `Delivery` is `(primarySourceId, saidAtEveryRegion)` and `publishSourceIdsOf` reads the targets off the diagnostic's own regions. There is no longer a way to name a file the report points at nothing in.

The boolean keeps the distinction I wanted between quoting and belonging — pointing at a definition in another module to explain a mistake here doesn't make that module mistaken — but the targets are derived, not restated. E1919 is `Delivery.atEveryRegionOf(recordedRowSource)`, and the second place it names is gone from the call site, so the two can't disagree.

`DiagnosticView.of` throws `IllegalArgumentException` now. I checked it can't be reached from a compile before making it throw: `diagnostics()` builds the buckets from `publishSourceIdsOf` and the `Located` from the same `primarySourceIdOf`, and both renderers call `of(d, own, own)`, where the primary spot always matches. So it's an invariant check, not a crash waiting for input.

## 3 — fixed

`SourceNames.of(paths)` gives each file the shortest tail of its path no other file in the compile shares, falling back to the whole path. Both `Main` and `SoutherProcessor` use it. One file, or files differing in their last segment, keep their basenames, so existing output is unchanged.

```
-- EXAMPLE (WARNING)  E1919 --------domain-a/model.sou:10:48

10|     | "m-1 is a member" : (MemberId("m-1")) -> Found { id = MemberId("m-1") }
                                                   ^

domain-b/model.sou:4:28
4|     | (MemberId("m-1")) -> Missing { why = "no such member" }
                              ^
the `fake findMember` row
```

I kept `file` as one field meaning one thing (the name a reader opens) rather than splitting display from identity. The compiler's source id is a list position on the command line and a document URI in an editor, so there is no stable identity to publish alongside it.

## 5 — fixed

The internal list keeps a null-tolerant copy and `ofAllInSources` documents `NO_SOURCE` entries. No caller passes one today; the type no longer contradicts its own Javadoc.

Also fixed the dead `Db.Found#publishSourceIds()` link. The `Delivery` invariant you flagged went away with the redesign — there is no list left to normalise.

## 4 — your call

`v0.1.0` is tagged and there is a Central publishing plugin, so this is a real question rather than a hypothetical. My reading is that deprecated adapters are the wrong trade here: `Map<String, List<Diagnostic>>` cannot express a diagnostic anchored in a file other than the one it is filed under, so keeping it would keep a shape that cannot say what this PR exists to say, permanently. `diagnoseModules` and `ofAllInSources` are seams between this project's own modules — they are public because Maven artifacts are, not because they are a supported surface.

If you want the adapters anyway, or a release note instead, say which and I'll add it. I haven't touched it either way.

## Verification

2475 + 113 + 73 + 1 pass (21 new since the last push). `souther-examples` still builds clean.

Issue #320, opened 2026-08-04
A signature position with no carrier after expansion loses what the signature said, and a function parameter the callee never applies is one


kawasima:
Split out of #313 on review. Same rule, different frontier — and #318 is the other half of it.

## The shape

An expansion writes a binding per value argument, and what the callee's signature said about those
survives on them. A signature position that leaves no binding behind has nowhere to put what it said.
There are two:

- **the declared return** — filed as #318
- **a function parameter the callee never applies**

For a function parameter, the lambda is registered as a scoped helper and β-reduced where the callee
applies it. If the callee applies it, the relation is reproduced by that application. If it does not,
the lambda disappears with the expansion and nothing carries what the signature said about it.

```souther
let witness (f: ('a) -> Bool, x: 'a) = true
```

```souther
let use (y) = witness((z) -> z == 1, y)
```

One application decides `'a` — the lambda says `Int` — so `y` is an `Int`. Whether the callee's body
uses `f` is not something a caller's typing should depend on. Today `x`'s binding carries a bare fresh
variable and the lambda is gone.

## Not reachable today

Every self-hosted combinator in the shipped core applies its function parameter — `List.map`,
`List.filter`, `List.foldFrom`, `Map.map`, `Set.map` and the rest — so the relation is reproduced by
that application in each of them. `witness` is not a function anyone has needed. I could not construct
a failing program without adding such a helper to core.

That is why this is filed rather than fixed: it is a hole in the implementation of a stated rule, not
a defect a program can reach.

## What it shares with #318

Both are the same statement — a single application instantiates each variable its signature wrote
once, and every occurrence of that variable denotes that one type — failing for the same reason: the
implementation distributes the instantiation to the AST fragments an expansion leaves, and some
signature positions leave none.

Closing either by hand is possible. Closing the class means keeping the application's instantiation as
something the expansion carries rather than something distributed into what it splits into. Worth
weighing which of the two is wanted before doing the first one.

Measured at `23af5a55`.

kawasima:
Closed by #321. Every function parameter boundary states its own type and is read against what the function handed to it is declared as — whether or not the callee applies it.

Issue #277, opened 2026-08-02
A named function in value position is an η-expanded block, and closure conversion has no Core form


kawasima:
`Core.FunctionRef` was considered while ADR-0086 was taken and left out. A function reference in value position becomes a function value; whether that value is folded into a compact IR node or expanded to a `Core.Block` every time is a representation choice, and it belongs with the explicit closure-conversion form `Core.Block`'s javadoc is waiting for.

This is the issue that was deferred to. What is left in it is that one question.

## What is settled

Reifying a name is one thing built one way. `HelperInliner.etaExpand` builds the block; `declarationArity` answers for a helper, a library function and a behavior alike, because what reification needs is how many arguments the block takes and not what stands behind the name. #271 closed with it.

Two things this issue used to say are gone. The rounding-mode exception left with ADR-0087, which made `RoundingMode` ordinary data. And the pipeline order it proposed — `Resolve` → the type checker → `Lower` → the backend — is not the order the compiler has: `Lower` runs *before* the type checker, deliberately, because a behavior's permission and `depends on` are defined on the inlined body (spec 12.5) and because inlining is what carries a settled parameter type onto the binding a call becomes (#178). Reading the proposal against the current `Lower` is reading it against the wrong stage.

## The question

A named function in value position is an `Ast.Block` the inliner writes, which the type checker turns into a `Core.Block` carrying the `Type.FnOf` the context fixed. Should it instead be an explicit function reference, with a closure-conversion form to emit it?

## Why it is not a node to add

A non-recursive helper is fully inlined at its call sites and never emitted, so it has no method for a reference to land on. `Core.FunctionRef(name)` alone would point at nothing. Something has to decide where the code lives:

- lift the helper body into a synthetic method, or
- have the reference carry the body and let closure conversion generate the method, or
- keep the η-expanded block and share or de-duplicate it inside the backend.

That decision also fixes function identity, what a capture is, how many classes are generated and what a generated method is a method *of*. It is an IR design, not an addition.

## Where it would go

Not in the current `Lower`, which is untyped and runs before the check. `Core` is produced by the type checker and `Core.Block` already carries its type, so the candidates are:

1. `Elaborator`/`TypeChecker` emits the reference form directly from `Ast.Var`.
2. A closure-conversion pass after the check, over `Core`.
3. A lift in the backend, taking today's `Core.Block` as input.

A design that wants a Lower-shaped stage after the typed Core has to introduce that stage; it is not the one that exists.

## When

[...]

PR #321, opened 2026-08-04
An expansion owns the instantiation of the application it stands for

### LINKED ISSUE #318: A declared return that names a type variable is not carried onto the binding an expansion writes, so a body that states nothing about what it answers loses the relation
Found while doing #313, and deliberately left out of it.

An expansion carries a helper's declared return onto a binding (`let $r: T = <body> in $r`) so that a
body whose own type is less specific than its declaration — an empty-collection seed the body cannot
type — takes the declared one. It refuses to do that when the declared return names a type variable:

```java
if (written == null || written.cases().size() != 1
        || !carriesCollection(written.cases().get(0))
        || mentionsTypeVar(written.cases().get(0))) {
    return body;
}
```

Before #313 the refusal was right: the variable was the library's own and said nothing at a call site.
After it, a call instantiates its callee's signature once, so the variable in a declared return is the
same one its parameters carry — `Map<'k, 'a>` says the result holds what the argument held. The
refusal now drops a relation the application decided.

## What it costs today

Nothing I could construct. Every self-hosted helper in the shipped core whose declared return names a
variable builds its result out of its arguments, so the relation reaches the caller through the body:
`Map.upsert`, `Set.filter`, `Map.filter` and `Option.withDefault` all do, and
`whatASignatureSaysBetweenItsArgumentsAndItsResultReachesTheCaller` pins one of them.

The shape that would lose it is a helper whose body says nothing about what it answers and whose
declaration is the only evidence:

```souther
let emptyLike (xs: List<'a>) : List<'a> = []
```

There is none in core today. So this is a latent gap rather than a defect anyone can reach, which is
why it is filed rather than fixed.

## Why it was not fixed with #313

Passing the instantiated return through is one line. The binding it lands on is `annotated = true`, so
it puts an open type through `Elaborator`'s annotation path — `checkLetAnnotation`, `annotatedType`,
`checkAnnotatedLambdaBindings` — which has never seen one. Measured: it breaks
`CompileHelperReturnPushdownTest`, `CompileMapLibTest`, `CompileSetLibTest`, `CompileMapUpsertFoldTest`,
`CompileLetAnnotationTest`, `CompileHigherOrderIntrinsicTest`, `CompileMapFixtureTest` and
`CompileExampleMismatchTest` — 17 failures across 8 classes.

That is a frontier of its own: what an annotation means when it names something the annotation cannot
state. It wants deciding on its own terms rather than as a rider on #313.

Also worth settling with it: `keepDeclaredReturn` carries only a collection-bearing return, because
"a scalar return has nothing to fix". With variables that is no longer obviously true — a scalar
return naming a variable says the result is what an argument holds.

kawasima:
Closes #318 and #320. They are the same rule failing the same way, so they close together.

## What was wrong

A signature is one statement. `emptyLike (xs: List<'a>) : List<'a>` says the result holds what the argument held, and it says it by the two occurrences of `'a` being one variable.

An expansion wrote that out as a binding per argument, each carrying its own declared type. Nothing was left saying the two were ever the same variable — and a signature position that leaves no binding at all has nowhere to say anything. Two positions leave none:

- the declared return (**#318**)
- a function parameter the callee never applies (**#320**)

#317 got the instantiation right — one application decides each variable once — but distributed it into the fragments the expansion leaves. These two are the fragments that do not exist.

## What is here

**`Ast.Expansion`** — the expansion is one node. It holds the callee's signature instantiated once, over its parameters and its result together, with the arguments and the body: `bound` for the value arguments, which become bindings, and `given` for the function arguments, which do not.

**`Type.MetaVar`** — what a signature is instantiated into: a variable one application decides. Apart from `Type.Var`, which is a declaration's and must hold for every use of it. Both sit under the new `Type.Open`, so a reader whose question is only "does this stand for something else" says that and not more. Identity is the pair `(application, spelling)` — `BindingOwner.Expansion` already named one application — so nothing depends on a name being made unique by hand.

**`Substitution`** — one application's answers, keyed by the variable itself. It reports a disagreement at the call rather than swallowing it on the grounds that another pass would report it too: it is the only reader holding two positions of one variable at once. A variable nothing pinned is read as the bottom (ADR-0028) rather than the declaration being thrown away — `Map<String, _>` still says the result is a map with `String` keys.

`keepDeclaredReturn` and the `$r` binding it wrote are gone. The declaration is on the node, so it reaches the body whether or not it names a variable, and the collection-only condition it carried has nothing left to decide.

## Three things this turned up

**A settled type may still be open, and writing it back generalizes it.** `let has (xs, y) = List.member(y, xs)` settles its parameters by reading its body, and what it settles them to names the expansion's variables. Written onto the helper's parameters those variables are the helper's now, not that expansion's — one decides at each call of `has`, the other decided once. `HelperParams.settle` converts them, with one map per helper so a variable two parameters share stays one.

**A declared result is not a demand.** The settling walk asks what a position demands of a parameter. A result the callee declared is what it promises its caller, so reading the body against it settles the parameter by a type the call site never asked for. Measured: it changes what `Map.upsert`'s callers settle to.

**A variable still open is not a type.** It may not reach code generation, which now refuses one, and `Elaborator.expansion` asserts that nothing it hands out holds one. `assignable` is left permissive about it: the settling walk flows open types on the actual side by design, and did so before this with inferred variables.

## What is unchanged

The invariant checker, at 2000-odd lines and seven readers of a binding. What may be discharged is a question about which value reached which position, and to that reader an expansion is the bindings it writes — `Ast.asBindings` gives it those, and the file is untouched.

## Tests

`CompileTypeVariableTest` holds both, in the reserved namespace a type variable may be written in.

For #318 the agreeing half proves nothing on its own: the empty-collection bottom fits every element type, so a call that answered it passes wherever the right answer does. Both halves are held — the call that answers a list of Strings is refused where the position wants Ints — and the refusing half was checked against the old behaviour restored, where it does not fire.

For #320, `witness (f: ('a) -> Bool, x: 'a)` now determines its caller's parameter although it never applies `f`; a lambda whose own body says nothing about its parameter still settles nothing, which is what says the answer came from the lambda.

Verified: souther 2493 + LSP 113, and `souther-lang/examples` in full.


kawasima:
Pushed d79c6feb. Verified each finding against `develop` first — the baseline matters for what this PR is answerable for.

## The root cause behind 1, 2a, 2b and 3

They are one thing. `Elaborator.expansion` read `bound` and `declaredReturn` and never `given`, and `Substitution.unify` collected answers without anyone asking, beside it, whether two types that name no variable agree. The node owned the whole signature and the one reader that decides types read part of it — which is the closing section of the review, and it is right.

So the fix is one statement rather than four patches: the expansion reads every position it holds.

## 1, 2a, 2b — `given` is read

`Elaborator.expansion` reads the function arguments, and the whole function type, parameters and result together. `(f: ('a) -> Bool, x: 'a)` relates the argument to what `f` takes; `(f: (Int) -> 'a, x: 'a)` to what it answers. Reading one position would carry one and drop the other, which is what `HelperParams.fromGiven` was doing — it reads the closure's result now too.

**The signal for "the callee never applies it" was wrong, and finding it cost a suite run.** I first asked whether the expansion's body still reaches the argument. It never does: applying a function parameter is what removes it, because the application β-reduces to the lambda's body. Every combinator in core came out as "never applied" and got its closure typed a second time in the wrong scope. It is asked of the callee **as written** now.

Named functions are read the same way, so `witness(positive, s)` with `s: String` is refused.

## 3 — a declared type that states one is held to

`ignores("not an Int")` was accepted on `develop` too, so this is not a regression — but the expansion is where it should have been caught, and it is now. `Substitution.unify` stays what it is: it collects what the application decided. The ordinary assignability question is asked beside it, of every argument whose declared type says what a value there is. A variable does not say that, whichever kind it is, so neither a `MetaVar` nor a `Var` is held against.

## Two things this turned up in `Substitution`

- It wrote a decided variable through before matching, so a later reading had nothing left to rebind. `Option.withDefault([], xs)` reads the variable as a list of nothing first, and the reading that says what the value holds has to win. Neither side is written through first now.
- The bottom widens at whatever depth it turned up, not only at the top — the same rule as ADR-0028, asked where the bottom actually was.

## 4 — not reachable

The example does not parse:

```
`List<'a>` cannot be a union member: a member is a type that is nominal, tells itself apart at
run time, and can be written as a `match` type arm
```

`'a` alone is refused for the same reason. A union member must be nominal, and the language has no generic data, so **no multi-case declared return can name a type variable** — there is no relation there to lose. The condition is left as it was, and its comment says why. If a ground multi-case return should fix an empty seed in a body, that is a separate question that predates this PR; say the word and I will file it.

## Tests

All four asked for, plus the ground one, in `CompileTypeVariableTest`:

- an unapplied function parameter holds the lambda to what it takes
- a variable in a function argument's result reaches the caller (both ways)
- a named function given to an unapplied parameter says what it takes (both ways)
- a parameter the callee never reads is still declared

Verified: souther 2497 + LSP 113, and `souther-lang/examples` in full.

kawasima:
Pushed 3875261a. All three confirmed against `develop` first, then fixed; the suite is green at 2501 + 113 and the examples repository in full.

## What the three had in common

An open type was being read as saying *nothing*, rather than as saying everything around the position the variable stands at. That is one mistake in three places, so it is one fix.

## 1 — a variable is a hole, and the type around it still states what it states

`Substitution.unify` collected answers and left agreement to whoever asked. Nobody asked. It is two operations now: `decide` records what a reading says about the variables, `constrain` holds the value to the type as well.

`fits` walks the declared type with a variable as a position that states nothing and everything else stating what it states — constructors, arities, ground positions at any depth. `ignores(1)` against `(xs: List<'a>)` is refused for not being a list; `ignores((x) -> x + 1)` against `(f: ('a) -> Bool)` for not answering a Bool.

One thing that turned up here: `TypeOps.assignable` compares no two function types — it has no `FnOf` case and falls through to the leaf-case test, so any two distinct function types were "not assignable". Handing function types to it would have been wrong in the other direction, so the shapes are walked in `fits`.

## 2 — one entry for an application's arguments

`arguments(ex, env, ctx)` is the only place an application's arguments are typed, and `expansion()` and `elaborateFunctionValue(Ast.Expansion)` both come through it, differing only in what they do with the body afterwards. What a callee answers decides that, and decides nothing about what it was given.

## 3 — who reads a function argument, not whether anyone does

`applied` no longer gates *whether* the signature is checked. It gates only where an open position is read from, and the actual gate on reading is whether the signature has **settled** the position.

Getting this wrong twice is worth recording, because both wrong versions passed the four tests you named and failed the suite:

- **Dropping the gate entirely** re-types every closure. `Map.fold`'s accumulator is `'a`, decided to `Map<_, _>` by an empty seed. That is an answer *so far* — reading the step at it fixes the step at what the seed happened to be, and the widening the step supplies arrives too late. So a type still carrying the bottom does not count as settled.
- **Reading an applied lambda anyway** breaks on scopes. Two things a scope cannot answer are left to the body that applies them: a name standing for a block the inliner is holding, which nothing binds at all, and a function reaching a declaration this scope does not reach — which declarations a scope reaches depends on where it was built.

Your case C works through this rather than around it: a named function reaches a function parameter **η-expanded**, so `use(wrong)` arrives as a block, its parameter type is stated, and it is read and refused for answering `Int` where `String` was declared.

## `HelperParams`

It uses the same kernel, in `decide` form — inference records what a reading says and reports nothing, which is that walk's contract. It no longer has a unification of its own.

## Tests

- `CompileTypeVariableTest` — a type naming a variable states what is around it; a function type naming one still states what it answers (both, both ways)
- `CompileClosureTest` — a helper answering a function still holds its arguments to what it declared
- `CompileFunctionValueFlowTest` — a function argument is held to what the parameter declares, named or not (both ways)

## 4

Withdrawn as you say — a union member must be nominal, so no multi-case declared return can name a variable.

kawasima:
Pushed da5103b5. Both confirmed by probe first: the two cases compiled, and now they do not.

```
what `use` is given to answer expects String, but got Int
```

## [P1] — the asymmetry is gone

The entry was all-or-nothing where the kernel behind it was not. `Substitution.constrain` already read a hole as one position; `reading` decided whether to call it at all, and answered no whenever any parameter was unsettled.

Two readings now, and the second is the one that was missing:

- **The positions it takes are settled** — read the argument at them and hold the whole function type. Unchanged.
- **They are not** — read what it answers on its own. Each parameter the declaration left open is read off the lambda's own body, the body is typed with those in force, and what comes back is held to the result the declaration states.

The parameter types read that way are deliberately **not** written back into the substitution. They came off the lambda alone, and the application may settle them wider — a fold's step read in isolation would fix the accumulator at whatever the step happens to build. They answer this one question and no other. That is what keeps `Map.fold` working while `('a) -> String` is checked.

A named function reaches the parameter η-expanded, so `use(wrong)` goes through the same reading with no separate path.

## [P2] — the catch is narrowed

`catch (CompileException | Unanswerable _)`. Those are the two ways a body has no type to give: it does not type, or it depends on a name that denotes nothing. Anything else is this compiler being wrong about itself, and reading that as "no answer" would leave a signature unread over it — which is the same class of silence the rest of this PR is about. It matters more than it did: `readFromBody` is called from the check now, not only from the settling walk.

## Tests

`CompileTypeVariableTest` — what a function argument answers is held even where what it takes is open, in four combinations: lambda answering the declared type and not, named function answering it and not.

Verified: souther 2502 + LSP 113, and `souther-lang/examples` in full.

kawasima:
Pushed 19e658bc. All three probed first — they compiled, and now they do not:

```
what `use` is given to answer expects String,  but got Int      (unread parameter)
what `use` is given to answer expects List<_>, but got Int      (open result)
what `witness` is given to answer expects Int, but got String   (result tied to a value argument)
```

## The diagnosis was exact

The kernel read a hole as one position; what handed it evidence still had to have the whole function type or none. So a signature was read or not read for reasons beside what it said. Three such reasons are gone.

**A parameter the lambda never reads is not read at all.** What a function answers cannot turn on something it does not look at. Before, an unread `x` made `answering` give up before it got to the result.

**A result is held to what it states however many variables stand inside it.** The `settled(want.result())` gate is gone. `fits` was already the right reader for this — it answers a variable as a position that states nothing and everything around it as stating what it states.

**A bottom in a declared position is a hole too.** `fits` now answers `Type.Nothing` on the declared side the way it answers a variable. That is your point 4, and it is also what keeps `Map.fold` working once the gate is gone: the accumulator settles to `Map<_, _>` from an empty seed, and that is a reading so far — but the shape around it is not, and it is now read.

## Hold and decide are separate

`Substitution` has `hold` beside `constrain`. Evidence read off an argument on its own **holds** it and **decides** nothing: a lambda's parameters read off its own body may be narrower than what the application settles them at, so what its body answers under them is enough to refuse a value the declaration does not admit and not enough to decide a variable by.

This is where I did not follow your point 3. You asked that a bare `MetaVar` result be decided from the evidence. I hold it and do not decide it, because the evidence was gathered under parameter types the application has not agreed to — deciding a fold's accumulator from its step read in isolation is the failure I hit two rounds ago. Nothing is lost by not deciding here: the settling walk learns that relation through `fromGiven`, and by elaboration the value arguments have decided it. If you want it decided here as well, say so and I will make the evidence carry which positions were the application's and which were guessed.

## Point 5

Constraints are written against the declaration **as it stands**, not against what it has been read through to. `givenFunctions` keeps the raw `Type.FnOf` and uses the zonked one only to ask what is settled.

## Tests

`CompileTypeVariableTest`, both directions each:

- a hole at one position of a function type does not silence the others — an unread parameter with a ground result, and an open result that still states its constructor
- what a value argument decided is held against what the function answers

Verified: souther 2504 + LSP 113, and `souther-lang/examples` in full.

kawasima:
Pushed a56238fa. Both cases fixed, and this time by removing the thing that made them possible rather than by adding a reader for them.

## The cause was upstream of every check

A lambda given to a function parameter was registered as a helper with **no parameter types and no declared return**:

```java
new Ast.FnDef(f.name(), lparams, null, new Ast.FnBody.Written(lambda.body()), lambda.pos())
```

So β-reducing it threw the signature away at the point it was reduced. From there nothing between the caller's function and what was declared of it survived, and every check of a function argument had to be reconstructed *outside* the position it belonged to. That reconstruction is where all four rounds of gates came from — one per way it could fail: is every parameter settled, is the result ground, is every parameter inferable, does the type hold a variable anywhere.

**The lambda carries the signature now.** It is registered under what the callee declared of the parameter, this application's variables written in, so where the callee applies it that application expands like any other call and is read against the signature *there* — in the one place the types this application decided are in force. No new node kind, no new check: it is `Ast.Expansion` doing what it already does.

Deleted with it: `answering`, the parameter reading it needed, and every gate. `givenFunctions` now reads only what the callee never applies — the case with no application to be read at, which is #320's.

## Two things that fell out

**An application reads the enclosing one's decisions.** A reduced function carries the variables of the call that supplied it, so `Substitution` records a reading where the variable belongs rather than shadowing it, and `Scope` carries what is in force over it the way it carries bindings. That is what makes `(f: ('a) -> 'b, x: 'b)` hold what the function answers to what the value argument decided — your E3, which the previous round only got by reading out of position.

**Your second case was a live bug in `fits`.** The `if (!Type.mentions(want, MetaVar)) return assignable(...)` early-out was still there, so `List<Nothing>` never descended. It is gone: every position is asked on its own, with no test for whether the type holds a hole somewhere first — that test *is* the question, and asking it about the type as a whole is exactly what let one position silence the rest. A bottom is a hole too.

## On the test sprawl

Nine tests that each pinned one way of losing a position are now one table: ten declared positions, run both ways — what the declaration admits and what it does not. The rows vary the things that used to change the answer and no longer do (whether the callee reads the parameter, whether it applies the function, whether the lambda reads what it was handed, lambda or name). Net −211 lines of tests, +98.

If a new way to lose a position turns up, it should be a row, not a test.

Verified: souther 2514 + LSP 113, and `souther-lang/examples` in full.

kawasima:
Pushed 56db84e2. Confirmed by probe first — the applied relay compiled, and now it does not:

```
what `g` was given to answer expects String, but got Int
```

## The invariant now holds at every boundary

You named it exactly: the signature-bearing representation was built at one kind of boundary and not the other. A lambda given to a function parameter was registered under what that callee declared of it; a **name** was substituted through, so the receiving parameter's declaration was never written anywhere and the boundary between the two held nothing.

Both are registered the same way now. A name is handed on as a function of the receiving parameter's declared type, applying the one it was given:

```
use.f  boundary:  $f : (Int) -> Int
inner.g boundary: $g : (Int) -> String,  body = $f(x)
```

What applies it expands like any other call and is held to the declaration it arrived at rather than the one it came from. No new check — the same `Ast.Expansion` path.

## One name is deliberately not wrapped

A `Var` this expansion will not reduce — a binding holding a function chosen at run time — is left substituted as before. It is applied as a *value*, its own type is what that application reads, and there is no boundary to make; wrapping it named a binding where it does not stand, which `anUnannotatedFunctionAppliedOnlyInsideALambdaIsToldToStateItsType` caught immediately.

## Tests

Two rows in the table, not two new tests — applied and never applied, each with a pair that agrees and a pair where the two boundaries disagree (`use` takes `(Int) -> Int` and hands it to `inner`, which declares `(Int) -> String`). Checked non-vacuous: with the inliner change reverted, the applied row fails.

The table is 22 cases now, in the file that states the rule once.

Verified: souther 2516 + LSP 113, and `souther-lang/examples` in full.

kawasima:
Pushed 89f4f9e4. Your case is closed, and the wrapper is gone with it.

```
`a function` of `inner` expects (Int) -> String, but got (Int) -> Int
```

## The wrapper was the same mistake one level up

You wrote that `expands(...) == null` is a special case. It is worse than that: it is the *same* condition as before, wearing a different coat. The wrapper made a boundary by **building** something, so it could only be made where something could be built. `let bad (f: (Int) -> Int) = inner(f)` has nothing to build from, and lost the receiving declaration exactly as before. Making the wrapper unconditional would have hit the next thing that cannot be wrapped.

So I stopped building.

## A boundary already has both declarations at it

- **what this callee wants of the position** — on the parameter
- **what the function handed to it is** — wherever it is bound: a helper's own parameter, a binding, a function an enclosing call supplied (whose declaration that call wrote on it, carried along on the boundary)

Both are written down. The boundary reads them against each other. Nothing is built, nothing is re-typed, and it holds wherever a name reaches a function parameter — at a call site, in a helper checked on its own, through as many hands as it is passed. The `expands` condition is gone because there is nothing left for it to gate.

Only a **lambda written at the call** has no declaration of its own, and that is the one case still left to the application that decides it — which is where it belongs, since that application is what decides what it takes.

## Two things that fell out

**Held, not decided.** What the function is declared as is another declaration's, and its variables are that one's to settle. The two are read against each other and neither decides the other's.

**A variable is a hole whichever kind it is.** `fits` treated `MetaVar` as a position that states nothing and sent `Type.Var` to ordinary assignability, where a rigid variable compares equal to nothing. Both stand for something a use decides; both are holes now. That was the last place the rule was stated for one kind and withheld from the other.

## Tests

Two rows in the table, run with no call site at all — your exact shape — plus the call-site pair already there. The table is 24 cases.

Verified: souther 2518 + LSP 113, and `souther-lang/examples` in full.

kawasima:
Pushed aabbd6b0. Both cases confirmed by probe first, and both now refused:

```
`a function` of `inner`: expected Int but got String
```

## This was the branch's own defect, at the last place it was left

`('a) -> 'a` is a function answering what it was given, and neither of its positions says that alone. Read a position at a time it admits one taking an `Int` and answering a `String` — the one thing it refuses. That is exactly "one variable read at two positions with nothing saying they are one", which is what #313 and #318 were about; the boundary was the last reader still doing it.

Each side is instantiated **once** and the two are read against each other. The receiving declaration already was, when the call was expanded; the arriving one is instantiated at the boundary — one variable per variable, not per occurrence. So the readings of one variable meet, and disagreeing is what they do when the declaration cannot be satisfied.

Read in a layer of its own, as you suggested: what the two declarations settle between them is that reading's. A variable the receiving side wrote twice must be read at one type *here*, and that says nothing about what the call as a whole decides it to be.

## And it uncovered the same mistake one level down

Making the boundary decide turned up spurious conflicts in the invariant tests:

```
BIND var='a@demo.V/List.map#1  held=INT  at='a@demo.V/List.allUniqueBy#0  zonkedAt=INT
```

`bind` compared the variable **as written** rather than what it stands for. `INT` against `'a@allUniqueBy#0` — which *is* `INT`. So every reading through a variable another application had already decided disagreed with every other. It never fired before because nothing reached `bind` with a foreign variable on the right.

That is the same failure as the boundary's, one level down: reading a variable instead of following it.

## Tests

Two rows, one per side that writes the variable twice — the receiving declaration (`('a) -> 'a` given `(Int) -> String`) and the arriving one (`('a) -> 'a` where `(Int) -> String` is wanted) — each with the pair that agrees. Checked non-vacuous: instantiating per occurrence instead of per variable makes the refusing half pass. No call site in either, so it is the boundary being read and nothing else.

The table is 26 cases.

Verified: souther 2520 + LSP 113, and `souther-lang/examples` in full.

kawasima:
Pushed e8691d92. The ownership finding is right and the layer is gone.

One correction on the reproducing shape, because it changes what the test has to be. Your two examples are already refused today:

```
emptyFrom(positive) ++ [ "wrong" ]   →  list elements disagree on type: INT vs STRING
witness(positive, y),  use("wrong")  →  `y` of `use` expects Int, but got String
```

A **named** function reaches a function parameter η-expanded, so it arrives as a block and never takes the `arrivesAs` path at all — it goes through the reading that already settles on the application. The boundary layer only ever applied to a function arriving under a **name**, which is a function parameter handed on:

```souther
let emptyFrom (f: ('a) -> Bool): List<'a> = []
let relay (g: (Int) -> Bool) = emptyFrom(g)
let taking (xs: List<String>) = true
let bad (g: (Int) -> Bool) = taking(relay(g))   -- accepted before, refused now
```

There the layer threw away `'a := Int` and the position wanting `List<String>` got one.

## The layer was compensating for the bug I fixed one commit earlier

I introduced it because deciding at the boundary raised conflicts against readings that *agreed* — `INT` against `'a@…` which is `INT`. That was `bind` comparing a variable as written rather than what it stood for, and the layer hid it by keeping the readings apart. Fixing `bind` on its own terms left the layer with nothing to protect, so it is gone and the boundary settles on the application like every other reading.

Two readings of one variable meeting is not a hazard to be contained. It is the only thing that says a signature is one statement.

## On the ownership table

The freshened arriving variables do land in the application's substitution, under a spelling of their own. Nothing else reads them and they cannot collide with the receiving ones, so no separate owner is needed for correctness. The union-find you describe is already how it behaves: `bind` links two open variables and `zonk` follows the chain, so a decision reaching either reaches both.

## Tests

Two rows — a named function and a function handed on — each asserting that a position wanting a list of Strings is refused and one wanting Ints is not. Checked non-vacuous: restoring the layer makes the refusing half pass. The table is 28 cases.

Verified: souther 2522 + LSP 113, and `souther-lang/examples` in full.

kawasima:
Pushed 1ac3d232. I checked this one before changing it, and it goes the other way — the reported case does not reproduce, and the proposed fix introduces a false acceptance. Evidence for both.

## The reported case is accepted today

```souther
let identity (x: 'a): 'a = x
let use (f: (Int) -> Int, g: (String) -> String) = true
let call = use(identity, identity)      -- COMPILED
```

A named function reaching a function parameter is **η-expanded to a block** (#277), so it never arrives under a name and never takes the `arrivesAs` path. Two readings of `identity` are two blocks, each read at the position it was given to. The pair also refuses what it should:

```souther
let use (f: (Int) -> Int, g: (String) -> Int) = true
let call = use(identity, identity)      -- `a function` of `use` expects (String) -> Int, but got (String) -> String
```

## Per-boundary freshening would break a real relation

The path that *does* arrive under a name is a helper handing its own function parameters on. There, two parameters written in one signature share their variable — that is what one signature means:

```souther
let use (f: (Int) -> Int, g: (String) -> String) = true
let call (p: ('a) -> 'a, q: ('a) -> 'a) = use(p, q)
```

`q` **is** `p`, so a call wanting two different function types must be refused. It is, today. With freshening moved to `application × boundary × name` it is accepted — I built that version and measured it. So the boundary is the wrong unit here; it would be the mirror of the defect you found last round, in the other direction.

## What was actually wrong: the unit was implicit

You are right that `(application, spelling)` is under-specified on its face. The unit is the **scope of the declaration that wrote the variable**, and within one expansion that is one scope — the enclosing definition's signature. What arrives from anywhere else arrives already instantiated, carrying the variables of the call that supplied it, with no spelling left to share. Nothing in the code said so, which is why the namespace looked arbitrary.

That is stated now, and pinned in both directions rather than left to be re-derived:

- two function parameters of one signature share their variable — a call wanting two different types is refused
- two readings of a declaration written elsewhere are independent — `use(identity, identity)` at two element types is accepted, and a reading that cannot be satisfied on its own is refused

The table is 30 cases.

Verified: souther 2524 + LSP 113, and `souther-lang/examples` in full.

If you can point me at a shape where two genuinely independent declarations arrive under names in one call, I will take it — I could not construct one, because every route that produces two independent copies instantiates them before they get here.

PR #325, opened 2026-08-04
Every position carries the source it was read from

### LINKED ISSUE #309: A name error in an  file is reported against the model file
An unresolved value name written in an `examples for` file is reported with the position from that file but the *name* of the module file. The line and column are right; the file they are attributed to is not.

## Reproducing

Two files. `shippingfee.sou` declares the model, `shippingfee.examples.sou` holds the rows. Rename a case in the model — say `北海道沖縄` becomes `北海道` — and leave a row still naming the old one:

```
-- 名前が見つかりません -------------------------shippingfee.sou:25:14

25| data 一般
                 ^^^^^

`北海道沖縄` という名前の値がスコープに見つかりません。
```

`北海道沖縄` does not appear in `shippingfee.sou` at all. It is at `shippingfee.examples.sou:25:16`, and line 25 of the model file is `data 一般`, which the caret then underlines. The line number matching is a coincidence of the two files: the position comes from the examples file and the source id from the model file.

The quoted source line is read from whichever file the diagnostic names, so what a reader is shown is an unrelated declaration with a caret in the middle of it.

## Where it is

Not from the adequacy work — it reproduces on `develop` with the same output.

Renaming a case is exactly when this fires, and it is the moment the rows most need to be pointed at. `example` rows in an attached file are the only ones affected; a name error in the model file itself is attributed correctly.

kawasima:
Closes #309.

## What was wrong

A line and a column are enough while one file is being read. They stop being enough
the moment a module's writings are not all in one file, which is what an attached
`examples for` file is for: its rows, fake tables and values join the module they
are written for, and after that nothing about a coordinate says which file it came
from.

Everything downstream then had to work the file out again from something else, and
each place that did so got it wrong in its own way:

- `Compilation.primarySourceIdOf` fell back to the module's own source, so a name a
  row could not resolve was reported against the model file at the row's line
  number — an unrelated declaration with a caret in the middle of it. Renaming a
  `data` case is exactly when this fires.
- A helper expanded from another module keeps that module's coordinates, so a
  problem found in one was reported on this module at a line number belonging to
  the other one's file.
- `Names.spans` matched an editor's cursor by coordinate alone over the merged
  module, so line 8 column 14 was two places and the question was answered about
  whichever the merge reached first. Go-to-definition, hover and rename described a
  name the cursor was not on.

## What this does

The fix is not to attribute these diagnostics better. It is to make a position
carry the source it was read from, and then to stop reconstructing that fact from
anything else. ADR-0044's last consequence deferred exactly this — *"until a file
handle is added to the position"* — and this adds it.

**A position carries its source.** `SourcePos` gains a `sourceId`, given where a
position is made from a text (`LineIndex`). That is the one place with both halves
to hand, so it reaches every position a parse makes without an AST walker and
nothing downstream has to work it out again. `CstFrontend.firstError` builds its own
index, so it is threaded too — otherwise a syntax error would be the single kind of
mistake still reported against whatever file the reader guessed at. A position read
from no source of this compile — a synthesized node, the prelude, a module read back
off the module path — names none, and says so.

The source is part of what makes two positions the same position. Line 25 of two
files is one coordinate and is not one place, and a value whose identity denied one
of its components would leave "the same position" meaning something different in
every container that held one — which is the original mistake, re-expressible. It
costs almost nothing to say so: no main-source change at all, and two assertions in
`ResolvedValueNamesTest` that now name the file they were already about.

**Where a report is said is read off where it points.** A key says which input was
asked about; a position says where the caret sits, and the line under the caret is
quoted out of the file the report is filed under. Where the two disagree, answering
with the key's shows a reader a line they did not write. So the order is now the
report's own claim (`Report.Delivery`, for a problem that belongs somewhere other
than where its caret sits), then the position, then the key, then the module.

The claim is honoured only if the compile has that source. A position may come from
the prelude or from a module off the path, and filing a report under a name
`diagnostics()` has no bucket for drops it in silence — a problem the author never
sees is worse than one on the wrong file. The guard reads `Front.Ids`, not the
index a compile of a plain list keeps, because a workspace never fills that one and
reading it there would have left the editor unfixed.

Read before the reports are collected, so one problem two questions found stays one
and two problems at one coordinate in two files stay two.

**A cursor is on a place, not a coordinate.** A name covers a cursor when it is on
the same line of the same file, and the editor says which file it is asking about. A
caller with only a coordinate to give is still answered by coordinate. Where a
declaration was written is read off its position too: working it out from the module
sent a row's `let` to the module's own file at the coordinate it has in the other
one — which, for a row naming a value declared beside it, is the cursor's own line,
so the cursor landed back where it started.

Three concepts that were one name are now three: what a report or key explicitly
named (`Report.Delivery.primarySourceId`), what a report claims
(`Db.Found.claimedSourceId`), and what this compile can actually show
(`Compilation.locatedSourceIdOf`).

Nothing changes in `HumanRenderer`, `JsonRenderer`, `DiagnosticView`, `Located` or
`LabeledRegion`. The renderer already took the file from `Located.primarySourceId()`
and the line from `Diagnostic.pos()`; the whole of this is making those two agree
upstream.

## Where it is asked

`APositionKnowsItsFileTest` is about the position alone, so a failure in the
parse-time plumbing is not mistaken for one in what reads a position — including
that two files' one coordinate is two places, which pins the identity decision.

`AMistakeInAnAttachedFileIsSaidOnThatFileTest` is the issue, one method per way of
getting it wrong: an unknown value named by a row, an unknown value and an unknown
type and a type error in the attached file's own fixture, a constant construction,
and a syntax error. Each asserts the file **and** the line, because either alone
passed for the wrong reason before — the two files' line numbers happened to agree,
which is what made the original report so confusing. The model file is long and the
attached file short on purpose, so a misattribution lands on a line about something
else. `theQuotedLineIsTheLineTheAuthorWrote` renders through `HumanRenderer` and
asserts the quoted text; it is the only one here that would have caught the bug on
its own.

`WhereAReportIsSaidIsReadOffWhereItPointsTest` fixes the order as a contract rather
than as whatever the implementation happens to do.
`AFileThisCompileDoesNotHaveIsNotAFileToFileUnderTest` takes the guard apart per
compile entry point, and includes the availability half — a report must not vanish.
`WhereADiagnosticIsSaidTest` gains dedup in both directions: one problem two
questions found is one, two problems at one coordinate are two, and one coordinate
in two files is two.

`AMarkerLandsInTheFileTheMistakeIsWrittenInTest` is the editor: the marker is on the
attached file, the model file is clean, and it moves when the attached file is
renamed without its contents changing. That last one runs on one `Analyzer`, so it
goes through the workspace compile's incremental path, which is where carrying
provenance beside a position rather than in it would have gone wrong.

`ACursorIsOnAPlaceNotACoordinateTest` — one at the query, one at the editor. The
query one is built so both files write a name at line 8 column 14 and the two names
denote different values, because otherwise a wrong answer still looks right.

## What this costs

`Compilation.sourceIds()` is a `List`, so the guard's membership test is linear in
the number of sources. Diagnostics times sources is small enough to leave alone.

`Compiler.compileModules` orders its first error by which source it is in, so a
report that crosses a file boundary moves in that order and a compile with two
errors can raise the other one. Nothing in the suite depended on it.

## Left out

Navigation from a cursor **inside** an attached file. `Compiler.moduleNameFromHeader`
answers nothing for `examples for m`, so the resolution-based path is skipped there
and navigation falls back to matching by spelling — which is why a local is out of
reach in a companion file. That is a gap of its own rather than this defect, and
wants its own issue.

## Specification

No rule changes. ADR-0044 gets an amendment: the file handle its last consequence
deferred is now on the position, and why the source counts towards a position's
identity.

### DIFF
--- a/docs/adr/0044-diagnostics-as-data-with-renderer-layer.md
+++ b/docs/adr/0044-diagnostics-as-data-with-renderer-layer.md
@@ -76,3 +76,25 @@ Two decisions bound the scope.
 - A multi-file build has no per-diagnostic file identity (SourcePos carries no file), so a snippet is
   quoted only for a single-file compile; a linked build renders the frame and message without the
   source line until a file handle is added to the position.
+
+## Amendment: the file handle is on the position (issue #309)
+
+The last consequence deferred a file handle on the position, and a compile worked out which file a
+report belonged to from the question that found it instead. That holds while each question is about
+one file and stops holding where it is not: a module's `example` rows, fake tables and values are
+written in the module's own source and in any number of attached `examples for` files, and once they
+are gathered under one name a question asked about the module can only answer with the module's own
+file. What the author was then shown was the right line number quoted out of the wrong file — an
+unrelated declaration with a caret in the middle of it.
+
+So `SourcePos` now carries the source it was read from. It is given once, where a position is made
+from a text (`LineIndex`), so no later pass has to reconstruct it, and it reaches every position a
+parse makes without an AST walker. A position read from no source of this compile — a synthesized
+node, the standard library, a module read back off the module path — names none, and a reader falls
+back as it did before.
+
+The source is part of what makes two positions the same position. Line 25 of two files is one
+coordinate and is not one place, and a value whose identity denied one of its components would leave
+"the same position" meaning something different in every container that held one — which is the
+original mistake, re-expressible. Where coordinates alone are wanted, they are compared as
+coordinates.
--- a/souther-compiler/src/main/java/souther/compiler/frontend/AstBuilder.java
+++ b/souther-compiler/src/main/java/souther/compiler/frontend/AstBuilder.java
@@ -39,21 +39,24 @@ public final class AstBuilder {
     private int patternCounter = 0;
     private int spreadCounter = 0;
 
-    private AstBuilder(String source) {
-        this.lines = new LineIndex(source);
+    private AstBuilder(String source, String sourceId) {
+        this.lines = new LineIndex(source, sourceId);
     }
 
     /**
      * Builds a module from a parsed source file. A header-less source is named
-     * {@code defaultModuleName}; a {@code null} default makes the header required.
+     * {@code defaultModuleName}; a {@code null} default makes the header required. Every position
+     * the module carries names {@code sourceId}, which is null for a source the caller has no name
+     * for — a module read off the module path, which is in no source of the compile reading it.
      *
      * <p>Requires a CST the parser accepted: every node it reads has the children the grammar gives
      * it, and a missing one is dereferenced rather than reported. {@link CstFrontend#parse} is the
      * one caller and raises the parser's first error before building — which is why this is
      * package-private, and what a second caller would have to guarantee.
      */
-    static Ast.Module build(SyntaxNode sourceFile, String source, String defaultModuleName) {
-        return new AstBuilder(source).module(sourceFile, defaultModuleName);
+    static Ast.Module build(SyntaxNode sourceFile, String source, String defaultModuleName,
+                            String sourceId) {
+        return new AstBuilder(source, sourceId).module(sourceFile, defaultModuleName);
     }
 
     // --- module ---
--- a/souther-compiler/src/main/java/souther/compiler/frontend/CstFrontend.java
+++ b/souther-compiler/src/main/java/souther/compiler/frontend/CstFrontend.java
@@ -26,13 +26,20 @@ public final class CstFrontend {
     }
 
     /** Parses one compilation unit, naming a header-less source {@code defaultModuleName} (a
-     * {@code null} default makes the {@code module} header required). */
+     * {@code null} default makes the {@code module} header required).
+     *
+     * <p>The positions it makes name no source. A caller that has a name for the text it is handing
+     * over — a compile reading one of its own sources — parses through
+     * {@link #parseWithSlices(String, String, String)} instead, so that what it gets back says which
+     * file each position was read from. A caller here has no such name: the standard library and a
+     * module read back off the module path are in no source of the compile that is reading them. */
     public static Ast.Module parse(String source, String defaultModuleName) {
         CstParser.Result result = CstParser.parse(source);
         if (!result.errors().isEmpty()) {
-            throw firstError(source, result.errors().get(0));
+            throw firstError(source, null, result.errors().get(0));
         }
-        return ImplicitUnits.expand(AstBuilder.build(result.root(), source, defaultModuleName));
+        return ImplicitUnits.expand(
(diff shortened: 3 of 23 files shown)


kawasima:
レビューありがとうございます。3点とも対応しました（`f7f2973d`）。

## `Names.sameSource` — 修正しました

ご指摘のとおりです。「caller with only a coordinate」の判定対象は `at` であって `start` ではなく、`start.sourceId() == null` を通していたのは、このPRが `SourcePos` に持たせた意味論と矛盾していました。非対称にしました。

```java
private static boolean sameSource(SourcePos start, SourcePos at) {
    return at.sourceId() == null || at.sourceId().equals(start.sourceId());
}
```

ご提示の2本に、ファイル違いの1本を足してテスト化しています（`ACursorIsOnAPlaceNotACoordinateTest`）。

現状の回帰有無も確認しました。model + attached の compile で `Names.Resolution` の候補集合を数えたところ、source を持たない候補は **values=0 / denotations=0** です。`new SourcePos(` はリポジトリ全体で5箇所しかなく、後段のパスは既存ノードから位置をコピーするだけなので、合成ノードも読まれた位置の source を引き継ぎます。つまり今日の挙動は動かず、変わったのは「2つの absence のうちどちらが caller の licence か」を契約として書いたことです。コミットのタイトルはその一文にしました。

## secondary region の二重表現 — 不変条件を入れました

`LabeledRegion` のコンストラクタで、両方が非nullかつ不一致なら拒否します。片方がnullのケース（手書き位置から作った region、`null` = 「診断自身のファイル」を意味するラベル）はどちらも表現可能なままです。

**リアクタ全体のテストはこの不変条件を入れた状態で通ります**。つまり現時点で label と region が食い違っている cross-file secondary は存在しません。`Output.SaidDisagreements` の `elsewhere` は両者が一致するとき `null` にする書き方なので、この不変条件と整合しています。

`Region` の source を正とする最終形については同意です。`LabeledRegion.sourceId` の削除は `Spot.secondary` / `publishSourceIdsOf` / renderer を巻き込むので、このPRには入れていません。

## attached file 内からの navigation — issue を作りました

#326 に切りました。原因（`Compiler.moduleNameFromHeader` が `examples for m` に対して何も返さないため、`typeUnderCursor` / `valueUnderCursor` がコンパイラに問い合わせる前に抜ける）と、必要なもの（`Front.Layout.exampleFileTargets` / `Front.AttachedTo` が既に持っている「寄与先のモジュール」）まで書いてあります。

ご指摘のとおり「A cursor is on a place, not a coordinate」の完成条件とは別です。そのコミットが直したのは、**同じ座標が2つの場所になったときにどちらが答えられるか**であって、attached file から問い合わせられるようにすることではありません。#326 の下地にはなっていて、モジュール名さえ分かれば残りは既に動きます。

Issue #327, opened 2026-08-04
Equal Set and Map values can encode differently after a hash collision


kawasima:
`a == b` does not imply `encode(a) == encode(b)` for a `Set` or a `Map`. Boundary encoding is not
extensional over value equality.

## Mechanism

Iteration order is decided by the element/key hashes — except inside a hash-collision bucket.
`HashCollisionNode.pairs` appends on insert (`PersistentHashMap.java:598-603`) and `EntryIterator`
walks that array in order (`:656-701`). So for any two elements sharing a full 32-bit hash, the
iteration order — and therefore the encoded JSON array or object — follows the order the collection
was built in.

Two ways to reach it:

- **Colliding strings.** `"Aa"` and `"BB"` both hash to `2112`. `Set.fromList(["Aa","BB"])` and
  `Set.fromList(["BB","Aa"])` are equal values that encode to different arrays. The same applies to a
  `Map` with those two keys.
- **Any enumeration, systematically.** `ValueClassGen` emits `hashCode` as `iconst_1` folded over the
  fields (`:425-446`), and a unit data has no fields, so **every unit data hashes to `1`**. Every
  `Set` holding two or more enumeration cases is therefore in a collision bucket, and
  `Set<Stage>` decoded from `["Won","Lost"]` and from `["Lost","Won"]` encode differently.

## Why it matters

An author who assumes an idempotent response body gets one most of the time. This reaches golden
tests, response diffs, cache keys, ETags, content addressing, and any comparison made after
encoding.

## What this issue has to decide

#299 removed the specification's claim that the same members always yield the same order, because the
implementation does not keep it — the specification now promises only an implementation-defined order
that nothing may depend on. That closed the documentation inconsistency and deliberately left this
question open:

1. **Accept the non-extensionality** and state it, so `a == b` with `encode(a) != encode(b)` is a
   documented property rather than a surprise.
2. **Canonicalize** — either the collision bucket or the encoding. Note that a plain `Comparable`
   sort does not do it: a `Set`'s element only needs *equality*, so a bucket can hold product data,
   sums with field-bearing cases, `Bool`, or nested collections, none of which have an ordering
   (`TypeOps.isOrdered`). This needs a total order over arbitrary Souther values, which puts it
   adjacent to canonical-encoding work.

A narrower fix — giving a unit data a deterministic per-type hash — would remove the systematic
enumeration case but not the general one, since a finite-width hash always has collisions.

Found while resolving #299.

[...]

PR #328, opened 2026-08-04
What a collection promises about its order is what it can keep

### LINKED ISSUE #299: A Set's order at the boundary is unstated, and with an Int element it looks sorted while with a String it does not
The spec says what a `Set` looks like at the boundary — "a JSON array, deduplicated on decode" — and
says nothing about the order of that array. For `Map` it does say: iteration order is a deterministic,
implementation-defined hash order, reproducible but not to be depended on (§stdlib-map). A `Set`
crossing the boundary behaves the same way, and nothing states it, so what an author concludes depends
on which element type they happened to try.

## What it actually does

```souther
data Tags = { tags: Set<String> }
data Out = { tags: Set<String> }

behavior echoTags : (t: Tags) -> Out
    constructs Out
let echoTags (t) = Out { tags = t.tags }
```

```
$ souther run h1.sou --behavior echoTags --input '{"tags":["zebra","apple","mango"]}'
{"tags":["zebra","mango","apple"]}
$ souther run h1.sou --behavior echoTags --input '{"tags":["apple","mango","zebra"]}'
{"tags":["zebra","mango","apple"]}
$ souther run h1.sou --behavior echoTags --input '{"tags":["apple","apple","zebra"]}'
{"tags":["zebra","apple"]}
```

Reproducible, independent of the order the elements arrived in, deduplicated. That is the `Map` rule,
and `PersistentHashSet` says so in its own javadoc — "deterministic, implementation-defined hash order
(not first-seen order)". The behaviour is not the problem; the silence is.

## Why the silence costs something

The same model with an `Int` element:

```
$ souther run h2.sou --behavior echoRanks --input '{"ranks":[3,1,2]}'
{"ranks":[1,2,3]}
$ souther run h2.sou --behavior echoRanks --input '{"ranks":[30,7,19,4]}'
{"ranks":[4,7,19,30]}
```

Sorted — or rather, hash order for small integers is ascending, which is indistinguishable from sorted
until the element is a `String` and it is not. An author who checks the shape of a response with an
`Int` set learns a rule that does not hold, writes the assertion the output invites, and finds out when
the element type or the membership changes.

In `souther-lang/examples`, `catalog`'s `Requirement.purchased` is a `Set<Sku>` in a behavior's output,
and the smoke test asserts membership rather than the array because that is the only assertion the spec
supports. That is the right test to write; nothing in the spec tells the author it is the only one
available.

## Proposal to discuss

Say the rule for a `Set` at the boundary where the `Set`'s external representation is defined, in the
same words `Map`'s iteration order already uses: deterministic, implementation-defined hash order,
reproducible for the same elements, not to be depended on. That alone closes the finding.

The larger question is whether the boundary should commit to more than that. A JSON array *reads* as
ordered to whoever consumes it, and a response body is a contract with them: a golden test, an ETag, a
diff between two runs of a report. Encoding a `Set` whose element type has an ordering (<<equality>>
already requires equality for a `Set`'s element and defines which types have an ordering) in that
order at the boundary would make a response body stable in a way an author can state, and leave the
in-memory iteration order exactly as it is. The cost is a sort per encode of a set, paid only where a
set actually crosses the boundary.

Worth deciding as one question, since "unordered in the model, ordered on the wire" is a position the
spec can hold, and "implementation-defined both places" is also one — but only one of them is written
down today, and it is written for `Map`.

Found while reviewing the open findings in `souther-lang/examples` (F29 there, whose own wording —
"not byte-stable" — the probes above correct). Verified against develop at 280f7be0.

kawasima:
Closes #299.

A `Set` crosses the boundary as a JSON array. Seven places in the repo described what order that
array is in, at three different strengths — and the strongest of them is not true.

## What was wrong

- **`specification.adoc:533` (§collections)**, where a `Set`'s external representation is *defined*
  ("a JSON array, deduplicated on decode"), said nothing about order. This is the silence #299
  reports.
- **`docs/adr/0039-set-collection-type.md`** said the opposite of the implementation: "Encode and
  `toList` use first-seen (insertion) order". It was left behind when the runtime collections became
  persistent tries (`e45ad66d`, recorded as a baseline in ADR-0042).
- **`specification.adoc:2945`, `:2949`, `:2912` and four javadocs** promised that the same members
  always yield the same order.

That last one is the load-bearing mistake. Iteration order is decided by the element hashes *except
inside a hash-collision bucket*, where `HashCollisionNode` holds its pairs in the order they were put
(`PersistentHashMap.java:598-603`, `:656-701`). Two `Set`s that are equal can therefore iterate — and
encode — differently: `"Aa"` and `"BB"` share a hash, and every unit data hashes to `1`
(`ValueClassGen.java:425-446` folds `iconst_1` over the fields), so *any* `Set` of two enumeration
cases is in a bucket.

## What this does

Removes the promise rather than qualifying it. A finite-width hash always has collisions, so no
better hash closes the gap; only a canonical order over Souther values would, and that is a separate
question — filed as #327.

What a reader can now rely on: the order is implementation-defined, `toList` and the boundary use the
one order and not two, and an input array's order carries no meaning and is not guaranteed to
survive. "Deterministic" is dropped from the specification because it invites the deleted reading
back — the order is deterministic *for the trie*, not for the value. The javadocs may name the trie,
and now say which of the two they mean.

The javadocs are fixed here rather than deferred: leaving them asserting what the specification just
dropped would move the inconsistency instead of closing it, which is the mistake ADR-0039 already
made once.

## Not done, deliberately

- **No sort at the boundary.** A `Set`'s element only needs *equality*, so sorting only where an
  ordering exists would make wire semantics depend on the element type — preserving the trap #299
  reports rather than closing it. A natural order is also not a canonical encoding: that needs `Map`
  key order, field order, `Decimal` notation, and the rest, and belongs with #327.
- **No input-order-independence test.** It would promote into a collection-wide contract a property
  that holds only for non-colliding values; choosing `String` elements does not make it safe.
- **No characterization test** pinning the `"Aa"`/`"BB"` divergence — it would fix a known defect as
  expected behaviour and have to be inverted by #327. The reproduction lives in that issue.

## Verification

`PersistentHashSetTest` (9), `PersistentHashMapTest` (14), `CompileSetCodecTest` (1),
`CompileSetLibTest` (3), `CompileNestedCollectionCodecTest` (1) — all pass.

Behaviour is unchanged by construction: every edit outside `specification.adoc` and the ADR is a
comment, which `git diff` over the Java files confirms — no non-comment line changed.


kawasima:
Pushed `a303c2bf`. All three points were right; the first was a defect I introduced in this branch, not pre-existing.

## 1. Map's boundary order — fixed

Confirmed against the code before changing anything. `Maps.mapKeys` rebuilds the entries into a fresh `PersistentHashMap.Builder` indexed by the *rendered* key, and `needsKeyRender` is true for everything except a bare `String` (`CodecGen.java:196-198`, `:1418-1422`). So `Map<ProductId, V>`, `Map<Date, V>` and `Map<Stage, V>` are written out in the rendered map's traversal, not the source map's — exactly the `"@"` / `"A"` inversion you describe, since a newtype hashes as `31 + value.hashCode()` and the string it renders to hashes as itself.

`and it is what the boundary encodes in` is gone. §stdlib-map now says the boundary order is unspecified too, and that it is not `toList`'s wherever a key is rendered, naming the three key kinds.

The `Set` half stands and is kept: `SetEnc` calls `Sets.toList` with no re-keying (`CodecGen.java:1411`, `:1506`), so `toList` and the boundary really are one order there. I added the contrast explicitly in `Sets`/`Maps` so the asymmetry is visible from either side, and put the explanation on `Maps.mapKeys` itself, which is where the divergence is created.

## 2. Terminology — unified

You're right that these are not synonyms, and mixing them let the removed guarantee back in through the other word. The specification now says the language specifies no order — no `implementation-defined` remains in `specification.adoc`, the four javadocs, or the ADR. The javadocs keep `deterministic` but say what it is *for*: a particular trie, not the value.

## 3. ADR-0042 — corrected

Agreed, and I had already hit this while implementing: ADR-0042's subject is deferring RRB and the transient accumulator, with the CHAMP backing as inherited context. The switch is `e45ad66d`, whose own message records "Map/Set iteration is now a deterministic hash order rather than insertion / first-seen order". The status line and the reference now name that commit, and the reference says ADR-0042 records the backing as an existing baseline.

## Verification

`PersistentHashSetTest` (9), `PersistentHashMapTest` (14), `PersistentHashMapCanonTest` (3), `CompileSetCodecTest` (1), `CompileSetLibTest` (3), `CompileNestedCollectionCodecTest` (1), `CompileMapKeyTest` (3), `CompileMapKeyBoundaryTest` (15) — all pass. Still doc-only: no non-comment line changed in any Java file.

Issue #306, opened 2026-08-03
A fake and an example row can state opposite answers for one behavior, and the report then points at the wrong arm


kawasima:
A `fake` and an `example` row can state opposite answers for one behavior, and nothing notices. The
compile is clean and the adequacy report — reading the fake, not the rows — then points the author at
an arm that its own model says is reached.

This became possible on #305. Before it, an `example` row on a behavior with no `let` was E1902, so a
`fake` had nothing to contradict; ADR-0088 records those rows instead, and a fake for the same
behavior is now a second, independent statement about what it answers.

It is not a rare shape. All 19 faked behaviors in `souther-lang/examples` are injected, and harvesting
a legacy system's answers into rows on exactly those behaviors is what #305 was built for.

## Repro

```souther
module example.clash

data MemberId = String
data Found = { id: MemberId }
data Missing = { why: String }

data Order = { by: MemberId }
data Placed = { by: MemberId }
data Refused = { why: String }

// Injected: what it owes is written as rows, harvested from the system being replaced.
behavior findMember : (id: MemberId) -> Found | Missing

example findMember
    | "m-1 is a member" : (MemberId("m-1")) -> Found { id = MemberId("m-1") }

behavior place : (order: Order) -> Placed | Refused
    depends on findMember
    constructs Placed, Refused

let place (order, findMember) = match findMember(order.by) with
    | Found    -> Placed { by = order.by }
    | Missing  -> Refused { why = "unknown" }

// The fake says the opposite of what the rows above say `findMember` owes.
fake findMember
    | (MemberId("m-1")) -> Missing { why = "no such member" }

example place
    | "m-1 cannot order" : (Order { by = MemberId("m-1") }) -> Refused { why = "unknown" }
```

`souther examples clash.sou` exits 0 and says nothing about the contradiction. What it does say is
worse than silence:

```
  place                    implemented   rows 1    pending 0
    signature   out specified 1/2  observed 1/2  verified 1/2
    branch      1/2
      · no row goes through `case Found` (21:33)
```

`case Found` is reported as an arm no row reaches. In the model's own terms a row does reach it —
`findMember(m-1)` is written to answer `Found`. The arm is unreached only because the fake says
otherwise, and an author following the report writes a row that already exists.

## What to report

That the two disagree, and where each is written. Not which one is right: a migration may deliberately
run a behavior against a stand-in while the real answer is still being harvested, and the compiler has
no way to tell that from a mistake. Naming both sides is the whole of what is known.

[...]

PR #329, opened 2026-08-04
The fixture reader is what wants extracting from ExampleVerifier

### LINKED ISSUE #316: The fixture reader is what wants extracting from ExampleVerifier, not either of the two readers on top of it
`ExampleVerifier` is 2348 lines. Its own class doc already states the rule that would have kept it
smaller:

> Three things a row needs are not here, because they are not about what an example means:
> `NeutralForm` says what a value looks like in the form a decoder reads, `HelperInvoker` runs a
> helper as the method its module emits, and `RowEvaluation` owns the state one row builds up while
> it is evaluated.

Two things are in there now, and the rule says they are two things. One evaluates a source's rows and
reports what did not hold. The other (#306) reads a module's written statements against each other and
reports where two of them disagree. They answer different questions, for different keys, and neither
reads the other's state.

## Why it was not split when it was written

The reading sits on the fixture builder: `built`, `raw`, `decode`, `fixtureType`, `caseOnly`,
`constructedCase`, `builtExpectedOrNull`, `describeExpected`. Those are instance methods over
`module`, `symbols`, `sigs`, `values`, `helpers`, `neutral`, `loader` and `bindings` — roughly the
middle third of the file — and every one of them is also what the row path uses.

So the seam is not between the two readers. It is under both of them: **a fixture reader** that owns
those fields and answers "what value does this text state, against this type", with the row evaluator
and the statement comparison as two callers of it. Pulling the comparison out without pulling that out
first would either copy the builder or reach back into the class it came from, and neither is a split.

## Why this is its own change

Extracting the fixture reader means moving the densest and most-depended-on part of the file, which
the row path runs on every compile. That is a refactor with its own risk and its own review, and
putting it under a feature would mean reviewing both at once. #312 left it and said so.

## What would be true afterwards

- One class owning "what does this written value mean", with `module`/`symbols`/`values`/`helpers`/
  `neutral`/`loader` as its state, and no diagnostics of its own — it throws `FixtureException` and
  the callers say what that means where they are.
- `ExampleVerifier` down to evaluating rows and reporting them.
- The statement comparison as a class of its own, holding `Answered`, `differs`, `Standins`,
  `standingIn` and the `Disagreement` shape.

`standingIn` is worth calling out: it is the fake's dispatch rule, and the reason #312 shares one
copy of it is that a second reading of the same table is what E1919 exists to catch. Whichever way
this is split, that has to stay one method both the runner's proxy and the comparison call.

## Not this issue

Anything about what E1919 says or when. This is where the code lives, not what it decides.

kawasima:
`ExampleVerifier` was 2709 lines. Its own class doc already stated the rule that would have kept it
smaller, and two things were in it that the rule says are two things: one evaluates a source's rows
and reports what did not hold, the other reads a module's written statements against each other and
reports where two of them disagree (#306).

But the seam is not between those two. Both sit on the same fixture builder — `built`, `raw`,
`decode`, `fixtureType`, `caseOnly`, `constructedCase`, `builtExpectedOrNull`, `describeExpected` —
instance methods over `module`, `symbols`, `values`, `helpers`, `neutral`, `loader`, `bindings` and
`expanding`, roughly the middle third of the file. Pulling the comparison out first would have
copied that builder or left it reaching back into the class it came from, and neither is a split.

So the thing that wanted a name is underneath both.

| | lines | answers |
|---|---|---|
| `ExampleVerifier` | 1024 | evaluates a row, reports what did not hold |
| `FixtureReader` | 1035 | reads an `Ast.Expr` as a value of a given `Type` |
| `ExampleStatements` | 846 | reads what a module wrote, reports where two writings disagree |

## The part that is not a file split

`FixtureReader` is not a shared service. It is a **disposable interpretation session** — one row's,
or one written statement's — dropped when that reading ends.

That was already true and already handled, but it was said by cloning the whole `ExampleVerifier`:
one copy per row, one per statement, each carrying twelve fields for the sake of three. A worker
that runs out of its budget is asked to stop and cannot be made to — a fixture reaches no interrupt
point — so it may still be inside `expandedValue` holding a binding or a half-walked expansion long
after its answer was given up on. What the copy was *for* could not be read off it.

Now the reading is the reader, and it is passed as an argument at every step. One reader for the
whole row, deliberately: the inputs, the expectation, each `with` and every fake table the row
resolves go through the same one, so a row that hangs inside a fake table's helper is still a row
inside a helper and E1910 can still name it.

## The rule that must not get a second copy

`standingIn` is gone as a static utility and is now `Standins.answering` — the table's own
operation. A second reading of one fake table is exactly what E1919 exists to catch, so sharing the
rule by asking the thing it is about is harder to undo than sharing it by both sides calling the
same helper. `standins` likewise stays one method, building the whole table or none of it, called
both by the proxy a row runs against and by the reading that holds the table to the recorded rows.

## Four commits

1. **`FixtureReader` extracted mechanically** — no renames, no message changes, no public-API
   changes; the verifier keeps one reader and delegates, so the diff is a move.
2. **The reader's lifetime made explicit** — both verifier clones gone; the reader threaded through
   `checkRowNow` / `resolveFakes` / `resolveFake` / `tableProxy`; `within` takes a
   `Function<FixtureReader, T>`.
3. **`ExampleStatements` extracted** — `disagreements`, `fakeTables`, the comparison, the fake
   table, and the public result records. `within` and `Read` leave the verifier entirely.
4. **Bridges removed** — `Construction`/`constructing` onto the reader, `requirements` dropped from
   the two statement readers, `Asserted` → `Answered`.

## No behaviour change

Every diagnostic keeps its code, position, secondary region and hint. Explicit-row vs `_` fallback
precedence, the second table answering nothing, a broken table answering nothing at all, and the
timeout paths are unchanged.

`CompileFakeExampleDisagreementTest` already pinned the boundaries this is most likely to break
(explicit/default dispatch, the second table, the broken table, timeout, runtime-absent). All 32
pass with **no expectation changed** — the only edits to it are a qualified type name and one
dropped argument.

`mvn -B --no-transfer-progress verify` — BUILD SUCCESS, all modules.

## Two things left as they are, on purpose

- **`fixtureType` has no callers anywhere.** It was already dead before this work; it is moved
  verbatim rather than deleted inside commits that claim to be moves. Trivial follow-up.
- **`Output` and `Adequacy` still ask for `requirements`** and refuse to proceed when it is null,
  though they no longer pass it. That is a readiness condition rather than an input; removing it
  changes when those queries recompute, which is a behaviour change this PR should not smuggle in.
  Commented at both sites. Whether those readers should still wait on it is its own question.

## One thing the plan did not anticipate

`FixtureReader` is `public`, not package-private: `Adequacy` lives in `souther.compiler.query`, so
`Construction` cannot be reached from a package-private class. Same reason `MemoryClassLoader` is
public. Its constructor and every reading method stay package-private — a reading belongs to
whichever reader started it.

Closes #316

### DIFF
--- a/souther-compiler/src/main/java/souther/compiler/ExampleStatements.java
+++ b/souther-compiler/src/main/java/souther/compiler/ExampleStatements.java
@@ -1,7 +1,6 @@
 package souther.compiler;
 
 import souther.compiler.ast.Ast;
-import souther.compiler.check.BehaviorRequirement;
 import souther.compiler.check.Sig;
 import souther.compiler.check.Symbols;
 import souther.compiler.check.TypeOps;
@@ -106,7 +105,6 @@ public final class ExampleStatements {
      */
     public static Readings disagreements(Ast.Module module, Symbols symbols,
                                          Map<String, Sig> sigs, Map<String, byte[]> classes,
-                                         Map<String, List<BehaviorRequirement>> requirements,
                                          ClassLoader parent, Map<String, Ast.FnDef> values,
                                          List<String> exampleOrigins,
                                          List<String> fakeOrigins, long budgetMs) {
@@ -157,7 +155,6 @@ public final class ExampleStatements {
      */
     public static List<Diagnostic> fakeTables(Ast.Module module, Symbols symbols,
                                               Map<String, Sig> sigs, Map<String, byte[]> classes,
-                                              Map<String, List<BehaviorRequirement>> requirements,
                                               ClassLoader parent, Map<String, Ast.FnDef> values,
                                               List<String> fakeOrigins, String sourceId,
                                               long budgetMs) {
@@ -334,7 +331,7 @@ public final class ExampleStatements {
     /** One recorded row, read as far as it can be without running it. What it says is left as written
      * — rendering it builds the fixture a second time, and only a row that turns out to disagree is
      * ever shown. */
-    private record RecordedRow(SourceRef at, Ast.Expr expected, Object[] arguments, Asserted answer) {}
+    private record RecordedRow(SourceRef at, Ast.Expr expected, Object[] arguments, Answered answer) {}
 
     private Readings collectDisagreements(List<String> exampleOrigins,
                                           List<String> fakeOrigins, Set<String> contested) {
@@ -391,8 +388,8 @@ public final class ExampleStatements {
                 if (arguments == null) {
                     return null;
                 }
-                Asserted answer = readExpected(reader, row.expected(), sig.out(), cases);
-                return answer instanceof Asserted.Unreadable ? null
+                Answered answer = readExpected(reader, row.expected(), sig.out(), cases);
+                return answer instanceof Answered.Unreadable ? null
                         : new RecordedRow(new SourceRef(origin, row.expected().pos()),
                                 row.expected(), arguments, answer);
             }, "a row of `" + ex.target() + "`");
@@ -454,7 +451,7 @@ public final class ExampleStatements {
             if (answering == null) {
                 continue;   // the table answers nothing for this input; E1909's where it is used
             }
-            Asserted stood = new Asserted.Whole(answering.answer());
+            Answered stood = new Answered.Whole(answering.answer());
             if (differs(row.answer(), stood)) {
                 // The output, not the row: what disagrees is the answer, and the marker lands on it
                 // the way a row's does on its expected.
@@ -503,10 +500,10 @@ public final class ExampleStatements {
                 // As with a recorded row: a `with` is written on a row, and that row is evaluated
                 // where the example is checked, installing this same value. What did not finish here
                 // did not finish there, and there is where it is said (E1910).
-                Asserted constant = within(
+                Answered constant = within(
                         reader -> readStandIn(reader, w.value(), depSig.out(), outCases(depSig.out())),
                         "a `with " + w.dep() + "`").orNull();
-                if (constant == null || constant instanceof Asserted.Unreadable) {
+                if (constant == null || constant instanceof Answered.Unreadable) {
                     continue;
                 }
                 for (RecordedRow recordedRow : rows) {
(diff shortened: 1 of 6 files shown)


kawasima:
All three addressed, in two commits.

## [P2] `Construction` reusing one reader — fixed (b58a7f1c)

Verified: `Adequacy` builds `Construction` once and asks it about every candidate of every behavior
it measures, so the lifetime the code had was one reader per adequacy query while the class doc
claimed one per reading. `constructing` now makes a reader per value asked about and shares only the
`MemoryClassLoader` — which has to be shared, since it caches what it has defined and loaded and a
fake's subclass is generated once.

On "not a behaviour change": confirmed. `NeutralForm` and `HelperInvoker` constructors are field
assignments, so a `FixtureReader` is two assignments plus a map and a deque — negligible against a
call that loads classes and runs helpers, even in `Generator`'s candidate loop. No measurable cost.

Worth stating plainly why this was worth fixing when nothing misbehaves today: every place a fixture
puts something in force takes it back out in a `finally`, so a sequential caller does see an empty
reader each time. But that makes the guarantee hold only because nothing has yet been added to the
reader that outlives a `finally`, and only for one caller at a time — which is exactly the kind of
thing this extraction existed to stop being traceable-only-by-reading.

## [P3] `ExampleVerifier.rendering` — fixed (f5dec10b)

Confirmed dead: constructed, never read, and its `{@link #shown}` named a method that left with
commit 3. Removed, along with the two stale doc neighbours you flagged — the `budgetMs` javadoc
pointing at `#within` and `newFixtureReader` offering a reader for a written statement this class no
longer reads.

## [P3] `ExampleStatements.sourceId` — fixed (f5dec10b)

Confirmed: the field is assigned in `fakeTables` and never read — the filtering beside it reads the
parameter of the same name, which shadows it. Both the field and the assignment are gone.

## Also in f5dec10b

A defect of mine that neither of us had flagged: the extraction script joined moved regions without
a blank line between them, so eight member boundaries in `FixtureReader` and `ExampleStatements` had
no separator. Fixed.

`mvn -B --no-transfer-progress verify` — BUILD SUCCESS, all modules, `CompileFakeExampleDisagreementTest`
still passing unchanged.

PR #339, opened 2026-08-04
A limit found by running out of stack is not the same limit twice

### LINKED ISSUE #336: An expression that nests too deeply is a diagnostic when compiled and a stack overflow when formatted
`Compiler` decides that an expression nesting deeper than the compiler can walk is a thing to report,
and `tooDeep()` says why (`Compiler.java:79-93`):

> Every phase descends an expression by recursion — parsing, inlining, checking, emitting — so past
> some depth the stack runs out. That is not a `CompileException`, so left alone it passes through
> the recovery boundary and reaches the author as a stack trace, and takes the language server's
> dispatch loop with it. It is reported like any other thing the compiler cannot accept.

`Formatter` descends an expression by recursion too, and has no such guard. The same source that
`compile` turns into `check.expr.toodeep` overflows the stack when formatted.

## Repro

```java
String deep = "module a\nlet f = " + "1+".repeat(5000) + "1\n";   // parses cleanly

CstParser.Result first = CstParser.parse(deep);      // errors=0
String once = Formatter.format(first.root());        // ok, 20024 chars on 4 lines
CstParser.Result again = CstParser.parse(once);      // errors=0
Formatter.format(again.root());                      // StackOverflowError
```

```
StackOverflowError @ souther.compiler.fmt.Formatter.expr(Formatter.java:542)
```

Measured at 13dcc307, on a thread with the JVM's default `-Xss`:

| additions | first format | second format |
|---|---|---|
| 3000 | ok | ok |
| 5000 | ok | **StackOverflowError** |

The threshold depends on the stack the caller happens to have, which is the part worth noticing:

- A thread the JVM creates gets `-Xss` (1 MB by default), and overflows between 3000 and 5000.
- The main thread gets `ulimit -s` (8 MB here), and survives 5000 — it needs a bigger source, not a
  different kind of source.

So `souther fmt` from a shell and the language server formatting the same file do not fail at the
same size, and the server — which handles requests on threads it created — is the one with less room.

The first format is what arms the second: it rewrites the chain onto one line, so re-parsing it
yields a deeper right spine than the source had.

## Where it comes from

`Formatter` has no depth guard of any kind — `grep -n "StackOverflowError\|depth\|tooDeep"` over
`souther-fmt/src/main/java/souther/compiler/fmt/Formatter.java` is empty — while `Compiler` catches
`StackOverflowError` and converts it in two places (`Compiler.java:74-76`, `Compiler.java:239-241`).

All three public callers invoke it bare:

- `souther-lsp/src/main/java/souther/lsp/analysis/Analyzer.java:528` — the language server's format
  path
- `souther-compiler/src/main/java/souther/compiler/Main.java:315` — `souther fmt`
- `souther-compiler/src/main/java/souther/compiler/report/GeneratedRows.java:139`

## Notes for the fix

The design question is which side owns the policy. `Compiler` already owns it for the same input
class and states the reason, so having `Formatter.format` answer the same way — a `CompileException`
carrying `check.expr.toodeep`, or whatever the formatter's equivalent should be called — keeps one
answer for one question and needs no change at the three call sites. Guarding at the call sites
instead would put the same decision in three places, and `GeneratedRows` is not a place an author is
looking at.

Either way the precondition `Formatter.format` documents today — "Assumes the source parses without
syntax errors" — is not the only one it has, and the second one is not stated.

---

Found while reviewing an external property-based harness: the property is "on a source that parses
cleanly, formatting yields a source that still parses cleanly, and formatting is idempotent". The
harness deliberately does not treat a formatter `StackOverflowError` as in-contract, since `compile`
does not either. Verified at 13dcc307 with a clean worktree build.

kawasima:
Fixes #336.

## What was wrong

`Compiler` refused an over-deep expression by catching the `StackOverflowError` its walks raised. `Formatter` did not, so the same source that `compile` turned into `check.expr.toodeep` reached the author of `souther fmt` as a 1025-line stack trace.

Guarding `Formatter` — what the issue proposed — would not have fixed it. `souther fmt` on a parenthesis nest overruns inside `CstParser`, before `Formatter` is reached at all.

## Root cause

`StackOverflowError` was being used as the depth-limit *mechanism*. Everything observed follows from that:

| Consequence | Observed |
|---|---|
| Nobody chose the limit (`-Xss` ÷ the frame the JIT picked) | `souther fmt --check` on one unchanged file: **7 crashes / 7 clean out of 14 identical runs**. Overflows under C1, not under `-Xint` or C2 |
| No position can be reported | `Compiler.tooDeep`'s own javadoc: *"The depth is not a written limit, so no position is claimed"* |
| Every path must independently catch an `Error` | `Compiler` ×2, `Analyzer` ×8, `LspServer` ×1 — an obligation written down nowhere |
| So it gets missed | `GeneratedRows` guards with `catch (RuntimeException)`, which **cannot catch an `Error`**; `Analyzer.format` is the one entry point lacking the guard its eight neighbours have |

Two corrections to the issue's account, both measured: the reparsed tree is **not** deeper (3003 before and after — reformatting is not what arms it), and the language server neither dies nor has less room (it creates no threads, and `LspServer.java:70` already holds the loop).

## The fix

**Write the limit down; demote `StackOverflowError` to a backstop.**

The design that matters is *where* the limit is applied. The first attempt put it in `expr` and in `wrap`, on the reading that every expression written inside another descends through `expr`. It does not — and review caught it:

- `unaryExpr` calls itself, so `-----1` closes a cycle that reaches `expr` once and never again
- `typeRef` closes one through its type arguments, another through a tuple type
- `pattern` closes one through a tuple pattern

**36 productions take part in a grammar cycle.** A limit each of them has to remember is a limit the next one written will not. So the refusal goes where every one of them *has* to pass: **`start()`** — to nest, a production must open a node. It is unwound by a private `TooDeep` caught in `parseSourceFile`, which closes the frames still open into their parents and takes the tokens the descent never reached as they were written. A production added later inherits the bound without knowing it is there.

`wrap` keeps a refusal of its own: a left-nested chain grows the tree without nesting a frame, the one thing `start` cannot see.

`TooDeep` never leaves the class, so the parser's promise not to throw holds.

Supporting pieces: `Green` folds a `depth` as it already folds `width`, at `Green.node` — the one place a node is built. The `StackOverflowError` catches stay, and `Formatter`, `Main`'s `fmt` (which had no recovery boundary at all) and `Analyzer.format` gain the ones they lacked.

`MAX_DEPTH = 64` is measured, not picked: parse-then-format costs ~13 frames per level and manages 91 on the 256 KB stack these tests pin. It is `public`, because a bound a caller inherits is a bound a caller should be able to read: **`Green.depth()` of a tree from `parse` never exceeds it** — counted as the tree a frame leaves behind, not as the frames themselves, which was wrong by one.

## Before / after

| | before | after |
|---|---|---|
| `souther fmt` — chain, parens, unary, type, pattern | 1025-line stack trace | `syntax error: this source nests too deeply to read …` |
| `fmt --check`, 14 identical runs | crash 7 / clean 7 | crash 0, **14/14 identical** |
| LSP `textDocument/formatting` | `-32603 (StackOverflowError)` | `result: []` |
| LSP diagnostic | none | **positioned at the token that reached the limit** |

## Tests

`DeepSyntaxTest` runs **six shapes** — descent, wrap, unary, generic type, tuple type, tuple pattern — through the same questions: it does not throw, it says the depth is why, the tree covers the whole source, and it comes back within the bound. Every shape is also swept across **each depth either side of the bound**, because a source that comes back *accepted* is where the bound is load-bearing, while sources far past it are refused and flattened.

The depth diagnostic is asserted not to say "expression" or "let": neither applies to `List<List<…<Int>…>>`.

The existing `CompileDeepExpressionTest` and `LspSurvivesADeepExpressionTest` pass unchanged; both pin a 256 KB stack precisely because of the nondeterminism this removes.

Full suite: **2868 tests, 0 failures**, `mvn -o clean install` green across all 8 modules (including `CorpusTest`, so the limit does not trip on real code).


kawasima:
指摘3点とも再現し、いずれも正しいことを確認しました。とくに指摘1はテスト上の理念違反にとどまらず、**#336 のユーザー可視の障害がそのまま残っていました**。

```
$ souther fmt unary.sou      # let f = -----…-1  (20000)
Exception in thread "main" java.lang.StackOverflowError   … 1025 lines
```

`MAX_DEPTH = 64` を導入したのに、`parse` が返す木の実測深さ:

| 形状 | 修正前(このPR初版) | 現在 |
|---|---|---|
| 単項連鎖 `-----1` | errors=0, **depth=5006** | errors=1, depth=8 |
| ジェネリック型 `List<List<…>>` | errors=0, **depth=6005** | errors=1, depth=4 |
| タプル型 | errors=1, **depth=3005** | errors=2, depth=4 |
| タプルパターン | errors=0, **depth=3005** | errors=1, depth=8 |
| 括弧入れ子（対照） | errors=1, depth=65 | errors=1, depth=8 |
| 二項連鎖（対照） | errors=1, depth=64 | errors=1, depth=61 |

256KB スタックでは前者4形状は `CstParser` 内で `StackOverflowError` を投げていました。ご指摘のとおり「The parser still never throws」は成立していませんでした。

## 何を変えたか

`unaryExpr` に個別チェックを足す案は採りませんでした。文法の循環に関与するプロダクションを機械的に数えると **36個** あり、同じ誤りを規模だけ大きくするだけです。

ご指摘の「再帰する全プロダクションが必ず通る境界」を探したところ、**`start()` がそれでした**。入れ子になるには必ずノードを開く必要があり、`unaryExpr`・`pattern`・`typeRef`・`field` はいずれも `start(...)` の直後に自分を呼びます。

そこで拒否し、private な `TooDeep` で `parseSourceFile` まで巻き戻します。巻き戻し先で、開いたままのフレームを親に畳み、descent が到達しなかったトークンをそのまま回収するので、木は元ソース全体を覆ったままです。プロダクション側の合意は不要で、後から追加される再帰プロダクションも自動的に境界を継承します。

`wrap()` の拒否は残しています。左結合の連鎖はフレームを入れ子にせずに木を深くする唯一の経路で、そこだけは `start()` から見えないためです。

`TooDeep` はクラス外に出ないので「never throws」は維持され、スタックトレースは埋めていません。

## 指摘3について

`stack.size()` と `Green.depth()` のずれはご指摘のとおりでした。`MAX_DEPTH` を `public` にし、契約を「`parse` が返す木の `Green.depth()` はこれを超えない」と定義したうえで、**その不変条件自体をテストで検証**するようにしました（以前は損失性とエラー件数しか見ていませんでした）。

## テスト

6形状 × 4性質のパラメータ化テスト（例外を投げない / 深さが理由だと言う / 元ソース全体を覆う / 上限内に収まる）＋ 実行間の一意性 ＋ 通常ソースが影響を受けないこと = 27件。全形状を 256KB スタックで実行しています。

全体: **2862 tests, 0 failures**、`mvn -o clean install` 通過。CLI も3形状すべてが1行の診断を返し、`fmt --check` 14回で crash 0 です。

なお初版のレビュー時点の測定値のうち、木の深さに関する主張（"bounded"）は誤りでした。指摘に感謝します。

kawasima:
3点とも修正しました（`55cec0f5`）。

## 1. [P1] off-by-one — 確認、修正

ご指摘のとおりでした。境界を走査すると、拒否されずに受理される入力で不変条件が破れていました。

```
shape            deepest ACCEPTED tree   first refusal
unary chain      65  (n=59)              n=60     <-- MAX_DEPTH=64 超過
paren nest       65  (n=59)              n=60     <-- 超過
generic type     65  (n=30)              n=31     <-- 超過
tuple pattern    65  (n=60)              n=61     <-- 超過
binary chain     64  (n=58)              n=59     ok
```

`binary chain` だけ無傷なのも、ご指摘のとおり `wrap()` が新しいラップノードの1階層を含めて数えていたためです。

`start()` を、フレーム数ではなく**そのフレームが残す木の深さ**で数えるようにしました。開こうとしているフレームで1階層、それが持つもの（トークンだけでも）でもう1階層です。

```java
if (stack.size() + 2 > MAX_DEPTH) {
    refuseAsTooDeep();
}
```

修正後は全形状が上限内（最深の受理木 63〜64）に収まります。

**テストが境界を見ていなかった点も、ご指摘のとおりでした。** 巨大入力は必ず拒否され、拒否は読み終えた分を平坦化するので、受理されて返る入力を一度も通っていませんでした。ご提案の走査を6形状すべてに対して追加しています。

teeth の確認として `start()` を元の判定に戻すと、`theBoundHoldsEitherSideOfIt` が該当4形状でちょうど落ちます。

## 2. [P2] 診断が式専用 — 修正

`List<List<…<Int>…>>` に `let` を勧めるのは誤りでした。カテゴリ別には分けず、一般化しています。

```
$ souther fmt gen.sou
gen.sou: syntax error: this source nests too deeply to read; break the nesting into named parts to flatten it
```

レンダリング後（位置付き、キャレットは深さに達したトークン）:

> This nests deeper than I read, and every later reading of it — checking, formatting, compiling — descends it the same way. Break it into parts that have names, so each one is read on its own and says what it is: a binding for a value, a declaration for a type.

日本語カタログも同様に一般化しました。`contains("deep")` しか見ていなかった点も、`"expression"` と `"let"` を含まないことを固定しています。

## 3. [P3] PR本文とテストクラス名 — 修正

PR本文を現在の設計（`start` を共通境界とし、`TooDeep` で一括アンワインド／6形状）に全面更新しました。「各プロダクションが個別に制限を覚える設計を廃止した」という判断が本文の中心になっています。テストクラスは `DeepSyntaxTest` に改名しました。

## 検証

全体 **2868 tests, 0 failures**、`mvn -o clean install` 通過。

2回続けて、私の側の検証がレビューで補われる結果になりました。とくに今回は「不変条件を主張しながら、その不変条件そのものをテストしていなかった」ことが原因で、これは指摘されるまで自分で気づけていませんでした。ありがとうございます。

Issue #326, opened 2026-08-04
Navigation from a cursor inside an `examples for` file falls back to matching by spelling


kawasima:
Split out of #309 (PR #325), where it was left out deliberately.

## What happens

Go-to-definition, hover and find-references from a cursor **inside** an attached
`examples for` file do not use resolution. They fall back to matching by spelling, so
a local — a row's binding, a parameter — is out of reach, and a name bound in several
places resolves to whichever the fallback finds.

The reverse direction works: a cursor in the model file on a value the attached file
declares now navigates to that file (`ACursorIsOnAPlaceNotACoordinateTest` in
`souther-lsp`).

## Where it is

`Analyzer.typeUnderCursor` and `Analyzer.valueUnderCursor` both start with

```java
String module = text == null ? null : Compiler.moduleNameFromHeader(text);
if (module == null) {
    return null;
}
```

`Compiler.moduleNameFromHeader` matches `^\s*module\s+(...)`. An attached file's header
is `examples for m`, so it answers nothing, and both fall out before asking the
compiler anything. `Analyzer.definition` then reaches its spelling-based path.

## What it needs

The module an attached file contributes to, rather than the module it declares — which
is what `Front.Layout.exampleFileTargets` already holds, and what
`Front.AttachedTo(id)` answers. The parse-side reading is `Ast.Module.exampleFileTarget()`.

Once the module is known, the rest already works: #309 made `Names.spans` file-aware
and the `Analyzer` pass the document's URI with the cursor, so a question asked from an
attached file is answered about that file's own names.

## Why it was not in #309

#309 is about which file a diagnostic is said on. This is a navigation gap of its own —
it predates that change and is not made worse by it — and wants its own tests.

kawasima:
Fixed by #340, merged into develop.

A cursor inside an attached `examples for` file now reaches the compile: which module a source is part of is `Front.ModuleOf`'s answer rather than a regex over the header, and the editor's questions are asked of a place rather than of a module and a place.

Two corrections to what this issue says. For a type, navigation did not fall back to matching by spelling — it answered nothing, since an attached file declares no `data` and writes no imports. And the same reading turned up a second defect on the other side: find-references filed every use under the module's own file, so a use written in an attached file was published in the model file at a line and column belonging to the other one.

[...]

PR #340, opened 2026-08-05
Which module a cursor is in is the file's to say, not the header's

kawasima:
Issue #326.

## What was wrong

A cursor inside an attached `examples for` file reached no compile. The editor
worked out which module a document was about with a regex over its text —
`Compiler.moduleNameFromHeader`, matching `^\s*module\s+X` — and an attached
file writes no such header. Every navigation entry point that starts from a
document fell out before asking the compiler anything.

The issue says navigation then falls back to matching by spelling. For a type it
does not: an attached file declares no `data` and writes no imports, so the
fallback finds nothing either. Go-to-definition and find-references answered
empty. For a value the fallback can land right by accident, when the name
happens to be declared in the same attached file.

The compiler side needed nothing. Asked with the module the file contributes to,
`Names.TypeAt`, `Names.ValueDenotedAt` and `Names.ValueDeclaredAt` all answer for
cursors in the attached file, and the positions carry that file's source id.

## What changed

**A question about a cursor is asked of a place.** `Front.ModuleOf(id)` answers
which module a source is part of — the one it declares, or the one an
`examples for` file's rows are for. `Names.TypeAt`, `DenotedAt` and
`ValueDenotedAt` now take only a position and read the module off the file it
names. A caller that had to name the module first was a caller that could name
the wrong one; there is no longer one to name.

`Compiler.moduleNameFromHeader` keeps one job, now stated on it: naming a module
whose file a compilation could not parse. Both remaining uses in the analyzer are
that case.

**An answer is filed where it was written.** Find-references asked each module
for every place it names a type and put all of them under the module's own
source. A use written in an attached file was published in the model file at the
line and column it has in the other one. In the fixture here that lands on the
`v` of `data D = { v: Int }` — a place in a file the use is not in, holding
something else. The reading of an answer's file is now one function that the
declaration, the type uses and the value uses go through. The same reading fixes
a code lens and the `write the rows` offer, which read another file's declaration
position against this document's line index.

**A declaration is named where its name is written.** `Names.TypeAt` compared the
cursor against the declaration's start — its `data` keyword — so it answered for
a cursor on the first columns of `data` and never for one on the name. That is
not specific to attached files: go-to-definition, find-references and rename from
a type's own declaration name fell back to spelling everywhere. `Ast.Def` now
carries where its name is written.

**What the fallback still does.** With the compile answering about types wherever
they are written, matching the spelling is no longer reached for one. It is still
what answers in the value namespace: the resolve pass records the value names
written in a body, and a composition's `>->` stages and a declaration's own name
are not among them, so nothing found there is not evidence of nothing. That is
written where the decision is made. `Names.ValueUsesOf` is new, which is what
lets find-references answer about a local at all — a spelling match could only
ever answer about top-level names.

**One removal.** `Names.spans` no longer answers a question that names no file.
That leniency was added in #309 for callers with only a line and a column to
give, and place-keyed queries leave none. The half that matters is kept: a name
that names no source is still not under a cursor in an open file.

## Tests

- `NavigationInsideAnAttachedFileTest` — a type, a row's type, a value and a
  local, from a cursor in the attached file, plus find-references and rename.
- `AUseIsFoundInTheFileItIsWrittenInTest` — every place found really spells the
  name there.
- `EverySourceKnowsWhichModuleItIsPartOfTest`, `ADeclarationIsNamedWhereItsNameIsWrittenTest`.

Against the unchanged code, 8 of the 10 new LSP assertions fail; the two that
pass are the fixture check and the value case that spelling gets right by
accident.

Full build green: runtime 109, compiler 2644, lsp 132, cli 1. The examples
repository builds and runs green against this compiler.

## Left out

Making find-references in the value namespace fully resolution-based needs the
resolve pass to record a composition's stages and a declaration's own name, which
is compiler work of its own. Worth its own issue.

### DIFF
--- a/souther-compiler/src/main/java/souther/compiler/Compiler.java
+++ b/souther-compiler/src/main/java/souther/compiler/Compiler.java
@@ -396,9 +396,16 @@ public final class Compiler {
                                                              ModulePath path) {
         return Compilation.ofDocuments(sourcesById, brokenModuleNames, path).diagnostics();
     }
-    /** The module name from a source's {@code module <name>} header, for identifying a source that will
-     * not parse (so its importers can be skipped, and the LSP can map a broken file to its module).
-     * {@code null} if no header token is found. */
+    /**
+     * The module name from a source's {@code module <name>} header, for identifying a source that
+     * will not parse (so its importers can be skipped, and the LSP can map a broken file to its
+     * module). {@code null} if no header token is found.
+     *
+     * <p>Only for a source a compilation does not have. Which module a source is part of is
+     * {@link souther.compiler.query.Front.ModuleOf}'s to answer, and this cannot answer it: an
+     * {@code examples for} file writes no header and is part of a module all the same, so what
+     * comes back here is nothing rather than the module its rows are for.
+     */
     public static String moduleNameFromHeader(String source) {
         java.util.regex.Matcher mt = java.util.regex.Pattern
                 .compile("(?m)^\\s*module\\s+([\\p{L}\\p{N}_.]+)").matcher(source);
--- a/souther-compiler/src/main/java/souther/compiler/ast/Ast.java
+++ b/souther-compiler/src/main/java/souther/compiler/ast/Ast.java
@@ -56,9 +56,19 @@ public interface Ast {
          */
         SourcePos namePos();
 
-        /** A binder as the parser read it from a name the author wrote at {@code pos}. */
-        static Binder written(String name, SourcePos pos) {
-            return new Written(name, pos, pos);
+        /**
+         * A binder read off a name the author wrote, taken from the name itself so its spelling and
+         * its place cannot come from two different things.
+         *
+         * <p>There is no form here that takes the two apart. A caller holding a spelling and a
+         * position separately is a caller nothing can stop from pairing a name with somewhere it is
+         * not written, and this branch shipped that mistake three times: the parameter {@code .v}
+         * desugars to, anchored at the {@code .}; a lambda's parameters, all anchored at the
+         * lambda; a {@code match} arm's binding, anchored at the {@code |}. The parser reads a
+         * binder off its token; a pass that needs a name of its own says {@link #desugared}.
+         */
+        static Binder of(Name written) {
+            return new Written(written.written(), written.pos(), written.pos());
         }
 
         /**
@@ -356,11 +366,6 @@ public interface Ast {
         /** A parameter whose type, if any, the author wrote (the common case). */
         public FnParam(Binder binder, RetType type) {
             this(binder, type, false);
-        }
-
-        /** A parameter as the parser read it. */
-        public FnParam(String name, RetType type, SourcePos pos) {
-            this(Binder.written(name, pos), type, false);
         }
 
         public String name() {
--- a/souther-compiler/src/main/java/souther/compiler/ast/Ast.java
+++ b/souther-compiler/src/main/java/souther/compiler/ast/Ast.java
@@ -374,6 +374,15 @@ public interface Ast {
     sealed interface Def extends Ast permits Data, SumData, UnitData {
         String name();
 
+        /**
+         * Where the name is written, which is not where the declaration starts — {@code data} comes
+         * first. A reader asking what a cursor is on has only the name to compare against, and a
+         * declaration that answers from its keyword answers about the keyword.
+         *
+         * <p>Null for a declaration nobody wrote: a unit data a construction implied.
+         */
+        SourcePos namePos();
+
         SourcePos pos();
     }
 
@@ -393,6 +402,7 @@ public interface Ast {
                 List<InvariantClause> invariants,
                 Optional<DecoderDef> decoder,
                 Optional<EncoderDef> encoder,
+                SourcePos namePos,
                 SourcePos pos) implements Def {}
 
     /**
@@ -424,6 +434,7 @@ public interface Ast {
                    List<Name> cases,
                    Optional<Discriminate> decoder,
                    Optional<SumEncoder> encoder,
+                   SourcePos namePos,
                    SourcePos pos) implements Def {}
 
     /** {@code encoder discriminate on "key" { Case -> "tag" ... }} — the inverse of discriminate. */
@@ -432,7 +443,7 @@ public interface Ast {
     record EncVariant(Name caseType, String tag, SourcePos pos) implements Ast {}
 
     /** A unit data definition {@code data U} with no fields. */
-    record UnitData(String name, SourcePos pos) implements Def {}
+    record UnitData(String name, SourcePos namePos, SourcePos pos) implements Def {}
 
     /** {@code decoder from Object discriminate on "key" { "tag" -> Case.decoder ... }} */
     record Discriminate(String key, List<Variant> variants, SourcePos pos) implements Ast {}
--- a/souther-compiler/src/main/java/souther/compiler/ast/Ast.java
+++ b/souther-compiler/src/main/java/souther/compiler/ast/Ast.java
@@ -728,7 +728,10 @@ public interface Ast {
             this(binder, value, null, false, null, body, pos);
         }
 
-        /** The same, as the parser read it. */
+        /** The same, as the parser read it, where the name is written at {@code pos} too — a
+         * binding a pass minted rather than read off a source. A binding the source wrote is built
+         * from a binder of its own, because a {@code let} statement starts at its keyword and the
+         * name it binds is somewhere after it. */
         public LetIn(String name, Expr value, Expr body, SourcePos pos) {
             this(Binder.written(name, pos), value, null, false, null, body, pos);
         }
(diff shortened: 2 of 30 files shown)


kawasima:
Both blocking findings confirmed by reproduction before changing anything, and both are the same gap rather than two defects: the type namespace has one question for what a cursor is on, one for where a name is declared and one for where it is used, and I had given the value namespace only part of that. Fixed as the missing third, not as two patches.

**P1 — a behavior's `let` line was left behind.** Confirmed: renaming `g` from a use gave `{exposing, behavior g, the use}` and dropped `let g`, which leaves the module uncompilable. A behavior is genuinely declared twice, so where a value is declared is now `Names.ValueDeclarationsOf`, answering with every line that writes the name. `ValueDeclaredAt` remains as its first element — that is the one to send a reader to, and the signature says more than the body does. `references(includeDeclaration)` uses the plural one.

**P2 — a binding was reachable from one end.** `ValueDenotedAt` reads the names a body uses, so a cursor on a use of a local answered and a cursor on the `let` binding it did not. `Names.ValueAt` is `TypeAt` for the value namespace: the binding under the cursor, or what the name there denotes.

That turned up the reason it could not already work. A `let` statement starts at its keyword, and the binder built from it carried the keyword's position, not the name's — the same distinction this PR already drew between a declaration's position and its name's, one level down. Go-to-definition on a local looked correct only because it searched forward for the name from there. `Resolve`'s binder map now records what a binding is called along with where that is written, and the parser builds a source-written binder at its name.

I took the option of making local rename real rather than refusing it, and updated the `Analyzer` and `LspServer` contracts, which said top-level only.

**On the coverage boundary.** Rather than leave a list of exceptions, it is now one line: resolution reads the names a body uses and the bindings it makes, and a pipeline's `>->` stages and a behavior's own declaration lines are written outside any body. That is what the spelling path still answers, and it is stated where the decision is made. Closing it means recording stage uses and giving `BehaviorDef`/`FnDef` a name position — worth its own issue, not this one.

**Non-blocking finding accepted as stated.** An attached file with a syntax error is out of the compile set, and the header fallback cannot read `examples for m`, so semantic navigation is unavailable there. Not a regression and not #326's subject; worth its own issue.

New test: `RenamingAValueAgreesWhereverItIsAskedTest` — renaming a behavior from its signature, from its body and from a use all produce the same edits, and renaming a local from its binding and from its use likewise. Against the code as reviewed, the two cases you named fail and the other three pass.

Full build green: runtime 109, compiler 2644, lsp 137, cli 1. The examples repository builds and runs green.

kawasima:
Both reproduced before changing anything, and both are one cause rather than two: `Binder.Written` meant "the parser produced this", which covers a name read off the source and a name a desugaring invented. Generalizing over that map was the mistake.

**P1a — a name nobody wrote, at a width that is not its.** Confirmed with `let total (ds) = List.sum(List.map(.v, ds))`: columns 36–39 of that line answered `$g0` (a `Local`), and the three characters at column 36 are `.v,`. Renaming from there wrote over the source.

**P1b — a field renamed at its declaration and nowhere else.** Confirmed: `data D = { v: Int }` answered `v` as a `Local` at its declaration, while the recorded value uses held only the `v` inside the `invariant`. `d.v` and `D { v = ... }` never appear, so the edit set was the declaration plus the invariant.

**The fix.** A binder now carries where the author wrote its name, or says no one did — the same distinction this branch already drew between where a declaration starts and where its name is, one level further down. `Binder.desugared(name, anchor)` is what a desugaring builds; `Binder.written(name, pos)` is a name read off the source. The resolve pass records only the second kind, which is what makes that map answer the question an editor asks of it. `ValueAt` answers a binding only when it is in that map.

A field is the other half and is not about who wrote the name, so it is a separate condition with its own reason: a field's reads are resolved by the type of what they are read from rather than by what is in scope, and only the ones inside an `invariant` are recorded. Excluded, stated where the decision is made. This restores what `develop` does — a field was never renameable there either, since a spelling match only scans top-level declarations.

**The untested source-written binders you listed are now written where they are written.** Chasing the same distinction turned each of them up:

- a lambda's parameters shared the lambda's own start; each is now its own binder
- a `match` arm's `as x` and an attempted construction's `as x` shared the form's start
- a pattern parameter takes a name nobody wrote, anchored on the pattern, and is not offered

Two convenience constructors that could only build a binder at the wrong place — `IfConstructed(…, String binder, …)` and `Case.written` — are gone rather than left for the next caller.

New tests. `AnEditorIsOnlyToldAboutNamesTheAuthorWroteTest` covers the getter span, a field declaration, the invariant read, both lambda parameters, and both `as` arms. `RenamingAValueAgreesWhereverItIsAskedTest` gains the editor-level half: renaming from `.v` and from a field declaration produces no edits at all.

One thing worth flagging that this turned up: `mvn compile` reported success while leaving a stale class for a real call site whose source had not changed. The clean build catches it; incremental does not.

Full clean build green: runtime 109, compiler 2651, lsp 139, cli 1. The examples repository builds and runs green.

kawasima:
You are right, and so is the observation that this needed a root fix rather than a fourth round of the same repair. Three reviews have found the same defect in three places; the way a binder is built is what changes.

**Why it kept happening.** A binder was made from a spelling and a position handed over separately — `Binder.written(String, SourcePos)`. Nothing can check the two came from the same token, because the tree has no text, so all twenty-odd callers had to be careful and three were not. Every round I fixed the careless ones and left the API that makes carelessness the easy path.

**What changed.** That form is gone. A binder read off the source is built from the token that spells it, which carries both halves as one thing, and the parser has a single place that turns a token into a binding. A name a pass needs for itself says outright that it is written nowhere. `FnParam(String, RetType, SourcePos)` and the two convenience constructors that could only anchor a name at the wrong place went with it.

**And a check, since no type can enforce it.** `ANameIsWhereItIsWrittenTest` parses a module exercising every binding form the language has — parameters written and patterned, a lambda's, a block `let`, a destructuring `let`, an attempted construction's `as`, a `match` arm's `as`, its field destructuring and its constructor destructuring — walks every binder in the result, and reads each claim back against the source. Seventeen bindings; one that names a place has to be the name written there, a `$` name has to name none, and a name the author wrote has to name one.

It is not a test that passes by accident. Putting the pattern-parameter defect back fails it with the diagnosis rather than a symptom:

```
`$p2` claims 29:11, which holds `Amount`
```

That is the tool that would have found all three rounds at once.

**The two you found are fixed as the two directions of one missing distinction.** A pattern parameter's carrier was claiming the pattern's own place; the names a destructuring binds — a record pattern's, a `match` arm's fields, the innermost name of a newtype pattern — were dropped to lowering and left with nowhere to be found at. Both now go through the token.

**Tests you asked for.** `RenamingAValueAgreesWhereverItIsAskedTest` gains the pattern forms: `{ left = l }` and the shorthand `{ right }`, a `match` arm's `{ f }`, and `Amount(inner)` — each renamed from the binding and from the use, giving the same edits. The destructured parameter itself carries nothing to rename, and a pattern's type name renames the type rather than what the parameter is called.

Full clean build green: runtime 109, compiler 2655, lsp 145, cli 1. The examples repository builds and runs green.

kawasima:
Confirmed and fixed, and the observation about what the tests were checking is the one that matters most here.

**The defect.** `{ right }` names the field to read and binds the value under that same spelling. Writing the new name over that token gives `{ r }`, which reads a field that does not exist. Same in a `match` arm. The edit set was right about where; only what to write there was wrong.

**Why it could not be fixed where it was.** Rename answered with ranges, and the server wrote one name over all of them. The shape had no room to say that one occurrence needs different text. So an edit now carries its text: a binding written as a field's own name is renamed by naming the field it reads — `right = r` — and everything else is the new name, as before. Recognising the shorthand is the syntax tree's job, since it is a question about what characters are there.

**On the tests.** You are right that reading the places an edit set touches fixed the wrong property. That check says the editor found the occurrences it was looking for, which is a claim about the search rather than about the result. `RenamedSourceStillCompilesTest` carries the rename out and compiles what comes back, over the parameter destructuring, the `match` arm's fields, the newtype pattern, an explicit `{ left = l }` and an ordinary parameter. It failed on its own fixture first — the fixture matched on a product type — which is the check doing its job before it was asked to.

**On the four rounds.** These have not been four unrelated defects. Each one is the same shape: a datum that is adequate for the compiler's own use, borrowed for an editor question that needs a strictly stronger property, without the property being stated.

- `moduleNameFromHeader` is enough to name a broken file's module; not to say which module a file is part of.
- `binders` is enough to give a binding an identity and a place for a diagnostic; not to say where a name is written.
- `Binder.Written` is enough to say the parser produced it; not to say the author wrote it.
- `Range` is enough to highlight something; not to say it can be replaced by a new name.

Each review found the next borrowed datum whose guarantee was too weak. The general consequence is the one this round makes concrete: an editor answer is verified by applying it, not by inspecting it. That is why the apply-and-recompile check is the one worth keeping, and why the position checks stay only as statements about reach.

Full clean build green: runtime 109, compiler 2655, lsp 151, cli 1. The examples repository builds and runs green.

PR #349, opened 2026-08-05
The compiler does not hand out a class the JVM will not load

### LINKED ISSUE #341: A data with 128 Int fields compiles into a class the JVM will not load, and a large literal reaches the author as an ASM stack trace
`Backend` emits with the JDK `java.lang.classfile` API and never asks whether what it is emitting is a class the JVM will accept.
Three of the JVM's structural limits are reachable from ordinary source, and not one of them is a
diagnostic. One of them does not even fail at compile time: the compile succeeds and the class it
wrote cannot be loaded.

Measured on develop at f89b41d9.

## A data type whose constructor exceeds 255 argument slots

```java
String wide = "module demo\n\ndata Wide = { "
        + IntStream.range(0, 128).mapToObj(i -> "f" + i + ": Int").collect(joining(", "))
        + " }\n\nbehavior f : (w: Wide) -> Wide\n\nlet f (w) = w\n";

Map<String, byte[]> classes = Compiler.compile(wide);   // succeeds, reports nothing
loader.defineClass("demo.Wide", classes.get("demo.Wide"), ...);
```

```
java.lang.ClassFormatError: Too many arguments in method signature in class file demo/Wide
```

`Int` is carried as `long`, which takes two slots, so for a record of `Int` the limit falls at 128
fields. Every other field type is a reference and takes one slot — `Decimal` is a `BigDecimal`, not a
`double` — so those fall at 255. The generated record constructor is an instance method, so `this`
takes one of the 255 slots and the fields have 254 between them:

| field type | last accepted | first refused |
|---|---|---|
| `Int` | 127 | 128 |
| `String` | 254 | 255 |
| `Decimal` | 254 | 255 |
| `Bool` | 254 | 255 |

This is the one worth taking first. The compiler says the module is fine. The failure moves to
whoever loads the class — a test run, `souther run`, or the application the model was compiled into —
and what it says there names a JVM rule rather than the declaration that broke it. A hundred and
twenty-eight columns is a wide table, not an absurd one.

## A list literal whose lowered method exceeds 64K of code

```java
String big = HEAD + "let f (n) = Out { n = List.sum(["
        + IntStream.range(0, 5955).mapToObj(String::valueOf).collect(joining(", "))
        + "]) }\n";
Compiler.compile(big);
```

```
java.lang.IllegalArgumentException: Code length 65539 is outside the allowed range in apply(Object)Object
```

Last accepted 5954 elements, first refused 5955.

## The same literal where the constant pool goes first

At 40000 elements the pool overflows before the code length does, and says so differently again:

```
java.lang.IllegalArgumentException: 80031 is not a valid index.
    Entry: 11 java/util/List.copyOf-(Ljava/util/Collection;)Ljava/util/List;
```

## And the derived codec, not the constructor

A `data` with 2000 fields passes the constructor and overflows in the codec:

```
java.lang.IllegalArgumentException: Code length 147953 is outside the allowed range in decode(Object,Path)Result
```

So the limit is not reached in one place that could be guarded once at the end. Four generated
members — the constructor, a lowered body, the derived decoder, the constant pool they share — each
reach a different one.

## Where it comes from

#336 already described this failure shape, in its own words:

> Every phase descends an expression by recursion — parsing, inlining, checking, emitting — so past
> some depth the stack runs out. That is not a `CompileException`, so left alone it passes through
> the recovery boundary and reaches the author as a stack trace, and takes the language server's
> dispatch loop with it. It is reported like any other thing the compiler cannot accept.

It closed that for one cause, `StackOverflowError`, and for the one reader that had no guard at all,
`Formatter`. The boundary catches nothing else. `Compiler.java:75` and `Compiler.java:240` are
`catch (StackOverflowError _)`; `Main.java:141,144`, `:244,247` and `:375` catch only
`CompileException` and `IOException`. The classfile API's `IllegalArgumentException` is neither, so `souther build`
prints a stack trace for the second and third cases above.

The first case does not reach a boundary at all. The classfile API writes a signature of any length without
complaining, so there is nothing to catch.

## Proposal

Two parts, and they are not the same part.

Report the limits where they are decided, as diagnostics with a position:

- the number of argument slots a generated constructor takes, at the `data` declaration, counting
  `Int` as two and every other field as one — this is the one that is currently silent
- the code length of a generated method, at the definition whose body it is
- the constant pool, at the definition being emitted

Each says what the author's move is: split the record into nested ones, split the work into another
behavior so it is emitted as another class, or move the table out of the source. Naming parts with
`let` is not one of them — a non-recursive helper is inlined before the backend sees it, so it does
not divide the generated method.

Then widen the recovery boundary so an emitter that fails for a reason nobody listed is a compiler
diagnostic rather than a stack trace. #336 argued that for `StackOverflowError`; the argument does
not depend on which throwable it was.

kawasima:
Closes #341.

`Backend` emitted with the class file writer and never asked whether what it was writing was a class that could be loaded. Three of the JVM's structural limits were reachable from ordinary source and none was a diagnostic; the first did not fail at compile time at all, so the compile succeeded and the failure moved to whoever loaded the class.

## What is answered where

**Before anything is emitted (E2101).** The argument slots a generated method takes are decided by the declaration, so they are counted there. The width each may reach differs because the emitted form differs:

| declaration | becomes | limit |
|---|---|---|
| a `data`'s fields | the record constructor's arguments (an instance method) | 254 slots — an `Int` field is a `long` and takes two, every other field one |
| a `behavior`'s parameters | `apply`'s arguments | 254 — every parameter is a reference whatever it was declared as |
| a `behavior`'s dependencies | its implementation constructor's arguments | 254 |
| a recursive helper's parameters | a `static` method's arguments on `$Fns` | 255 — nothing is spent on a receiver |

A composition declares neither its parameters nor its dependencies, so both are counted from the signature and the requirement set it will be emitted from, not from what it wrote.

**While it is emitted (E2102, E2103).** How much code a method becomes and how many constants a class refers to are not known until it is written, so they are reported as the definition that was being emitted when the writer refused. The message names the generated method, because one declaration becomes several and which of them grew is not readable from the declaration. The recursive helpers share one class, so a code-length refusal is traced back to the helper whose method it names, and a constant-pool refusal — which names no method — belongs to the module.

**Only what is past a limit.** The writer refuses an empty method body with the same sentence as an oversized one (`DirectCodeBuilder`), and an index of zero with the same sentence as one past the pool's last entry (`BufWriterImpl`). The number is read, so a refusal that is not one of these limits goes on unchanged rather than being reported to the author as something they wrote too much of.

**And the boundary.** `Main` caught `CompileException` and `IOException` and nothing else. A failure nobody listed is now one line naming what was thrown, after those two, and `RuntimeException` rather than `Throwable`.

## Two things worth knowing

The issue's fourth case — a `data` with 2000 fields overflowing its derived decoder — **is no longer reachable**, because E2101 caps a `data` at 254 slots first. At 254 fields the decoder does not come near the code-size limit even with deeply nested collection types, so there is no test for that path; E2102 is reached by a behavior's body and a recursive helper's body.

The language server needed no new boundary. `LspServer.run` already catches per message and reads on, and `Analyzer` catches on each of its diagnostic paths, so a refusal has never cost the session.

The issue body has been corrected: it said to count `Decimal` as two slots, which contradicted its own table — `Decimal` is a `BigDecimal` reference and takes one — and it named ASM, which this backend does not use.

## Verification

The whole reactor: 2683 compiler, 151 lsp, 109 syntax, 1 cli, no failures. The parameter-slot tests load the accepted side through a class loader rather than only compiling it, so the width that is allowed is shown to be a width the JVM takes. The classifier's thresholds are pinned in unit tests; the source-level tests only assert that a small one is emitted and a large one is reported, so neither side is tied to a number the emitter could reasonably change.

Not covered by a test: the `catch (RuntimeException)` clauses in `Main` themselves. The line they print is unit-tested, but each path calls `System.exit`, so the wiring is verified by reading.


kawasima:
Fixed in 76413f3e — all four, and the first one was a real gap.

**1. `Constant pool is too large N` is now E2103.** Verified in the JDK: `SplitConstantPool.writeTo:163` throws it when `size() >= 65536`, on a path `DirectClassBuilder` always takes after writing fields, methods and attributes. The classifier only read `BufWriterImpl`'s refusal, so that path fell through to a raw `IllegalArgumentException`.

The two are one code and two messages, because the number means different things — the entry that was wanted, against how many entries there are. `Limit` now carries which refusal it was (`CONSTANT_POOL_INDEX`, `CONSTANT_POOL_SIZE`) and the message key with it, so neither says the other's number. Both series are pinned on both sides:

```
65535 is not a valid index    → not a limit      Constant pool is too large 65535 → not a limit
65536 is not a valid index    → E2103            Constant pool is too large 65536 → E2103
```

Checked that the new tests bite by making the pattern match nothing and watching them fail, rather than only watching them pass.

**2. The boundary is on the dispatch.** `Main.main` is now `guarded(() -> dispatch(args))` and the three duplicated catches are gone. `fmt`'s per-file catch stays, since one file it cannot read should cost that file and not the run. A `CompileException` reaching the boundary is rethrown rather than reported as a fault of the compiler.

This also made the wiring testable, which it was not before: `guarded` takes the command, so a test can throw an `InvalidPathException` at it — a failure from reading the arguments, which is exactly the case that was escaping — and assert both the exit code and the line. The one thing still not covered is `main`'s three-line call into it, because it calls `System.exit`.

**3. The result union, its encoder, and the injection base are said at their behavior.** Wrapped with `behaviorOf(module, resultName)` and `spec.pos()`. The bridge case stays the module's, and now says why in a comment: it is shared by every union that reaches its member, so no one behavior owns it.

**4. `\R`.** `said.replaceAll("\\R", " ")`, with `"first\r\nsecond\rthird"` in the test.

Whole reactor still green: 2689 compiler (six more than before), 151 lsp, 109 syntax, 1 cli.

Issue #343, opened 2026-08-05
The invariant checker carries its own typer and its own name identity, and is the one file the cleanups that retired both did not reach


kawasima:
`check/InvariantChecker.java` is 2397 lines, with 75 `instanceof`, 23 `return null` and 13 `default`
arms. `TypeChecker` was 4490 lines and is now 466 across eleven classes. The discharge analysis never
had that pass, and the reason it is worth doing now is not the line count — it is that the two
cleanups which stopped a class of recurring corrections elsewhere are both stated as rules the rest of
the compiler follows, and this file is the one place neither reached.

**The cause is the other way round.** This was written as two cleanups that did not reach one file.
Measured, it is one cause with two consequences: this analysis reads the only representation in the
compiler that the checker cannot type, so it had to rebuild typing and name resolution locally, and
that is what kept it outside every correction the rest of the compiler received.

Elaborating the discharge tree fails on 30 of 75 behaviors across the invariant tests, always the same
way — `List.map`, `List.filter`, `List.all`, `List.allUniqueBy`, `List.member`, `Bool.not` are "not a
standard-library function". `CallElaborator` treats a name arriving at call elaboration unexpanded as
an internal error: typing is defined only over the fully-expanded vocabulary, and `InliningPolicy`
keeps this representation's operations standing on purpose. So proposal 1 below cannot be done as
written; asking `Elaborator.typeOf` is not available to this walk today.

What is available: every local reference in the discharge tree carries a `BindingId` already —
measured over 136 behaviors, none missing — so the identity the fact store needs is there as soon as
the tree is typed. The order is therefore

    the discharge representation becomes one the checker can type
      -> typeExpr is unnecessary rather than delegated
      -> BindingId is in hand
      -> the string fact key is unnecessary

and the change is to typing and name resolution, not to this file alone.

**A location is derived once from typed Core and is identified by resolved declaration identity, not
spelling. `GuardThresholds` and the invariant facts consume the same location representation and the
same projection-normalization rules.** `GuardThresholds.pathOf` and `TermPath` already implement
`chainKey`'s rule over Core — its own comment says so — with a spelling for the root. Once this
analysis reads Core too, nothing separates the two, so they become one rather than three.

## It re-derives types instead of asking the checker

```java
// InvariantChecker.java:2292
// --- a minimal local typer (enough for atom/affine detection) ------------------------------

private Type typeExpr(Ast.Expr raw, Scope scope) {
```

[...]

PR #351, opened 2026-08-05
The invariant check reads a representation the checker typed

kawasima:
Closes the work on #343.

The invariant-discharge check carried its own typer and its own name identity. The issue read that as two cleanups that had not reached one file; measured, it is one cause with two consequences — this analysis read the only representation in the compiler the checker could not type, so it had to rebuild typing and name resolution locally, and that is what kept it outside every correction the rest of the compiler received.

So the order is: the discharge representation becomes one the checker can type, and the rest follows.

## The representation is typed

A call the analysis keeps standing had nowhere to be typed: `CallElaborator` treated a standard-library name arriving unexpanded as an internal error, because typing was defined only over the fully-expanded vocabulary.

- Which calls a representation keeps is that representation's to say (`Preserved`), and the checker asks only whether this call is one of them. An operation kept with no discharge rule types like any other and simply states nothing, so being able to type a call and being able to reason about one stay independent.
- A kept call is typed from its declared signature onto a node of its own, `Core.PreservedCall`, carrying what the name resolved to rather than how it was written. Refused by node kind rather than by operation name — a `sealed` interface caught only one of the four readers that had to change; two exhaustive switches in `BodyGen` and one `default ->` in `GrowingFold` passed it silently.
- An operation kept standing must declare what it answers, raised where the representation says what it keeps rather than at whichever call was reached first. 28 operations in the `.sou` library gained their return types.

Five things a preserved call had been hiding came out of this — a signature's application is not a fold's, a lambda not reduced away is a lambda given a name, the argument that states nothing must not settle the variable, a bound block needs an expected type, and a function-typed `let` reads its parameter types from being handed over and not only from being applied.

## The analysis reads it

The walk is over typed Core, so nothing in the check derives a type: the local typer is gone and so is the scope of names it needed. Which value a fact is about is the binding a name was answered with (`Location`), so a body that binds one spelling twice states two things by construction — nothing has to be forgotten when a name is bound again, and the walk no longer drops every fact whose text mentioned it.

The clause had to be typed too. A clause read over one tree and a body over another are two term grammars that must keep producing the same key for the same value, which is the agreement discharge rests on and nothing would have held them to. A declaration's fields are bindings already, so a clause is elaborated once and read at a construction by putting the value each field is being given where that field is read. That deletes the machinery that used to carry a clause's names to a site: the bindings record, the two path rules, the term-and-rule pair a container resolved to, the splice that let one key hold a term of each tree, and the string identity underneath all of it.

Measured over the suite, 1050 clauses type and 34 do not — every one of the 34 in a program that is already an error where it is declared.

## What is left is separated

One file of 2397 lines becomes the walk (662) and five files it reads with: what the language's operations keep (`DischargeRules`), a declaration's invariant read at a value (`Clauses`), where a value is and what can be said of it (`Terms`), what a clause owes and what a guard settles (`Predicates`), and the values threaded through all of it (`Known`, `Denotations`).

`GuardThresholds` asks `Location` which fields are steps instead of restating the rule. Its own path stays its own: a partition is derived from what a behavior declares, and a declared parameter is not a binding — a behavior with no implementation has axes all the same. So the issue's proposal to delete `TermPath` is not taken, and the reason is written where the two meet.

## What this leaves for later

- `preservedCall` does not go through `applySignature`: the two differ in how a function argument is typed, and joining them is a change to how a signature is applied.
- `Elaborator.handedOver` settles a kept signature without the bottoms-last ordering `preservedCall` has, so an empty collection argument can settle a variable differently in the two places.
- `DischargeRules`' combinator table and `TotalityChecker`'s are 24 of 25 identical rows under two key types.

## Verification

The whole suite is green — 2709 in the compiler, 151 in the language server, 109 in the syntax, 1 in the CLI. Warm compile is 144.5 ms for crm and 18.1 ms for issuetracker, against 144.8 and 18.1 on `develop`.

### DIFF
--- a/souther-compiler/src/main/java/souther/compiler/check/CallElaborator.java
+++ b/souther-compiler/src/main/java/souther/compiler/check/CallElaborator.java
@@ -125,7 +125,7 @@ public final class CallElaborator {
         // it is not this walk's business, and a kept call with no rule types like any other.
         CompleteSignature kept = ctx.preserved().signatureOf(call.denotes());
         if (kept != null) {
-            return preservedCall(call, kept, env, ctx);
+            return preservedCall(call, kept, env, ctx, expected);
         }
         CallArgs ca = new CallArgs(call.args(), env, ctx);
         Type result = typeOfCall(ca, call, env, ctx, expected);
@@ -150,38 +150,13 @@ public final class CallElaborator {
      * try to emit.
      */
     private static Core preservedCall(Ast.Apply call, CompleteSignature kept, Scope env,
-                                      CheckContext ctx) {
+                                      CheckContext ctx, Type expected) {
         List<Type> params = kept.params();
         CallArgs ca = new CallArgs(call.args(), env, ctx);
         if (call.args().size() != params.size()) {
-            throw CompileException.of(
-                    Diagnostic.of(null, "check.arity").title("check.arity.title")
-                            .at(call.pos(), call.written().length())
-                            .args(call.written(), params.size(), call.args().size()).build(),
-                    "`" + call.written() + "` takes " + params.size()
-                            + " argument(s) but is applied to " + call.args().size());
+            arity(call, params.size());
         }
-        Map<String, Type> bind = new HashMap<>();
-        // The values first, so what they settle is in force when a function argument is read at the
-        // parameter types this call gives it — `List.map(f, xs)` decides `'a` from `xs` and hands `f`
-        // the element type rather than a variable.
-        // An argument that answers no value states nothing about one — an empty collection carries a
-        // bottom, and letting it fix a variable would hold every other argument to the element type
-        // of nothing. The ones that state something go first, and a bottom then widens to what they
-        // settled instead of the other way round.
-        List<Integer> stating = new ArrayList<>();
-        List<Integer> bottoms = new ArrayList<>();
-        for (int i = 0; i < params.size(); i++) {
-            if (params.get(i) instanceof Type.FnOf) {
-                continue;
-            }
-            (Type.mentions(ca.type(i), BottomInfer::answersNoValue) ? bottoms : stating).add(i);
-        }
-        stating.addAll(bottoms);
-        for (int i : stating) {
-            TypeOps.unify(params.get(i), ca.type(i), bind, ctx.symbols(),
-                    call.pos(), "argument " + (i + 1) + " of " + call.written());
-        }
+        Map<String, Type> bind = settledByValues(call, params, kept.result(), expected, ca::type, ctx);
         for (int i = 0; i < params.size(); i++) {
             if (params.get(i) instanceof Type.FnOf declared) {
                 Type.FnOf at = (Type.FnOf) TypeOps.substitute(declared, bind);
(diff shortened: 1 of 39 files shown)


kawasima:
Fixed in `88cb4d4b`. All three points taken; the first one was real.

## 1. The blocker — measured, then fixed at the shared level

Your example does not reproduce at source level: `List.filter(x -> true, [])` in a position expecting `List<Int>` is rejected on `develop` too, so it is a pre-existing hole in the expansion path rather than something this branch introduced. I probed eight shapes (closure answering an empty list, empty map, a fold seeded with `[]`, `Option.withDefault([], …)`, nested maps) against both `develop` and this branch and found no divergence.

But the mechanism you point at is real, and it shows the moment the call is elaborated directly:

```
List.filter : (('a) -> Bool, List<'a>) -> List<'a>
List.filter(x -> true, [])   expected List<Int>

before: List<Nothing>, predicate typed (Nothing) -> Bool
after:  List<Int>,     predicate typed (Int) -> Bool
```

Two tests hold it and both fail without the fix (`OneCallSettlesOneSignatureTest`).

Rather than thread `expected` into `preservedCall` alone, I did what your "究極的には" paragraph asks for. Applying a declared signature was written three times with three different subsets of the rules, and they had already drifted in both directions — `applySignature` pinned from `expected` and did **not** order its arguments; `preservedCall` ordered its arguments (bottoms last) and never saw `expected`; `handedOver` did neither. That is the same shape of defect #343 is about, so the settlement before any function argument is typed is now one method:

```java
static Map<String, Type> settledByValues(Ast.Apply call, List<Type> params, Type result,
                                         Type expected, IntFunction<Type> argType, CheckContext ctx)
```

1. pin the result's variables from the expected type, where the signature takes a function at all;
2. unify each value argument, the ones that state something before the ones that answer none.

`applySignature`, `preservedCall` and `handedOver` all ask it. What each does *after* stays its own — the fold reads its step's result as the accumulator to grow, the ordinary application does not — which is why I did not merge the two call paths outright.

One thing I could not give `handedOver`: there is no expected type at the point it asks. It walks a body looking for what a binder was handed to, and no position has been reached yet, so it passes none and learns only what the values settle. That is now a stated fact in the code rather than an omission, and the parameter type it learns can no longer differ from the call's in the two respects that were drifting.

## 2. `Preserved.NONE` was building the DISCHARGE signatures

Correct, and for the reason you give — a class is initialized whole. It is a lazy holder now, so `FULL` no longer depends on the `CompleteSignature.of` contract and the demand is raised where a representation actually says it keeps something.

## 3. The clause-typing property is fixed as a test

`EveryClauseADeclarationPassesTypesInTheDischargeRepresentationTest`: for seven declarations — newtypes, spreads, an optional field, a quantifier, a projection, a module helper inside an invariant — every clause of a declaration this compiler accepted must become typed Core. The property, not the count, as you suggested. It asserts a clause was read at all, so the test cannot pass by finding nothing.

I left `Clauses.typed` catching `RuntimeException` broadly: fail-open is the existing policy and narrowing it would turn an analysis bug into a rejected program. What the test fixes is the thing that made the swallow dangerous — that nobody would notice.

Suite: 2714 in the compiler (5 new), 151 in the language server, 109 in the syntax, 1 in the CLI. Warm compile 143.3 ms for crm and 19.3 ms for issuetracker, unchanged.

Still outstanding, as follow-ups rather than in this PR: `DischargeRules`' combinator table and `TotalityChecker`'s are 24 of 25 identical rows under two key types; and `DischargeRules` answers with argument indices that ten call sites each bounds-check, where answering with the argument itself would remove the guard.

kawasima:
Fixed in `0a5dcdda`, one level below where you pointed.

You are right that this is not a performance question, and right that it reached the ordinary application path — `applySignature` asked once per value argument before the settlement was shared, and the shared version asked twice.

I did not fix it by holding the types in an array at that one call site, because the site is not what guarantees anything. Two things now hold, and each has a test that fails without it.

**The settlement reads each value argument once**, in the order it is written. What its ordering decides is which argument settles a variable first, and nothing about how many times one is read. That covers `handedOver`, whose reader (`j -> typeOf(...)`) is not idempotent and never will be.

**And the arguments of a call answer for that themselves.** `CallArgs.type` elaborates on the first ask and remembers it. The class already said "each elaborated once", and `requireTyped` already rested on it — it reads the stored Core *on the grounds that* nothing is elaborated twice. That was a claim no code enforced. Now it is true by construction, so a rule may ask in whatever order it types in, which is the reason the arguments are held in an object rather than in a list at each rule.

The failure mode you described was exactly right, and it is worth writing down because it is the general shape: elaborating an argument is not a pure function of the expression. It can decide a variable of the application it stands in, and that decision is written into the enclosing substitution. So "ask again" and "ask once and remember" are not the same question, and the difference is invisible — the classification, the unification and the Core kept in the tree simply stop being three readings of one thing.

Tests: a settlement over a counting reader (a function argument read not at all there, a value argument exactly once) and one that asks an argument's type twice and finds one Core.

Suite: 2716 in the compiler, 151 in the language server, 109 in the syntax, 1 in the CLI. Warm compile 139.2 ms for crm and 17.3 ms for issuetracker.

kawasima:
Follow-ups filed rather than left in this thread: #353 (the combinator table kept twice, once per checker) and #354 (the rules answering with argument indices that nine readers each bounds-check). #352 is the larger one this unblocks — eliding the run-time check where the compiler proved the invariant.

PR #356, opened 2026-08-05
A rule about a library operation is read off the operation, or its absence is written down

### LINKED ISSUE #353: Two checkers keep the same table of what a combinator hands its closure, and a row added to one is silence in the other
`TotalityChecker.COMBINATORS` (`TotalityChecker.java:301`) and `DischargeRules.COMBINATORS` (`DischargeRules.java:42`) are the same table: the same record shape — `(closureArg, elementParam, containerArg)` against `(closureArg, elementParam, listArg)` — and the same 24 rows, in the same order, with the same two comments about `List.indexedMap` and `Map.fold`. They answer the same question: which argument of a stdlib combinator is the closure, which parameter of that closure is the element, and which argument is the container it comes from.

They differ in one row and in how they are keyed.

`List.fold` is in the totality table and not in the discharge one, correctly: totality reads the surface tree, where an author's `List.fold` is still spelled that way, and the discharge tree is read after the rewrite to `List.foldFrom`, so a rule under that name could never fire (#230 is that mistake, already made once and now held by `InvariantCombinatorRulesTest`). And the discharge table is keyed by `ValueName.Stdlib` — what the name resolved to — while the totality one is keyed by the string `call.reaches()`.

## What it costs

A combinator the library gains has to be added twice, and the failure of adding it once is silence rather than an error: the totality check quietly stops crediting an element as a sub-term (a valid recursive helper is rejected), or the discharge check quietly stops binding the element (a construction inside the closure is reported as unproven, or a guard on the element stops discharging). Neither says anything about a missing row. `InvariantCombinatorRulesTest` holds the discharge table to being reachable and to having a program that fires each rule; nothing holds the two tables to each other.

The split is not old. Both were one table in `InvariantChecker` until #343 gave the discharge rules a file, which moved them further apart rather than closer.

## What would settle it

One table, holding what is true of the operation — which argument is the closure, which parameter is the element, which argument is the container — with each reader asking it under the identity it has. That means answering a `ValueName.Stdlib` lookup for the discharge side and a spelling for the totality side, or moving totality onto resolved names too, which it may already be able to do.

The one real question is `List.fold`: the row belongs to the operation as an author writes it, and the discharge tree never sees that spelling. That is a fact about which representation a reader holds, not about the operation, so it wants to be stated as one — the table says what `List.fold` hands its closure, and a reader of a tree where the rewrite has already happened simply never asks. Worth confirming that a rule keyed by a sugared name cannot be looked up in the discharge tree by construction rather than by test, which is what `Prelude.sugared` already knows.

kawasima:
Closes #353.

## What a combinator hands its closure is read off its declaration

The two tables said nothing the signature does not: the argument that takes a function is the closure, and the parameter whose type is what a container argument holds is the element — which `Terms.elementType` already answers for the walk that reads it. So `Combinators` reads the rules off `Prelude` and there is no table. The derivation reproduces every one of the twenty-five rows the discharge table had, and adds `List.distinctBy`, which was in neither. A signature admitting more than one reading raises where it is read rather than answering half; today's library has none.

The totality check moves off the spelling onto what the name denotes. The one name the two readers do not share is the sugar — `List.fold` is still written where totality reads, and gone where discharge does — so what it rewrites to is said once by the library and the rewrite reads it there. That the discharge side cannot look a sugar up is not a rule anyone keeps: what a representation keeps standing is built from declarations, and a sugar has none.

Measured rather than assumed: dropping the `List.fold` row from the totality table fails `CompileTotalityTest` twice, so that reader does hold the written spelling.

## A question a rule answers says which operations it is asked of

Deriving the combinator rules removes one table, not the mechanism that produced it. Of the thirty-three operations that answer a container built from a container, fourteen had a rule and nineteen had nothing — most because a shape genuinely cannot say what `Map.union` or `Set.insert` keeps, and `List.distinctBy` because the row was never written. Which was which was recorded nowhere.

Each question (`BUILT`, `PREDICATE`, `SIZE`, `OPERATOR`) now states its range, settled by what an operation is declared to be, and an operation in range answers — with a rule, or by being named among the ones there is nothing to say of, with the reason. Adding an operation to the library fails the build until someone decides which. The other direction is held too: a name said to have nothing to say must be one the question is asked of.

Answering the nineteen added eight rules. A rule is worth having only where something can travel through it, and the check names a statement by its container's kind — `Set.toList` and `Map.values` answer the same elements, which is true and carries nothing, because a statement about a set and one about a list are not the same statement. What is left is the eight that stay within a kind: `Map.remove`, `Set.remove`, the two intersections and differences, `Map.updateIfPresent`, and `List.distinctBy`.

## Held to

- `InvariantCombinatorRulesTest` — every derived rule the discharge check can reach has a program that fires it, and a sugared name's rule is absent from what the representation keeps standing
- `AnOperationTheLibraryGainsIsAnsweredForTest` — both directions of each question's range
- `ARuleAboutABuildingCarriesSomethingTest` — each new rule has a construction that discharges only because of it; dropping all eight fails all eight
- `CompileTotalityTest` — a recursion through the closure of a combinator nobody registered

ADR-0097; the preservation table in `specification.adoc` gains the eight rows. Full suite green, and green at the first commit alone.


kawasima:
All three are right, and they are the same omission: the range-and-answer mechanism lived inside `DischargeRules`, so the question whose answer is derived sat outside it. Fixed in 2b9e8c4a. The mechanism is now a `Question` of its own, and every question — the derived one included — says the same four things: what it asks, which operations it is asked of, where its answers are, and where its silences are written.

## A derivation that gets nothing out of a signature (P2)

`Question.COMBINATOR`'s range is an operation that takes a function; its answer is the derived rule, and its silence is `Combinators.HANDS_ITS_CLOSURE_NOTHING` — empty today, with the reason. An operation that derives no rule fails unless it is written down as one whose closure is handed nothing a container holds.

Mutating `ruleFor` to answer null for `List.map` reports `List.map — what it hands its closure`.

## PREDICATE was one OR over independent rules (P2)

The questions are now as fine as the rules: `PREDICATE_CARRY`, `EMPTINESS`, `QUANTIFICATION`, `PROJECTION`. Each range is read off the declaration alone and none of them consults another table — `QUANTIFICATION` asks any operation that answers a `Bool` about a container by applying a predicate to what it holds, and `PROJECTION` any that computes something other than a truth value from each element.

So `List.any` is in `QUANTIFICATION`'s range and answers by being named in `NOT_A_QUANTIFIER`, with what it says instead. Dropping `List.all` from the quantifiers, `List.allDistinctBy` from the projections and `List.isEmpty` from the emptiness table reports all three:

```
[List.all — whether it states its predicate of every element,
 List.isEmpty — which size call it means,
 List.allDistinctBy — which argument is the projection it is stated over]
```

A third check went in with them: a rule must answer a question its operation is in range of, so a rule keyed by a name nothing asks is caught from the other end.

## The rewrite was general in what it recorded and specific in what it did (P2)

`Rewrite(String target, List<Integer> supplied)`, with `keptArgs()` derived from the target's own count. `desugar` writes out what the rewrite supplies, so a second sugar is not rewritten as a fold. That a sugar supplies constants and nothing else is what the record's shape says; one that had to supply anything else could not be written down as this.

## The witness test claimed more than it held

`CARRIES` is pinned to `DischargeRules.builtNames()`, and all twenty-two rules have a witness. Dropping thirteen of the pre-existing rules reports all thirteen, and dropping `Set.map` alone reports it.

Holding them that way turned up something: **the two `collapses` rules had no witness anywhere**. With both removed, all 2786 tests passed. They do carry the upper bound — `bounds()` applies to every shape that does not keep the size — so the programs that show it are now written.

ADR-0097 and the index are updated. Full suite green.

kawasima:
Both are right. Fixed in 39c8d906.

## The blanket skip on a declaration with no return type (P2)

You are right that it is the same silence, narrowed to the declarations nobody wrote a return type for — and `signatureOf` allows exactly that for a Souther-bodied helper with parameters. `askedOf` no longer decides anything on behalf of the questions; each reads the part of the declaration it is about, and a null result falls out false everywhere it matters (`holdsElements(null)`, `== Type.Prim.BOOL`, and `OPERATOR` short-circuiting before `result::equals`).

Held to a signature written in the test rather than to one the library declares, since the point of a range is to hold an operation nobody thought of:

```java
Prelude.Signature leavesItsResultToItsBody = new Prelude.Signature(
        List.of(new Type.FnOf(List.of(Type.Prim.INT), Type.Prim.INT), Type.Prim.INT), null);
assertEquals(List.of(Question.COMBINATOR), Question.askedOf(leavesItsResultToItsBody), ...);
```

Putting the early return back fails it: `expected: <[what it hands its closure]> but was: <[]>`.

## The reverse direction over the tables (P3)

Also right, and I have done it now rather than deferring it — it is the same defect from the other end, and leaving it would have meant the ADR claiming more than the tests hold. Each question answers with `answeredOperations()`, and both a rule and a silence must be written against an operation the question is asked of.

The two mutants it now catches:

```
[String.trim — what it keeps of the container it is built from (a rule),
 List.every — whether it states its predicate of every element (a rule)]
```

— a rule outside its question's range, and a rule for a name the library does not have. The two directions are one test now (`everyOperationWrittenDownIsOneTheQuestionIsAskedOf`), since a rule and a silence go wrong the same way.

A sugar is asked what the call it becomes is asked, so `Combinators`' rule for `List.fold` is checked against `List.foldFrom`'s declaration rather than being skipped for having none of its own.

ADR-0097 records both. Full suite green.

Issue #346, opened 2026-08-05
A `partial` helper handed over as a function value leaves the call graph


kawasima:
```
module demo
data Item = { n: Int }
data Out = Int
behavior run : (i: Item) -> Out constructs Out
partial let spin (n: Int): Int = spin(n)
let apply1 (f: (Int) -> Int, n: Int): Int = f(n)
let run (i) = Out(apply1(spin, i.n))
```

`apply1` has no call edge to `spin`, and its parameter type `(Int) -> Int` (ADR-0076) says nothing
about termination. This compiles, and on its own that is consistent — nothing is being claimed about
`apply1`.

It matters once #345 makes a helper answer for what it reaches: a walk over the call graph does not
see this, so the guarantee would hold on the direct path and not through a function value.

On the invariant path the hole is currently closed by expansion rather than by a rule. With a
non-recursive higher-order helper,

```
let anyPositive (f: (Int) -> Int, items: List<Item>): Bool = List.all(i -> f(i.n) >= 0, items)
data X = { items: List<Item> } invariant anyPositive(spin, items)
```

is rejected, because substitution puts `spin(i.n)` back into the clause. Through a recursive one it
would not be (#344).

## What could be done

Cheapest first.

- **Bar naming a `partial` helper as a value.** It may be called, not handed over. Small change, and
  what the corpus hands to combinators is closures, not `partial` helpers.
- **Leave function-typed positions outside the guarantee and say so.** This is today's behaviour; the
  change is that the hole is written down in the spec instead of being found by reading the checker.
- **Carry partiality in the function type.** `(Int) -> Int` splits into a total and a partial form,
  every higher-order signature picks one, and ADR-0076 grows a second colour throughout. Precise, and
  more machinery than the corpus has asked for.

The first, with the third recorded as rejected for now, looks like the right trade — but this is a
language surface decision, so it should be settled together with #345 rather than implemented ahead
of it.

kawasima:
Closed by #358 (merged into `develop` as 45b121f8). An unmarked helper now carries the termination guarantee for everything it reaches; `partial` declares that a helper does not carry it, rather than that it diverges. See ADR-0098.

Issue #344, opened 2026-08-05
An invariant reaches a `partial` helper through a total recursive one


kawasima:
`TotalityChecker` proves a recursive helper terminates, and `HelperTyping.rejectPartialHelperInInvariant`
bars a `partial` one from an invariant. Between them an invariant is meant to be a computation that
always finishes — it runs on every construction (spec §invariant-expressions). It does not hold: a
total recursive helper hides everything it reaches.

```
module demo
partial let spin (n: Int): Int = spin(n)
data Tree = { child: Tree?, n: Int }
let depth (t: Tree): Int =
    match t.child with
        | Some c -> depth(c) + 1
        | None -> spin(t.n)
data X = { root: Tree } invariant depth(root) >= 0
```

This compiles. Constructing an `X` never returns.

## Where it comes from

The check walks the *inlined* clause and looks for a call whose name is a `partial` recursive helper:

```java
static void rejectPartialHelperInInvariant(Ast.Expr e, String data, Set<String> partial) {
    if (e instanceof Ast.Apply call && partial.contains(call.reaches())) {
```

Inlining is what makes a reachable call syntactically visible, and inlining stops at recursion. A
non-recursive wrapper is caught, because its body is expanded into the clause and the call to `spin`
appears there. So is one handed the helper as a value: with

```
let anyPositive (f: (Int) -> Int, items: List<Item>): Bool = List.all(i -> f(i.n) >= 0, items)
data X = { items: List<Item> } invariant anyPositive(spin, items)
```

the clause reads `spin(i.n)` after expansion and substitution, and is rejected. A *recursive* wrapper
is not expanded, and nothing else looks inside it.

## What it costs today

Nothing catches this at compile time except an example that happens to construct the data. Adding a
row that builds the `X` above reports E1910 (`did not finish within 2000ms`) — the example runner's
deadline, not the invariant rule. Without such a row the module compiles clean and the construction
hangs at run time, in the one place ADR-0003 makes a construction answer to.

## Proposal

Decide the question by reachability rather than by what inlining left visible: from the invariant
clause, walk the call graph of every helper it names and reject when a `partial` one is reachable,
reporting the path.

Everything the walk needs is already there. This module's helpers are in `HelperInliner.helpers()`,
and a published helper's body travels as source in the jar (`ModuleMetadata.invariantHelpers`), so it
crosses a module and a project boundary the same way the current check does — an imported `partial`
helper called directly from an invariant is rejected today, naming it `pricing.spin`.

[...]

PR #358, opened 2026-08-06
An unmarked helper may not reach a `partial` one

### LINKED ISSUE #345: A helper the totality check certifies may call a `partial` one, so `total` does not mean it terminates
ADR-0052 states the guarantee as "User recursion is total unless explicitly `partial` ... This is
Souther's totality guarantee". What the compiler proves is narrower: that the descent of *this*
recursion is structural. A certified helper may call a `partial` one, and then it does not terminate.

```
module demo
partial let spin (n: Int): Int = spin(n)
data Tree = { child: Tree?, n: Int }
data Out = Int
behavior run : (t: Tree) -> Out constructs Out
let depth (t: Tree): Int =
    match t.child with
        | Some c -> depth(c) + spin(t.n)
        | None -> 0
let run (t) = Out(depth(t))
```

This compiles. `depth` is certified total and does not terminate on any tree with a child.

## Where it comes from

`TotalityChecker.check` analyses the strongly-connected group of the helper under test, and nothing
else:

```java
Set<String> group = cycleMembers(name, ownEdges);
...
// `partial` opts out; a `partial` anywhere in a mutual group opts the whole group out.
if (group.stream().anyMatch(m -> own.get(m).partial())) {
    continue;
}
```

`spin` is not on `depth`'s cycle, so it is not in the group. The size-change proof is about the
arguments of the recursive calls and is sound about that; what is missing is that the callees off the
cycle terminate too.

## Proposal

Propagate. A helper not marked `partial` that reaches a `partial` one, transitively over the call
graph, is an error naming the path — its caller has to be `partial` as well. This is Idris2's rule (a
`total` function may not call a partial one), and it fits the existing asymmetry that `partial` only
disclaims totality and nothing claims it.

Two things to decide first.

**Where propagation stops.** A behavior's implementing `let` is the outward boundary and publishes no
totality claim, so it is a candidate terminal. Without a terminal, one numeric-recursion helper puts
`partial` on every helper above it. Which is right depends on who the marker is written for: if an
author of the model is meant to read it, spreading it is the point; if it is bookkeeping for the
checker, a terminal keeps it readable.

**Whether `partial` becomes part of what a published helper promises.** Today the flag crosses a jar
because the body travels as source (`ModuleMetadata.invariantHelpers`) and the reading side sees the
word — an imported `partial` helper called from an invariant is rejected by name. Under propagation
it becomes a rule about an interface and should be stated as one.

The cost is not small: `partial` spreads along every call chain out of a numeric-recursion helper.
That is the honest reading of what is being disclaimed, but it changes how much of a corpus carries
the word, so it is worth measuring on `examples` before committing.

## Related

- #344 — the invariant boundary, which can be closed on its own without deciding this.
- A helper handed over as a function value has no call edge, so the walk proposed here does not see
  it; filed separately.

kawasima:
Closes #345, #344 and #346, which are one rule between them.

## What the word means

`partial` declares that a helper does not carry Souther's termination guarantee. It does not say the helper diverges. The same word covers three things and used to be read as only the first:

- a recursion that loops (`spin`)
- one that terminates and that size-change cannot prove (the specification's own illustration, `partial let firstOver` — a linear search that `List.find` already does, replaced here)
- one the size-change closure was too large to decide (`tooComplex`)

Read as "may loop" the word is wrong about two of them. Read as "the compiler is not answering for this one" it covers all three, and propagating it does not spread a lie.

## The rule

An unmarked helper may not reach a `partial` one, through any chain of calls. E2001, reported at the helper's own name and naming the path:

```
`let depth` does not carry the termination guarantee it claims: it reaches the
`partial` helper `spin` (depth -> measure -> spin). Mark `depth` `partial` as
well, or stop reaching it.
```

Before this, the guarantee held only inside a recursion's own strongly-connected group. A structurally-certified helper could call a `partial` one off its cycle and not terminate (#345), and an invariant — checked on every construction — could reach one behind a recursive helper the inlining left standing (#344).

Around it:

- **Mutual groups.** Every member of a group holding a `partial` member needs the word, each reaching the others. The size-change skip is unchanged; the implicit exemption from saying so is gone.
- **The behavior boundary.** A behavior's implementing `let` is not a helper and publishes no termination guarantee, so the rule is not asked of it. `HelperInliner.helpersOf` already says this by excluding behavior names. It is not a boundary the walk stops at — a helper between the behavior and the `partial` one still needs the word.
- **Function values** (#346). A `partial` helper may be applied but not written where a value goes: a function type says nothing about termination, so a name handed over leaves the call graph. Colouring the function type instead is recorded as rejected on scope, not on correctness.
- **Values.** Reading a `let` with no parameter list is reaching it — the value is substituted where it is named. That is an edge in the walk rather than a value-position rejection, since there is no function there. Without the split, `let a = spin(1)` / `let b (x) = a + x` leaves `b` unmarked.
- **Across a jar.** The answer is the word on the declaration being called; the reader does not walk what is behind it. That holds only if the writer enforced the rule, so `BOUNDARY_VERSION` moves 5 → 6 and an older jar is refused rather than believed.

## How it is built

`PartialReachability` is its own graph, kept apart from the one `TotalityChecker` groups a recursion by. That one has the module's own helpers as nodes because a strongly-connected group is a group of bodies this module can see; this one reaches past the module, with an imported helper as a terminal. Paths are shortest, breadth-first with callees in name order, so a module always reports the same path.

One walk answers both questions — the helper's and the invariant's — so a clause and the helper above it are reported in the same build rather than one per fix. Both are asked per declaration, so a module needing the word in several places is told about all of them at once.

`TotalityChecker` keeps to size-change analysis; `ownCallGraph` is untouched.

## Cost, measured

`souther-examples` and the shipped core contain no `partial` declaration at all — every module that could have needed one is written the other way round, and the READMEs say so. The corpus builds clean, so the spread costs nothing today.

In the compiler's own fixtures, 20 tests failed and needed:

- 15 fixtures where a caller had to say the word
- 1 fixture whose `partial` was decorative (non-recursive) and was removed
- 1 test whose intent changed, split into "every member needs the word" and "all of them accepts"
- the boundary version, and four fixtures where E2001 had come to sit in front of the error the test was about

25 new tests. 3092 in the repo, all passing.


kawasima:
Pushed 84b93a8b. The root cause is as diagnosed. Neither of the two reported reproductions fails today, though — and *why* they do not is the finding.

## The two reported cases

Both were run.

```
P1-a  invariant -> imported partial, called directly
      REJECTED: the invariant of `X` reaches the `partial` helper `maths.spin`
                (invariant -> maths.spin)

P1-b  unmarked helper -> bare-imported zero-parameter partial value
      REJECTED: E2001 `let offset` ... (offset -> maths.seed)
```

Instrumenting the graph construction says why.

```
P1-a  [DBG] own=[maths.spin] partial=[maths.spin]
P1-b  [DBG] own=[offset]     partial=[maths.seed]
```

- **P1-a**: an imported *recursive* helper arrives in the reader's own fns through `injectedRecursiveHelpers`, so it is in the set as if the reader had declared it. The only imported helper an invariant can name is a recursive one that survived inlining, so this path always looks like that.
- **P1-b**: found as a side effect of walking `offset`'s edge, exactly as described.

So the checks hold, but they hold **because of how another pass behaves**. A soundness rule resting on that is the thing to fix, and it is.

## What changed

**The `partial` set is gone.** The marker is read off the declaration.

```java
private boolean isPartial(String name) {
    Ast.FnDef declared = inliner.helper(name);
    return declared != null && declared.partial();
}
```

This one is not a no-op. In the previous commit `anImportedPartialHelperMayNotBeNamedWhereAValueGoes` failed with "nothing was thrown", and this is what fixed it. I reported at the time that the qualified-name fallback fixed it; removing that fallback afterwards left all 25 tests passing, so it was the set. Correcting that here — the PR body said the same thing and is wrong.

**Name resolution goes through one function, `HelperInliner.keyIn`,** which reads the key off what a name denotes rather than off how it was written.

```java
public static String keyIn(String module, ValueName.Helper helper) {
    return helper.module().equals(module) ? helper.name() : qualified(helper.module(), helper.name());
}
```

This is the rule `publishedClosure` already used, named and shared rather than invented. Call edges, value edges and the value-position check all go through it, so they cannot disagree about which declaration a name reached. The two-key fallback in `isPartialFunctionNamed` is gone with it.

I could not construct an end-to-end failure for this one. No path reaches the check before `qualifyImports` has run, as far as I could get. **It removes a dependence on another pass rather than fixing a reproducible defect**, and it is pinned by a test on the seam itself, not by a compile.

**A library name is a node now.** Leaving it out would have made "the standard library declares nothing `partial`" a premise of the check rather than a fact about the library — one it could stop honouring in silence. No answer changes today.

## The value-position rule's range

All three surfaces were measured. Each rejects, by two different mechanisms.

| surface | result | what rejects it |
|---|---|---|
| fn body | REJECTED | the value-position rule |
| invariant, through a non-recursive higher-order helper | REJECTED | what the clause reaches (`invariant -> spin`) |
| invariant, through a **recursive** higher-order helper | REJECTED | the same |
| example row | REJECTED | a behavior cannot take a function |

The invariant case works because the argument `spin` is eta-expanded into `(x) -> spin(x)` during inlining, so the clause holds a call. That is true whether or not the helper it was handed to is expanded with it. An example row has nothing to hand over, a function having no external representation to cross the boundary with.

Of the three options offered, this takes the second: each surface's rejection is pinned by a test, so the premise breaking breaks a test.

## Tests

Six added, 31 in the class, including the two named as merge conditions.

- `anInvariantMayNotCallAnImportedPartialHelperDirectly` — `invariant -> maths.spin`
- `readingABareImportedPartialValueIsReachingIt` — `offset -> maths.seed`
- `aHelperIsKeyedByWhatItDenotesRatherThanByHowItIsWritten` — the keying rule on its own
- `anInvariantHandingOverAPartialHelperIsRejectedByWhatTheClauseReaches`
- `andThroughARecursiveHigherOrderHelper`
- `anExampleRowHasNoFunctionToHandOver`

3098 across the repository, and `souther-examples` builds clean.

Issue #354, opened 2026-08-05
The discharge rules answer with an argument index, so every reader bounds-checks a table that knows the arity it was written for


kawasima:
`DischargeRules` answers `Built.from()`, `Combinator.closureArg()/elementParam()/listArg()`, `Carried.container()` and `projectionOf` as `int`. None of them is meaningful on its own — each is "the argument of this call" — so every caller turns it back into an argument and guards the turn:

```java
Built built = DischargeRules.builtFrom(call.operation());
if (built == null || built.from() >= call.args().size()) {
    return;
}
Core source = call.args().get(built.from());
```

That shape appears at `Predicates.java:65`, `:108`, `:428`, `:458`, `:467`, `:499`, `:505`, at `InvariantChecker.java:269`, and the same guard on the totality side at `TotalityChecker.java:378`. Nine sites, one guard, and the table that knows the arity performs none of it.

## What it costs

The guard is not defending against anything an author can do — a call reaches these rules already typed against the signature the operation declares, so an index the table wrote for that operation is in range. It is defending against the table and the signature disagreeing, which is a real possibility exactly once: when the library changes an operation's parameter order and the table is not updated. Nine copies of a bounds check answer that by doing nothing quietly, at whichever site is reached first, which is the same silence #353 describes.

A reader also has to know which index means what. `built.from()` is the container, `combo.listArg()` is the container, `carried.container()` is the container — three names for the argument each rule reads its container from, and a caller that swaps two of them is a compiling program with a wrong answer.

## What would settle it

Have the rules resolve against the call rather than return positions into it:

```java
// the container this call built its result from, and what the building kept — or null
DischargeRules.Source builtFrom(Core.PreservedCall call);   // record Source(Core container, Shape shape)
DischargeRules.Closure combinator(Core.PreservedCall call); // record Closure(Core.Block step, Ast.Binder element, Core container)
```

Then no caller indexes anything, the "which argument" question is answered once per rule, and a table that disagrees with a signature is a null answer in one place — where it can be said out loud rather than swallowed nine times.

The awkward part is `combinator`, whose closure argument may be a name bound to a block rather than a block (`Terms.blockOf` resolves that today), so it needs the denotations to answer fully. Either it answers the argument and leaves the resolving to the caller, or the resolving moves with it. Worth deciding before the rest, since it is the rule with the most readers.

Follow-up from #351, where the rules were given a file of their own.

kawasima:
Done in #359 (merged into develop as 1fb134a).

PR #359, opened 2026-08-06
The discharge rules resolve against the call, and are bound to the declarations they are about

kawasima:
Follow-up from #354.

## What was wrong

`DischargeRules` answered `Built.from()`, `Carried.container()` and `projectionOf` as `int`, and `Combinators` its three positions the same way. None means anything on its own, so every reader turned one back into an argument and guarded the turn — twelve sites, fourteen bounds checks.

Measured before changing anything: thirteen of the fourteen are dead. Making every guard throw and running the whole suite (3098 tests) fires none of them, and every arity mismatch an author can write — argument count, closure parameter count, intrinsic or inlined — is reported before the discharge check reads a call. A `PreservedCall` is built only where the signature accepted the arguments, so one that exists has them.

What the guards were defending against is a row disagreeing with a declaration, and each answered that by returning quietly at whichever site was reached first. Nothing said the row was wrong: adding `String.contains -> Carried(7, PERMUTES)` passed all 2822 tests, including both `Question` ones, which range over whether a row exists and not over what it says.

## What this does

**A rule answers a call.** `builtFrom`, `carried` and `projectionOf` take one and answer with the container, the shapes, the projection; `Combinators.handedTo` answers with the block, the parameter the element arrives on, and the container it comes from. No reader indexes anything.

**Which argument a rule reads is written as the part it plays**, wherever the signature already answers it (`Reads.TheContainer`, `Reads.TheClosure`) — fifteen of the twenty-nine rows, every one of which was restating a position `Combinators` derives. `List.all` had its container written in one table and derived in another; now it has one. The fourteen that remain are decisions a signature does not settle: an operation given two containers, or one that takes no closure.

**Every row of every table is held to the operation it is about, on the first ask** — as `Combinators` and `Preserved` already are, not eagerly. The operation is declared, the argument is one that declaration has, what stands there is the kind of thing the rule is about, and a position the signature already answers is refused. `String.contains` reading argument 8 now fails the build.

**One guard survives, and it is not about the tables.** The totality check reads the tree an author wrote, beside the checks that report an arity rather than after them, so it does reach a call already known to be wrong — `List.sortBy(x -> go([x]))` inside a recursive helper gets there. That is a fact about the surface tree, so it is answered once in `Combinators.handedTo(Ast.Apply)` with the reason, and pinned by a test.

**A carrying rule is held to what each shape does, from both sides.** `BUILT_FROM` and `Combinators` each had a test holding every row to a program; `CARRIED` had none. A rule too narrow loses a discharge; a rule too wide accepts a construction that can abort, and no program that discharges can detect that. Every shape of every carrying rule now has a program on the side it belongs to, written in the test rather than read off the table it holds.

## Two things writing the cells out surfaced

`Set.contains` and `Map.containsKey` carry through `PERMUTES`, and no set or map construction permutes — a set holds its elements in no order. Both rules are true and reach nothing any program can write. Now said, rather than left looking witnessed.

A mapping over a map leaves its keys alone, so `Map.containsKey` does survive one. Not stated, because `MAPS` is about the elements and not about what they are held under; saying it would need a shape that does.

## Held to

- a row reading an argument the declaration does not have, one that is not a container, one naming a part of something the signature says the operation does not hand, one for an operation the library does not declare, and one writing a position the signature already answers — each fails the binding
- widening `List.contains` to carry through `SUBSET`, and narrowing `List.all` to drop it — each fails the carrying witness
- `Set.intersection` reading the other of its two containers — fails the building witness
- a combinator of the wrong arity in a recursive helper is reported as what is wrong with it

3112 tests, all green.

### DIFF
--- a/souther-compiler/src/main/java/souther/compiler/check/Combinators.java
+++ b/souther-compiler/src/main/java/souther/compiler/check/Combinators.java
@@ -83,19 +83,27 @@ final class Combinators {
      * The same, off the tree an author wrote — where a closure written as anything but a block is one
      * this says nothing about, the tree not yet having been read for what names denote.
      *
-     * <p>Nor is a call written with fewer arguments than the operation takes. That is not this table
-     * disagreeing with a signature — it is what the surface tree still holds: the walk that reads it
-     * runs beside the checks that report an arity, not after them, so a call already known to be
-     * wrong reaches here. Nothing about the arguments it does not have is true, so nothing is said,
-     * and the arity is reported by the check whose question it is. The tree a representation keeps
-     * standing needs no such answer: a {@code PreservedCall} is built only where the signature it was
-     * applied to accepted the arguments, so one that exists has them.
+     * <p>Nor is a call the operation would not have accepted — written with fewer arguments than it
+     * takes, or handed a block with fewer parameters than it applies one to. That is not this table
+     * disagreeing with a signature: it is what the surface tree still holds, the walk that reads it
+     * running beside the checks that report an arity rather than after them, so a call already known
+     * to be wrong reaches here. A sugar is how both arrive — it has no declaration of its own, so
+     * what is said about the arity of {@code List.fold} is said against the call it becomes, and by
+     * then this has already read it. Nothing about arguments or parameters a call does not have is
+     * true, so nothing is said, and the arity is reported by the check whose question it is.
+     *
+     * <p>The tree a representation keeps standing needs no such answer: a {@code PreservedCall} is
+     * built only where the signature it was applied to accepted the arguments and typed the block it
+     * was handed, so one that exists has both.
      */
     static Written handedTo(Ast.Apply call) {
         Combinator rule = of(call.denotes());
         if (rule == null || rule.closureArg() >= call.args().size()
                 || rule.containerArg() >= call.args().size()
                 || !(call.args().get(rule.closureArg()) instanceof Ast.Block step)) {
+            return null;
+        }
+        if (rule.elementParam() >= step.params().size()) {
             return null;
         }
         return new Written(step, step.params().get(rule.elementParam()),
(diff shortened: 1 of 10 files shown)


kawasima:
Both fixed in `20064ab`. CI green.

## P1 — you are right, and the reachable case is sugar

I had measured this before and concluded the surface tree could not hold it. The measurement was wrong: I only tried the parameter case over combinators spelled under their own name (`Map.mapValues`, `List.mapIndexed`, `Map.filterEntries`, `List.foldRight`, `Map.fold`). Every one of those is reported before the totality walk, which is why the probe never fired.

The two that do get through are the sugars:

```
let go (xs: List<Int>) : Int = List.fold(acc -> go(xs), 0, xs)
let go (s: Set<Int>)  : Int = Set.fold(acc -> go(s), 0, s)
```

A sugar has no declaration of its own, so what is said about the arity of `List.fold` is said against the call it becomes — and by then `handedTo(Ast.Apply)` has already read it. Without the guard, `step.params().get(rule.elementParam())` throws `IndexOutOfBoundsException`. Removing the guard again now fails two tests with exactly that, so the boundary is pinned rather than argued.

Fixed as you wrote it, in two stages for the pattern variable's scope. The class doc and `TheSurfaceTreeStillHoldsACallOfTheWrongArityTest` now state both incompletenesses — too few arguments and too few parameters — and the sugar as how each arrives.

## P2 — also right, and it was the same flaw twice

`aShapeSaidToHaveNoConstructionHasNone` compared two tables written in the same file. It held them to each other and to nothing else. That is the flaw the class doc warns about, one level up: I moved the expectations out of `CARRIED` and left them reading a catalogue I had also written.

`DischargeRules.constructionKinds()` now answers which `(kind, shape)` pairs the library builds, read off the building table — whose every row already answers to a program that fires it. The test holds **both** the matrix and its own program catalogue to that.

Simulated the case you named by giving `Set.filter` the shape `PERMUTES`:

```
Set.contains through PERMUTES: the library builds this and no answer is written
Set.PERMUTES: the library builds this and no program is written for it
```

## Minor

`carriedThrough` removed — it was the leftover from the witness reading the table it holds. `bind` now refuses a negative position.

3113 tests, clean build.

PR #360, opened 2026-08-06
A row is held to what it goes through, not to how long it takes

### LINKED ISSUE #347: Whether a model compiles can depend on how fast the machine is: E1910 is a wall clock
Every row gets 2000ms of wall clock, and E1910 is what the compiler says when the row does not come
back inside it:

```java
private static long readTimeoutMs() {
    String override = System.getProperty("souther.example.timeout.ms");
    if (override == null) {
        return 2000L;
    }
```

So a row that terminates but is slow — a large fixture, a loaded machine, a cold JIT — is reported as
one that did not, and the same model compiles clean on one machine and fails on another. That is a
property users have, not only a test problem.

#331 removed the race from the tests by making the deadline a seam they can decide. Its closing note
left the product side open:

> Counting steps rather than milliseconds would make the *product* deterministic too — today the same
> model can compile clean on one machine and report E1910 on another [...] That means instrumenting
> the classes generated for example evaluation, and is a much larger change than the seam.

ADR-0052 names the same fix and defers it: "a deterministic step budget (a counter in the generated
loop) is the complete fix, deferred as a follow-up". The spec is already written as though it were in
place — §E1910 says "the step budget the example runner holds it to", which is not what runs.

## Proposal

Count steps, not milliseconds. Instrument the classes generated for example evaluation with a counter
and report E1910 when the budget is spent. The reading then answers the same on every machine, and a
model that compiles somewhere compiles everywhere.

The wall clock stays as the outer guard for what a counter cannot reach — a runaway inside a runtime
or library call, with no generated loop on the stack — but it stops being what decides an ordinary
row.

Whichever way this lands, §E1910 and the diagnostic text should then describe the mechanism that
actually holds the row.

Independent of #344, #345 and #346: those ask what `partial` claims, this asks that the safety net
under `partial` code answer the same way twice.

kawasima:
Closes #347.

Every row got 2000ms of wall clock, and E1910 was what the compiler said when the row did not come back inside it. So whether a row was reported as non-terminating was a property of the host: the same row failed on a busy machine and held on a quiet one.

## What decides a row now

The classes generated for an evaluation count what they go through. Every loop the emitter emits counts its backward branch, and every recursive helper counts the frame it adds. A row that passes more counted points than the policy allows, or recurses deeper than it allows, is reported — and how many counted points a row passes is a property of what the row does.

Counting in the code is also what stops the code. A worker asked from outside to stop reaches no interrupt point in a pure computation and goes on burning a core for the life of the JVM.

## What is guaranteed, and what is not

Whether a row is reported as **E1910** does not depend on the machine, on how loaded it is, on what the JVM has compiled, or on what any row before it did.

Whether the compiler answers at all is weaker, deliberately. Work the compiler owns rather than generates — loading a class, decoding a value through a derived decoder, comparing two of them — is not counted, and an evaluation that stops answering within its wait is **E1923**. That work is bounded by what the counted work produced, since a value only gets large by being built and building it costs counted points; but bounded is not counted, so it is still measured in time. Counting it would mean instrumenting decoders this compiler does not generate.

## The three axes, measured before and after

| | before | after |
|---|---|---|
| a terminating row under CPU load (64 burners, 16 cores) | 1.5s quiet, compiles → **E1910** loaded | 34ms quiet, 33–38ms loaded, **compiles** either way |
| `count(10000)` under `-Xss512k` / `1m` / `8m` | **E1910 / E1910 / compiles** | **compiles** at all three |
| workers left after 17 non-terminating rows | **17 alive**, indefinitely | **0** |

## Three things told apart that were all E1910

- **E1910** — the model reaches something that goes round more than an example may, or recurses deeper. Decided by counters.
- **E1923** — the evaluation stopped answering. The compiler failing to decide, not the model failing to terminate.
- **E1924** — the JVM stack ran out before the counted depth limit. A fact about how large the frames are, with the settings to change named in the hint.

E1920 and E1921 carry the same distinction for a `fake` table that could not be read. Every one of these is classified where the failure arrives — one place per boundary — rather than where it happened to cross a reflection call.

## Scopes, and one evaluation per compile

Coverage stays on the module whose report reads the arm numbers; counting covers every module in the source closure, because an evaluation goes wherever the row goes.

A row also runs once per compile now. Asking what the rows covered used to run all of them a second time against different bytecode on a second wait. `Output.Probed`, `Output.ProbedLinked` and `Adequacy.ProbedExamples` are gone.

The classes an evaluation runs are the ones this compile generated: `MemoryClassLoader` looks in itself for the names it holds, so a class file left on the module path from an earlier build no longer answers in their place. The set is whole or absent — never short a module.

## Defaults, measured rather than picked

`RowOutcome` carries `stepsSpent`. A fold over 8000 elements costs 8000 steps against a limit of 100,000,000, and a check keeps the corpus inside that margin.

The depth limit is 50,000 (was 10,000). On the 64MB stack an evaluation is given, a plain recursion and one carrying eight extra parameters both reach 100,000 counted frames before the stack, and the wide one loses at about 149,000. Too low reports a legitimate deep walk as a recursion that does not terminate — a wrong answer about the model; too high reports that the stack ran out first, which says the settings are wrong for this model and what to do about it.

## Settings

`souther.example.timeout.ms` is replaced by `souther.example.step.limit`, `souther.example.recursion.depth`, `souther.example.outer.timeout.ms` and `souther.example.worker.stack.bytes`, read when a compilation is set up rather than once into a static — a build daemon and a language server both outlive the compile a setting was written for.

## Dependencies

A module read from the module path is not regenerated. What a published module carries is what an importer needs to read its declarations, so a behavior's body stays in the jar it was built into — and regenerating what travels while taking the rest from the jar splits a module across two loaders, which makes two types under one binary name and fails a cast on a model that is fine.

Instead its classes are taken whole and given a counted point on every backward branch as the evaluation loads them. One loader, one version of every type, and a row that loops inside a dependency spends the budget there. Nothing anyone ships changes and a jar built before this existed is counted like one built after.

What rewriting cannot add is the recursion-depth count, which the emitter provides by moving a helper's body aside and wrapping it. A non-tail recursion inside a jar is bounded by the stack the evaluation runs on and reported as E1924 rather than by the depth limit.

## What is left

Process isolation for uncounted code that will not stop is a separate question, and worth deciding from what E1923 and E1924 actually turn up rather than in advance.


kawasima:
All five confirmed against the code and fixed at the mechanism. Pushed in 1f01bac.

## P1 — ModulePath class wins over the counted class

Reproduced. `MemoryClassLoader` and the `ModulePath` loader both override only `findClass`, so `loadClass` is parent-first and the path answers first. The `ShadowsPath` check refuses a path module only when `PublishedModule.read` finds its declarations, so anything the check cannot recognise as a module — a directory of loose class files, a jar built before the declarations were stamped on — slips through.

Measured with the stale module's `$Module` class removed: the row was compared against the earlier implementation and a correct model reported `E1905: This example does not hold`.

`MemoryClassLoader.loadClass` now looks in itself for the names it holds and leaves every other name to the parent, so the JDK, the runtime and the compiler's own classes are still loaded once by whoever already has them. `ModulePath.loader`'s javadoc claimed the ordering; it now points at where the ordering actually lives. Regression test in `AnEvaluationRunsTheClassesThisCompileGeneratedTest`, both directions.

## P1 — raw `StackOverflowError` escaping as a compiler failure

Correct, and it is the placement that was wrong rather than a missing case. `Deadline` delivers every `Throwable` to `Threw`, and that branch had no `Error` arm, so anything not converted at one of the two reflection boundaries fell through to `IllegalStateException`. `ExampleStatements.within` was worse: it rethrew `Error` outright, so a table that overflowed took the build down instead of saying what it could not check.

I could not build a model that reaches it — the parser's nesting guard fires before a deep literal reaches a decoder, and helper and behavior routes both cross a boundary — so it is stated through the deadline seam rather than raced for. Removing the arm and rerunning gives `IllegalStateException: java.lang.StackOverflowError`, which is the failure mode you described.

Both boundaries now classify it. The reflection-site conversions stay, but only to name the helper; they are no longer the only classification. New `Unread.StackRanOut` variant for the fake-table reading, with its own messages.

## P1 — "machine-independent" was stated more strongly than delivered

You are right, and this is a defect in the text I wrote rather than in the mechanism. The two-tier form was already settled earlier in the design discussion and I wrote the strong version into the spec anyway.

I did not take option 1. The large uncounted work is decoding and comparing values that helpers produced, which runs inside derived decoders this compiler does not generate; counting it means instrumenting third-party decode paths. Worth recording that the exposure is bounded rather than unbounded — a value only gets large by being built, and building it costs counted points — but bounded is not counted, so it is still measured in time.

So the guarantee now reads: whether a row is reported as E1910 does not depend on the machine; whether the compiler answers at all is weaker, and an evaluation that stops answering is E1923. Changed in `specification.adoc` §e1910, in ADR-0052's amendment and consequences, and in the PR description above.

## P2 — `EvaluationLinked` swallowing a missing module

Correct, and it made the `PROBE_MAPPING_LOST` branch unreachable, which is worse than not having it. The set is now whole or absent, with reports propagated. `AnEvaluationSetIsWholeOrAbsentTest` covers both that a reaching set goes absent and that a whole one is still answered.

## P2 — `stepsSpent` zero for a budget-exhausted row

Correct. Measured: a row stopped at a 10,000-step limit reported `stepsSpent=0`. The read was on the success path only; it is now in the `finally`, so it covers every exit rather than the two exception types — a row that leaves any other way records what it cost too. The `RowOutcome` contract said zero meant the row did not run, which was wrong for a second reason: a row whose body has no loop in it legitimately passes no counted point. It now says that, and points at `disposition` for telling them apart.

## On the tests

Each fix has a test that fails without it, verified by removing the fix and watching the test fail rather than by assuming it would. The two routes with no model that reaches them go through the deadline seam.

Full clean build across all modules is green.

kawasima:
Three findings, three outcomes: one confirmed and fixed, one confirmed and fixed with a correction to what the PR itself claimed, one where the reported breakage did not reproduce but the underlying rule was still wrong and is now fixed. Pushed in 69e2c8a.

## [P1] budget exhaustion is not proof of non-termination — confirmed, fixed

You are right, and the demonstration is short. A helper that counts down to zero, held to a hundred steps:

```
E1910: This example did not terminate (spent its budget of 100 steps) —
       a `partial` recursion it reaches may not stop. Make it structural, or reduce it.
```

The model terminates. The counter never measured what the sentence asserted, and the wall clock had the same defect — counting made the reading the same on every machine without making it true.

E1910 now says what the counting shows, with the two budgets said apart because what to do about them differs and because either can legitimately be raised:

```
The evaluation of this row did not finish within its budget of 100000000 steps.
Hint: What the counting shows is that the evaluation did not finish, not that the model
does not: a `partial` recursion that never stops reaches this, and so does one that would
have stopped shortly after. If the model does finish and needs more, raise
`souther.example.step.limit`; otherwise bound the loop or make the recursion structural.
```

The specification section is renamed to "An example exceeded its evaluation budget", and the claim that a helper carrying the termination guarantee cannot reach it is gone — it is bounded, not free. ADR-0052 says the same.

On the sub-claim that the wrapper and the backedge counter are inserted without regard to `partial`: correct, and it stays that way. Distinguishing them would mean the budget depended on which helper you happened to be inside when it ran out, which is not a better answer. What was wrong was the sentence, not the placement.

## [P1] ModulePath modules refused as missing classes — did not reproduce, rule still wrong, fixed

The breakage did not reproduce. A dependency compiled and put on the `ModulePath`, imported by a source module with an example, evaluates fine — `Evaluated(lib.money)` is **present**, not absent.

The reason is worth recording, because it also corrects something this PR claimed. A published module travels as the source it was written as, so `Classes.inputs` answers for a path module and it is regenerated — with the counting. Measured: a `partial` helper that loops **inside a published module**, reached from an importing module's row, spends the budget in there and is reported for it, `spent=50000`. So the residual I wrote into ADR-0052 — that a `partial` helper inside a jar has no counted points — was wrong, and is now written as what it is: only a dependency this compiler cannot re-derive at all is uncounted.

The rule was still wrong in the direction you describe. Whole-or-absent now applies only to modules this compilation declares (`Front.Layout.idOfModule`); a module from the path that cannot be regenerated is left out rather than refused, because its classes are on the loader the set is put over — the name resolves and only the counting is missing. Refusing there would stop evaluating the rows of every module importing such a dependency.

New `ADependencyOnThePathIsCountedLikeAnyOtherTest`: a real published dependency on the path, a consumer with an example that runs, a row that loops inside the dependency and is stopped by the budget there, and a dependency that cannot be read not making the set absent. That is the coverage gap you identified — the previous tests had everything in one source compilation.

## [P2] `NonTerminationException` unclassified at the reading boundary — confirmed, fixed

Correct. A stack that runs out inside a helper crosses the reflection boundary and arrives named; anywhere else it arrives raw. Only the raw form was recognised, so the named one fell to `throw runtime` and left the compilation as a failure of the compiler rather than a report about the table.

Taken the first of your two options: both forms are read as the same thing at the boundary. The second would have cost the helper name in the row diagnostic, which is the one piece of information an author can act on.

## Confirmed from the previous round

All five carried over as you describe. Full clean build across all modules is green.

kawasima:
All three fixed, and the first two turned out to be one bug with one fix. Pushed in 95c5d7d.

## The two P1s were the same thing

Your second finding explains your first. I reproduced the split you described, and it is not hypothetical — it is what the earlier code did to *every* dependency with a behavior in it:

```
set holds lib.svc.$Fns?      false
set holds lib.svc.Spin$Impl? false
rows: [FAILED/INVOCATION spent=0]
compile says: E1905: This example does not hold.
```

`lib.svc` is regenerable, so its types were defined here; its behavior body does not travel, so that came from the parent. The row handed this loader's `Amount` to the parent's `Spin$Impl` and the cast failed — reported as an example that does not hold, about a model whose dependency loops forever. Wrong answer twice over.

You are also right that my test proved nothing. `spun` was exposed, so it travels as source and is re-emitted here; what the test showed was this compilation counting its own code.

## The fix

A module from the path is no longer regenerated at all. Its classes are taken whole and given a counted point on every backward branch as the evaluation loads them — `Recounted`, a ClassFile transform, applied by the evaluation's loader.

One loader defines every one of them, so there is one version of every type. The jar's behavior bodies are counted, so the case above is now:

```
rows: [INCOMPLETE/STEP_LIMIT spent=50000]
compile says: E1910: The evaluation of this row did not finish within its budget of 50000 steps.
```

Rewriting rather than publishing bodies, deliberately: publishing would change what an artifact carries and make an implementation readable by everyone who imports it, for the sake of a measurement, and would do nothing for a jar built before this existed. Rewriting works on both.

What it cannot add is the recursion-depth count, which the emitter puts in by moving a helper's body aside and wrapping it. A recursion inside a jar is bounded by the stack and reported as `E1924`. That is written down rather than glossed.

The all-or-none rule from the previous round is gone — it was a workaround for the split, and the split is now closed at its cause. `EvaluationLinked` carries only what this compilation declares, and remains whole-or-absent for those.

## [P2] `NonTerminationException` contract

Renamed to `StackExhaustedException`, with the contract rewritten: it carries a stack overflow that crossed a reflection boundary with the name of what was running still in hand, and the boundary that reads everything an evaluation ends with is what classifies it. It is enrichment, not classification — nothing decides anything by its absence. Left named for the old classification, the next person to touch it would have put the decision back where it was, which is your point.

## Tests

`ADependencyOnThePathIsCountedLikeAnyOtherTest` now covers what it claimed to: a published dependency on the path, a row that loops in an exposed helper, a row that loops in a **jar-resident behavior body**, and a dependency that cannot be read not making the set absent. The behavior-body case asserts the row spends its budget inside the jar, which is the assertion the previous version was missing.

Full clean build across all modules is green, 2868 tests.

kawasima:
All three documentation inconsistencies fixed in b46f819, and the PR description above is synced.

- `ADependencyOnThePathIsCountedLikeAnyOtherTest`'s class javadoc now describes the rewriting, not the reach it had before it. One sentence inside a test method said the same thing and is fixed with it.
- `EvaluationLinked`'s javadoc says what it now holds — only the modules this compilation declares — and points at `evaluationLoader` for the path's.
- The PR body's "What is left" no longer claims a published module's `partial` helper is uncounted. It now describes the rewriting, and states the one thing it cannot add: the recursion-depth count, so a non-tail recursion inside a jar is E1924 rather than the depth limit.

These were the previous round's design surviving in prose next to code that had moved, which is how the next person to edit the file learns the wrong contract from it.

Full clean build across all modules is green.

Issue #375, opened 2026-08-06
Warnings cannot fail a build, and E2011 still fires where a size can only grow


kawasima:
A compile that warns exits 0. `souther compile` has no `--warnings error`, no `--no-warn`, and no
per-site suppression: the exit code is decided by whether a `CompileException` was thrown, and the
warning list the command already holds is never read (`Main.java:277-296`).

Separately, E2011 still fires on constructions that provably cannot violate their invariant.

The original report tied these together through #363. That link is gone — and the case for both is
stronger without it.

## The example in the original report is fixed; the class is not

#363 was fixed by #382. Its repro now compiles with no diagnostic. This one does not:

```souther
data NEL = List<Int>
    invariant List.length(value) >= 1

let appendL (a, b) = NEL(List.append(a.value, b.value))
```

```
-- INVARIANT (WARNING)  E2011 --------------
Constructing `NEL` here may violate its invariant: the guards do not establish it.
```

`length(append(a, b)) = length(a) + length(b) >= 2` holds unconditionally. `S(a.value ++ b.value)`
under `invariant String.length(value) >= 1` warns the same way. Neither remedy the message names can
be written: there is no relation between the inputs to reify, and a guard would need a departure case
for a failure that cannot occur.

So #363 was never the cause. It was one entry point into an incomplete discharge domain, and
`append`, `++`, `union` and `insert` are others. The original report's example is stale; its claim is
not.

## Why the domain cannot say it

`DischargeRules.Shape` carries two properties in one enum: where the result's elements came from, and
how its size relates to the source's. For `map` / `sort` / `filter` the two point the same way.
`append` separates them — it is not built from `a`'s elements alone, yet `size(r) >= size(a)` holds
regardless.

Cardinality on its own is a two-valued model:

```
keepsSize() == true   ->  size(r) =  size(src)
keepsSize() == false  ->  size(r) <= size(src)      (Predicates.bounds)
```

There is no way to state `size(r) >= size(src)`. So `List.append`, `Map.union`, `Set.union`,
`Map.insert` and `Set.insert` sit in `NOTHING_KEPT`, whose comment states the fact and then discards
it:

> They put in what the container they read did not hold. Nothing that held of every element still
> does, and the size can rise; a shape says neither of those.

A lower-bound clause over such a result is expressible, so it is owed; nothing establishes it, so
`verdictOf` answers `UNKNOWN`, so E2011.

## What is sound for these five operations

| operation | what actually holds |
| --- | --- |
[...]

Issue #369, opened 2026-08-06
stdlib-list names zip among the list operations, but the function is zipShortest


kawasima:
`souther doc stdlib-list` lists the operations as:

```
map  filter  fold  foldRight  all  any  length  get  mapIndexed
take  drop  zip  unzip
```

There is no `List.zip`:

```
$ souther api --search zip
List.zipShortest(xs: List<'a>, ys: List<'b>) : List<('a, 'b)>
List.unzip(ps: List<('a, 'b)>) : (List<'a>, List<'b>)
```

The same section explains further down *why* the name is `zipShortest` — that whether a bare `zip` truncates or pads is a thing languages disagree on, so the name carries the answer. The operations block above it contradicts that paragraph and sends a reader to a function that does not exist.

kawasima:
Fixed in #393. The block now names `zipShortest`, and `EveryOperationAModulePublishesIsListedInItsSectionTest` holds every module's list of operations against what the module publishes, so the two enumerations cannot drift apart again.

Issue #364, opened 2026-08-06
E2011 is the one diagnostic with no section to look up


kawasima:
Every other code the compiler emits is addressable through `souther doc`. E2011 is not.

```
$ souther doc e2011
no section `e2011`
`souther doc` lists every section and topic
```

`e2010`, `e2012` and `e2014` all have sections. E2011 is the gap, and it is the invariant-discharge warning — the one most likely to be hit by someone who cannot tell whether their model is wrong or the checker is conservative. Its message already carries a hint, but the hint is not always followable (see #363), which is exactly when a reader goes looking for the section.

kawasima:
Fixed in #391.

`souther doc e2011` now answers, and so does every other code the compiler prints. Walking the whole set first showed E2011 was one of eight, and that two neighbouring contracts were broken the same way: `E2010` did not resolve although `e2010` did, and seven anchors the specification cross-references resolved to nothing. All three are now invariants with a test each.

Issue #361, opened 2026-08-06
A spread override that sets an optional field to None is dropped in an example fixture, so the row passes on a value nobody wrote


kawasima:
## What happens

In an `example` fixture, a spread construction that overrides an optional field with `None` keeps the spread source's value instead. The override is discarded silently.

```souther
module none

data Note = { body: String, tag: String? }

behavior identity : (n: Note) -> Note

let identity (n) = n

let tagged = Note { body = "b", tag = "t" }

example identity
    | "an override to None must survive the spread" :
        (Note { ...tagged, tag = None }) -> Note { body = "b", tag = None }
```

```
$ souther compile none.sou -d out --lang en
-- EXAMPLE  E1905 ----------------------------none.sou:13:45

This example does not hold.

It is:
    Note { body = "b", tag = "t" }
But it needs to be:
    Note { body = "b" }
```

`tag` is `"t"`. The `tag = None` override never took.

## Where the line falls

- `{ ...base, optField = "other" }` in a fixture — the override applies.
- `{ ...base, optField = None }` in a fixture — the override is dropped.
- `{ ...n, tag = None }` in an ordinary `let` body — the override applies. Confirmed by running the behavior: the result carries no `tag`.

So this is the fixture construction path, and only for `None`.

## Why this one matters more than its size suggests

The failure direction is silent. Here the expected side differs so E1905 fires, but a row written to assert the `None` case against an expected value that also spreads from the same source passes — and passes without ever exercising `None`. An example system whose rows can hold for a reason the author did not write is worse than one that is merely incomplete.

This surfaced because `souther examples` kept reporting that no row reaches `None` after two rows had been written that were believed to set it. The adequacy report was right and the rows were wrong, which is the only reason it was caught.

kawasima:
A second, independent reproduction characterises this more sharply than the report above: **the expected side of a row drops the override too, so two mutually contradictory expectations both pass.**

```souther
// ordinaryClaim has costCenter = "cc-42"
(Submitted { ...ordinaryClaim, costCenter = None }, EmployeeId("boss-1"))
    -> Approved { ...ordinaryClaim, costCenter = None,    approver = EmployeeId("boss-1") }   // passes
    -> Approved { ...ordinaryClaim, costCenter = "cc-42", approver = EmployeeId("boss-1") }   // also passes
```

[...]

PR #376, opened 2026-08-06
The toolchain answers what the language, its standard library and its dependencies are

kawasima:
Reading a Souther model means answering three kinds of question — what a construct means, what the standard library offers, and what a dependency at the Java boundary exposes. All three are answered today by opening a checkout of this repository or by disassembling a jar. A coding agent that has only the compiler cannot do either, and a person should not have to.

## What this adds

```
souther doc                    every specification section and shipped topic, name<TAB>title
souther doc newtype            one section, by its anchor
souther doc cli/run            a topic the command line ships about itself
souther doc --search decoder   ranked hits, each with the line it matched on
souther api                    the whole standard library surface, one screen
souther api --search fold      the names that answer a term
souther api --source Option    a module's own source, design comments included
souther japi net.unit8.raoh.Issues          a dependency's public API, with javadoc and throws
souther japi net.unit8.raoh.Issues#add      one member of it
souther mcp                    the same answers over MCP stdio
```

Three properties are worth stating because they are what make the answers trustworthy rather than merely present:

- The specification travels inside the jar, so a section is answered from the copy the compiler was built from.
- `api` answers with the signatures the type checker resolved, and includes the names that exist only as sugar over a private helper — `List.fold` among them, which the specification tells a reader to call and which a listing of declarations alone denies.
- `japi` parses class files without loading them, so no dependency's static initializer runs for the sake of printing its surface.

A library ships its own documentation inside its own jar under `META-INF/souther-docs`, so a reader gets the docs of the version actually on the class path. The command line's own reference is carried the same way and is nothing special — it is a doc set like any other. The companion change is souther-lang/raoh#122.

## How this was checked

Beyond the tests: six agents that had never seen Souther were each given a domain-modelling task, the CLI, and a prohibition on opening this repository or any jar. All six produced a model that compiled, and all six answered "no" when asked whether they would rather have had the checkout. Two rounds were run, the second after fixing what the first exposed.

The first round found real defects in this work, all now fixed and covered by tests:

- `japi` promised javadoc and delivered none on the ordinary invocation. The shaded jar has its dependencies inside it and nothing beside it, so the sibling `-sources.jar` lookup found nothing every time. The boundary libraries' sources are now bundled, and the source declarations name the parameters that the class files, compiled without `-parameters`, had left as `arg0`.
- `api` denied that `List.fold` exists.
- `--search` returned 131 of 207 sections for `type`, flat and unranked.
- A tool that threw took the whole MCP session down with it.
- The `souther run --input` encoding — a single argument is passed bare, several are an array — was written down nowhere. Three of three agents guessed at it. It is now the section headed "the part that is easy to get wrong", and in the second round three of three got it right first try.

The second round found the remaining complaint that four of six readers had made independently — that a list of section titles is a filtered table of contents and not a search — which is why each hit now carries the line it matched on.

The dogfooding also turned up defects in the compiler and specification that are outside this change; they are filed as #361 through #375, each with a reproduction taken by hand.

## Not included

#364 (E2011 has no section) and #369 (`stdlib-list` names a `zip` that does not exist) are specification gaps this branch could carry. They are left out so this stays one change with one purpose.


kawasima:
All five items in the merge line are addressed in cbf794e1, along with the multi-release read. Each was reproduced against the built jar before being changed.

**1 — the blank term.** Confirmed as described, and worse than a hang in one process: the server answered nothing at all and lost every request behind it.

```
$ printf '%s\n' '{"jsonrpc":"2.0","id":1,"method":"tools/call","params":{"name":"doc_search","arguments":{}}}' \
                '{"jsonrpc":"2.0","id":2,"method":"tools/list"}' | souther mcp
(no output; still running after 8s)
```

Fixed in two layers as you set out. `rank()` refuses a blank term and `count()` cannot be entered with an empty needle; arguments are checked against the tool's own published schema before the tool runs, and absent, non-string or empty is `-32602`. Now:

```
{"jsonrpc":"2.0","id":1,"error":{"code":-32602,"message":"`doc_search` needs `term`"}}
{"jsonrpc":"2.0","id":2,"result":{"tools":[...]}}
```

Regression tests cover `souther doc --search ""`, `souther api --search ""`, the missing `term`, the empty `term`, an argument of the wrong type, and that a request after each is still served. They are time-bounded, since the failure they guard against is a hang.

**2 — raoh is not in the artifact.** Correct, and the README was the thing overpromising. Waiting on a raoh release would block this indefinitely, so the README now says no published release carries the guides yet and names the change that adds them. The mechanism is what this PR delivers, and it is now tested against the real artifact rather than a synthesized one: a new `souther-cli` integration test runs the shaded jar and asserts the specification, the doc sets and their topics survive the shade, that the set registries were merged rather than one replacing the other, and that a bundled dependency's javadoc comes back with no sources jar beside the artifact. Six tests, run under `verify`.

**3 — overloads.** Correct, and it mattered more than it looked: the parameter-name recovery is only worth having because it is true. Javadoc is now filed under name and parameter count. Where two overloads take the same count, neither doc is attached — this reader does not resolve types, so printing nothing is the only other honest answer. Tested with different-arity overloads, same-arity overloads, and a class compiled without `-parameters`.

**4 — package listing.** Correct. `net.unit8.raoh.package-info` was being reported as public API. The class's own access flag now decides, and `package-info`/`module-info` go with the rest. A type asked for by name is still shown, with a line saying it is not part of what the jar publishes — inspecting one is legitimate, presenting it as published is not.

**5 — ranking.** Correct, and it defeated the point of the change: `souther doc --search input` returned no `cli/run`, which is the topic written to answer exactly that. Both sides now score into one type and sort once.

**Multi-release.** Taken, not deferred — the shipped jar is itself `Multi-Release: true`, so the base-copy read was the ordinary case rather than an exotic one. Class lookup and package enumeration both go through `JarFile.runtimeVersion()` now.

**Protocol revision.** The pin is gone. What this server serves is the same under every revision it knows, since it declares only `tools`, so the client's revision is echoed when it is one of `2026-07-28`, `2025-11-25`, `2025-06-18`, and the newest is offered otherwise. The README states the set. `server/discover` is not implemented and is optional in that revision; if you would rather it were tracked, say so and I will file it.

`mvn clean verify` is green: 2939 + 166 + 6 + 1.

kawasima:
All three High items and both non-blocking ones are addressed in 7fe4500a. Each was reproduced first; none of the three was a false positive.

**1 — the one-sided overload.** Correct, and it is the part of the previous fix that was doing the wrong work. Detecting a collision between two doc comments cannot see an overload that has no comment at all, which is exactly the case that mislabels. The count now comes from the class file, where an undocumented member is as visible as a documented one:

```java
// both are `read/1`; only the first is documented
public String read(String text)     // @param text
public String read(byte[] bytes)
```

Neither takes the comment now, and the undocumented one prints `read(byte[] arg0)` rather than borrowing `text`. Your suggested key — `(name, JVM parameter count) → overload count`, and refuse when it exceeds one — is what it does, and it stays inside the existing decision not to resolve types.

**2 — the version-crossing fallback.** Correct, and it broke the claim this PR is built on. Rather than check identity between the artifact and the tool, the lookup no longer leaves the artifact: sources carried inside are read from the same jar the class bytes came from. A jar that carries none and has no `-sources.jar` beside it gets no javadoc, which is the third of the options you listed. Containment does the work an identity check would have, and there is no configuration in which the two can disagree.

Verified both directions against real jars: `japi net.unit8.raoh.Issue` still returns the javadoc and the recovered parameter names from the shipped CLI jar, and the same class read from an older `raoh-0.4.1.jar` on `-cp` comes back with no prose at all. The regression test builds an old jar and a differently-documented same-FQCN source elsewhere, and asserts nothing from the second reaches the first.

**3 — `throws`.** Correct. It reads the generic signature's throws clause when there is one and the `Exceptions` attribute otherwise, for methods and constructors alike, and a thrown type variable prints as written. On a real class:

```
public tools.jackson.databind.JsonNode readTree(tools.jackson.core.JsonParser p) throws tools.jackson.core.JacksonException
```

Tests cover a method, several thrown types, a constructor, a generic exception parameter, and a method that throws nothing.

**Medium — a broken index.** Fixed. A topic an index names but the jar does not carry is a broken doc set rather than a topic with nothing in it: `read` distinguishes a missing resource from an empty one, and the command exits 2 saying which name failed instead of printing nothing successfully.

**Low — the PR body.** Updated; the `raoh/tutorial` line is gone.

`mvn clean verify` is green: 2947 + 166 + 6 + 1.

kawasima:
Addressed in 8e2c72a2. One correction to a premise, and agreement with the conclusion.

## The MCP revision

The recommendation is taken: the declared revisions are now `2025-11-25` and `2025-06-18`, an unknown one is answered with `2025-11-25`, a client asking for `2026-07-28` is not echoed back its own, the README says the same, and the test is narrowed to legacy revisions.

On the premise, the TypeScript SDK's own migration guide reads the other way: the `server/discover` probe is *"strictly optional"* and runs only when a client sets `versionNegotiation: { mode: 'auto' }` or pins a modern era, and the 2025-era handshake persists as a fallback. So a server without `server/discover` is not thereby non-compliant.

The conclusion holds regardless, for the reason further down that same page: on a modern connection the era is settled in the opening exchange and handlers read each request's identity from its envelope rather than from an `initialize` phase. This server does neither. Naming `2026-07-28` in an `initialize` answer claimed an era it does not speak, whatever the status of the probe — and a modern-only client is better off failing to connect than being told it succeeded. That is the same standard the rest of this PR is held to, so it should not have been declared. The previous round's "optional, so non-blocking" call was mine to get wrong.

## The mechanism behind the last three rounds

Three rounds found the same defect in three places — documentation attached to a member it was not written for — and each fix was another key: name, then name and count, then name and count with nesting refused and bridges excluded. Each key was a guess at identity, so each round found another way for it to miss. Rather than add a fourth, the source is now read with `JavacTask` and `DocTrees`, and a member is found by its enclosing type, its name, and the erased types of its parameters.

What that changes beyond the reported items:

- Nesting is a scope the parser tracks, so `Outer$Inner` takes its own documentation and none of `Outer`'s. The bail-out added last round is gone — nested types are documented again rather than skipped.
- Two overloads of the same count are told apart by their parameter types instead of both being given up on. `decode(String)` and `decode(byte[], Charset)` each keep their own, and so do `encode(String, String)` and `encode(String, int)`.
- A bridge is not in the source, so it cannot be mistaken for an overload in it. The special case for it is gone with the counting it existed for.
- There is no fallback to a name and a count. That fallback is precisely how one overload's documentation reaches another, and it is what made the one-sided case in the last round possible. A comment that cannot be matched goes unprinted.
- Where `jdk.compiler` is not present, nothing is read. The README already requires a JDK at run time; on anything less the API is still printed, without documentation.

## raoh

0.7.1 is released and carries its guides, so the dependency is that release rather than 0.7.0. `souther doc raoh/composition-patterns` now reads raoh's own guide from the built jar, and the integration test holds the shaded jar to serving both doc sets — which is what appending their registries exists for, and was previously only asserted against synthesized jars.

`mvn clean verify` is green: 109 + 2951 + 166 + 7 + 1.

The commit was amended after an initial push to describe the whole change rather than only the parser; the tree is identical, and the SHA to look at is 8e2c72a2.

kawasima:
All three are addressed in d7ed2294. Both blockers reproduced exactly as described.

```
$ souther japi acme.Use -cp use.jar
  /** Uses the b one. */
  public void use(a.Foo arg0)

  /** Uses the b one. */
  public void use(b.Foo arg0)

$ souther japi acme.Join -cp join.jar
  public String join(String right, String left)      # declared (String left, String right)
```

## 1 — the same simple name

Nothing is spelled into a key any more. A class file's parameters are their erased types in full; a declaration's are every fully-qualified type its file's imports allow them to be. A declaration answers for a member when each of its parameters could have erased to the one the class file has, and when no sibling declaration could.

Resolving the imports outright is what `task.analyze()` would give, and it needs the library's own class path — which this does not have and cannot assume, since it is reading a jar someone handed it. Keeping every reading the file allows, and refusing a member that two declarations could both be, gets the same guarantee without needing the types to resolve: a candidate set is a singleton whenever the type is written qualified, explicitly imported, or in the file's own package, which covers the reported case in both directions. Where it is not a singleton and two declarations overlap, neither is used.

## 2 — parameter names

Taken from `node.getParameters()`, in declaration order. The `@param` regex is gone. Two things follow: a file whose tags are written out of order no longer reorders the parameters, and a parameter that no tag mentions is still named — so a library compiled without `-parameters` names its parameters wherever its sources are at hand, documented or not. `read(byte[] bytes)` in the regression test is an undocumented overload that now names its parameter, where before it printed `arg0`.

## 3 — the nested constructor

Fixed. The lookup used `Outer$Inner` where the source calls the constructor `Inner`. Both sides now use the type's own last segment. A documented static nested constructor comes back with its doc and its parameter names.

A non-static inner class is not handled: its class file carries a synthetic leading outer-instance parameter the source does not declare, so the arity does not agree and nothing matches. That is a miss rather than a mistake, and it is left as one — dropping a leading parameter on the assumption of what it is would be the kind of guess this whole sequence has been about removing. Say the word if you would rather it were tracked.

## Also

The stale comment about counting by name and parameter count is gone with the counting. The nested-type test is renamed and now asserts the stronger thing the parser makes true: the nested type takes its own documentation, not merely none of its enclosing type's.

`mvn clean verify` is green: 109 + 2954 + 166 + 7 + 1.

kawasima:
949227ce. The reported case reproduced, and rather than tighten the rule again the rule is gone.

```
$ souther japi acme.Use -cp use.jar        # before
  /** Uses the local number. */
  public <T extends Number> void use(T local)

  /** Uses the local number. */
  public void use(acme.Number local)
```

## Why not the recommended fix

Both suggestions are correct and would have closed this case. I did not take them, because five rounds have now found the same defect five times, and each fix was a narrower rule standing in for identity: the simple name, then the name and a count, then the set of qualifications a file's imports allow, and next the same set made bidirectional. Every one of them was another pair of members that some rule read as one. Bidirectional uniqueness would have caught this instance; it would not have made spelling into identity, and the sixth case would be about whatever the candidate generator misses next — the `AnnotatedTypeTree` in your own note is one that is already visible.

## What it does instead

The source is compiled against the class path the class file was read from, which `japi` already has — it is the `-cp` the caller gave, or the CLI's own jar when they gave none. A parameter's type comes back from `Types.erasure()` as a binary name, which is the same string the class file holds. Lookup is equality on that.

There is nothing left to disambiguate. Two members with the same name and the same erased parameters *are* the same member, so no uniqueness rule is needed in either direction. The candidate sets, the import table, the `java.lang` probe, the type-variable bound handling, and the simple-name normalisation are all deleted rather than fixed — `AnnotatedTypeTree` with them, since a resolved type is the type whatever annotations were written on it.

A declaration whose types did not resolve — a dependency of the library that is not on the path it was read from — is not registered at all. That is the one case a guess would have filled in, and the answer to it is silence rather than a near-miss.

Confirmed on the shipped jar: the two `use` overloads each carry their own sentence, and raoh's `Issue.of` overloads still come back documented with their declared parameter names.

`mvn clean verify` is green: 109 + 2955 + 166 + 7 + 1.

The PR body's stale `@param` sentence is corrected.

PR #377, opened 2026-08-06
The inliner holds one kind of state, and the facts it used to rebuild are asked once

kawasima:
`HelperInliner` was 2015 lines and the most-changed file in the compiler — 73 of the last 400 commits.
Reading those commit subjects, most of them are the same defect coming back rather than new work.
The cause is that one object held three kinds of state with three different lifetimes: facts about the
module, the frame one body is expanded in, and the substitution one call writes. What lived on it
followed from that rather than from what it does.

Each of the five commits takes one of those apart. Nothing about the language changes.

## What moves out

`HelperNames` answers how a definition is named outside the module that declares it, and carries the
marks a body takes with it. `HelperInvariants` answers what a declaration's invariant becomes, in both
representations that read one, and settles once for the two rather than once each. Neither reads an
expansion table.

## What changes hands

A value reaching itself was refused wherever an expansion table was built, which is eleven places in a
compile — so the refusal came from whichever ran first, and a question that does not expect a refusal
from a table it is merely building passed it on as a failure of its own. `Bodies.RecursiveHelpers`
had no `try` around it at all, so the refusal left `compute` as a thrown exception rather than as an
absent answer. It is now `Shapes.NoValueReachesItself`, asked by name by the two questions that expand
a body. It underlines the name the value was declared under rather than the `let` in front of it.

## What is asked once

The table a body is expanded against, and the call graph over it, are facts about the module's
declarations. Both were rebuilt by every expansion — walking each of the standard library's hundred
and thirty-odd bodies again each time. They are `HelperTable` and `HelperGraph` now, held by
`Bodies.Expanding` and keyed by the policy as well as the module, because the two policies are two
tables and one answer shared by both would put the fold that `List.map` is into the tree the discharge
rules read.

Type checking is shorter for it. On the bench corpora, median of twenty walks: crm 26.3 → 22.7 ms,
issuetracker 3.8 → 3.2 ms.

## What was two and is one

Three pairs held one fact between them: a substituted parameter (a name, and what the name means — with
a run-time check standing between the two maps to catch one being written without the other), a scoped
lambda (a body, and where the author wrote it), and what one call writes for its arguments (counted
twice, once as a list and once as the keys of the map beside it). The renaming a copy carries travels
as one value too, so the rule about which position a rebuilt node takes is written once rather than at
every node kind.

The four things true only while a body is being expanded are one `Writing`, made whole when one starts
and dropped when it ends — refused or not. Expanding can no longer be asked for without saying which
body it is written into, so the run-time check that said so is gone.

## What was found on the way

Two defects turned up that were not what the work set out to do, and both are fixed here.

The number an expansion is written under came from one counter for the whole pass, but a pass expands
many bodies. A body's bindings moved when a body beside it gained an expansion — or lost one to a
refusal — and a query answer that moves is a query answer everything below it is recomputed for.
`BindingOwner.Expansion` says of its own ordinal that expanding something elsewhere moves nothing here;
it did. Counted per body now. Not per writing either: two writings can share one body, and a count that
restarted with each was already handing the second the first one's bindings, from a hardcoded zero.

## Tests

Five new test classes, thirteen tests, each stating a rule rather than pinning a symptom.

- a value reaching itself is one refusal, at the name, and a second mistake beside it changes nothing about it
- a module is told about its helpers in the order it wrote them — the table's order decides which report the author reads, and nothing made that true
- a parameter hides the declaration it is named like, and hides nothing else
- a body expanded after a refusal is what it would have been, and two writings into one body do not write the same binding

Three of those exist because the way the rule breaks is not a red test. Hiding nothing does not fail,
it fails to finish. A counter shared across bodies does not fail, it moves an answer.

Every step was checked with a mutant that has to make a named test fail. Two of the mutants the plan
predicted did not: substituting a parameter's spelling while keeping its denotation survives the whole
suite, because spelling decides nothing (which is what ADR-0067 was for), and a scoped lambda outliving
its writing is unobservable once bindings are numbered per body.

## Verification

`mvn clean verify` green at every commit: compiler 2881, LSP 166, runtime 109, syntax 109.
`souther-examples` builds green against this branch — through a split local repository, so the shared
`~/.m2` was not written to.

Follow-up, not in this PR: the table is still keyed by a string that `HelperNames.keyIn` builds, and
its own javadoc says the spelling is a key only after `qualifyImports` has run. `TotalityChecker:57`
tells an injected prelude helper from a module-own one by looking for a dot in that string, so a
module-own helper with a dot in its name would skip the totality check. I will file that separately.

### DIFF
--- a/souther-compiler/src/main/java/souther/compiler/check/HelperGraph.java
+++ b/souther-compiler/src/main/java/souther/compiler/check/HelperGraph.java
@@ -1,0 +1,95 @@
+package souther.compiler.check;
+
+import souther.compiler.ast.Ast;
+
+import java.util.ArrayDeque;
+import java.util.Collection;
+import java.util.Deque;
+import java.util.HashSet;
+import java.util.LinkedHashMap;
+import java.util.LinkedHashSet;
+import java.util.Map;
+import java.util.Set;
+
+/**
+ * Which declaration calls which, over one {@link HelperTable}, and which of them recurse.
+ *
+ * <p>A function of the table it was built from and of nothing else. Two bodies of one module are
+ * expanded against one table and so read one graph — before this each expansion built its own, which
+ * meant walking every one of the standard library's bodies again for each, and eleven answers that
+ * had to agree.
+ *
+ * <p>A helper recurses iff it can reach itself through helper calls; every member of a mutual cycle
+ * is reached from itself, so all are marked. Both a module's own helpers and the shipped prelude ones
+ * are walked: {@code List.foldFrom} is a recursive prelude helper and has to be left standing —
+ * lowered to a method, not inlined — exactly as a module-own recursive helper is, or its self-call is
+ * expanded forever.
+ *
+ * <p>Built over the table as it stands. A table narrowed for an expansion ({@link HelperTable#hiding})
+ * narrows what a call reaches and changes nothing here: a graph taken over the narrowed table would
+ * find the very helper being expanded non-recursive.
+ */
+public record HelperGraph(Map<String, Set<String>> callsOf, Set<String> recursive) {
+
+    /** The graph of {@code table}: what each declaration in it calls, and which of them recurse. */
+    public static HelperGraph of(HelperTable table) {
+        Map<String, Set<String>> callsOf = new LinkedHashMap<>();
+        for (Map.Entry<String, Ast.FnDef> e : table.reachable().entrySet()) {
+            Set<String> called = new LinkedHashSet<>();
+            HelperInliner.helperCallsIn(e.getValue().writtenBody(), table.reachable(), called);
+            callsOf.put(e.getKey(), called);
+        }
+        Set<String> recursive = new LinkedHashSet<>();
+        for (String name : table.reachable().keySet()) {
+            if (reaches(callsOf, name, name, new HashSet<>())) {
+                recursive.add(name);
+            }
+        }
+        return new HelperGraph(Map.copyOf(callsOf), Set.copyOf(recursive));
+    }
+
+    /** Whether {@code name} is on a call cycle, so a call of it is left standing rather than
+     * expanded (spec 13.1). */
+    public boolean recurses(String name) {
+        return recursive.contains(name);
+    }
+
+    /** What {@code name}'s body calls directly, or an empty set where it calls nothing this table
+     * reaches. */
+    public Set<String> calls(String name) {
+        return callsOf.getOrDefault(name, Set.of());
+    }
+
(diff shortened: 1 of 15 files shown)


kawasima:
Both confirmed and fixed in f5e09e61.

## P2 — the blocking one

Reproduced before changing anything: asking `Shapes.InvariantCapabilities` for a module whose invariant reads a cyclic value throws `StackOverflowError`, while `NoValueReachesItself` answers correctly for the same module. Nobody was asking it.

I did not add the dependency to that one query. Counting the sites first: eight places construct an expansion, and five reach the check through `Derived`. The three that read the module straight from `Names.Resolved` — `Derived`, `InvariantsForDischarge`, `InvariantCapabilities` — each stated the condition for themselves, and one of the three did not. Adding it to the third leaves the fourth to forget it, and this rule had already been moved once for exactly that reason: it used to be enforced wherever an expansion table was built, which was eleven places.

So the condition is no longer something to state. `NoValueReachesItself` is now `Shapes.Expandable`, and it answers the module rather than a boolean: a question that expands a body asks for a module to expand and gets one or gets nothing. There is no answer that hands a module over without having checked it, and nothing beside it left to remember.

The test is a row per question that expands — `Derived`, `InvariantsForDischarge`, `InvariantCapabilities`, `Prepared`, `LoweredBody` — asked by name rather than through a compile, because two of them are not on the path a compile takes to a report. That is why compiling the module went on saying the right thing while this was wrong, and it is why your point about the existing value-cycle tests not covering this path was right. Adding a question that expands means adding a row.

Verified by putting the reported bug back (`InvariantCapabilities` reading `Names.Resolved` again): two of the three tests fail with `StackOverflowError`.

Your reading of the cause was exact — `HelperGraph` collects `Ast.Apply` edges while `valueOf` recurses through bare value references, so the graph never sees the cycle. The up-front check is what stands in for that, which is why forgetting to ask it is unbounded rather than merely wrong.

## P3

Fixed. `Map.copyOf` left every set the mutable one that was built, and both `callsOf()` and `calls(name)` hand one out. Each set is fixed now as well as the map.

## Not in this PR

Worth naming, since it is the reason the precondition has to exist at all: expanding a value has no cycle detection of its own, so it is only safe because something checked in advance. Detecting re-entry inside the expansion would make it safe wherever it is called from — but it would report at the use rather than at the declaration, and many times over, so the report would want to stay where it is. I would rather file that than fold it in here.

`mvn clean verify` green: compiler 2884, LSP 166, runtime 109, syntax 109. `souther-examples` rebuilt green against this branch.

PR #381, opened 2026-08-06
An optional field reads None from a written null, not only from an absent key

### LINKED ISSUE #362: An optional field decodes None from an absent key but rejects an explicit null, which is not what the specification says
## What the specification says

`souther doc decoder-derivation`:

> `T?` (Option; §optional) is `None` if the key is absent **or the value is `null`**, and `Some`-wrapped after reading `T` if a value is present.

## What happens

```souther
module opt

data Note = { body: String, tag: String? }

behavior echo : (n: Note) -> Note

let echo (n) = n
```

Absent key — decodes, as documented:

```
$ souther run opt.sou --input '{"body":"b"}' --lang en
{"body":"b"}
```

Explicit `null` — rejected:

```
$ souther run opt.sou --input '{"body":"b","tag":null}' --lang en
input #1 could not be decoded — /tag: is required
```

## Two things to settle

**Which is right.** A JSON boundary receives `null` for an absent optional routinely — it is what most encoders emit for one. If the specification is the intent, the derived decoder needs to accept `null`; if the compiler is the intent, `decoder-derivation` needs correcting.

**The message is wrong either way.** `tag` is declared `String?`. Saying `is required` about an optional field sends the reader looking for a missing `?` in their own source before they think to doubt the tool. Whatever the decision on `null`, this should say what it actually wanted — a string or an absent key.

kawasima:
Fixes #362. The specification is what holds: `T?` is `None` for an absent key **and** for a written `null`.

## What was wrong

`CodecGen.emitObjectDecode` emitted Raoh's `optionalField` for a `T?` field. That combinator answers *is the key there* — absent gives `Optional.empty()`, present hands the value to the element decoder:

```java
// raoh-json JsonDecoders.java:283-294
var node = in.get(name);
if (node == null || node.isMissingNode()) return Result.ok(Optional.empty());
return dec.decode(node, fieldPath).map(Optional::of);   // a NullNode arrives here
```

A JSON `null` is present, so it reached `string()`, which has no `None` to answer with and failed `required`. The `is required` text the report calls out is Raoh's, from the element decoder that should never have been handed the value; only the `/tag: ` prefix is ours.

Raoh separates the three readings deliberately — `optionalField` → `Optional` (about the key), `optionalNullableField` → `Presence` (keeps absent apart from present-null), `nullableField` → collapses both to a null. The third is what the specification describes, and `Option` has no third state to lose by it. The one that was picked is the one whose name matches the concept's name.

## The change

Emit `nullableField` and bridge with a new `Option.ofNullable` instead of `Option.ofOptional`. All three sources are affected the same way and all three are changed: a JSON `null`, a map entry holding a null, and a NULL column. The now-dead `ofOptional` / `MTD_ofOptional` / `CD_JavaOptional` go, and the descriptors are renamed to the combinator that is emitted.

Nothing says `is required` about an optional field any more. A value of another shape still fails and still says what it wanted (`expected string`), and a *required* field written `null` still gets `is required`.

## Tests

`OptionalFieldReadsAWrittenNullTest` — one row per generated decoder, because the bug was per decoder, plus the end-to-end `souther run` case as reported. Reverting the codegen change fails four of the six; the two that survive are the guards on what did **not** change.

The rule was already written in `specification.adoc:1252`, in two comments in `FixtureReader`, and in the javadoc of `CompileOptionalFieldTest` ("An absent (or null) key decodes to `None`") — and asserted nowhere. `DecoderPathAgreementTest`, which exists to catch exactly this drift, has no `Option` row: it is keyed by type and `T?` is only a field. That is why the new rows are their own class rather than a row there.

`mvn clean test` is green across all modules (2894 in the compiler).

### DIFF
--- a/souther-compiler/src/main/java/souther/compiler/codegen/CodecGen.java
+++ b/souther-compiler/src/main/java/souther/compiler/codegen/CodecGen.java
@@ -79,7 +79,7 @@ final class CodecGen {
 
     private static MethodTypeDesc srcFieldMtd(Src s) { return s == Src.JOOQ ? MTD_fieldJooq : MTD_field; }
 
-    private static MethodTypeDesc srcOptFieldMtd(Src s) { return s == Src.JOOQ ? MTD_optFieldJooq : MTD_optionalField; }
+    private static MethodTypeDesc srcNullableFieldMtd(Src s) { return s == Src.JOOQ ? MTD_nullableFieldJooq : MTD_nullableField; }
 
     /** Leaf value decoders: JSON reads a JsonNode, the map/jOOQ column value is an Object. */
     private static ClassDesc srcLeafOwner(Src s) { return s == Src.JSON ? CD_JsonDecoders : CD_ObjectDecoders; }
@@ -1041,7 +1041,7 @@ final class CodecGen {
             code.loadConstant(bind.key());
             if (bind.ref() instanceof Ast.OptionDecRef opt) {
                 emitDecoderObject(code, opt.element(), src);
-                code.invokestatic(srcFieldOwner(src), "optionalField", srcOptFieldMtd(src));
+                code.invokestatic(srcFieldOwner(src), "nullableField", srcNullableFieldMtd(src));
             } else {
                 emitDecoderObject(code, bind.ref(), src);
                 code.invokestatic(srcFieldOwner(src), "field", srcFieldMtd(src));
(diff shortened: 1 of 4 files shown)


kawasima:
**1. `ofNullable` without `@Nullable` — fixed** (0c97a107).

The package doc of `souther.runtime` states the rule this broke: "The few places that do admit a null are marked one by one." `ofNullable` is one of them and said nothing, so a marked package declared a parameter that refuses the value the method exists to accept. It is now `ofNullable(@Nullable T value)`, matching how the rest of the package marks such a position (`PersistentHashMap.get(@Nullable Object)`, `InvariantFailure`'s `@Nullable String clause`).

The erased descriptor is unchanged — the annotation lands as `METHOD_FORMAL_PARAMETER` in `RuntimeVisibleTypeAnnotations`, and `(Ljava/lang/Object;)Lsouther/runtime/Option;` is what the generated `invokestatic` still resolves against. Full suite green.

**2. `ofOptional` as a compatibility shim — declined, by decision.**

The condition you named for declining is the one that holds: compiler and runtime move in lockstep, and a runtime update requires recompiling the Souther sources. That is the project's compatibility policy, not an oversight, so a class emitted by an earlier compiler is not expected to run against a later runtime.

Your reading of the mechanism is right — `provided` scope does mean the two are deployed separately, and a retained `ofOptional` would be reachable by old bytecode. It is being declined on the policy, not on the mechanics. Keeping the shim would leave a method whose javadoc has to describe a combination we do not support, and `Option` is small enough that such a method reads as current API rather than as a grave marker.

PR #382, opened 2026-08-06
A binding is named and seeded by one act, so there is no place the check names and knows nothing about

### LINKED ISSUE #363: An invariant is not assumed for a term reached through a match arm's binding, so E2011 fires on an addition the same types discharge everywhere else
## What happens

Adding two values of the same non-negative newtype discharges when the terms are parameters, and warns when one of them is reached through a `match` arm's `as` binding.

```souther
module e2011

data Q = Int
    invariant value >= 0

data Box = { q: Q }
data M = A | B
data A = { q: Q }
data B = { q: Q }

behavior direct : (a: Q, b: Q) -> Q

let direct (a, b) = a + b

behavior viaMatch : (x: Box, m: M) -> Q

let viaMatch (x, m) = match m with
    | A as a -> x.q + a.q
    | B as b -> x.q + b.q
```

```
$ souther compile e2011.sou -d out --lang en
-- INVARIANT (WARNING)  E2011 ---------------e2011.sou:18:21

18|     | A as a -> x.q + a.q
                        ^

Constructing `Q` here may violate its invariant: the guards do not establish it.
```

`direct` — no warning. `viaMatch` — one warning per arm. `a >= 0 ∧ b >= 0 ⟹ a + b >= 0` holds in both.

## Why it reads as a seeding gap

`souther doc invariant-discharge-terms` enumerates what counts as a location — a parameter, a field chain, a newtype's wrapped value — and says explicitly that a `let` binding given a location *is* that location. A `match` arm's `as` binding is not mentioned, and it is the case that does not work. Destructuring instead of `as`, re-binding through a `let` inside the arm, and extracting a helper all still warn.

## Why the hint cannot be followed

> Guard it with `guard`, or reify the relation into `Q`'s (or an input's) invariant so it is assumed here.

`Q`'s invariant is already `value >= 0`. There is no relation between the two operands to reify — the implication is unconditional. Guarding would require inventing a departure case for a failure that cannot occur, which changes the behavior's output type to appease a warning. Both suggested exits are unavailable, so the only remaining option is to live with the warning, and warnings that cannot be cleared cost the ones that should be acted on.

If the limit is deliberate, `invariant-discharge-terms` should name `match` bindings among what is not seeded, so a reader stops looking for a mistake in their own model. Reported alongside #364.

kawasima:
Fixes #363. Adding two non-negative values discharges when they are parameters and warned when one is reached through a `match` arm. The arm was not the difference — the seeding was.

## What was wrong

`Denotations.of` answered a binding nobody entered with the location it would be:

```java
Denotes of(BindingId binding) {
    Bound given = bound.get(binding);
    return given != null ? given.denotes() : new Denotes.At(Location.of(binding));
}
```

`Terms.readable` grants `Denotes.At` unconditionally, so **every** binding was nameable and a clause reading one was owed. Facts, though, came only from `seedAt`, called from three places: a behavior's parameters, an attempt's binder, and a combinator's element parameter. Naming was open-world and seeding was a list of call sites, and the two sets differed.

A `match` arm's binding was in neither, so `a` in `| A as a -> x.q + a.q` was a location the check named and knew nothing about: the clause `value >= 0` was owed over `x.q + a.q`, `x.q >= 0` was seeded through the parameter, and `a.q` had nothing. Hence a warning with no guard that clears it. The same gap covered the destructured and re-bound spellings, since both reach the arm's binding, and a fold's accumulator, which `walkCall` also left unseeded.

The comment at the site justified not seeding the *scrutinee* ("A sum has no fields of its own"). The binding beside it was never considered.

## The change

The record is closed. An unentered binding denotes `Nothing`, so a clause naming one is left to the run-time check — which is what `InvariantChecker` already promises ("An invariant naming something it cannot name is left opaque ... so every flagged construction has a guard that discharges it"). That promise now holds by construction instead of by everyone remembering.

The walk reaches a binding three ways, and each says which of the two halves it owes:

| | name | type's invariant |
| --- | --- | --- |
| fresh location — a parameter, what an arm binds, what a closure is handed | new `Denotes.At` | seeded |
| alias — `LetIn` | inherited from the initializer | none |
| alias that also held — `IfConstructed`'s success branch | inherited from the construction | seeded |

`enter` is the first row, and it returns `Entered(Known, Denotations)`. Returning only the facts and leaving the caller to record the name would reintroduce the same defect in another spelling. The pair is local to the transition — `Terms.bodyKey` and the rest still take whichever environment they ask about.

Closure parameters are entered from the block's `Type.FnOf`, not just the element, so a fold's accumulator carries the type it was seeded with. `capabilityOf` reads a clause over the declaration's own fields, which stand for a value rather than hold one; those are entered as locations with nothing seeded, since what a clause owes is the question being asked.

## Tests

`CompileBindingIsEnteredTest`. Closing the record turns a missed site from a false positive into a false negative, so silence cannot be the assertion — a test that a valid program is silent passes whether the binding was seeded or dropped entirely. Each site is held instead by a construction that must violate:

| state | diagnostic |
| --- | --- |
| entered and seeded | E2010 |
| entered, not seeded | E2011 |
| not entered | none |

Measured, not argued. Dropping `at.location(...)` from `enter` fails five of the seven; dropping `seedAt` fails six. Both halves are pinned by the assertion.

The remaining two are the report's own cases: the same addition reached as a parameter and through an arm, and a fold that stays in the newtype.

## Also

`specification.adoc` enumerated what a term can name as "a parameter, a field chain, a newtype's wrapped value", which is what the report read and reasonably took for a deliberate limit. It now names what an arm binds and what an operation hands its closure, with an example that compiles clean.

`mvn clean install` is green across all modules (3271 tests). The reported spellings — `as`, destructured, re-bound through a `let`, the fold accumulator, and the guard the report showed could not be written — all compile with no diagnostic.

### DIFF
--- a/souther-compiler/src/main/java/souther/compiler/check/Denotations.java
+++ b/souther-compiler/src/main/java/souther/compiler/check/Denotations.java
@@ -3,6 +3,7 @@ package souther.compiler.check;
 import souther.compiler.core.Core;
 import souther.compiler.types.BindingId;
 
+import java.util.Collection;
 import java.util.HashMap;
 import java.util.Map;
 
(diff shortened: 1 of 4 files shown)


kawasima:
Both fixed in fdebe7f0.

## P1 — confirmed, and it was mine

The alias inherited whatever the construction denoted, and an attempt is written precisely where the check could not say enough about that construction. So the one case attempts exist for is the one where the binding ended up naming nothing, dropping what reaching `then` had established.

The success binding is now entered as a location through `enter`, which is what it was before this branch. Held by the same refutation as the other sites, as `anAttemptsSuccessBindingIsEnteredAndSeeded`:

```souther
let use (x) = if Q(x * x) as q then Q(0 - q.value - 1) else Nope
```

Measured both ways: with the alias the test reports `no diagnostic`, with `enter` it reports E2010.

Keeping the alias identity *and* the success postcondition would need a `Denotes` state that carries both, which does not exist and is not worth inventing here — nothing in the suite depends on a guard written about the construction reaching the binding.

## P2 — correct

The `null → elem` fallback had no grounding. `Combinators.handedTo` only answers for a `PreservedCall`, which exists only where the signature accepted the arguments and typed the block, so an untyped closure parameter is an internal inconsistency and not a case to interpret. Substituting the element's type there could seed a different newtype's invariant, which is the wrong direction for a fail-open check.

It now raises, which `analyze` catches — the whole body falls back to the run-time check, which is the documented behaviour for an analysis inconsistency and matches what `Combinators` already does with a signature it cannot read.

## Verification

`mvn clean install` green across all modules; the discharge suite is 2989 with the added witness.

One correction to my earlier report on this branch: the `if constructed q = ...` spelling I used to check the attempt path is not the language's syntax, and my grep for diagnostics did not match `SYNTAX ERROR`, so those two checks passed on files that never compiled. Re-run with `if X(v) as n then` and unfiltered output — the reported spellings and the ordinary attempt case all compile with no diagnostic and emit classes.

PR #383, opened 2026-08-06
Reaching a behavior from outside the module is a second question, asked where the first one was

### LINKED ISSUE #366: Running a behavior the module does not expose fails with a raw reflection exception naming a compiler internal
## What happens

```souther
module exposed exposing ( Note )

data Note = { body: String }

behavior hidden : (n: Note) -> Note

let hidden (n) = n
```

```
$ souther run exposed.sou --behavior hidden --input '{"body":"b"}' --lang en
`hidden` failed: java.lang.IllegalAccessException: class souther.compiler.Runner cannot access a member of class exposed.Hidden$Impl with modifiers "public"
```

Adding `hidden` to `exposing` makes it run.

## The problem

The rule — `run` drives an exposed behavior — is a fine rule, but it is never stated. What the author gets instead is a JVM reflection failure that names `souther.compiler.Runner`, a compiler internal they have no business knowing about, and that contradicts itself by reporting it cannot reach a member "with modifiers public". The natural reading is that codegen produced something broken, so the next step is investigating one's own model rather than adding one name to `exposing`.

This should say that the behavior is not exposed and that `run` needs it to be.

kawasima:
Closes #366.

## What was wrong

`run` decided what it could drive from the module's own workings — the behavior has a `let` and needs nothing injected — and never asked whether it could reach it. It reaches a compiled behavior by reflection from `souther.compiler` into the generated package, which the JVM allows only for a class marked public, and codegen marks public exactly the names on the `exposing` list (`CodegenContext.pub`). A behavior the module keeps to itself passed the first test and failed the second, as a JVM access error:

```
$ souther run exposed.sou --behavior hidden --input '{"body":"b"}' --lang en
`hidden` failed: java.lang.IllegalAccessException: class souther.compiler.Runner cannot access a member of class exposed.Hidden$Impl with modifiers "public"
```

Naming the compiler's own class and saying a public member is out of reach reads as a codegen defect, so the author goes through their model for what is one name missing from `exposing`.

The same root reached a second site. With the input's data hidden too, the decoder was asked for first and failed the same way, reported as `cannot obtain a decoder for `Note`: java.lang.IllegalAccessException: ...`.

## What changed

Runnable is unchanged and still means what it meant: a `let`, no `depends on`. What `run` can drive is that *and* exposed, and the two are separate questions — one about the module's workings, one at its edge, which is where `run` stands. Selection, the sole-behavior pick and the list to choose among all read the second.

A behavior the module does not expose is refused as that, ahead of the reasons that read its workings — those describe the inside of the module, and following one leads back to the same refusal:

```
`hidden` is not exposed, and `run` drives a behavior the module exposes — add it to `exposing`. Runnable: none.
```

`run.behavior.none` was reworded with it. It said `A runnable behavior has a `let` and depends on nothing.` — a condition the hidden behavior meets, so on the very module this fixes it contradicted itself.

The decoder site needs no separate fix and cannot be reached: the behavior is refused where it is chosen, and an exposed behavior may not rest on a type the module hides, which `SpecChecker.checkExposedSurface` settles at compile time.

Separately, `invoke` reported the behavior throwing and the runner being unable to start it under one message. They are told apart now. After the exposed check, the second is codegen and the runner disagreeing rather than anything the module did, and calling it a failure of the behavior sends the author through a body that was never entered.

A file with no `exposing` list keeps nothing to itself, so a header-less `.sou` drives everything in it as before.

## Verification

`mvn -o clean test` green across all modules. The issue's exact command was run against a built CLI in both languages, along with the exposed and header-less cases.

Seven tests in `RunnerTest`: refusing a named hidden behavior, refusing it as the sole runnable one, picking the exposed one over a hidden runnable one, offering only exposed names to choose among, refusing a hidden pipeline, refusing before the input type is asked for a decoder, and driving a behavior exposed alongside its type. Six of them failed on the old code with the reflection text; the seventh is a regression guard that passed both before and after.

## Adjacent, not fixed here

When nothing can be driven, the `Runnable: {1}` argument is the English literal `none`, so the Japanese message reads `実行できる behavior: none`. It predates this change; fixing it means holding the list rather than the rendered string until the locale is known, which is a change to how `RunException` carries its arguments.


kawasima:
All three are real. Fixed in c13022ad.

**The list under the wrong label.** `available` is built from `drivable` and every refusal appended it as `Runnable:`, so on the module in the issue it said `Runnable: none` about a behavior with a `let` and nothing injected. It reads as `Available to run:` now (`run` で動かせる behavior: in Japanese), across `notexposed`, `unknown`, `noimpl`, `depends`, `pipeline.noimpl`, `pipeline.depends` and `runnable`. `run.behavior.several` had the same fault in its own words — `several runnable behaviors` for a list that is not all of them — and now says `several behaviors can be run`.

A test pins it: an unknown name in a module whose only runnable behavior is hidden, asserting the message does not say `Runnable`. It fails on the previous commit with `no behavior named \`typo\`. Runnable: none.`

**`run.behavior.none` losing pipelines.** Right, and it was the same fault turned around — I re-listed the conditions and reached only `SpecBehavior`. It now names the two things instead of restating one of them: `run` runs a behavior that is runnable — implemented in Souther and needing nothing injected, a `>->` pipeline when every stage is — and that the module exposes.

I kept the conditions rather than dropping to `No exposed runnable behavior in this module can be driven`, because this message is the only place an author learns what runnable means; naming the concept without saying it leaves them nowhere to go.

**The counting rule in the docs.** Both said `exactly one runnable behavior` and I had appended the exposure rule beside it instead of into it, which is exactly the module in your example: two runnable behaviors, one exposed, documented as stopping and actually running `shown`. `souther doc run` now states it once —

> `run` runs a behavior that is both runnable and exposed. … With `--behavior` the named one runs. Without it, a module with exactly one behavior that is both runs that one; if there are several, the run stops and lists them, and you pick one. A runnable behavior the module does not expose is not among them, and naming it says so.

— and the README paragraph is rebuilt the same way. I did not introduce `drivable` as user-facing vocabulary; it stays the name in the code, and the docs say "both runnable and exposed", which needs no glossary.

`mvn -o clean test` green (2997 + 166 + 1). Every message above was read off a built CLI in both languages, along with the exposed and header-less cases.

PR #385, opened 2026-08-06
A decode failure says which rule broke, in the reader's language

### LINKED ISSUE #365: A one-sided invariant renders its violation with an unfilled {max} placeholder
## What happens

```souther
module inv

data Amount = Int
    invariant value >= 0

behavior echo : (a: Amount) -> Amount

let echo (a) = a
```

```
$ souther run inv.sou --input '-1' --lang en
input #1 could not be decoded — (root): must be between 0 and {max}
```

The invariant is one-sided. The renderer picked a two-sided range template and left `{max}` unsubstituted, so the message names an upper bound that does not exist and shows the reader a placeholder.

A reader at the boundary — which is where this message lands, in front of whoever sent the bad value — cannot act on it. It should say the bound the invariant actually states.

### LINKED ISSUE #371: A named invariant clause is not reported when it fails, though two sections promise the name comes back
## What the specification promises

`souther doc invariant-clause-name`:

> An attempted construction departs by the named clause that did not hold, and a decode failure reports the name.

`souther doc decoder-error` repeats it: the failure carries, *"where the clause was named, that name as `clause`"*.

## What comes back

```souther
data BillingPeriod =
    { startsOn: Date
    , endsOn: Date
    }
    invariant ordered = startsOn <= endsOn

data PastDue = { failedAttempts: Int }
    invariant atLeastOneFailure = failedAttempts >= 1
```

```
$ souther run billing.sou --behavior issueInvoice --lang en --input '<reversed period>'
input #2 could not be decoded — (root): validation failed: invariant_violation

$ souther run billing.sou --behavior issueInvoice --lang en --input '<PastDue with 0 attempts>'
input #1 could not be decoded — /state: validation failed: invariant_violation
```

Neither `ordered` nor `atLeastOneFailure` appears. Nor does the `type` or `module` metadata the same sections describe. Every clause in the model was named on the strength of that promise, and naming them bought nothing observable.

Related: the path for the record invariant is `/state`, not `/state/failedAttempts`. The documentation stresses that which field broke it should be statable through `path`.

## What needs settling

Whether the metadata is not attached at construction, or attached and not rendered by `souther run`. From the CLI these are indistinguishable, and `souther run` is documented as being itself the Java boundary — so it is the one surface where a reader can check whether the promise holds, and from there it does not.

kawasima:
A decode failure told the reader less than the decoder knew.

```
$ souther run inv.sou --input '-1' --lang en
input #1 could not be decoded — (root): must be between 0 and {max}
```

`invariant value >= 0` has no upper bound, and `{max}` is a placeholder. The issue the decoder
built said `must be non-negative`; resolving it through Raoh's catalog replaced that with a
template for a two-sided bound and left the end the constraint never had as written.

Every one-sided `Int` and `Decimal` bound read this way, in both locales, and so did an
invariant written with `&&` — its clauses are separate constraints, and the one that fires
reports alone. A rule no constraint states fared worse: `invariant_violation` has no entry at
all, so resolving yielded `validation failed: invariant_violation` and discarded
`invariant violated on demo.T: nonzero`, which is #371 — the module, type and clause were
attached at construction all along and the runner threw them away.

## What is here

Two commits, because the answer moved.

**bc9dc7bf** made the runner non-destructive: a catalog entry applies when its placeholders are
all ones the issue's metadata fills, and the decoder's own message stands when it does not. That
stopped both symptoms, at the cost of a one-sided bound reading in English under `--lang ja` —
the catalog had no way to state such a bound.

**d8b67313** follows Raoh 0.7.2, which fixes that upstream (kawasima/raoh#123,
kawasima/raoh#124). Every constraint now carries a message key beside its code; the resolver
tries the key before the code, uses a template only when the metadata fills all of it, and
returns the message stored at decode time when neither applies. That is the same guard, in the
place that can do it properly — a key names which constraint failed, whereas a code cannot:
`out_of_range` covers a lower bound, an upper bound and a range, and `value > 0` and
`value >= 0` carry the same `min = 0`.

So the guard goes. Keeping it would have been the rule written twice, and worse: looking up
`raoh.out_of_range` alone skipped the entries the release added, so `value >= 0` still read
`must be non-negative` under `--lang ja` where the catalog now says 0以上で入力してください.

## What a reader sees now

| invariant | `--lang en` | `--lang ja` |
|---|---|---|
| `value >= 0` | must be non-negative | 0以上で入力してください |
| `value <= 100` | must be at most 100 | 100以下で入力してください |
| `value > 0.0m` | must be positive | 正の値で入力してください |
| `value >= 0.0m` | must be non-negative | 0以上で入力してください |
| `List.length(value) >= 1` | must not be empty | 空にはできません |
| `nonzero = value /= 0` | invariant violated on demo.T: nonzero | (unchanged) |

The strict bound and the inclusive one read differently in Japanese too, which the metadata
alone cannot tell apart. An empty list is rejected as empty rather than as one element short of
a bound.

A rule no constraint states stays in Souther's own English: the catalog carries no
`invariant_violation` entry, and the message is Souther's rather than the decoder's. What #371
asked for — the clause name coming back — is there.

## Tests

`DecodeMessageLocalizationTest`, 13 cases.

Each bound is compared whole, in both English and Japanese: a lower bound alone, a lower bound
away from zero, an upper bound alone, the inclusive Decimal bound against the strict one, and an
empty list. A message that names one bound too many contains everything a substring check would
look for, so `contains` would not have held these — `must be between 0 and 100` mentions 100 and
carries no placeholder.

The `&&` case asserts both sides separately: `-1` reads as the lower bound alone and `101` as the
upper bound alone, in both locales. That is the rule the case exists for — each side of an `&&`
is its own constraint and reports alone.

The strict and inclusive Decimal bounds each assert the rule they are rather than merely
differing from each other, which a swapped pair would satisfy. Nothing else pins the `messageKey`
correspondence for Decimal: `CompileInvariantConstraintTest` covers codes and `Int` metadata.

`invariant_violation` keeps its module, type and clause; an unnamed clause keeps its type. Three
cases pin what the runner relies on the catalog for — entry selection by constraint, an entry the
metadata cannot fill leaving the message alone, and a custom message never being touched.

Extracting the reason runs a placeholder sweep on the way through, so every reason a case reads
is swept without the case asking. The sweep's grammar is the decoder's own — a letter or
underscore then letters, digits, underscores, dots or hyphens — so `{_min}`, `{range.min}` and
`{min-bound}` are caught, and the JSON prose the decoder deliberately does not treat as a
placeholder is not.

The two failures are kept apart on purpose. #365 is "the key exists but the metadata cannot fill
it"; #371 is "the key does not exist and a specific message is lost". One test for both would let
a missing-key fix pass while #365 stayed broken.

`mvn clean test`: green across all modules, souther-compiler 2996 tests, 0 failures, 0 errors.

## Also

`specification.adoc` <<decoder-error>> said a `MessageResolver` "keyed on the standard codes"
resolves and localises these. It cannot — that is #123. It now says the resolver is keyed on the
issue, and why the code alone does not decide.

`raoh.version` 0.7.1 → 0.7.2.

Closes #365, closes #371.


kawasima:
Review addressed in ea8ad49b — all three were right.

**1. The `&&` case did not hold what it was for.** `must be between 0 and 100` mentions 100 and carries no placeholder, so it satisfied both old assertions. Both sides are now compared whole, in both locales: `-1` reads as the lower bound alone and `101` as the upper bound alone.

**2. The Decimal case passed a swapped pair.** Comparing for difference was the wrong question — the contract is that each reads as the rule it is. Each now asserts its own wording, English and Japanese. As noted, `CompileInvariantConstraintTest` pins codes and `Int` metadata and nothing pinned the `messageKey` correspondence for Decimal.

**3. The sweep read a narrower name than the decoder writes.** Confirmed against `net.unit8.raoh.Placeholders`: `[A-Za-z_][A-Za-z0-9_.\-]*`. Checked all four forms against both patterns —

| fragment | raoh | new | old |
|---|---|---|---|
| `{max}` | true | true | true |
| `{_min}` | true | true | false |
| `{range.min}` | true | true | false |
| `{min-bound}` | true | true | false |
| `an object like {"id": 1}` | false | false | false |

The last row is the prose Raoh deliberately does not treat as a placeholder, and the copied grammar keeps that. `Placeholders` is package-private, so copying is the only option today; noted in the field's comment.

Also took the non-blocking points. Reason extraction is a helper that says what is missing when the marker is not there, and it runs the sweep on every reason a case reads, so no case has to remember to ask. The PR body's test description was written for a shape the tests never had — rewritten to what is there (13 cases, whole-reason equality, which cases cover both locales).

`mvn clean test` green across all modules, souther-compiler 2994 tests, 0 failures, 0 errors.

Issue #372, opened 2026-08-06
A diagnostic code is optional, though it already is the public identity of a diagnostic


kawasima:
A diagnostic code is currently optional, despite already serving as the public identity of a compiler diagnostic.

`souther doc <code>` works because the codes are anchors, and the toolchain advertises the affordance — `souther doc cli/start-here` tells a reader to run `souther doc e1918` on any diagnostic they hit. Two readers went looking and found diagnostics with nothing to type: `-- BOUNDARY TYPE ---` (an anonymous union in a parameter position) and `-- TYPE MISMATCH ---` (reading a field off a sum). Those two are the symptom. The defect is that nothing requires a diagnostic to have a code at all.

## What the audit found

- 200 emission sites carry no code, across 164 distinct message keys. 55 codes are in use.
- The title is not an identity and cannot stand in for one. `check.type.mismatch.title` is written by 25 sites, `check.module.title` by 24, `check.boundary.title` by 20. A reader holding a `BOUNDARY TYPE` report is holding a category shared by twenty unrelated rules.
- Eight codes that *are* emitted had no section to resolve to: E1922, E2011, E2013, E2015, E2016, E2017, E2018, E2019.
- Nothing enforced any of this. `MessageCatalogFormatTest` checks keys, arguments, duplicates and the two catalogs against each other, and never looks at a code. No test related a code to the specification.
- It is not only `souther doc`. `JsonRenderer` emits `code` only when it is non-null, and `Analyzer` branches on `"E1922"` to mark an unused import. A diagnostic without a code is unidentifiable to tools and to the LSP as well as to a reader.

## The rule

> A diagnostic code identifies one normative rule of the language, not one message template and not one title category. Each such rule has a canonical anchor in the specification. Diagnostics that report different manifestations of the same rule share a code; diagnostics enforcing different rules use different codes, even when they share a title or a specification section.

Three names, kept apart:

- `titleKey` — the category a reader sees. Shared, not unique, says nothing about identity.
- `DiagnosticCode` — the identity a reader looks up and a tool keys on. One per rule.
- message key — the wording for one way that rule was broken. Several belong to one code.

The discriminating case is `constructs`. `E1002` (constructs something the clause omits) and `E1006` (declares something it never builds) can be written as one equality, and are still two independent obligations the specification states separately — so they are two codes with two rule anchors, not one. Going the other way, `E1908` covers "no fake supplied" and "fake supplied but its fixture will not build", which are one obligation broken two ways, so they are one code.

[...]

Issue #370, opened 2026-08-06
A newtype product is refused with a message saying the newtype cannot be an arithmetic operand, which is not the rule


kawasima:
## What happens

```souther
data Yen = Int
    invariant value >= 0

let square (plan: Plan) : Yen = plan.monthlyPrice * plan.monthlyPrice
```

```
-- TYPE MISMATCH -------------------------------pD.sou:78:23
78|     plan.monthlyPrice * plan.monthlyPrice
                          ^
An arithmetic operand must be Int or Decimal, but got Yen.
```

## Why the message is wrong

`Yen` is a perfectly good arithmetic operand. `monthlyPrice * 30` compiles and is the documented form — `newtype-arithmetic` says `*` and `/` scale a newtype by a bare scalar of its base and stay in the newtype.

The rule actually broken is that a product of two newtypes changes dimension, which the language does not model, so the operands must be unwrapped and the result re-wrapped. The message never says this. A reader who believes it will delete the newtype — the opposite of the fix.

Two reproductions independently reached for a workaround the message misdirected them toward. The caret also sits on the operator rather than on the second operand, which is the one that makes the expression illegal.

## What would help

Say the rule that was broken: two newtypes cannot be multiplied because the product is a different dimension; compute on `.value` and re-wrap. Every other diagnostic in the toolchain names the fix — this one is the outlier.

The message also carries no error code, so there is no `souther doc` entry to reach from it. See #372.

kawasima:
Fixed by #387 (840a6b7a). Every arithmetic refusal now names the rule it broke and a fix that holds for every operator, and the generic operand message is left to the operand that is not a number at all.

PR #387, opened 2026-08-07
Which rule refused the operands is what the message has to say

kawasima:
## What was wrong

`BinaryElaborator` decided arithmetic by trying two boolean predicates and letting everything they did not admit fall through to the check written for a non-numeric operand. That check then wrote the message. Measured on `develop`, seven distinct rules arrived as one sentence, and one arrived as a plain type mismatch:

| expression | rule actually broken | message before |
|---|---|---|
| `Amount * Amount` | a product leaves the dimension | An arithmetic operand must be Int or Decimal, but got Amount |
| `Amount * Quantity` | same, and the other operand was never named | same |
| `Amount / Amount` | a quotient is a value in neither | same |
| `Amount - Quantity` | two unlike newtypes do not combine | same |
| `Amount + n` (`n: Int`) | only a literal is read as the newtype | same |
| `Rate * n` (`Rate = Decimal`, `n: Int`) | a newtype reads its own base | same |
| `Outer + Outer` (`Outer = Inner`) | arithmetic reaches one wrapping | same |
| `n / Amount` | a number over a newtype inverts it | operand of arithmetic does not have the type it needs here … needs to be Int |

Told that a newtype is not an arithmetic operand, a reader deletes the newtype — the opposite of what every one of these rules asks for.

## What changed

`ArithmeticCheck` answers a binary expression in one of three ways:

- `Allowed(resultType)` — the operator combines them, and this is what it gives back
- `Refused(refusal)` — a rule of arithmetic refuses them, and the refusal carries its message, its arguments, the operand it is about and the fix
- `DeferToPlainTypeCheck(left, right)` — arithmetic has nothing of its own to say: two plain numbers of unlike bases, or an operand whose type states nothing. The plain type check answers as it always has, which may end in a found-versus-expected block or in absorbing the operand and letting the expression stand.

The diagnostic is chosen from that answer instead of being written by the next check to fail. `operandBeside` puts its question to the same rules rather than deciding a second time.

The rules are asked in the order they are answerable, so that no rule is reported for a question a nearer one had already settled: an operand that states nothing is set aside, then whether arithmetic can read each operand at all, then what two newtypes make of each other (settled by the operator, the bases not consulted), then that the bases agree, then a newtype paired with a value of that base.

Nothing that compiled before stops compiling, and nothing refused before is now admitted; only the explanations changed.

```
6| let square (a: Amount) : Amount = a * a
                                       ^
6| ...                                a
                                      ^   this is Amount
6| ...                                    a
                                          ^   this is Amount

A product of Amount and Amount is a quantity in neither of them, and the
dimension it would be in is not one this language models.
Hint: Compute on `.value` and build the newtype the product belongs to.
```

A refusal owns its fix as well as its rule, and a rule that every operator asks owns a fix that holds for every operator — a fix written for one of them sends the author from this refusal into the next.

## Tests

`EveryArithmeticRejectionNamesTheRuleItBrokeTest` holds the table above as 13 rows. It fixes both directions of `String * Int`, `Int * Decimal` as the plain mismatch it always was, a two-module case where one name stands for two types, and that every newtype rule carries a fix. `String * Int` is there so that widening the newtype rules cannot quietly absorb the operand rule they were carved out of.

The admission set was also compared against the previous implementation over every pair of Int, Decimal, String, Bool, four newtypes and two literals under all four operators — 484 combinations, identical in what compiles, 120 of them naming a different rule.

Full clean run green: 3052 + 109 + 166 + 1.

## Not in this PR

Error codes for these diagnostics (#372). A code is a question about a diagnostic's identity, independent of which rule it names.

Refs #370 — this PR targets `develop`, so the issue is closed by hand after the merge.


kawasima:
All five were reproduced before touching anything, and all five are real. Fixed in 5be59afa.

Points 1, 3 and 6 turned out to be one fault, not three: the rules were asked in a single line of conditions, so whichever stood first claimed everything that reached it — the same shape this PR set out to remove, one level down. Adding a branch per case would have left the shape in place, so the order is now the one the questions answer in:

1. an operand whose type states nothing (`BottomInfer.answersNoValue`) breaks no rule — set aside first
2. can arithmetic read this operand at all — asked of **each** side, not just the left
3. what two newtypes make of each other — settled by the operator, the bases not consulted
4. the bases must agree
5. a newtype paired with a value of that base

That reordering closed two holes the review had not listed. `String + Amount` and `Bool + Outer` were reported as rules about the newtype when the fault was that String and Bool are not numbers at all.

| point | before | after |
|---|---|---|
| 1. `Amount + Decimal` | Scaling Amount reads a bare Int … | `Amount wraps Int, and this Decimal is not one.` — one refusal, asked by every operator, not a rule of scaling |
| 2. `Amount / Quantity` | … is a bare ratio, not a value of either | `… is a value in neither of them, and the dimension it would be in is not one this language models.` One refusal, wording true of both alike and unlike newtypes |
| 3. `Int * String` | plain mismatch against Int | `An arithmetic operand must be Int or Decimal, but got String.` — the operand named is the one that is not a number, whichever side it stands on |
| 4. `Amount + Quantity` hint | Compute on `.value` and build … | `If both measure the same thing, convert one to the other's newtype and combine those. If they do not, there is nothing to combine.` No longer teaches unwrapping as the way around a nominal distinction |
| 5. two modules, one name | `Amount and Amount are different newtypes` | `one.Amount and two.Amount are different newtypes` — every pair message writes each type against the other |
| 6. `messageKey()` throwing | one `Refusal` was a partial function | not a `Refusal` at all. `PlainTypeCheck` is a third answer of `ArithmeticCheck`: what remains is one type against another, which the plain check already states. Every `Refusal` carries a message |

Also confirmed the concern behind the review's own verification. Enumerating every pair of Int, Decimal, String, Bool, four newtypes and two literals under all four operators — 484 combinations — against the previous commit: **what compiles and what does not is identical**, and 120 combinations name a different rule. All 120 fall into four transitions, each moving toward the operand actually at fault:

| transition | count | example |
|---|---|---|
| `scalarbase` → `operand` | 48 | `String + Amount` |
| `scalarbase` → `base` | 32 | `Amount + Decimal` |
| `type.mismatch` → `operand` | 24 | `Int + String` |
| `nested` → `operand` | 16 | `Bool + Outer` |

The full run also caught two regressions the 484-combination table could not express, because the fixture has no syntax for them: `unreachable` as an operand had lost its E1307, and a fold over `[]` whose element type is not yet fixed was being refused. Both came from asking the readability question of a type that answers no value. `BottomInfer.answersNoValue` already exists to decide exactly that, and says in its own javadoc that a type with the property belongs there rather than in each reader's exclusion list, so that is what the guard asks.

Tests are 13 rows now, adding `Amount + Decimal`, `Amount / Quantity`, `Int * String` beside `String * Int`, `Int * Decimal`, and the two-module case asserting the message arguments are written out. `CompileHelperBodyTypingTest.anErrorBesideAnOpenElementIsReportedAsTheErrorItIs` expected the old key for `Int + String`; what it is testing — that the `+` is blamed and not the parameter — is unchanged, so its expectation moved to the new key.

Full clean run green: 3051 + 109 + 166 + 1.

kawasima:
Both hint cases reproduce, and the diagnosis is right: the reason is operator-independent while the fix was written for one operator. Fixed in 3af85c1f.

Following the old hints landed on the next refusal:

```
a + d  ->  Use Int here, or compute on `.value`.
a + n  ->  Only a literal is read as Amount beside it, and this Int is not written out.

Outer { value = a.value * b.value }  ->  A product of Inner and Inner is a quantity in neither of them.
```

I did not make the hint depend on the operator. `reason × operator → fix` would put the operator back into a refusal that is right not to have it, and multiply the keys for a fix whose content barely differs. A rule that every operator asks owns a fix that holds for every operator, so where they want different things the fix says both:

```
Hint: `*` and `/` scale by Int; `+` and `-` read Int only where it is
written out as a literal. Otherwise build Amount, or compute on `.value`.

Hint: Arithmetic does not open Outer on its own. Write `.value` to reach
Inner, and the rules of Inner apply from there.
```

The nested one no longer promises the answer is at `.value` — it says where `.value` leaves you, which is true for `Outer + Outer` and for `Outer * Outer` alike. That contract is now written on `Refusal.hintKey` so the next rule added has it in view.

The other refusals were checked against the same test rather than assumed: `scalar`, `incompatible`, `product`, `quotient` and `reciprocal` all name a fix that compiles. `Amount { value = n } + a`, `Amount { value = q.value } + a` and `a.value * b.value` wrapped at the end are each admitted.

On finding 2, I took the rename: `PlainTypeCheck` is now `DeferToPlainTypeCheck`, and its javadoc says what you pointed out — it is neither an allowance nor a refusal, an absorbed `Never` or bottom leaves the expression standing, and handing the question on is itself the answer, which is what separates it from falling through.

Hints were indeed unpinned. They are pinned structurally now: `everyNewtypeRuleSaysWhatToDoInstead` walks the newtype refusals and fails on one that carries none. Verified by removing the hint from `NoDirectNumericBase` and watching it fail. A test cannot judge whether a sentence is true, so the contract for that is on `hintKey` where the next author will read it.

PR body brought in line: three answers rather than two, 13 rows, 3052 + 109 + 166 + 1.

kawasima:
Reproduced and fixed in ad010c14.

```
Decimal / Amount  ->  Amount wraps Int, and this Decimal is not one.
                      Hint: `*` and `/` scale by Int; ...
Int / Amount      ->  Dividing Int by Amount inverts it, ...
```

The hint now says which way `/` goes:

```
Hint: `*` scales by Int on either side; `/` scales only as Amount / Int, and
Int / Amount is a reciprocal. `+` and `-` read Int only where it is written
out as a literal. Otherwise build Amount, or compute on `.value`.
```

Since the same class of error could hide in the others, I read every remaining hint against every operator and direction its refusal is reachable from, rather than fixing this one and stopping:

| refusal | reachable from | hint holds |
|---|---|---|
| `base` | `+ - * /`, newtype on either side | now, with the direction of `/` stated |
| `scalar` | `+ -` only, value on either side | `Amount { value = n } + a` compiles |
| `reciprocal` | `/`, number on the left | `n / a.value` compiles |
| `product` / `quotient` | `*` / `/`, both newtypes | `a.value * b.value` wrapped at the end compiles |
| `incompatible` | `+ -` only | converting one operand to the other's newtype compiles |
| `nested` | `+ - * /` | says where `.value` leaves you, promises nothing |

`base` was the only one left with a direction it did not state.

The test row follows the author instead of the rule: it takes the refusal for `Decimal / Amount` and, beside it, the refusal that agreeing the base alone arrives at. Whoever changes either message reads both.

Full clean run green: 3052 + 109 + 166 + 1.

Issue #378, opened 2026-08-06
Prepared flattens imported definitions into `m.fns()`, and what is lost there is read back out of the name


kawasima:
A module emits the recursive helpers it reaches as its own methods, so `Shapes.Prepared` puts them into `m.fns()` under the names this module reaches them by — a prelude `List.foldFrom`, a helper another module published as `maths.spin` — beside the module's own `let`s:

```java
List<Ast.FnDef> fns = new ArrayList<>(m.fns());
fns.addAll(injected.values());
```

`HelperInliner.helpersOf` reads `m.fns()` and nothing else, so from that point a module's helpers are declarations of several modules under one set of names, and nothing records which is which. Printing what the totality check is handed:

```
module=demo   own=[firstOver, anyOver, List.foldFrom]  recursive=[List.foldFrom, firstOver]
module=order  own=[maths.spin]                         recursive=[maths.spin]
```

Two defects follow, and they are different defects.

## 1. The declaring module was discarded, and read back out of the name

`HelperTable.own()` says "the module's own helpers" and `declares(name)` says "whether the module declares `name` itself". After the flatten they answer whether the name is in this module's fn set, which is another question. Every rule that is about the declaring module then had to reconstruct it, and the only trace the join left was the dot:

```java
// A qualified name (`List.foldFrom`) is a prelude recursive helper injected into the module
// (trusted total); only bare-named user helpers are checked.
if (h == null || name.indexOf('.') >= 0) {
    continue;
}
```

The `h == null` branch never fires on this path — `Prepared` put the prelude helper into the fn set, so `own.get("List.foldFrom")` is not null. The dot test carried the whole rule.

`PartialReachability` has the same shape. Its javadoc states ADR-0098 outright — an unmarked published helper summarises its whole call closure, so a reader never walks it — and its node set is built from `inliner.helpers()`:

```java
/** Module-own helper -> the helpers it reaches, in name order. A name absent as a key is a
 * terminal: an imported or prelude helper, whose declaration answers for its own closure. */
```

Those helpers are keys after the flatten, so their bodies were walked. The rule was written down and not run.

[...]

PR #389, opened 2026-08-07
Which module declared a helper is the declaration's to say, not the name's

kawasima:
The first half of #378: a helper's declaring module is recorded on the declaration, so the rules that are about it stop reading the name.

## What was wrong

A module emits the recursive helpers it reaches as its own methods, so `Shapes.Prepared` puts a prelude `List.foldFrom` and a helper another module published into `m.fns()` under the names this module reaches them by. `HelperInliner.helpersOf` reads `m.fns()` and nothing else, so from there a module's helpers are declarations of several modules under one set of names, with nothing recording which is which. What the totality check is handed:

```
module=demo   own=[firstOver, anyOver, List.foldFrom]  recursive=[List.foldFrom, firstOver]
module=order  own=[maths.spin]                         recursive=[maths.spin]
```

`HelperTable.own()` said "the module's own helpers" and answered "in this module's fn set". `TotalityChecker` needed the first and had only the dot to read:

```java
if (h == null || name.indexOf('.') >= 0) {
    continue;
}
```

`h == null` never fires there — the prelude helper is in the fn set — so the dot carried the whole rule. And the dot does not hold the answer anyway: `List.foldFrom` is reached under the library's alias and declared in `souther.list`, so the module a helper came from is not recoverable from the name it is reached by.

`PartialReachability` had the same shape. Its javadoc states ADR-0098 — an unmarked published helper summarises its whole call closure, so a reader never walks it — and builds its nodes from `inliner.helpers()`, where those helpers now are. The rule was written down and not run.

## What this does

`Ast.FnDef` carries `declaredIn`, written once in `AstBuilder` where the source is parsed and copied by every pass after. Renaming goes through `reachedAs` and a body swap through `withBody`, so no caller is in a position to pair a name with an origin that is not its own; a lambda is `FnDef.lambda`, which no module declares. `declaredIn == null` says "not a module-level declaration", never "from somewhere else".

With that recorded:

- `TotalityChecker` requires a descent of what this module declared, asked of the declaration
- `PartialReachability` makes a helper this module took on a terminal, which is what its javadoc and ADR-0098 say
- `TypeChecker` runs `rejectReachingPartial` over this module's own declarations, so another module's mistake is not reported here
- `HelperTable` stops claiming to answer provenance: `own()`/`declares()` became `fns()`/`holds()`, which is what its three callers were asking for

`HelperNames.keyIn`'s javadoc is corrected too. It blamed `qualifyImports`, but its one caller hands it a pair and it never reads a spelling; what it returns is a key within one module, and it does not hold the declaring module.

## What changes for a build

One thing. A module published under an older `Backend.BOUNDARY_VERSION` can hold an unmarked helper reaching a `partial` one. The reader used to walk into it and report, in its own module, what the declaring module was supposed to have answered for. It no longer does.

Nothing else: a module-own helper's reach name cannot contain a dot, since `CstLexer.identifier` takes `Character.isJavaIdentifierPart`, so the old test and the new rule agree on everything a source can spell.

## Witnesses

`WhichModuleDeclaredAHelperIsAskedOfTheDeclarationTest`. No source can spell the two apart, so the tables are built where a table is built. Each was checked against the old rule:

| | under the old rule |
|---|---|
| a qualified name is not an exemption | `Expected CompileException to be thrown, but nothing was thrown` |
| a bare name is not a declaration | `E2001` raised for a definition this module never wrote |
| a helper taken on to emit is a terminal | `Optional[[maths.wrapped, maths.spin]]` |

The fourth states why the name cannot be asked at all: `List.foldFrom` is declared in `souther.list`.

`mvn -o clean test` — 109 + 3043 + 166 + 1, no failures.

## What is left

The rest of #378, which the rewritten issue splits out: a reach name is still a `String`, and it is still decoded downstream. That is a representation change and reviews on its own.

Refs #378

### DIFF
--- a/souther-compiler/src/main/java/souther/compiler/ast/Ast.java
+++ b/souther-compiler/src/main/java/souther/compiler/ast/Ast.java
@@ -412,31 +412,72 @@ public interface Ast {
      * return type and names a primitive: {@code let trim (s: String): String = intrinsic
      * "string.trim"} — its body is {@link FnBody.Intrinsic}, written only in the {@code souther}
      * namespace. What the modifiers say is in {@link Modifiers}.
+     *
+     * <p>{@code declaredIn} is the module that wrote the {@code let}, and it is written once, where
+     * the source is parsed. Every pass after that copies it. It is here because a module emits the
+     * recursive helpers it reaches as its own methods, under the names it reaches them by, so from
+     * that point a module's fns are declarations of several modules under one set of names — and
+     * which module declared one decides what may be checked of it (ADR-0098). The name cannot answer
+     * that: {@code List.foldFrom} is reached under the library's alias and declared in
+     * {@code souther.list}.
+     *
+     * <p>Null says this is not a module-level declaration at all — a block standing where a function
+     * goes, which has parameters and a body so that applying it expands like a call. It does not say
+     * "from somewhere else": a helper this module took on to emit was written by some module and says
+     * which, and every rule here reads that rather than the absence. Anything a later pass mints that
+     * <em>is</em> a declaration says which module it belongs to, so that these stay one question.
      */
-    record FnDef(WrittenName written, List<FnParam> params, RetType declaredReturn, FnBody body,
-                 Modifiers modifiers, SourcePos pos) implements Ast {
+    record FnDef(WrittenName written, String declaredIn, List<FnParam> params,
+                 RetType declaredReturn, FnBody body, Modifiers modifiers, SourcePos pos)
+            implements Ast {
         /** A fn with no modifier (the common case; totality-checked if recursive, published). */
-        public FnDef(WrittenName written, List<FnParam> params, RetType declaredReturn, FnBody body,
-                     SourcePos pos) {
-            this(written, params, declaredReturn, body, Modifiers.NONE, pos);
+        public FnDef(WrittenName written, String declaredIn, List<FnParam> params,
+                     RetType declaredReturn, FnBody body, SourcePos pos) {
+            this(written, declaredIn, params, declaredReturn, body, Modifiers.NONE, pos);
         }
 
-        /** A fn a pass wrote — an expanded copy, a helper filed under its module's qualifier —
-         * named but written nowhere the author would recognise. */
-        public FnDef(String name, List<FnParam> params, RetType declaredReturn, FnBody body,
-                     Modifiers modifiers, SourcePos pos) {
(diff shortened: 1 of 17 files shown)


kawasima:
All three, at `10208b7c`.

## 1. `rejectNamedAsValue`

You are right that the scope was missing, and chasing the reachability moved where it comes from.

A closed body cannot carry a partial-as-value. `HelperInliner.valueOf` eta-expands a helper named where a value goes (`HelperInliner.java:1265`), so a definition that went through `closeAcross` arrives with the name standing in a call. My first attempt at a witness failed on exactly that, which is what pointed at the real route: `injectedRecursiveHelpers` and `injectedExampleHelpers` take `table.reached(name)`, and for a library helper that is `Prelude.helpers()` — parsed, never closed — put into the reader's fns as it stands.

So the route exists and does not depend on eta-expansion at all. What stops a report today is that the library declares nothing `partial`, which is the premise `PartialReachability` already refuses to rest on:

> The standard library declares nothing `partial` today, so keeping it changes no answer; leaving it out would make that a premise of the check rather than a fact about the library, and one the library could stop honouring in silence.

Both rules now carry their own scope rather than leaving it to the caller's loop, and `rejectNamedAsValue` takes the fn instead of the expression so it has one to ask. It stays over `module.fns()` and not `inliner.helpers()`, as you said — a behavior's own `let` may call a `partial` helper and may not hand it over either.

The fifth witness builds the declaration as its own module wrote it, which is the shape those two injections produce for a library helper. Dropping the scope again:

```
Unexpected exception thrown: CompileException: 5:33 E2001: the `partial` helper `spin`
is written where a value goes.
```

reported against a module whose author never wrote `spin`.

## 2. `keyIn`

Replaced rather than appended to. It now says what it converts — a declaration identity to this table's reach key — that the result is a key and not an identity, and that the library does not come through it at all, since a standard-library name denotes a `ValueName.Stdlib` that already holds its reach name. The phase-dependence paragraph is gone; that belongs to `Ast.Apply.reaches()` and to #378's remaining half.

## 3. The retired invariant

Both were still there — thank you, the PR description was ahead of the diff.

`HelperNames` and `Shapes.Prepared` now say that writing the qualified spelling settles the name a reader reaches a definition by, that this is not the definition's identity, and that provenance is on the declaration. Naming `List.foldFrom` / `souther.list` in both, since that is the case where the two visibly part.

`mvn -o clean test` — 109 + 3043 + 166 + 1, no failures.

Issue #374, opened 2026-08-06
bool.sou says not is auto-imported, and the compiler requires Bool.not


kawasima:
`souther api --source Bool` prints the module's own source, comments included:

> `not` is auto-imported into every user module and, being a non-recursive helper, is expanded inline at each call site

Writing it that way does not compile:

```
65|     guard not(List.isEmpty(claim.receiptIds))
              ^^^
`not` is a standard-library function and must be called qualified, as `Bool.not`.
```

The diagnostic is right and helpful; the comment that sent the reader there is stale. (`import Bool ( not )` does work, so the unqualified form is reachable — just not automatically.)

This matters more than an ordinary stale comment because `souther doc cli/start-here` points at `api --source` as the fastest answer to why a function is shaped the way it is, so its comments read as authoritative. A reader who trusts it writes code the compiler rejects.

kawasima:
Fixed in #394, merged to `develop`.

`bool.sou` no longer describes how a caller reaches the library — that is a rule about every library name, not a fact about `not`. The rule itself is recorded in ADR-0099, with the day it was implemented (2026-07-18) and the day it was written down both in the status line; ADR-0028 is superseded in the auto-import half and stands in every other. The specification now states the import-versus-declaration conflict the checker has always reported.

Two things came out of it that the issue did not ask about. The hand-written table behind the "did you mean" report had gone stale for `insert`, `remove`, `append` and `addDays` — each published by two modules with one named — so the report is now derived from the published surface and the table is gone. And the report only ever reached calls, so a bare `empty` got "unknown identifier" and a bare `not` handed to a combinator got a guess at a nearby binding; both are answered now.

PR #394, opened 2026-08-07
bool.sou described a rule the compiler had refused for three weeks

kawasima:
`souther api --source Bool` printed a comment saying `not` is auto-imported into every user module. Writing it that way does not compile: the standard library stopped being auto-imported on 2026-07-18 and is reached through its qualifier, or by importing the name. `souther doc cli/start-here` points at `api --source` as the fastest answer to why a function is shaped the way it is, so what it prints reads as authoritative — and it was published documentation for a rule the compiler refuses.

## What was actually wrong

The reversal was implemented, migrated across the examples and the specification, and never recorded. `list.sou` and `check/Exposing` were rewritten to the new rule at the time; `bool.sou` was not, and neither were a Javadoc in the inliner, two test names, and the bullet in ADR-0028 that ties the privilege to a namespace the compiler "ships and auto-imports". A reader of `docs/adr` still learned the old rule.

So the same sentence about how every library name is reached had been written into five places, one of which is printed to users. Four of the five went stale together and nothing noticed, because nothing ties any of them to what the compiler does. The `--source` test asserts only that the module's comments come along.

## What is in this PR

**The rule is recorded.** ADR-0099 states it: qualified access always works, an import elides the qualifier and settles nothing else, an import beside a declaration of that name is a conflict, and a bare library name is reported as one. Its status line carries the day it was implemented and the day it was written down, because an ADR three weeks late is worth less than one written with the change and should say so. ADR-0028 is superseded in the auto-import half and stands in every other — the namespace and its privileges are unchanged.

**The prose that outlived the rule is gone.** `bool.sou` now says why `not` is written in Souther and why no class is generated for it, and says nothing about how a caller reaches it — that is a rule about every library name, not a fact about this one. `HelperInliner` describes the keys its table holds instead of the name resolution that ran before it.

**The specification says what the checker does.** It described an import beside a declaration of that name as the declaration winning. The checker refuses it on the import line, and has all along; what the spec was describing is what recovery leaves standing in a module that will not be emitted.

**The report is derived, not maintained.** `Prelude` kept a hand-written map from a bare name to a qualified suggestion. Four entries had gone stale: `insert`, `remove`, `append` and `addDays` are each published by two modules and the table named one, so a reader reaching for `Set.insert` was told about `Map.insert`. That map is gone; the candidates are the published surface grouped by bare name. The order they are offered in, the order the prelude sources load in, and the qualifiers that exist are one list in `Reserved` rather than three that had to agree. The catalog now receives the candidates as a list and writes the sentence around them — the previous message had English `or`s assembled in Java and handed to the Japanese report.

```
`insert` is a standard-library function. Write it qualified (`Map.insert`, `Set.insert`) or import the name you mean (spec §stdlib).
```

## Tests

`AStandardLibraryNameIsReachedQualifiedOrImportedTest` fixes the rule as programs: qualified with no import, imported and written bare, bare with no import refused and offered both routes, a module's own `not` beside `Bool.not`, and an import beside that declaration a conflict.

The candidate check reads the published surface out of the bundled `.sou` text rather than out of the tables the compiler built from it, so it does not move when the thing it checks does; what that leaves resting on `Reserved.MODULES` is the module list, so the `.sou` files shipped are listed and compared against it. Mutating the grouping to keep one candidate per bare name fails it on `length` and fails the four pinned names on `insert`.

The boundary is pinned by output: the two spellings of one call generate byte-identical classes, all of them except `$Module`, which records the imports as written for a reader of the jar. Resolution answers what a bare name means with the imports consulted last, and nothing downstream reads an import.

`CompileDidYouMeanTest` asserted the joined phrase and now asserts both names; it was the test that already held this intent, and it asked about `filter` alone, which is why the other four names drifted under it.

`mvn clean install` is green across the modules; 3363 tests, no failures.

Refs #374


kawasima:
All three are in, at `a2958a64`.

**1 — the report reached only calls.** Confirmed, and wider than the review says. `Resolve.unknownIdentifier` never consulted the surface, so besides a bare library value there is the case of a function's own name handed over rather than applied. Before the fix:

```
bare `empty` as a value:        unknown identifier `empty`
bare `not` handed to List.all:  unknown identifier `not` (did you mean `go`?)
```

The second is worse than silence — the nearby-name guess names an unrelated behavior in the same module while the answer is known. `unknownIdentifier` now asks `StdlibNames.writtenBare` after the library-member check and before the nearby-name fallback, so a bare library name wins over a guess at a binding, as it already did for a call. The catalog says "standard-library name", which is what `StdlibNames` claimed to be about and what the surface actually holds.

```
`empty` is a standard-library name. Write it qualified (`Map.empty`, `Set.empty`) or import the name you mean.
`not` is a standard-library name. Write it qualified (`Bool.not`) or import the name you mean.
```

Three tests: the value reached both ways, the bare value answered with both candidates, and the name-as-a-value case asserting no `did you mean` appears.

**2 — the registry checked one half.** Fixed where you say, at the registry. The reasons differ and both are written down: two entries writing one qualifier publish two modules under one name; two entries naming one module load that source twice and publish its declarations under two qualifiers. The second is the one nothing downstream could report, since everything downstream derives from this list.

Your point about the test is right, and the fix needed a second assertion rather than a stronger one. A repeated module names a file that is there, so the derived set and the actual set stay equal — the test now counts as well as compares, and 11 entries against 10 files is a length the sets cannot see.

**3 — the ADR sentence.** Correct, and the two conditions are independent as you say. It now reads "what needs neither a qualifier nor an import", lists module-own declarations among them, and says the library publishing that spelling too does not enter into it. I added a paragraph in the same section saying the rule is about names and not calls, since that is what 1 turned out to be.

`mvn clean install` green: 3366 tests, no failures.

PR #395, opened 2026-08-07
A diagnostic code names a rule, and a rule can be read about

kawasima:
Refs #372.

`souther doc <code>` is the affordance the toolchain advertises, and 200 of the emission sites had no code to advertise. The two diagnostics the issue named are the symptom; the defect is that a code was optional while already serving as the identity — `JsonRenderer` emits it, `Analyzer` branches on `E1922`, and a diagnostic without one is unidentifiable to a tool as well as to a reader.

## The rule

A code identifies one normative rule of the language, not one message template and not one title category. Each rule has a canonical anchor in `specification.adoc`, and no two codes name one anchor. Diagnostics reporting different manifestations of one rule share a code; diagnostics enforcing different rules take different ones, however alike their titles read.

Three names are kept apart:

| | what it is | shared? |
|---|---|---|
| `titleKey` | the category in the header — `BOUNDARY TYPE` | yes |
| `DiagnosticCode` | the identity a reader looks up and a tool keys on | one per rule |
| message key | the wording for one way the rule was broken | several per code |

`constructs` is the case that decides the granularity. E1002 and E1006 can be written as one equality and are still two obligations the specification states separately, so they are two codes with two anchors. Going the other way, E1908 covers "no fake supplied" and "fake supplied but its fixture will not build", which is one obligation broken two ways, so it is one code.

## What changed

`DiagnosticCode` is a closed enum carrying `(code, ruleAnchor, titleKey)`. `Diagnostic.uncoded`, `Diagnostic.of(String, String)` and the code→title fallback in `HumanRenderer` are gone, so a check raises through `of(DiagnosticCode, key)` and there is no builder that does not take one. `RetiredDiagnosticCode` holds the five retired numbers separately, so one can neither be emitted nor reused.

55 codes became 143. The rule anchors were written into the specification before any number was assigned, which is what kept the numbering from following the message keys. No existing code was renumbered: each of the 55 turned out to name a single rule.

Some sites joined a rule that already had a code, which is the check that the unit is the rule and not the site — a pipeline composing with itself is a behavior reaching itself (E1608); a data taking a built-in `Option` case name is a declaration taking a name the core has (E1502); a literal violating its invariant is a construction violating its invariant (E2010), the same obligation with a folded value instead of a proved path.

Others split. `MODULE ERROR` was eleven rules, and the split worth naming is E1506 against E1507: a name a module never declared is a mistake about what is there, a name it declared and kept is a mistake about what it publishes, and the two are fixed in different files. `FUNCTION` and its neighbours were fourteen; `BOUNDARY TYPE` and `CODEC` were nine.

The thirteen legacy `CompileException(pos, message)` sites split as they had to. Twelve are codegen asserting against itself — an unknown intrinsic, a node reaching the emitter untyped — and say so in their own comments; they are `IllegalStateException` beside their neighbours now. One was an author mistake and took E1608.

## Held

Seven invariants in `EveryDiagnosticCodeIsReadableTest`, each verified to fail when mutated:

```
code            -> rule anchor            exists in the specification
code            -> rule anchor            no two codes share one
code            -> [#eNNNN]               resolves to a section
retired code    -> [#eNNNN-removed]       resolves to a retirement note
[#eNNNN]        -> code                   names a code that exists
live ∩ retired  = ∅
Diagnostic.literal                        held to the one site that wraps a handed message
```

While the 200 sites were being mapped, a ceiling on the remaining `uncoded` count kept the set from growing back; it went with the last of them.

## Found on the way

Not looked for, and each its own small defect:

- six catalog messages defined and never named by the compiler, so nothing could show them;
- four title keys all reading `TYPE MISMATCH`, which existed only because each site picked its own;
- `check.case.discriminatorfield` and `check.member.discriminatorfield` — one rule shown under two titles;
- E1301 and E1501 rendering English under `--lang ja`, having never moved off the literal path;
- a test asserting that a diagnostic had no code.

## Meeting develop

Two changes landed on develop while this was being written, reaching the same thing from the other side, and both are kept.

`EveryDiagnosticTheCompilerEmitsCanBeLookedUpTest` found the codes by scanning the sources for `E`-and-four-digits literals. There are none left, so it enumerates `DiagnosticCode` instead — the whole domain rather than a sample of it — and asks about the retired codes as well.

`EveryNameTheSpecificationSendsAReaderToCanBeAskedForTest` holds that every `<<name>>` in the specification resolves to something `souther doc` answers, and `SpecDocument` already sends an anchor written inside a section to that section. It did not see an inline `[[id]]`, which is how a rule stated as a list item is named — a list item has nowhere to put a block anchor. It does now, so `souther doc a-match-has-arms` answers with `match`, and the 143 rule anchors are reachable rather than merely present.

One thing was dropped and is worth saying so. `also-named` gave one section the codes of several diagnostics, which was the right answer while eight codes had nowhere to resolve to. Each of those codes now has a section of its own, so the two attributes would give them a second answer; the attribute lines are gone and the mechanism stands. develop's E1922 and E2011 sections are fuller than the ones written here and are the ones kept.

## Verification

`mvn clean install` then `mvn test`: 3361 tests, no failures. `souther doc` answers for all 143 codes and the 5 retirement notes; checked end to end that a coded diagnostic prints its code in both locales and that the eight codes which previously had no section now resolve.


kawasima:
All four P1s and both P2s are fixed, plus one more of the same kind that the E1814 finding turned up. `fe6ac4a..HEAD`.

**1. `expect()` answering E2302 for everything.** Right, and it is the emitter breaking the invariant this branch is about. `expect` takes the rule from its caller now, since which part of the language did not read is a property of what is being read and not of the token that was missing; the 89 call sites pass the code of the construct they are in. `ASyntaxErrorSaysWhichPartOfTheLanguageDidNotReadTest` holds six inputs, one per code including E2306, and three of them fail when `expect` is put back to a single code. Two hand-written ones were wrong the same way: `AstBuilder`'s "unexpected pattern" was E2302 and "expected a literal" was E2302.

**2. E1814 changing the language.** Right, and the cause is worse than the symptom. That rule was written off `check.rechelper.fnparam` — a catalog message **nothing emits**. An unreachable message had become a language rule contradicting both the specification and the compiler. The anchor is narrowed to what is enforced: a recursive helper's own body does not call an injected behavior.

Three more dead keys were found the same way: `check.fn.predicatebool`, `check.fold.acctype`, `check.fold.shape`. E1805's section quoted one of them. All four are gone, and `noExampleMessageIsDefinedAndNeverShown` — which only ever looked at `check.example.*` — is now `noMessageIsDefinedAndNeverShown` over every namespace. That check would have caught all four before any rule was written. Three title keys the fallback removal left unreachable (`e1001.title`, `e1302.title`, `e1601.title`) are gone too.

**3. E2305 holding things that are not literals.** Took option B. `lex.unexpected` and `lex.typevar` are **E2306 `the-source-is-made-of-tokens`**: tokenizing answers before anything is parsed, so unlike the five syntax codes it says nothing about where in the language the text was. That is the difference that makes it a separate rule rather than a wider one.

**4. E1314 dropping enumerations.** Restored in both the anchor and the section. Agreed this one is the most serious of the four — `souther doc E1314` is now the canonical path to that rule, so narrowing it there narrows the language for whoever reads it.

**5. Generic sum case.** E1020.

**6. Not closed at the type level.** `Diagnostic` is a final class with a private constructor rather than a record, and `titleKey` is no longer a component — it is read off the code. `same code → different title` is now unconstructible rather than merely unwritten.

**7. What the tests prove.** Two added. `everyCodeIsEmittedBySomething` gives the live list the meaning "can be printed today", which is what separating the retired ones was for. `Diagnostic.literal` is held to an exact path and an exact count of one, so a second one in `Report.java` fails. Both verified by mutation.

Your closing point is the one that mattered: the tests held `code ↔ documentation` and left `emission site → correct rule` alone, and every P1 above came through that gap. The axis has moved there.

`mvn clean install` then `mvn test`: 3446 tests, no failures. 144 codes, 144 rule anchors, 144 sections.

kawasima:
Fixed, and the audit your point 2 implied was done rather than the two sites patched. `ad54b28..HEAD`.

**2 first, because it changed how the rest was handled.** You are right that ordering is a side condition and not a type mismatch, and the specification says so itself: `sort` / `max` / `min` / `sortBy` declare `List<'a>`, and "what stays in the compiler is what the type language cannot say". The argument fits its parameter; what is missing is a capability the signature had no way to ask for.

But the deeper reading — that the inventory is still partly message-key driven — is the one that mattered, so every one of the 144 codes was put beside its rule anchor and its message texts and read. Four sites reported a rule other than the one their code names:

| site | was | is | why |
|---|---|---|---|
| `check.ordered`, `check.ordered.key` | E1806 / E1805 | **E1816** `what-is-ordered-by-is-an-ordered-value` | a side condition, not parameter-type equality |
| `check.numeric` | E1806 | **E1817** `what-is-summed-is-a-number` | same shape, different obligation |
| `e1401.behavior` | E1401 | **E1818** `a-helper-does-not-call-a-behavior` | a helper not reaching a behavior is not an arbitrary Java call |
| `e1004.dup` | E1004 | **E1011** | a data declaring one field twice is a declaration made twice, not a spread collision |

Two of the four were visible in the key's own spelling. `e1004.dup` was being emitted as E1011 and `e1401.behavior` as E1818, and a key named for a code carries an assertion that goes stale the moment the diagnostic is found to report a different rule. `aKeyNamedForACodeIsEmittedUnderIt` checks that now, and both keys are renamed for what they say. That is the part of "the site reports the rule its code names" a machine can check — the rest was read, and the reading is what found the other two.

**1. The editor dropping the code.** Right, and it is the half of #372 about tools. Both routes pass `e.code().name()` now. Rather than a new LSP test, the six syntax sources join `DiagnosticPathAgreementTest`, which already holds that the compile path and the editor path answer the same code — syntax errors were the one family whose agreement was not implied by the others, since they reach the editor through the recovering parser rather than through a compile. Six of its cases fail if the null goes back.

**3. Not closed at the type level.** `Diagnostic.of` and `CstError` both took null without complaint, so the way out had moved from the record's constructor to the factory. Both require it now — verified by construction, not just by reading.

**4. What the test proves.** Renamed to `everyCodeIsNamedByCompilerSource`, and the javadoc says what it does and does not establish, and where the stronger statement comes from: an input that produces the diagnostic, which is what the six syntax tests do.

147 codes, 147 rule anchors, 147 sections. `mvn clean install` then `mvn test`: 3465 tests, no failures.

kawasima:
Fixed. `22fa242..HEAD`.

You are right, and it is the same defect as the last round wearing the other of the three names: the identity moved to the rule and the category stayed behind on the old one. E1816 read `TYPE MISMATCH` two lines under a paragraph I had written saying it is not a type mismatch and does not read as one.

- E1816, E1817 → **`UNSATISFIED CONSTRAINT`**, a category for a side condition the signature had no way to ask for.
- E1818 → **`BEHAVIOR NOT CALLABLE HERE`**.

Checked end to end in both locales rather than just in the enum:

```
-- UNSATISFIED CONSTRAINT  E1816 ----------------d4.sou:5:43
-- BEHAVIOR NOT CALLABLE HERE  E1818 ------------d5.sou:10:24
-- 満たされない制約  E1816 -----------------------d4.sou:5:43
-- ここからは呼べない behavior  E1818 -------------d5.sou:10:24
```

**On making it mechanical.** Took your suggestion. `aTitleNamedForACodeIsUsedByThatCode` fails when a code takes an `eNNNN.title` that is not its own, and E1818 fails it — verified by putting `e1401.title` back. As you say it cannot reach E1816: a shared title whose wording contradicts the rule is read for, like the rules themselves, and the javadoc says which of the two this test is. The enum's title keys are also all defined and all used, so a category cannot be left behind with nothing showing it.

**The non-blocker, taken.** The duplication is what let the two routes drift — one handed over the code and the other a null — so leaving it and testing around it would have kept the reason. There is one reader now, `syntaxOf(text, lines)`, which is what makes the agreement test's single-file coverage mean something for the workspace route.

147 codes, 147 rule anchors, 147 sections. 3466 tests, no failures.

PR #411, opened 2026-08-07
A call the specification writes is resolved against the signature it is written for

### LINKED ISSUE #406: The section that teaches passing a function writes List.all arguments in the order it forbids
The section that teaches passing a function writes `List.all`'s arguments in the order it forbids.

`<<blocks>>` (specification.adoc:1804) states the rule that a named function may be passed where a
block may, and writes both halves of the example with the collection first:

```
List.all(reasons, r -> isValid(r))   // wrap in an anonymous block
List.all(reasons, isValid)           // pass by name directly. Same thing
```

The signature is the other way round:

```
stdlib_api {name: "List.all"}
→ List.all(p: ('a) -> Bool, xs: List<'a>) : Bool
```

And the convention is stated outright at specification.adoc:3053:

> Standard library arguments put *the primary subject (collection, string, Map) last, functions,
> separators, and keys first*

So the one example whose entire point is "these two lines are the same thing" has both lines
wrong, and a reader who reaches `<<blocks>>` before `<<pipe>>` writes code that does not compile.

## Scope

`rg 'List\.(all|any|map|filter|fold|sum|find|sortBy|isEmpty|length)\(' specification.adoc` — these
two lines are the only ones in the file that break the convention. Every other call in the prose
puts the subject last.

## Why it survived

Nothing ties a call written in prose to the signature the type checker resolves. `stdlib_api`
already prints resolved signatures; a check that every `Module.name(` in the specification agrees
with the resolved arity and argument order would have caught this, and would keep catching it.

## Measured

Found by an agent that happened to read `<<stdlib-list>>` and `<<blocks>>` in the same turn and
noticed they disagreed. It resolved the disagreement with `stdlib_api {name: "List"}`. A reader who
opens only `<<blocks>>` has no signal at all.

kawasima:
Closes #406.

The section that teaches passing a function wrote `List.all`'s arguments in the order the convention forbids:

```
List.all(reasons, r -> isValid(r))   // wrap in an anonymous block
List.all(reasons, isValid)           // pass by name directly. Same thing
```

The signature is `List.all(p: ('a) -> Bool, xs: List<'a>) : Bool`, and `<<pipe>>` states the rule outright — the primary subject last, functions first. So the one example whose point is "these two lines are the same thing" had both lines resolving to nothing the checker would accept, and a reader who reaches `<<blocks>>` before `<<pipe>>` writes code that does not compile.

## The check

Nothing tied a call written in prose to the signature the checker resolves. `EveryStandardLibraryCallTheSpecificationWritesAgreesWithItsSignatureTest` reads every `Module.name(...)` out of the specification's own text — 148 of them, in prose and in listings alike — and asks `ApiCommand.surface()`, which is what `souther api` prints, what it declares. Neither side is written for the other: what is checked is that the side a reader reads and the side the compiler enforces say the same thing.

Three ways they can disagree:

- the name is not one the standard library publishes
- the argument count differs. A `|>` supplies the last argument, so a piped call is read with one more than it writes; a call quoted with an ellipsis (`String.length(...)`) stands outside the count
- a function is written where the signature declares something else. An argument is a function on its face when it is an anonymous block, a getter block, or a name declared as a function — by the library, or by the document that writes the call. A name nothing declares says nothing either way, so it is left alone

That last one is what catches both halves of the example. `r -> isValid(r)` is a block; `isValid` is a bare name the same listing declares as `let isValid (r: PreApprovalReason)`.

## It is not passing on an empty scan

Reverting the two lines makes the check name them:

```
specification.adoc:1804: List.all is passed a function as argument 2, where it declares xs: List<'a>
specification.adoc:1805: List.all is passed a function as argument 2, where it declares xs: List<'a>
```

Each kind of finding also has its own test over inline text, and one test asserts a call is found at all.

## What it does not cover

`Decimal.divide(dividend, divisor)` (specification.adoc:3154) is written in order to say that it is refused. The document is right and the checker is right to resolve it otherwise, so it is listed in `WRITTEN_TO_BE_REFUSED` with the reason rather than checked. It is the only entry.

Calls written without a qualifier — `singleton(k, v)`, `insert(k, v, m)` and the rest of the `<<stdlib-map>>`-style sections, which are where each module's argument order is described — are not read. Resolving those needs the section to say which module the bare name belongs to, which is a different scan.

## Visibility

`ApiCommand.Signature` and `surface()` widen from private to package-private for the test beside them. Reading the resolved surface any other way would be a second copy of what `souther api` answers.

### DIFF
--- a/souther-compiler/src/main/java/souther/compiler/doc/ApiCommand.java
+++ b/souther-compiler/src/main/java/souther/compiler/doc/ApiCommand.java
@@ -78,7 +78,7 @@ public final class ApiCommand {
     }
 
     /** One callable name's parameters, as written, and the type its call answers with. */
-    private record Signature(List<String> paramNames, List<Type> paramTypes, Type result) {}
+    record Signature(List<String> paramNames, List<Type> paramTypes, Type result) {}
 
     private static void listPublished(PrintStream out, String prefix) {
         surface().forEach((name, signature) -> {
@@ -97,7 +97,7 @@ public final class ApiCommand {
      * only the arguments its caller writes. Leaving it out would have this command contradict the
      * specification about what exists.
      */
-    private static Map<String, Signature> surface() {
+    static Map<String, Signature> surface() {
         Map<String, Signature> surface = new LinkedHashMap<>();
         for (Map.Entry<String, Prelude.PreludeEntry> e : Prelude.entries().entrySet()) {
             if (Prelude.published().contains(e.getKey())) {
--- a/souther-compiler/src/test/java/souther/compiler/CompileLibraryFunctionByNameTest.java
+++ b/souther-compiler/src/test/java/souther/compiler/CompileLibraryFunctionByNameTest.java
@@ -12,8 +12,8 @@ import static org.junit.jupiter.api.Assertions.assertThrows;
 
 /**
  * A named function may be handed over rather than applied, and that is one rule for every named
- * function ([#blocks]): {@code List.all(reasons, isValid)} and {@code List.all(reasons, r ->
- * isValid(r))} are the same thing. It held for a module's own {@code let} and not for the library's,
+ * function ([#blocks]): {@code List.all(isValid, reasons)} and {@code List.all(r -> isValid(r),
+ * reasons)} are the same thing. It held for a module's own {@code let} and not for the library's,
  * whose names could not be written without an argument list at all.
  *
  * <p>What a name is on the other side of — a Souther body, a shipped kernel — is not the name's
--- a/souther-compiler/src/test/java/souther/compiler/doc/EveryStandardLibraryCallTheSpecificationWritesAgreesWithItsSignatureTest.java
+++ b/souther-compiler/src/test/java/souther/compiler/doc/EveryStandardLibraryCallTheSpecificationWritesAgreesWithItsSignatureTest.java
@@ -1,0 +1,322 @@
+package souther.compiler.doc;
+
+import org.junit.jupiter.api.Test;
+import souther.compiler.Prelude;
+import souther.compiler.types.Type;
+
+import java.io.IOException;
+import java.io.InputStream;
+import java.io.UncheckedIOException;
+import java.nio.charset.StandardCharsets;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+import java.util.TreeSet;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertFalse;
+import static org.junit.jupiter.api.Assertions.assertNotNull;
+import static org.junit.jupiter.api.Assertions.assertTrue;
+
+/**
+ * A call the specification writes in its prose or in an example is what a reader copies. The
+ * checker resolves that call against a signature the standard library declares, so a call the
+ * document writes one way and the library declares another is the document teaching code that does
+ * not compile.
+ *
+ * <p>The calls are read out of the specification's own text and the signatures come from the
+ * resolver `souther api` prints. Neither side is written for the other: what is checked is that the
+ * side a reader reads and the side the compiler enforces say the same thing.
+ */
+class EveryStandardLibraryCallTheSpecificationWritesAgreesWithItsSignatureTest {
+
+    /** A qualified standard-library call, wherever the specification writes one. */
+    private static final Pattern CALL =
+            Pattern.compile("(?<![\\w.])([A-Z][A-Za-z0-9]*)\\.([a-z][A-Za-z0-9]*)\\(");
+
+    /** A function the specification declares, so a bare name it writes elsewhere is known to be one. */
+    private static final Pattern DECLARED_FUNCTION = Pattern.compile(
+            "(?m)^\\s*(?:partial\\s+)?let\\s+([a-z][A-Za-z0-9]*)\\s*\\("
+                    + "|^\\s*behavior\\s+([a-z][A-Za-z0-9]*)\\s*:");
+
+    /** An argument written as an ellipsis rather than as a value — the call is quoted, not made. */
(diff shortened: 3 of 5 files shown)


kawasima:
Three of the five findings hold and are fixed in 1f78bd04. Two do not, and I asked the compiler rather than settle it by reading.

## Fixed

**[P1] one-directional.** Right. Argument kind is three-valued now — `FUNCTION` / `NOT_A_FUNCTION` / `UNDECIDED`. A literal, a list literal, a construction and a library name declaring no parameter list are `NOT_A_FUNCTION`, and one of those at a parameter declared `('a) -> 'b` is a disagreement too. `List.map([1, 2], xs)`, `List.fold(0, step, xs)` and `List.filter("a", xs)` are each caught by a test.

`List.map(xs, ys)` still passes. Nothing in the text settles what either name holds, and the compiler catches it — `E1804: argument 1 of List.map is f, which takes a function: pass a named function or a lambda. Write List.map(f, xs).`

**[P2] document-global scope.** Right, and your `let n = 3` example was the shape. Declarations are collected per region now — a listing block, or a paragraph outside one — so one listing's `let n (x)` no longer makes another listing's `n` a function. `aFunctionDeclaredInOneListingDoesNotNameAnotherListingsValue` covers it.

**[P2] `Map.empty()`.** Right, and there is a diagnostic for it: `E1803: Map.empty is a value, not a function, so it takes no (). Write Map.empty on its own`. `Prelude.isEmptyCollectionValue` is the library's own answer to "was this declared with a parameter list", so the check asks that rather than reading zero arity as agreement. The specification writes `Map.empty()` and `Set.empty()` twice to say they are refused, so those join `Decimal.divide(dividend, divisor)` in `WRITTEN_TO_BE_REFUSED`, which now carries the sentence each one comes from.

**Name.** Right that it claimed more than the scan decides. It is `NoStandardLibraryCallTheSpecificationWritesContradictsItsSignatureTest`, and the javadoc says what is decided and what is left alone. It refutes; it does not verify.

## Not fixed

**[P1] `behavior` should come out of `DECLARED_FUNCTION`.** It compiles:

```souther
behavior norm : (s: String) -> String

let norm (s) = s

behavior go : (i: In) -> Out constructs Out

let go (i) = Out { ys = List.map(norm, i.xs) }
```

specification.adoc:1818 states the rule the quote at :202 does not contradict — "A behavior's name handed over is the *behavior*, not the `let` that implements it: a Java implementation replaces the behavior, so passing one to `List.map` and calling it by name reach the same place". E1024 fires where a name reaches a value position with nothing behind it, which is a different case. Treating a behavior as a function is what the specification and the compiler both do, so the scan keeps doing it. There is now a test asserting both directions of it.

**[P2] a `Type.Var` parameter can hold a function.** It cannot:

```souther
let bump (x: Int) = x + 1
let go (i) = Out { n = Map.size(Map.singleton("k", bump)) }
```

`E1809: a block is not a value: it may be passed as an argument or bound to a let and applied, but it cannot be returned or stored in a data (spec 12.5)`. A function does not go into `value: 'a`. Requiring `FnOf` is the language's rule, not a Java-class shortcut standing in for one.

## On the design note

Agreed that this must not grow into a second type system, which is why the three-valued split is the shape it takes: the scanner extracts calls, arity, pipe and argument text, and the only semantic question asked is one the text settles on its face. Everything else is `UNDECIDED` and belongs to the checker.

Also fixed: `CompileLibraryFunctionByNameTest`'s javadoc quoted `List.all` in the order this branch is fixing.

kawasima:
Both hold. Fixed in 1a0a9fca.

## `WRITTEN_TO_BE_REFUSED` is now an expectation

It skipped before resolving, which suppressed rather than stated. It now holds the contradiction each negative example is written to carry, and the document test asserts the whole list instead of emptiness:

```java
private static final List<String> WRITTEN_TO_BE_REFUSED = List.of(
        // "`Decimal.divide(dividend, divisor)` without scale and mode cannot be written"
        "stdlib-decimal: Decimal.divide(dividend, divisor) is written with 2 argument(s) and declares 4",
        // "a *value* and not a function — written `Map.empty`, never `Map.empty()`"
        "stdlib-map: Map.empty() is a value and takes no argument list",
        // "so it is a *value* and not a function — written `Set.empty`, never `Set.empty()`"
        "stdlib-set: Set.empty() is a value and takes no argument list");
```

Both directions you named now fail. If the library came to accept two-argument `Decimal.divide`, that entry goes missing. And writing `Map.empty()` anywhere else is an extra — I checked by adding one to `<<blocks>>`:

```
specification.adoc:1798 [blocks]    Map.empty() is a value and takes no argument list
specification.adoc:3260 [stdlib-map] Map.empty() is a value and takes no argument list
```

A finding is keyed by the anchor it stands under rather than the line, so an edit above it does not move the expectation. The line is still printed in the assertion message, where nothing compares it.

One limit worth stating: a second `Map.empty()` written *inside* `<<stdlib-map>>` would produce the same key. Using a `List` rather than a `Set` means the duplicate still shows up as an extra element, so that case fails too.

## Shadowing

Right, and the diagnosis is right — the problem was not the region size but using a name set instead of resolving. A region now also collects what it *binds*: helper and block parameters, and `let` value bindings. A name that is both declared as a function and bound as something else settles nothing, so it falls to `UNDECIDED` rather than being read as the function it shadows. That is the conservative option you suggested; growing a scope resolver here is exactly what this scan should not do.

Your example is now a test, along with a block-parameter version of it:

```text
let index (x: Int) = x + 1

let nth (index: Int, xs: List<String>) =
    List.get(index, xs)
```

```text
let p (x: Int) = x > 0

List.map(p -> List.get(p, xs), ys)
```

Neither is vacuous. Dropping the shadow guard fails both:

```
aParameterShadowingAFunctionOfTheSameListingSettlesNothing:216
  expected: <[]> but was: <[: List.get(index, xs) is passed a function as argument 1, where it declares index: Int]>
aBlockParameterShadowingAFunctionOfTheSameListingSettlesNothing:230
  expected: <[]> but was: <[: List.get(p, xs) is passed a function as argument 1, where it declares index: Int]>
```

And the original defect is still caught — reverting specification.adoc:1804-1805 names both lines.

`mvn clean test` is green across the reactor.

Issue #405, opened 2026-08-07
The MCP answers are the CLI answers, so they send the reader back to a shell they do not have


kawasima:
The MCP answers are the CLI's answers, so they send the reader back to a shell they do not have.

`McpServer.call` runs the subcommand and hands its captured text back as the tool result. The text
was written for a reader at a prompt, and nothing adapts it to the transport. So a server whose
purpose is to keep a reader out of the repository tells them, repeatedly, to go run a command.

## Every occurrence

| Answer | What it says | Is the content reachable over MCP? |
|---|---|---|
| `doc_search`, every truncated result | ``… 118 more; `--limit 0` for all of them`` | No. `doc_search` takes `term` and nothing else |
| `stdlib_api {name: "String.length"}` | ``souther api --source String` for what it means`` | No. `stdlib_api` takes `name` and nothing else |
| `doc_read` miss, and the tool description | ``souther doc` lists every section`` | No — #404 |
| `doc_read {name: "raoh/tutorial"}` | 55,904 bytes in one result | No. No offset, no page, no section argument |
| `cli/commands` § examples | `souther examples` — "the command worth running on a model you believe is finished" | No tool at all |

The first three are the same defect twice over: the answer names the escape hatch and the schema
forbids it. One agent tried the only thing left and got the flag offered back:

```
doc_search {term: "newtype --limit 0"}
→ no section says `newtype --limit 0` — searching for its words instead
  … 123 more; `--limit 0` for all of them
```

## Measured

Three agents wrote a domain model from the MCP tools alone. Every `doc_search` any of them issued
was truncated at 20 hits — 60 to 140 total — and none of them ever read a tail. The only mitigation
available was to re-search with a narrower term, which reorders the ranking rather than continuing
the page.

One reached `doc_read {name: "raoh/tutorial"}`, took a 2 KB preview of 55,904 bytes, and could not
ask for the rest. It got what it needed from the smaller `raoh/composition-patterns`; had the
content lived only in the tutorial, that was the end of the road.

One found `stdlib_api`'s deferral answerable by accident — `doc_read {name: "stdlib-string"}` holds
the semantics `souther api --source String` would have printed. Nothing in the `stdlib_api` output
points there. It points at the shell.

## Two things to separate

The text is one fix: what an answer says when it runs out of room has to be what *this* caller can
act on. The capability is another: `doc_search` needs a way to continue, `doc_read` needs a way to
read a large topic in pieces, and the content behind `--source` and `souther examples` is either
[...]

Issue #404, opened 2026-08-07
An MCP client reaches a section only by guessing its name, because nothing lists them


kawasima:
An MCP client reaches a section only by guessing its name, because nothing lists them.

`doc_read` requires `name`. There is no listing tool, and no no-argument form. So every section
an MCP client reads, it reached by mining a name out of a `doc_search` hit line — and `doc_search`
finds only what the reader can already name.

Two error messages and the tool's own description answer the question by naming a command an MCP
client does not have:

```
doc_read {name: "raoh"}
→ no section `raoh`
  `souther doc` lists every section

doc_read {name: "raoh/index"}
→ no doc topic `raoh/index`
  `souther doc` lists every section and topic
```

> `doc_search` and the no-argument `souther doc` listing name them.
> — the `doc_read` tool description

## Measured

Three agents wrote a domain model from the MCP tools alone, with the repository and every other
subcommand withheld. All three named this as the gap to fix first, and all three reached for a
listing before anything else. One agent's very first call was `doc_read {name: "index"}`.

The cost is concrete. `cli/start-here` is the page written for exactly this reader — *"If you have
never written Souther, read these four in this order"* — and the three reached it, in order: never;
by accident, through an unrelated search; and on the 35th call, after the model already compiled.
Over the CLI, in the previous round, all three found it first.

Two of the three found `doc_read {name: "example"}` — the single highest-value page, a complete
annotated model — by guessing the noun. The third found it as a hit line under a search for `data`.

One agent found a workaround for a doc *set*: `doc_search {term: "raoh/"}` returns exactly the six
raoh topics with no truncation. It does not scale to the specification — `doc_search {term: "/"}`
returns 20 of 107. Nothing documents the trick.

## What the listing is

`souther doc` with no argument prints 307 lines, 13,784 bytes — anchor, indent, title. That is one
tool result, well inside what a client can take at once.

kawasima:
Fixed in #412, merged to develop as ee88fc9d.

`doc_read` with no `name` answers the listing — the same 307 lines `souther doc` prints. The tools now publish capabilities rather than single argv forms, so `--limit`, `--search` and `--source` each have a spelling a client can write, and no answer sends an MCP client to a shell it does not have.

#405 stays open for what remains there: reading a large topic in pieces, and a tool for `souther examples`. Those are capabilities the commands do not have either.

PR #412, opened 2026-08-07
A tool publishes what the command can do, not one way of calling it

kawasima:
`souther doc` decides among three answers by the shape of its argument vector: none is the listing,
`--search` is the search, anything else is one section. The MCP layer bound `doc_read` to exactly
one of those shapes — always one argument — and declared `name` required, so the listing had no
spelling a client could write. The code that builds it runs at a prompt and was unreachable over
the wire.

That is why the error text and the tool's own description both named `souther doc`: an escape hatch
the schema forbids. `stdlib_api` said the same thing about `souther api --source`.

## What changed

The tools publish capabilities rather than the argv forms they happened to take.

| tool | arguments |
| --- | --- |
| `doc_search` | `term`, `limit` (count, optional) |
| `doc_read` | `name` (optional) |
| `stdlib_api` | `name` (optional) |
| `stdlib_api_search` | `term` |
| `stdlib_api_source` | `name` |
| `jar_api` | `name`, `classpath` (optional) |

`doc_read` with no name is the listing. `doc_search` takes the `limit` the flag takes. What `souther
api` selects with `--search` and `--source` is a tool each. Where MCP and the CLI still differ — the
stdlib search answers every match, so it takes no count — the tool description says so.

The schema a client reads is the one the server enforces. Each argument kind both publishes its
domain and decides membership in it, so the two cannot drift, and the domain is a set of values
rather than of ways to write them down:

- a count publishes `minimum: 0` and `maximum: 2147483647`, and is judged whole-or-not and
  small-enough-or-not, which are those two edges. `1`, `1.0` and `1e0` are one number and all three
  are counts;
- requests are read keeping decimals, so what the checks see is the number that was sent —
  `1.0000000000000000001` is not silently a `1`;
- a string publishes the negation of ECMA-262's whitespace, written in code points. The same
  literal is what gets published and what gets matched, so `\s` — which does not mean one thing on
  both sides — appears on neither.

An argument no tool declares is refused rather than dropped, and `additionalProperties: false` says
up front that there are none to guess at. Absent, `{}` and `null` are three different things at both
the `arguments` object and each argument in it. `params` as an alias for `arguments` is gone — it is
not in the protocol and reached the tool without validation.

What an answer offers next is chosen by who is asking. `Caller` holds both spellings of each offer,
so an MCP client is pointed at a tool call and a reader at a prompt keeps the flags and subcommands
they have.

## Measured, over a real session against the packaged jar

```
doc_read {}                             -> 307 lines, identical to what `souther doc` prints
doc_read {name:"nope"}                  -> no section `nope`
                                           `doc_read` with no `name` lists every section
doc_search {term:"/"}                   -> 20 hits + "... 87 more; `limit: 0` for all of them"
doc_search {term:"/", limit:0}          -> 107 hits, no truncation line
doc_search {term:"newtype", limit:1.0}  -> 1 hit
doc_search {term:"/", limit:"0"}        -> -32602 "`limit` must be a whole number"
doc_search {term:"/", limit:-1}         -> -32602 "`limit` must not be negative; 0 is all of them"
doc_search {term:"/", limit:1e-324}     -> -32602 "`limit` must be a whole number"
doc_search {term:"/", limit:1e400}      -> -32602 "`limit` must be no larger than 2147483647"
stdlib_api {name:"String.length"}       -> `stdlib_api_source {name: "String"}` for what it means
stdlib_api_source {name:"String"}       -> module souther.string ...
stdlib_api {source:true}                -> -32602 "`stdlib_api` has no argument `source`; it takes `name`"
doc_read {page:2}                       -> -32602 "`doc_read` has no argument `page`; it takes `name`"
doc_read arguments:null                 -> -32602 "`arguments` must be an object, or left out entirely"
doc_read {name:null}                    -> -32602 "`name` must be a string"
doc_read {name:"<U+00A0>"}              -> -32602 "`name` must not be empty"
```

Four test classes. `TheSchemaAClientReadsIsTheOneTheServerEnforcesTest` holds the published schema
as text and then sends every call that schema admits, closing both directions — a schema that
invites a call the server refuses, and a schema that hides a form the server would have answered.
The count's domain is fixed as a boundary table of values, each of which is sent, not as a handful
of representative calls.

Every assertion was checked against a mutant: `name` required again, the MCP wording reverted to the
CLI's, the unknown-argument check removed, the range check dropped, the null bypass restored, the
`\S` shorthand restored, decimals not kept on the way in. Each fails only the matching case. A
further case holds the CLI's own text in place, so adapting it for the other caller is not the same
as taking it away from this one.

`mvn clean test` is green across every module.

## Left out on purpose

Reading a large topic in pieces — `doc_read {name:"raoh/tutorial"}` is 55 KB in one result — is a
capability the commands do not have either, so it stays in #405 rather than becoming a projection
fix. So does the missing tool for `souther examples`. Documents that describe the CLI still
describe it; that is their subject. What changed is the line that says what to do next.

Extracting `DocService`/`ApiService` so the CLI and MCP share a capability instead of an argv and a
stdout is the follow-up refactor. `Caller` is the interim seam.

Refs #404, and the three MCP answers of #405 that named a CLI command.

### DIFF
--- a/README.md
+++ b/README.md
@@ -161,8 +161,10 @@ same way from 0.7.1 on, so `souther doc raoh/composition-patterns` reads raoh's 
 version this build depends on.
 
 The same answers are served over the Model Context Protocol: `souther mcp` speaks MCP on stdio,
-exposing `doc_search`, `doc_read`, `stdlib_api`, and `jar_api`, so agent harnesses that prefer
-tools over shell commands register one command. It answers under protocol revisions `2025-11-25`
+exposing `doc_search`, `doc_read`, `stdlib_api`, `stdlib_api_search` and `jar_api`, so agent
+harnesses that prefer tools over shell commands register one command. `doc_read` with no argument
+lists every section and topic, which is where a client with nothing else starts. It answers under
+protocol revisions `2025-11-25`
 and `2025-06-18` — the ones whose opening exchange is `initialize` — echoing the client's own when
 it is one of them:
 
--- a/README.md
+++ b/README.md
@@ -161,10 +161,10 @@ same way from 0.7.1 on, so `souther doc raoh/composition-patterns` reads raoh's 
 version this build depends on.
 
 The same answers are served over the Model Context Protocol: `souther mcp` speaks MCP on stdio,
-exposing `doc_search`, `doc_read`, `stdlib_api`, `stdlib_api_search` and `jar_api`, so agent
-harnesses that prefer tools over shell commands register one command. `doc_read` with no argument
-lists every section and topic, which is where a client with nothing else starts. It answers under
-protocol revisions `2025-11-25`
+exposing `doc_search`, `doc_read`, `stdlib_api`, `stdlib_api_search`, `stdlib_api_source` and
+`jar_api`, so agent harnesses that prefer tools over shell commands register one command.
+`doc_read` with no argument lists every section and topic, which is where a client with nothing
+else starts. It answers under protocol revisions `2025-11-25`
 and `2025-06-18` — the ones whose opening exchange is `initialize` — echoing the client's own when
 it is one of them:
 
--- a/souther-compiler/src/main/java/souther/compiler/doc/ApiCommand.java
+++ b/souther-compiler/src/main/java/souther/compiler/doc/ApiCommand.java
@@ -25,6 +25,10 @@ public final class ApiCommand {
     private ApiCommand() {}
 
     public static int run(String[] args, PrintStream out, PrintStream err) {
+        return run(args, out, err, Caller.CLI);
+    }
+
+    static int run(String[] args, PrintStream out, PrintStream err, Caller caller) {
         if (args.length == 0) {
             listPublished(out, null);
             return 0;
@@ -64,8 +68,7 @@ public final class ApiCommand {
             out.println(line(asked, signature));
             // A signature says what to pass, not what it means — whether a span counts both ends,
             // whether a division aborts or answers a case. That is in the module's own source.
-            err.println("`souther api --source " + asked.substring(0, asked.indexOf('.'))
-                    + "` for what it means");
+            err.println(caller.stdlibSource(asked.substring(0, asked.indexOf('.'))));
             return 0;
         }
         if (!Prelude.isQualifier(asked)) {
--- a/souther-compiler/src/main/java/souther/compiler/doc/Caller.java
+++ b/souther-compiler/src/main/java/souther/compiler/doc/Caller.java
@@ -1,0 +1,49 @@
+package souther.compiler.doc;
+
+/**
+ * Who a command is answering, so that what an answer offers next is something that caller can do.
+ *
+ * <p>The doc and api answers are written once and reached over two wires. A line that says how to
+ * see more has to name a way of asking, and the two callers ask differently: a reader at a prompt
+ * has subcommands and flags, an MCP client has tools and their arguments. Naming the other one's
+ * spelling sends the reader somewhere they cannot go.
+ *
+ * <p>Every such offer lives here rather than at the place that prints it, so that adding a caller
+ * is answering these questions once instead of finding every message that asks them.
+ */
+enum Caller {
+
+    CLI, MCP;
+
+    /** How to see every specification section and every shipped doc topic. */
+    String everySectionAndTopic() {
+        return switch (this) {
+            case CLI -> "`souther doc` lists every section and topic";
+            case MCP -> "`doc_read` with no `name` lists every section and topic";
+        };
+    }
+
+    /** How to see every specification section. */
+    String everySection() {
+        return switch (this) {
+            case CLI -> "`souther doc` lists every section";
+            case MCP -> "`doc_read` with no `name` lists every section";
+        };
+    }
+
+    /** How to ask a search for the hits it held back. */
+    String everyHit() {
+        return switch (this) {
+            case CLI -> "`--limit 0` for all of them";
+            case MCP -> "`limit: 0` for all of them";
+        };
+    }
+
+    /** How to read a stdlib module's own source, where what a signature means is written. */
+    String stdlibSource(String module) {
+        return switch (this) {
+            case CLI -> "`souther api --source " + module + "` for what it means";
+            case MCP -> "`stdlib_api {name: \"" + module + "\", source: true}` for what it means";
+        };
+    }
+}
--- a/souther-compiler/src/main/java/souther/compiler/doc/Caller.java
+++ b/souther-compiler/src/main/java/souther/compiler/doc/Caller.java
@@ -43,7 +43,7 @@ enum Caller {
     String stdlibSource(String module) {
         return switch (this) {
             case CLI -> "`souther api --source " + module + "` for what it means";
-            case MCP -> "`stdlib_api {name: \"" + module + "\", source: true}` for what it means";
+            case MCP -> "`stdlib_api_source {name: \"" + module + "\"}` for what it means";
         };
     }
 }
--- a/souther-compiler/src/main/java/souther/compiler/doc/DocCommand.java
+++ b/souther-compiler/src/main/java/souther/compiler/doc/DocCommand.java
@@ -21,10 +21,18 @@ public final class DocCommand {
     private DocCommand() {}
 
     public static int run(String[] args, PrintStream out, PrintStream err) {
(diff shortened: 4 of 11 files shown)


kawasima:
All three confirmed against the packaged jar before changing anything, and all three are fixed. A
fourth instance of the same class turned up while fixing the third.

## 1 — `limit` domain

Reproduced exactly as described: `limit: 2147483648` passed `isIntegralNumber()` and `asInt()` threw,
so a schema-valid call came back as `JsonNodeException: 'LongNode' method 'asInt()' cannot convert
value 2147483648 to 'int'`. `limit: -1` answered every hit through `limit <= 0`, by a spelling no
description mentions.

A count now publishes its bounds and is checked against the same ones:

```
limit {'type': 'integer', 'minimum': 0, 'maximum': 2147483647}

limit: 2147483648  → -32602 "`limit` must be an integer no larger than 2147483647"
limit: -1          → -32602 "`limit` must not be negative; 0 is all of them"
```

The two are no longer written apart. `Kind` now both publishes its domain and decides membership in
it, which is what let them drift. Strings picked up the same treatment — `pattern: "\\S"` is the
blank the server already refused, said before the call rather than after it.

## 2 — `stdlib_api {source: true}`

Took the split. `--source` is now `stdlib_api_source`, the way `--search` is already
`stdlib_api_search`. You named the real problem: the abstraction was bending for one argument, and a
conditional schema would have kept the bend while making it expressible. With the split there is no
cross-argument rule left to encode, and `Kind.BOOLEAN` had no other user, so it is gone too.

```
stdlib_api        | required: []       name
stdlib_api_search | required: [term]   term
stdlib_api_source | required: [name]   name
```

The deferral text follows: `` `stdlib_api_source {name: "String"}` for what it means ``.

## 3 — `arguments: null`

Fixed as you wrote it. Absent, `{}`, and `null` are now three different things, the last being
`-32602 "arguments must be an object, or left out entirely"`.

## 4 — the same slip one level down

Checking for leftovers turned up `{"name": null}`, which was also read as the argument left out —
the schema gives `name` a type and null is not of it. `doc_read {"name": null}` now says
`` `name` must be a string ``. Same rule, same reason.

## Tests

Added `TheSchemaAClientReadsIsTheOneTheServerEnforcesTest`, which holds the published schema as text
— a count's type and both bounds, a string's pattern, what each tool lists as required — and then
sends every call that schema admits and asserts the server answers it. That closes both directions,
not just the one you asked for.

Each new assertion was checked against a mutant: dropping `canConvertToInt()`, restoring the
`isNull()` bypass, and marking `doc_read`'s `name` required again. Each fails only the matching
case, and the last one fails `everyCallTheSchemaAdmitsIsOneTheServerAnswers` with
`{"name":"doc_read"} fits the published schema` — which is the original bug of this PR, now caught
by the schema test rather than by a hand-written call.

`mvn clean test` is green across every module (3215 in souther-compiler).

kawasima:
Both confirmed and both fixed. One sub-claim in 1 does not hold for the Jackson this build uses, and
it changes what the second predicate is there for — details below.

## 1 — `integer` is a value, not a notation

Reproduced: `{"limit": 1.0}` and `{"limit": 1e0}` were both refused, and both are integers under
JSON Schema. The check asked Jackson about the node's representation, which is not the question the
schema poses.

`COUNT` now asks whether the value is whole and whether it fits, which are the two edges the schema
publishes, and says which edge was fallen off:

```
limit: 1.0         -> 1 hit          (read as the count, not merely permitted)
limit: 1e0         -> answered
limit: 1.5         -> -32602 "`limit` must be a whole number"
limit: 2147483648  -> -32602 "`limit` must be no larger than 2147483647"
```

The two messages are deliberate: a client told only that its count was wrong cannot tell whether to
round it or shrink it.

### On `canConvertToInt()` alone

Measured against jackson-databind 3.2.1 rather than taken from the 2.18 docs:

```
1            IntNode      integral=true  exact=true  toInt=true
1.0          DoubleNode   integral=false exact=true  toInt=true
1.5          DoubleNode   integral=false exact=false toInt=false
2147483648   LongNode     integral=true  exact=true  toInt=false
2147483648.0 DoubleNode   integral=false exact=true  toInt=false
```

`canConvertToInt()` is already false for a value with a fraction here, so it alone would draw the
same accept/reject boundary — the 2.18 `DoubleNode` behaviour you cited has changed. I confirmed
this the hard way: my first mutant dropped `canConvertToExactIntegral()` and the suite stayed green.

It is still carried, because it decides *which* refusal is given: without it `2147483648` would be
reported as not a whole number, which it is. `aCountOutsideTheDomainIsToldWhichEdgeItFellOff` fails
without it. So the predicate stays, for a reason other than the one stated.

## 2 — `\s` is not one set

Also real, and I had it in the tree. With a `name` of one U+00A0:

```
before: answered, "no section `<U+00A0>`"
after:  -32602 "`name` must not be empty"
```

U+00A0 is ECMA-262 whitespace and is not `Character.isWhitespace`, so the published pattern refused
what the server took. Rather than weaken the claim, the set is now written once, as ECMA-262's own
whitespace spelled in code points — U+0009 to U+000D, U+0020, U+00A0, U+1680, U+2000 to U+200A,
U+2028, U+2029, U+202F, U+205F, U+3000, U+FEFF — negated into a character class. Both engines read
those identically, and that one literal is both what gets published and what gets matched.

`\s` appears on neither side, so no shorthand is left whose meaning depends on which engine reads
it. `aSpaceIsASpaceOnBothSidesOfTheSchema` runs U+0020, U+00A0, U+2003 and U+FEFF through the
published pattern and through the server, and requires both to refuse.

## 3 — the boundary table

Added as values rather than as representative calls, and each one is sent:

| value | schema | server |
| --- | --- | --- |
| `0`, `1`, `1.0`, `1e0`, `2147483647` | accept | accept |
| `1.5`, `-1`, `2147483648`, `"1"`, `null` | reject | reject |

`everyCallTheSchemaAdmitsIsOneTheServerAnswers` keeps its name and now carries `1.0` and `1e0` too.

Mutants: reverting to `isIntegralNumber()` fails four cases including the property test; restoring
`"\\S"` + `isBlank()` fails the space case; dropping `canConvertToExactIntegral()` fails only the
two-refusals case, as above.

`mvn clean test` green across every module (3219 in souther-compiler).

kawasima:
Confirmed and fixed. Probing it turned up a second case at a magnitude a client could plausibly
send, which is the one that convinced me this is P2 rather than a curiosity.

## Measured, before the fix

```
value                    plain mapper                    USE_BIG_DECIMAL_FOR_FLOATS
1.0000000000000000001    DoubleNode exact=true  int=true DecimalNode exact=false int=false
1e-324                   DoubleNode exact=true  int=true DecimalNode exact=false int=false
1e400                    DoubleNode exact=false int=false DecimalNode exact=true  int=false
1.0                      DoubleNode exact=true  int=true DecimalNode exact=true  int=true
1e0                      DoubleNode exact=true  int=true DecimalNode exact=true  int=true
```

`1e-324` is the one you named. `1.0000000000000000001` is the same defect at ordinary size: it
arrived as `1` and was honoured as a count of one, and it took no exotic exponent to get there.
`1e400` was refused for not being whole, which it is — what it is not is small enough.

## After

```
limit: 1.0000000000000000001  → -32602 "`limit` must be a whole number"
limit: 1e-324                 → -32602 "`limit` must be a whole number"
limit: 1e400                  → -32602 "`limit` must be no larger than 2147483647"
limit: 1.0                    → 1 hit
limit: 0                      → 46 hits
```

Nothing else in the domain moved. The two checks are unchanged; what changed is that they now see
the number that was sent.

The mapper is the one that writes responses too, but `USE_BIG_DECIMAL_FOR_FLOATS` is a
deserialization feature and every response node is built here rather than parsed, so nothing on the
way out is affected.

`aCountIsJudgedAsTheNumberSentAndNotAsWhatADoubleCanHold` covers the three, and the which-edge case
now also requires `1e400` to be refused for its size. Removing the feature from the builder fails
both.

## Your framing

Right, and worth restating because it is the third time in this branch:

```
wire value -> representation -> Kind -> command
```

The first round had the tools projecting one argv form instead of the command's capabilities. The
second had `Kind` asking about a node's notation instead of its value. This one had the reading
settle the question before `Kind` could ask it. Same defect, moving one layer earlier each time it
is chased out of the last one. I do not think there is another layer below this one — the bytes are
the request.

## PR body

Updated: the `pattern: "\\S"` line was a commit stale, and the measured table now carries these
cases.

`mvn clean test` green across every module (3220 in souther-compiler).

Issue #407, opened 2026-08-07
japi stops short of what the class file holds, and a near miss ends the search


kawasima:
`japi` stops short of what the class file holds, and a near miss ends the search.

Four defects in one surface, found by agents reading raoh through `jar_api` with the jar and
`javap` withheld.

## A constant is printed without its value

```
jar_api {name: "net.unit8.raoh.ErrorCodes"}
→ public static final String REQUIRED
  public static final String TOO_SHORT
  ...
```

21 names, no values. The class is documented as *"String constants for all built-in error codes"* —
the value is the whole content of the declaration, it is in the class file, and the reader is
comparing it against `Issue#code()`. One agent inferred `"required"` and `"out_of_range"` from
`Issue`'s javadoc and the specification's `<<decoder-error>>` section, and said so in its notes.
This was the single place in the round where an agent said it would have unzipped the jar.

`ConstantValue` is on the field. Printing `= "required"` is the fix.

## A near miss ends the search

```
jar_api {name: "net.unit8.raoh.decoder.Decoder"}
→ no class or package `net.unit8.raoh.decoder.Decoder` on the class path:
    /Users/kawasima/workspace/souther/souther-cli/target/souther.jar
```

The package is `net.unit8.raoh.decode`. Two of three agents made this exact miss. `japi` offers no
nearest name, though the compiler does it for every identifier in a source file — *"The nearest
name that is gets suggested, since a name out of scope and a name misspelled read the same"*. The
tool does not do for its reader what the compiler does for its author.

`doc_read` does suggest, and what it suggests is not usable either:

```
doc_read {name: "cli"}
→ no section `cli`
  did you mean: module-dependencies-are-acyclic
```

## The miss names a path into the compiler's own tree

The message above hands an absolute path inside `/Users/kawasima/workspace/souther` to a client
whose reason for existing is that it does not read that tree. Name the entry that was searched by
what it is, not by where this machine keeps it.

## Member selection works and nothing says so

```
jar_api {name: "net.unit8.raoh.Result#map2"}
→ public static <A, B, C> Result<C> map2(Result<A>, Result<B>, BiFunction<...>)
```

This is the `japi Class#method` form. The `jar_api` description says only *"Give a fully qualified
class or package name"*, so an agent asked for `net.unit8.raoh.json.JsonDecoders` and took 16
near-identical `combine` overloads, arities 2 through 16, each with a javadoc enumerating
[...]

PR #413, opened 2026-08-07
japi is read by someone who cannot open the jar

kawasima:
Fixes the four defects in #407. They are one thing four times: `japi` and the MCP tools over it are read by someone who cannot fall back on `javap` and a zip listing, and each defect is a place where that reader is left with less than the class file holds.

## A constant is printed with its value

`ConstantValue` is on the field and was never read. `fieldLine` now prints it as the source wrote it — the pool stores `boolean` and `char` as an int and says nothing about the width of an integer, so the field's descriptor tells them apart, and what Java cannot write plainly is escaped.

```
public static final String REQUIRED = "required"
public static final char SEPARATOR = '\n'
public static final long LIMIT = 10L
public static final float RATIO = 1.5f
public static final Object OPEN
```

A field an initializer computes has no value in the class file, and none is invented for it — the last line above.

## A near miss is answered with the name it almost spells

`Suggest.nearest` measures the edit distance the compiler already measures for every identifier in a source file, and takes its bound from the caller: a class identifier and a section anchor are not the same length, so the measure is shared and the threshold is not.

On a miss `japi` reads the entry names on the class path — names only, not the classes themselves — and offers the nearest.

```
$ souther japi net.unit8.raoh.decoder.Decoder
no class or package `net.unit8.raoh.decoder.Decoder` on the class path:
  souther.jar
did you mean: net.unit8.raoh.decode.Decoder, net.unit8.raoh.decode.Decoders
```

`souther doc` had a suggestion already, arrived at by asking whether one name occurred inside another. That is not a measure of nearness — `cli` sits in the middle of `acyclic` — and it now asks the same distance, with one case ahead of it: a name that is the whole of a section name's opening segment is what was asked for. `e1001` names the diagnostic `e1001-removed` is about, and no count of edits would put the two near each other. That case settles the question on its own, so `e1002` and its neighbours are not listed beside it.

```
$ souther doc E1001
did you mean: e1001-removed
$ souther doc cli
(nothing offered)
```

## The class path is named by what it is, not where this machine keeps it

A class path entry now carries whether the caller is the one who named it. A path they wrote is said back as they wrote it; a path the command fell back on is its own hosting and is named by the file it is.

```
$ souther japi net.unit8.raoh.decoder.Decoder        # fell back
  souther.jar
$ souther japi acme.Nobody -cp /tmp/nowhere.jar      # given
  /tmp/nowhere.jar
```

## `Class#member` is declared, not just implemented

The form worked and was written down nowhere. For a client whose whole knowledge of a tool is its schema, a form the schema does not mention does not exist. It is now in the `jar_api` description, in the description of the `name` argument that takes it, in the CLI usage line and in the command's doc.

```
souther japi <class-or-package>[#<member>] [-cp <path>]
```

## Tests

Every regression test was watched failing before its fix. Three assertions could not fail beforehand and are there to keep the new code from reaching further than it should: that a field without `ConstantValue` gets no value, that a name near nothing on the class path is offered nothing, and that an explicit `-cp` is not abbreviated.

The class path test asserts that no defaulted entry line contains a path separator, rather than that some expected filename appears — a positive assertion would not catch the leak coming back.

`mvn clean verify` is green across all modules.

### DIFF
--- a/souther-compiler/src/main/java/souther/compiler/check/Suggest.java
+++ b/souther-compiler/src/main/java/souther/compiler/check/Suggest.java
@@ -1,16 +1,53 @@
 package souther.compiler.check;
 
+import java.util.ArrayList;
 import java.util.Collection;
+import java.util.Comparator;
+import java.util.List;
 
 /**
  * "Did you mean" hints for unknown names. When a name the checker cannot resolve is a near-miss of
  * one that is in scope, the error carries the likely intended name, so a typo reads as a typo. The
  * closeness bound is Levenshtein distance ≤ 2 and strictly less than the name's length, so a short
  * name does not match every candidate.
+ *
+ * <p>What "near" measures is one thing across the compiler and the documentation tools: an edit
+ * distance, not a containment. How near is near enough is not — a class identifier and a section
+ * anchor are not the same length — so {@link #nearest} takes its bound from the caller while the
+ * measure stays here.
  */
 public final class Suggest {
 
+    /** The bound an unknown identifier is matched under. */
+    private static final int NEAR_ENOUGH = 2;
+
     private Suggest() {
+    }
+
+    /**
+     * The candidates within {@code bound} edits of {@code name}, closest first, at most
+     * {@code limit} of them.
+     *
+     * <p>A candidate must also be further than its own length is short — a two-letter name would
+     * otherwise be within two edits of everything and suggest whatever came first.
+     */
+    public static List<String> nearest(String name, Collection<String> candidates, int bound, int limit) {
+        record Near(String candidate, int distance) {}
+        List<Near> near = new ArrayList<>();
+        for (String candidate : candidates) {
+            if (candidate.equals(name)) {
+                continue;
+            }
+            int d = distance(name, candidate);
+            if (d <= bound && d < name.length()) {
+                near.add(new Near(candidate, d));
+            }
+        }
+        return near.stream()
+                .sorted(Comparator.comparingInt(Near::distance).thenComparing(Near::candidate))
+                .limit(limit)
+                .map(Near::candidate)
+                .toList();
     }
 
     /** {@code " (did you mean `X`?)"} for the closest candidate within bound, or {@code ""}. */
@@ -34,7 +71,7 @@ public final class Suggest {
                 best = candidate;
             }
         }
-        if (best != null && bestDistance <= 2 && bestDistance < name.length()) {
+        if (best != null && bestDistance <= NEAR_ENOUGH && bestDistance < name.length()) {
             return best;
         }
         return null;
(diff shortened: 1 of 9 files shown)


kawasima:
Addressed in afa283c2. All three were real; each was reproduced before being fixed.

## 1. `Files.walk` and `UncheckedIOException` — fixed

Reproduced: a class path directory holding a subdirectory this process cannot descend into ends the run, and ends it inside the suggestion lookup.

```
japi acme.Greetr -cp <tree>:<jar>
→ java.io.UncheckedIOException: java.nio.file.AccessDeniedException: .../tree/shut
```

`namesOn` now catches it and steps the entry over. The walk also collects into the list as it goes rather than through `toList()`, so the names visited before the failure are kept — one of them may be the name that was meant.

Test: `aDirectoryTheSearchCannotDescendIntoIsSteppedOverLikeAnyOtherEntry`.

## 2. The leak through the exception message — fixed

Reproduced. `FileNotFoundException` from `new JarFile(...)` on an unreadable file carries the path:

```
skipping `shut.jar`: /var/folders/.../shut.jar (Permission denied)
```

You are right that `Entry` should carry this to the end. `Entry.failure(IOException)` now decides it from the same `given` flag `shown()` uses: a caller who wrote the path gets the message, and an entry the command fell back on gets the kind of failure.

```
default   skipping `shut.jar`: cannot be opened
given     skipping `/tmp/.../shut.jar`: /tmp/.../shut.jar (Permission denied)
```

The invariant the tests state is now the one you named. `theClassPathThisToolFellBackOnIsNamedByWhatItIsNotWhereItIs` asks the whole of stderr rather than the entry listing, and `aFailureOnTheClassPathThisToolFellBackOnCarriesNoPathEither` covers the exception route specifically.

Testing this needed a seam: `JapiCommand.run` has an overload taking the class path to fall back on, so a test does not set a system property for every other test in the JVM. The `ClassLoader` overload stays — it is not consulted, and naming the loader that is not consulted is what `JavadocIsOnlyPrintedWhereItProvablyBelongs` uses it for.

## 3. Floating-point specials — fixed, contract kept as "Java"

`NaNf` and `Infinity` confirmed. I kept the stronger contract rather than falling back to javap-like display: what `japi` prints is read as Java by someone who is going to write Java against it, and `NaNf` is not something they can write.

```
public static final float NOT_A_NUMBER = Float.NaN
public static final double UNBOUNDED = Double.POSITIVE_INFINITY
public static final double BELOW = Double.NEGATIVE_INFINITY
```

`literal`'s doc now says a name is printed where Java has no literal, so the contract is written down rather than implied.

## One more, found while fixing 1 and 2

A miss walks the class path three times — for the class, for the package, and for a name to suggest — so one unopenable entry was reported three times. The third was mine. `Skipped` remembers what has been said and a reader is told once (`anEntryThatCannotBeOpenedIsSaidOnceHoweverManyPassesMeetIt`).

## On the two permission tests

Both assume rather than assert their fixture: a run as a user who may read anything reports them skipped, not green. On this machine they run — `Skipped: 0` in the module's surefire summary.

## Not done: the `Files.walk` cost

Agreed and left alone. A recursive scan of `.` on a one-character typo is real, and a per-class-path name index is the fix; it is worth its own issue rather than this PR.

`mvn clean verify` green across all modules (3208 in `souther-compiler`).

Issue #420, opened 2026-08-07
Diagnostic rendering conflates reader locale with locale-invariant legacy text


kawasima:
Split out of #410, which settled the default and deliberately left this. This replaces the original
framing of the issue — "five surfaces, five answers" — which the inventory below does not support.

## Inventory

Five call sites in main sources select a `Locale` for a diagnostic:

- `Main.java:547,640,648` — `Messages.resolveLocale(render.lang)`, so `--lang`, then `SOUTHER_LANG`, then English
- `apt/SoutherProcessor.java:134` — the same resolution, from the `souther.lang` processor option
- `lsp/analysis/Analyzer.java:1885,1902` — `Locale.ENGLISH`, written in
- `query/Report.java:152` — `Locale.ENGLISH`, written in
- `ExampleVerifier.java:162` — `Messages.defaultLocale()`

That is the whole list. `Runner.RunException.localized(Locale)` takes one as a parameter and decides
nothing.

## They are not five instances of one policy

The first three render for a reader and have to name a language. The last two build the `legacyBody`
argument of `CompileException` — the text `getMessage()` returns. `Report`'s own javadoc calls it
"the one-line English text `CompileException#getMessage()` has always returned for it", and
`asException()` gives the reason as "a test that reads it should not move with the reader's locale".
What those two require is invariance under locale, not English.

## Reader-facing adapters never read that text

Each of the three adapters prints `getMessage()` only when the exception carries no structured
diagnostic:

- `Main.reportCompileError:619`
- `apt/SoutherProcessor.render:125`
- `lsp/analysis/Analyzer.fromCompile:1845`

`ExampleVerifier.legacySummary` is always passed alongside diagnostics (`Compiler.java:209,212,379`),
and `Report.asException` always builds an exception that carries one. So neither string reaches a
reader through any adapter. What reads them is `getMessage()` itself: JUnit output, embedding
callers, tests.

Those two sites have no reader locale to choose. The API required one anyway.

## What made the wrong choice look right

`Locale` is an ordinary parameter on the leaf of the rendering API — `DiagnosticRenderer.body(Diagnostic,
Locale)`, `Messages.get(String, Locale, Object...)`. Any caller can produce a value: `Locale.ENGLISH`
is a constant and `Messages.defaultLocale()` is a no-arg static. Nothing at a call site separates
"resolved from the reader's flag" from "needed a value and typed one".

[...]

PR #421, opened 2026-08-07
A catalog that ships is complete on its own, and the build finds them all

### LINKED ISSUE #419: A newly added message catalog bypasses the catalog integrity checks
Split out of #410, which settled the default and deliberately left this. Rewritten around what
investigating it found: the defect is in the catalogs, and the premise this issue started from —
that a catalog may only name a language the documentation is written in — is not adopted. See
*What this replaces*.

## Observed

A `messages_fr.properties` of three keys, dropped beside the English and the Japanese catalog,
passes every catalog check the build has. `MessageCatalogFormatTest` (10 tests) and
`EveryDiagnosticCodeIsReadableTest` (9) stay green. The file defines none of the ~180 keys the base
defines and quotes a standard-library name that does not exist, and nothing says so.

Run with `SOUTHER_LANG=fr`, that catalog answers:

```
-- INVARIANT (WARNING)  E2011 -----------------probe.sou:8:5

8|     Eaches(m)
       ^

cette construction peut violer son invariant Eaches
Hint: If a relation between inputs guarantees this, declare it as an `invariant` on the input data
so it is checked once at its boundary and assumed here.
```

One diagnostic in two languages: the message body out of the catalog, the title and the hint out of
the base.

## Why

Every catalog check names its catalogs literally — `/souther/compiler/diag/messages.properties` and
`/souther/compiler/diag/messages_ja.properties`, two paths repeated per check. Nothing discovers the
catalogs that ship, so a third one falls outside all of them at once: key completeness, duplicate
keys, `MessageFormat` validity, and the standard-library names the prose quotes.

`ResourceBundle`'s key-level fallback to the base is what turns that incompleteness into a mixed
answer rather than a missing one. The fallback was written as the migration mechanism — a message
became Japanese as it was migrated and read English until then — which is also what makes an
incomplete catalog look like a legitimate thing to ship.

## Decision

1. Every catalog that ships defines the base's complete key set and is valid on its own: no
   duplicate key, every message a well-formed `MessageFormat` pattern, every standard-library name
   it quotes one the library publishes. The checks quantify over the catalog files found in the
   tree, not over a list of locales somebody remembers to extend.
2. The base fallback stays as a fail-safe — a half-migrated bundle on a branch still compiles, a key
   missing everywhere renders as itself — and no catalog that ships may depend on it.
3. A language named explicitly, through `--lang`, `SOUTHER_LANG` or an adapter's own option, is
   answered in whether or not documentation exists in that language. A diagnostic's code is the same
   string in every language, so the English documentation stays reachable from an answer in any of
   them.
4. The machine's locale never selects a diagnostic language, at any entry point.

## What this replaces

The original title — *A message catalog may name a language the documentation is not written in* —
took `CatalogLocales ⊆ DocumentationLocales` as the correctness condition and asked for a build-time
check enforcing it. That condition is not adopted. Rendering diagnostics in a language and
publishing documentation in it are separate capabilities, and the code already holds the link
between them that matters: the code a diagnostic carries is locale-independent, so a translated
diagnostic keeps its way back to the English text. The documentation's language is the reason
English is the *default* — it is not a ceiling on which languages may be rendered.

That leaves `messages_ja.properties` with a clear meaning. It declares that Japanese diagnostic
rendering is available, not that Souther publishes Japanese documentation.

The half of the original that does hold is 4, and it already holds in the code: after #418 the only
environment read for language is `SOUTHER_LANG`, in one place. `SOUTHER_LANG` is not `LANG` — it has
no value until somebody writes one, so a session answered in French is a session that was configured
for French, which is the second source and not the third. What was missing is that nothing held it:
the existing test asks `Messages.resolveLocale` what it answers, which an adapter resolving a locale
of its own never goes through.

## Not in scope

- Centralizing which component decides the language (#420). This issue says the machine's locale is
  never the source; that one says who owns the choice.
- Localizing the specification, the CLI topics or the library topics. Introducing a locale axis on
  the documentation is a separate design, and there is no non-English documentation to hang it on
  today.

kawasima:
Closes #419.

## The defect

A `messages_fr.properties` of three keys, dropped beside the English and the Japanese catalog,
passed every catalog check the build had. It defined none of the ~180 keys the base defines and
quoted a standard-library name that does not exist, and nothing said so. Run with `SOUTHER_LANG=fr`
it answered one diagnostic in two languages:

```
-- INVARIANT (WARNING)  E2011 -----------------probe.sou:8:5

8|     Eaches(m)
       ^

cette construction peut violer son invariant Eaches
Hint: If a relation between inputs guarantees this, declare it as an `invariant` on the input data
so it is checked once at its boundary and assumed here.
```

Every check named its catalogs literally — two paths repeated per check — so a third one fell
outside all of them at once: key completeness, duplicate keys, `MessageFormat` validity, and the
standard-library names the prose quotes.

## What changes

`MessageCatalogFormatTest` becomes `EveryShippedMessageCatalogIsCompleteAndValidTest`, quantified
over the catalog files found in the tree. Each is a `Catalog(path, locale)`, the locale read off the
name and spelled back through the control `Messages` looks a bundle up with, so a name
`ResourceBundle` resolves differently than the test reads it fails rather than leaving the checks
describing a locale nobody is answered in. Exactly one base catalog has to exist, which is also what
stops a scan that missed the tree from passing everything. The format check renders each pattern
from the catalog's own text rather than through `Messages.get`, which reads the bundle and so could
answer with the base's copy of a key.

Each check was shown to fire: a misnamed file, a catalog missing the base's keys, a duplicate key,
an apostrophe-quoted pattern, and a `List.nosuchthing` in a hint each fail the check that owns them.

`TheDiagnosticLanguageIsNeverDerivedFromTheMachineTest` reads the sources for `Locale.getDefault(`,
`"user.language"`, `"LANG"`, `"LC_ALL"` and `"LC_MESSAGES"`. It is a tripwire and documented as one:
a static import or a helper in between defeats it, and what would make the property structural is
one owner for the choice, which the adapters do not have yet (#420).

## What the decision is

The Revision of ADR-0044 no longer asks for a check that the catalogs offer only languages the
documentation is written in. Rendering diagnostics in a language and publishing documents in it are
separate capabilities, and the code holds the link that matters: a diagnostic's code is the same
string in every language, so `souther doc e2011` answers in English from an answer given in any of
them. The documentation's language is the reason English is the default, not a ceiling on what may
be rendered — so `messages_ja.properties` declares that Japanese diagnostic rendering is available,
and nothing about Japanese documentation.

What binds instead is each catalog on its own, which also demotes the base fallback. It was the
migration mechanism — a message becoming Japanese as it was migrated, reading English until then —
and that is what made an incomplete shipped catalog look legitimate. It stays as a fail-safe, and no
catalog that ships may depend on it. The `Messages` javadoc, the Decision and the Consequences that
stated the old standing are corrected with it, along with `cli/commands.md` and the ADR index.

Separately, `specification.adoc` still said `--lang` selects the locale "Japanese by default,
English available", left behind when the default became English in #418.

## Verification

`mvn clean test`: 3283 in the compiler, 184 in the language server, 109 in the runtime, 1 in the
benchmarks. BUILD SUCCESS.


kawasima:
All five are in `da1a2b9a`. Both blockers were reproduced before fixing, and the first one had
something under it.

**1. `MessageFormat` validity.** Confirmed: `new MessageFormat("bad {x}")` throws
`IllegalArgumentException: can't parse argument number: x` at construction, and `{10}` was arity 0
under `\{(\d)}`. Constructing every message, ahead of any arity question, found a live one:
`parse.case.record.direct` in Japanese, whose `| Some { field }` reads as an argument named
`" field "`. The English one parses only by accident — the apostrophe in `record's` quotes the rest
of the line.

Neither could be fixed in the catalog, which is what the check was really reporting. `Messages.get`
returned a message that took no arguments as written, so whether a template was a `MessageFormat`
pattern was decided by the call site and not by the catalog — and the file held messages written
under both assumptions, both rendering wrong:

- `check.match.notcase.online` quoted its braces to survive formatting, and was being shown to the
  reader as ``-> '{' match ... '}'``, telling an author to write something that is not Souther.
- `check.arith.newtype.incompatible.hint` was showing `other''s`.
- Ten more would have lost an apostrophe the moment anybody gave their diagnostic an argument.

So there is one interpretation now: every message is rendered as a pattern, a literal brace is
`'{'` and a literal apostrophe `''`, and the build refuses a catalog the formatter cannot read.
Twelve messages are escaped and the three above start rendering what they meant. Every message of
both catalogs was rendered before and after the change — 535 keys each, 0 refused, and those three
are the only text that differs. A pattern refused at run time is answered with the text as written,
for the same reason a missing key renders as itself.

**2. Duplicate keys.** Confirmed: `foo=first` / `foo:second` loads as one key defined twice and the
`=` scan sees one definition; whitespace separators and backslash continuations behave as you said.
Took the second option — the catalogs are held to `key=value`, one line each, which is what makes
reading a line as a definition sound. The probe that fails on `probe.dup:second` fails the format
check, not the duplicate check, which is the right division.

**3. One catalog per language.** Added, across every module, with the base under it — so a
`messages_en.properties` beside the base is the same collision, and `ResourceBundle` would prefer
the one this file does not call the base. Witnessed with a second `messages_ja.properties` under
`souther-syntax`.

**4. Base language vs default locale.** `Locale.ENGLISH`, as a fact about the file. Agreed on the
reasoning — the Revision's whole point is that those are not one axis.

**5. ADR.** `messages_ja` (the original default; revised above).

Each new check was shown to fail on the input it exists for: a stray brace, `a data's {10}`, a
`:`-separated redefinition, a duplicate key, a misnamed file, and two catalogs for one language.

`mvn clean test`: 3286 in the compiler, 184 in the language server, 109 in the runtime.

kawasima:
Both are in `7f420768`.

**Argument contract.** Confirmed, and one message had already drifted:
`check.type.mismatch.msg` names `{0}` in English — the position that did not type-check,
``argument 3 of `foo``` — and named nothing in Japanese, so a reader answered in Japanese was told
the type here does not match and not which of several places was meant. Every check around it
passed, exactly as you described: the Japanese message is well-formed and substitutes every
argument it names.

`everyCatalogUsesTheSameArgumentsAsTheBase` compares the sets, order being a translation's
business. Read through `MessageFormat` rather than off the text, so the `'{0}'` this branch
introduced for literal braces is not counted as an argument that has to appear elsewhere. Three
mutations were shown to fail it: dropping the argument (`takes []`), inventing one
(`takes [0, 1]`), and quoting the brace so the formatter sees none (`takes []`). And you are right
that this was already written down — the base catalog's own header asked for it.

**Headers.** Both were stale, and in the way you named. They now state what holds: the fallback is
a fail-safe rather than a way to ship an incomplete catalog, every value is a `MessageFormat`
pattern whether or not it takes arguments, and a key names the same arguments everywhere. The
argument-set rule is in the ADR's Revision too, since it is part of what a catalog has to be.

**One thing left, not fixed here.** What `{0}` is given is an English literal at the call site
(`operand of arithmetic`, ``argument 3 of `foo```), so a Japanese answer now renders an English
phrase inside a Japanese sentence. Making that read as one language means the argument becomes a
catalog key, which is a change to the checkers rather than to the catalogs. Happy to file it.

`mvn clean test`: 3287 in the compiler, 184 in the language server, 109 in the runtime.

PR #422, opened 2026-08-07
The MCP answers are written for the reader they reach

### LINKED ISSUE #405: The MCP answers are the CLI answers, so they send the reader back to a shell they do not have

kawasima:
`McpServer.call` ran a subcommand and handed its captured text back as the tool result, so the answers were written for a reader at a prompt. #412 fixed the offers a command composes; what was left was the text it prints without composing, and how much of it arrives at once.

Those turn out to be three separate guarantees.

## Where a document sends its reader

Not every mention of a command is the same thing. `souther compile <file.sou> -d <outdir>` states what the compiler's interface is, and a reader on any wire learns the same true thing from it. Thirty-seven mentions across the specification and the bundled `cli/` topics; thirteen of them tell the reader to go and read something else, and only those are followed by a reader who can spell them.

Those thirteen are written as the operation — `{{doc:example}}` — and spelled by the same `Caller` that picks a command's own offers. `Caller.stdlibSource` renders through it, so a command's deferral and a document's instruction cannot drift into two spellings of one call. Spelling happens as the document is loaded, before it is ranked or cut to a snippet, so a search answers over the words its own reader is going to be shown.

`cli/start-here` is the document a client is most likely to open first, and all nine of its instructions were shell commands with an existing tool counterpart. It now names none.

`cli/run` is a manual for a command a client cannot invoke. Its index entry names the caller whose listing carries it, so it is off a client's map and still read by name.

## How much of a document one answer carries

`doc_read {name: "compile-errors"}` was 80,199 characters and `doc_search {term: "a", limit: 0}` was 41,618. Naming a smaller thing says a reader can ask for less; it does not bound what comes back for the larger one, because a section's body runs on through its subsections.

Every tool now hands back at most 16,000 characters, the line saying how to carry on included — that line is part of the answer, so the room for it comes out of the count before the text is cut. Which answers are long is a property of the documents and the jars on the class path rather than of the tool table, so `cursor` is declared once for all six.

Where a part stops is read from the document: a heading, failing that a blank line, failing that the end of a line. The first two are only taken outside a block whose text is taken as it stands. A line or a block longer than one answer carries has no such place in it and is cut where the count runs out — a bound a document could talk this server out of by writing a long enough fence would not be a bound. Below half the count, a blank line and the end of a line lose to that cut too; a heading does not, being where the document itself starts something new.

The cursor carries where to resume and a SHA-256 prefix of what it was measured against, so a jar bundling another version of a library's docs is a cursor refused rather than an answer resumed in the wrong place.

A reader at a prompt is not cut at all.

## What is addressable is what the doc set named

The specification's sections are asked for one by one because every heading carries an anchor its author wrote. A shipped file can now say the same thing, with `<!-- souther-section: name -->` above a heading, read as `set/topic/name`.

A file that names no part of itself stays one document. Deriving names from headings was the alternative, and each thing such a name is made of moves: `## 0. Setup` renumbers, a reworded heading is a renamed identifier, `raoh/comparisons` shows heading level standing for emphasis rather than nesting, and `tutorial.ja` writes the same fifty headings in Japanese. A name invented here for a jar this compiler does not author would be published as though it were the library's.

`cli/commands` names its nine, so the contract has a reader in the repository that wrote it.

## Three ranges, not one

The same section model answers three questions, and they are not the same interval.

- **Addressable** is a tree: what has a name.
- **Read** is a subtree range: a part carries the parts under it, the way a specification section carries its subsections.
- **Searched** is a partition of the prose: a part's extent with the extent of every part named inside it taken out. It is a list of runs, not a range — an unnamed heading closes the named part before it, and the text under it goes on belonging to the part around them both. A file and its parts divide that file between them, so a term written once is answered once by the smallest thing that has it, and a term written anywhere is answered by something.

Where a block is, is one account and two grammars. `----` opens a listing in AsciiDoc and separates in markdown, so the specification asks in AsciiDoc and a shipped file asks in markdown. A reader choosing where to cut a long answer is not told which it holds and asks for either; over-reading a thematic break costs it some of the places it would rather stop and nothing else.

## Measured

| | before | after |
|---|---|---|
| shell commands `cli/start-here` names a client | 9 | 0 |
| `doc_read {name: "compile-errors"}` | 80,199 | 15,637 |
| `doc_search {term: "a", limit: 0}` | 41,618 | 15,938 |
| names a client can ask for | 306 | 315 |
| listing entries a client cannot act on | 1 (`cli/run`) | 0 |

`raoh/tutorial` arrives in four parts opening at `#`, `### One-to-many JOIN`, `## 15. Bulk decoding` and `## 25. Amount and currency`.

## Tests

Forty-seven tests in four classes. The ones that carry the weight:

- every code span naming a tool `tools/list` publishes, taken from every shipped document as a client reads it, is called over the protocol and has to be answered
- every name the listing gives is followed to the end of its chain, and the parts joined are held against what the same read prints at a prompt — to the character, for all of them
- every character of every shipped file is searched exactly once, the lines naming its parts excepted
- a name derived from a heading is *not* addressable

Sixteen mutants fail them, each caught by the test that names the property it breaks — among them: a character dropped at a seam, an instruction written back as a shell command, the count left unclamped, no room set aside for the line that carries on, a block closed by any delimiter, a shipped file read by AsciiDoc's rules, a part's own text written as one range, and what is searched being what is read.

## Deferred, on purpose

- **raoh has not named its parts.** `raoh/tutorial` is one document until it does, which is what the bounding makes survivable. Nothing here needs to change when it does.
- **`souther examples` has no MCP equivalent.** Its file-oriented interface cannot be exposed meaningfully to a repository-less client. That is an API to design, not a tool to add, and it is not what #405 is about.

Closes #405.


kawasima:
All four are real and all four are fixed in 68a5e017. Reproduced each before touching anything; one of them works differently from the account above, and it matters for the fix.

## 1. The count

Confirmed, with a different mechanism. A 20,000-character line with *no* trailing newline already came back as 16,000 — `end < 0` breaks the walk before `at` advances, so the last-resort clause fired. The same line **ending in a newline** came back as 20,001: `at` advanced past the count and was returned unclamped. A fenced block of 20,000 was worse — part 1 of 4 characters, part 2 of 20,006.

Every candidate the walk records is now inside `from + budget`, and what is returned when there is none is that limit.

The two conditions are indeed not both satisfiable, and transport bounding wins. Where there is no boundary the answer is cut where the count runs out — keeping a block whole would be a bound a document could talk this server out of by writing a long enough fence.

Also added a tier the review did not ask for: below half the count, a blank line and the end of a line lose to that cut. Without it the fenced case handed back 4 characters, then 5, then the rest. A heading is exempt — it is where the document itself starts something new, and a short part before one is worth it.

The suffix now comes out of the count before the text is cut, so the whole result is bounded rather than the text.

**And the bound is on every tool, not `doc_read`.** `doc_search {term: "a", limit: 0}` was 41,618 characters — larger than anything `doc_read` was refusing to hand over whole. Which answers are long is a property of the documents and the jars on the class path, not of the tool table, so `cursor` is declared once for all six. A tool whose answers fit never issues one.

| | before | after |
|---|---|---|
| `doc_search {term: "a", limit: 0}` | 41,618 | 15,938 |
| `doc_read {name: "compile-errors"}` | 80,199 | 15,637 |
| every other tool's largest | 14,834 | 14,834 |

## 2. Read range, rank range

Confirmed in both directions. `doc_search {term: "command reference"}` returned nothing, and `child text` under a nested named part was counted for both the part and the one around it.

`Topic` carries `ownTo` beside `to`, which is `ownTexts` beside the bodies. A file's own text is what comes before it names anything — the preamble, and for a file that names nothing, all of it. Every entry is ranked, so nothing is unsearchable and nothing is counted twice.

## 3. and 4. One account of where a block is

Both confirmed, and they are the same defect at two sizes: `named` did not ask at all, and `cut` asked a boolean. A sample showing how to name a part named one — `somelib/guide/fake` was addressable from inside a fenced code block, and the part containing that sample ended at the sample's `##`.

`TakenAsItStands` is now the one walk, and it remembers what opened the block: an AsciiDoc `----` is closed by `----` and not by `....`, and a markdown fence by a run of the same character at least as long with no info string after it. `SpecDocument.opaqueLines` delegates to it, so there is one account rather than three.

## Non-blockers

**Marker visibility.** Taken, in the form the spec already has. A part is read from its heading, so its own declaration is not in its text — exactly as a section's own anchor is not in its body. Read the *file*, and its children's markers are there, exactly as `[#e1303]` is visible inside `compile-errors`. Stripping them entirely would mean stripping the ones inside code samples too, or keeping two notions of position to tell those apart.

**Digest.** Taken. SHA-256 prefix rather than `String.hashCode`.

## Witness

Twelve new tests. Six mutants fail them, each caught by the test that names the property: the count unclamped, no room set aside for the line that carries on, a block closed by any delimiter, a fence not consulted where a file names its parts, what is searched being what is read, and a file that names parts not being searched itself.

The five cases suggested are all in: a 20,000-character line, a fenced block over the count, a parent preamble, a nested named part, and both a heading and a declaration written inside a fence.

CI green.

kawasima:
Both reproduced, both fixed in 5b8da252. The PR description is rewritten to the current semantics — every claim you listed as stale was.

## 1. Prose belonging to nobody

Confirmed in both shapes.

```
somelib/guide           search "unique-word" -> []      (after an unnamed sibling)
somelib/guide/parent    search "unique-word" -> []      (nested, one level down)
```

You are right that this is not fixable with another endpoint. `ownTo` is gone; `Topic` carries a list of runs — its extent with the extent of every part named inside it taken out. The comment that said "text nothing else claims" now describes what happens rather than contradicting it: the nearest named ancestor claims it, and the file is the root ancestor.

```
somelib/guide           search "unique-word" -> [somelib/guide]
somelib/guide/parent    search "unique-word" -> [somelib/guide/parent]
```

The witness is the partition itself rather than the two cases: for every file this compiler ships, every character is searched exactly once. The lines naming the parts are the one exception, and they are nobody's on purpose — a declaration is what a file calls something rather than something it says, and a reader looking for words should not be handed a file because of the notation it names its parts in. That is asserted explicitly, so it is a decision rather than a hole.

Two mutants fail it: a part's own text written as one range ending at the first name inside it, and a file's written the same way.

## 2. One account, two grammars

Confirmed, and worse than an unclosed block: it takes the declarations with it.

```markdown
# Guide

<!-- souther-section: first -->
## First

text

----

<!-- souther-section: second -->
## Second
```

`second` was not declared at all — the `----` opened a listing that never closed, so the rest of the file stopped being structure. In a file whose parts are the contract this branch is adding.

`TakenAsItStands.asciiDoc` and `.markdown` now; `SpecDocument` asks the first, `LibraryDocs.named` the second. `Continuation` asks `.either`, and its javadoc says why that is sound rather than lazy: a reader choosing where to *prefer* to cut cannot be wrong about a name, and over-reading a thematic break costs it some of the places it would rather stop, after which it still has the end of a line and the count.

I checked the raoh docs shipped today — no `----` at all, so nothing regressed there, but the input is as realistic as you say.

## 3. PR description

Rewritten. It now says the bound is on all six tools, that a line or block longer than one answer is cut where the count runs out, that a file is ranked beside its own parts over its own text, and it carries the three-range distinction — addressable is a tree, read is a subtree range, searched is a partition of the prose. Counts are current: forty-seven tests in four classes, sixteen mutants.

CI green.

kawasima:
Reproduced exactly as described, and fixed in 6cba02f8.

```
doc_search {term: "x" × 20000, limit: 0}
→ isError=false, result = 20,114 characters
→ …xxxxx", limit: 0, cursor: "MS41NDIy..."}` for what follows
```

The text was cut to one character and the line saying so went out at its full length.

## The fix

Neither of the two options, and not a validation contract either. I went for the third thing your framing implies: **the line names the cursor and not the call.**

```
… 64683 more characters; ask `doc_read` again with the same arguments and
`cursor: "MTU1MTYuM2RkODFjZjZjMDk5MWU5Mg"` for what follows
```

Everything in it is this server's — the tool name, the count, the cursor — so its length is this server's to know, and the reserve is a constant. No `maxLength`, no cross-field size rule, no call the schema invites and the server refuses.

The reason to prefer it over bounding the input is that bounding the input does not settle the question, it only moves it: whatever `maxLength` is published, the reserve has to be sized for the sum of every string argument after JSON escaping, and the budget for the text shrinks by that whether or not the caller uses it. `jar_api` has a classpath argument that is legitimately thousands of characters; sizing for it would cost every other call a quarter of its answer.

And the tension you named against option 2 turns out not to hold here. "The offer names an operation this caller can act on" is the point of this branch, and this line does: it names the tool, the cursor, and that the rest is the call just made. The client that sent those arguments still has them — which is what made writing them back out worth nothing to it in the first place.

```
doc_search {term: "x" × 20000, limit: 0}   →  15,960 characters, in parts
doc_read {name: "compile-errors"}          →  15,651 characters, in parts
```

## Witness

One test, as you said:

```java
int parts = counted("doc_search", "{\"term\":\"" + "x".repeat(20_000) + "\",\"limit\":0}");
```

`counted` follows the chain to the end and asserts every answer is inside the count. Two mutants fail it — the reserve dropped, and anything of the caller's appended to the line. The second one fails three tests, since a notice that is not a bounded, known shape also breaks the join-back-up witness.

`cli/commands` § mcp now says the same thing, including why the line names the cursor rather than the call.

CI green on 6cba02f8.

PR #434, opened 2026-08-07
A caller listing is the whole set, whatever that caller can invoke

### LINKED ISSUE #424: The topic that explains `--input` is the one the catalog does not list
`doc_read` with no `name` describes itself as the place to start:

> With no `name` it lists every section anchor and every shipped topic as `name<TAB>title`, one per
> line — start there.

It lists all but one. `souther doc` answers 316 names; the same listing over MCP answers 315. The
difference is exactly `cli/run`, and it is not a truncation — the listing runs
`cli/commands/shared-options` straight into `raoh/tutorial`, with nothing cut at either end.

```
$ diff <(souther doc | cut -f1 | sort) <(doc_read {} | cut -f1 | sort)
> cli/run
```

`cli/run` reads fine when asked for by name. What it holds is the arity table for `--input`, the
rule that a one-parameter behavior is not wrapped in an array, and the sentence that a newtype
crosses bare. The only other page about the command is `cli/commands/run`, whose body past the usage
line is one sentence:

> The `--input` encoding depends on the behavior's arity and is easy to get wrong — see
> `doc_read {name: "cli/run"}`.

So the page that explains the argument the documentation itself calls easy to get wrong is reachable
only by reading a stub carefully enough to follow a pointer out of it.

## Measured

Six agents wrote a Souther model from the MCP tools alone and then ran it. All six reached `cli/run`,
all six reached it through that pointer and not through the listing, and five of the six named the
omission unprompted. Two wrote down that they tried the name on faith because it did not appear in
the listing they had just read. One put it plainly: had it gone to `run` before reading the stub, it
would have had no way to learn the page existed.

Five of the six then built `--input` correctly on the first attempt, for behaviors taking up to
three parameters with nested records, newtypes and discriminated sums. The page works. It is the
listing that does not admit it.

## What this is

The same shape as #404: the CLI answers something the MCP projection drops. A listing that a tool
description calls the place to start has to be the whole set, or the strategy it recommends silently
misses whatever is outside it.

kawasima:
Closes #424.

`cli/run` was in the listing at a prompt and not in the one an MCP client gets — 316 names against
315. Its entry in `cli/index` named `CLI` as the caller whose listing carries it.

## Why the mark does not hold

**Its reason had already been taken away.** The commit that added the mark also turned the two
places `cli/run` sends its reader into affordances, spelled per caller. What is left in the topic is
descriptive shell text — the same kind `cli/commands` keeps and every caller is shown. The mark took
a reachable answer off one map rather than rewriting an instruction one caller cannot follow.

**It was the only topic marked.** `cli/commands` documents eight commands a client cannot invoke and
is listed for everyone. `cli/commands/run` is among them: a section about the same command, whose
body past the usage line is a pointer into the topic the listing had just left out.

**It reached one route of three.** `doc_search` ranked `cli/run` for a client and `doc_read`
answered it by name. Only the listing — the one `doc_read`'s description calls the place to start —
held it back.

## What changed

`Topic.listedFor` and the caller column in an index are gone, and `DocCommand`'s listing prints
every topic. `caller` still reaches `LibraryDocs` and `SpecDocument`, where it decides how an
affordance is spelled. `Caller`'s javadoc now says where that boundary is: how to ask for something
varies by caller, which documents there are and which names a listing gives do not.

## Tests

- `aClientThatKnowsNoNameIsGivenEveryNameThereIs` is back to the two listings being one set.
- `everyNameADocumentSendsThisClientToIsOnTheListingItStartedFrom` is new: every ``doc_read {name: "X"}``
  a document offers a client names something the listing gave. Placeholders standing in for a name
  the reader supplies are left out; names are compared the way `doc_read` resolves them, so
  `{{doc:E2011}}` matches the `e2011` the listing prints.

Two mutants fail them apart: the mark restored fails only the closure test, and dropping `cli/run`
from both listings passes the set-equality test and fails the closure test.

`mvn clean verify` passes. On the shipped jar, `souther doc` and `doc_read {}` both answer 316
names.

### DIFF
--- a/souther-compiler/src/main/java/souther/compiler/doc/Caller.java
+++ b/souther-compiler/src/main/java/souther/compiler/doc/Caller.java
@@ -14,6 +14,12 @@ package souther.compiler.doc;
  * <p>These are the offers a command composes. A document that sends its reader somewhere writes
  * the same kind of offer in its own text, as an {@link Affordance}, and this is what picks its
  * spelling too.
+ *
+ * <p>How to ask for something is the whole of what varies here. Which documents there are, and
+ * which names a listing gives, are the same over every wire: a caller with no way to invoke a
+ * command still gets the manual for it, with whatever that manual sends its reader to spelled as a
+ * call that caller can make. Holding a topic back from one listing would answer a reader who has no
+ * other map with a smaller world rather than with a spelling of its own.
  */
 enum Caller {
 
--- a/souther-compiler/src/main/java/souther/compiler/doc/DocCommand.java
+++ b/souther-compiler/src/main/java/souther/compiler/doc/DocCommand.java
@@ -44,13 +44,8 @@ public final class DocCommand {
                 out.println(s.anchor() + "\t" + "  ".repeat(s.level() - 2) + s.title());
             }
             for (LibraryDocs.Topic topic : shipped.topics()) {
-                // The listing is the map a caller navigates by, so it names what this one can use.
-                // A topic left out of it is still read by name, which is how a reader who is asking
-                // about the toolchain rather than about their next step still arrives at it.
-                if (topic.listedFor(caller)) {
-                    out.println(topic.name() + "\t"
-                            + "  ".repeat(Math.max(0, topic.depth() - 1)) + topic.title());
-                }
+                out.println(topic.name() + "\t"
+                        + "  ".repeat(Math.max(0, topic.depth() - 1)) + topic.title());
             }
             return 0;
         }
--- a/souther-compiler/src/main/java/souther/compiler/doc/LibraryDocs.java
+++ b/souther-compiler/src/main/java/souther/compiler/doc/LibraryDocs.java
@@ -63,21 +63,10 @@ public final class LibraryDocs {
      * finding it once and finding it at all the same guarantee.
      *
      * <p>{@code depth} is 0 for a document and the heading's level for a part of one, which is what
-     * a listing indents by. {@code listedFor} is null for a topic every caller is shown, which is
-     * nearly all of them. A topic that documents an interface only one caller has is named in that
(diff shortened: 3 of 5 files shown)


kawasima:
Both test points are in, as 65294016.

**1. The whole-set invariant is now checked directly.** `aClientThatKnowsNoNameIsGivenEveryNameThereIs`
builds the universe — every `SpecDocument.sections()` anchor and every `LibraryDocs.topics()` name — and
holds the listing against that. The CLI/MCP comparison moved out into
`aClientAndAReaderAtAPromptAreGivenTheSameNames`, so the three checks are one property each: the listing is
whole, the two wires name the same things, and nothing a document points at falls outside.

Your third mutant is the witness: dropping `cli/commands/fmt` — which nothing links to — from both listings
passes wire consistency and closure, and fails the whole-set check alone with
`expected: <[]> but was: <[cli/commands/fmt]>`. That the sets hold what their indexes promise stays where it
was, in `WhatIsAddressableIsWhatTheDocSetNamedTest`; the javadoc says so rather than leaving the reader to
find it.

Taking your point 3 with it: the comparison is over names, not over the rendered line. A title is spelled as
its text is, so requiring both listings to render identically would be stronger than the boundary this PR
draws.

**2. The placeholder rule is the sibling's.** `STANDS_IN = Pattern.compile("<[^>]+>")` with `find()`.

This one had a mutant too. Writing `{{doc:cli/<run}}` into a section body of the specification: the old
`contains("<")` rule passed it, `ADocumentSendsItsReaderWhereThatReaderCanGoTest` passed it — it scans
`LibraryDocs` topics and not the specification, which is the gap you named — and the new rule fails with
`[cli/<run]`.

`mvn clean verify` passes. Every mutant is out of the tree; `git status` is clean against the two commits.

PR #438, opened 2026-08-07
An arm that answers nothing is not an arm a row can be in

### LINKED ISSUE #426: An arm that answers `unreachable` is reported as an arm no row reaches
[#example-branch-coverage] rules this out in as many words:

> An `unreachable` (<<unreachable>>) is not an arm. Reaching one is already <<e1911>>, so counting it
> would leave every correct model permanently one arm short.

The measure counts it.

```
data Flag = On | Off
data Answer = Int

behavior pick : (f: Flag) -> Answer
    constructs Answer

let pick (f) = match f with
    | On  -> Answer(1)
    | Off -> unreachable "the probe only ever passes On"

example pick
    | "on" : (On) -> Answer(1)
```

```
$ souther examples probe.sou
probe                                                    status: complete
  pick                     implemented   rows 1    pending 0
                in #1 specified 1/2
      · no row uses `Off`
    partition   axes 1   single-axis 1/2
      · no row is in `Off`
    boundary    0/0
    branch      1/2
      · no row goes through `case Off` (8:16)
```

Three of the four measures ask for a row that E1911 forbids. `branch 1/2` is the one the quoted
sentence is about; `no row uses Off` and `no row is in Off` are the same demand arriving through the
signature and partition measures, which the sentence does not mention and which are equally
unsatisfiable.

## What it costs

A model that says where its combinations cannot arise reports a coverage it can never reach. A model
that invents a fallback answer instead reports full coverage. The report pays for the wrong one.

Measured: one agent, asked to state the impossible combinations rather than invent fallbacks, ended
at `branch 28/30` with both remaining gaps behind an `unreachable`, and wrote that the report can
never go green for the model as specified. It could not tell from the output which of its gaps were
closeable — see the sibling issue on boundary derivation, and #428 on the gates disagreeing.

## Note on the arm feeding it

In the probe the `Off` arm's body *is* the `unreachable`. In the larger model the same shape appeared
one level up: a `match` arm whose only content is a nested `match` whose single non-`unreachable`
path is already covered. Whatever rule is written should say what it does with an arm that leads only
to an `unreachable`.

kawasima:
Closes #426.

Three measures asked for a row the compiler refuses. An arm answering `unreachable` was numbered as an arm, and the case behind it was counted at the signature and in the partition, so a model that states where its combinations cannot arise reported a coverage it could never reach — while one that invented a fallback answer reported full coverage. `--generate` offered the row that is E1911.

```
                in #1 specified 1/2              →   in #1 specified 1/1   excluded 1
      · no row uses `Off`                              · `Off` is declared unreachable
    partition   axes 1   single-axis 1/2         →   partition   axes 1   single-axis 1/1   excluded 1
      · no row is in `Off`                             · `Off` is declared unreachable: the probe only ever passes On
    branch      1/2                              →   branch      1/1
      · no row goes through `case Off` (8:16)
```

## The arms

The rule was written where it could not apply. `CoverageSites` skipped an `unreachable` while walking a body, and a body walk never numbers one — arms are numbered where a fork is made. The test guarding it asserted that no site's label contains "unreachable"; a label is `case Off`, so it could not fail.

It is decided at the fork now, and by whether the arm can answer a value (`NormalReturn`) rather than by the shape it is written in. An arm built out of an `unreachable` and an arm whose every path ends in one answer as little as a bare `unreachable`. An arm with one answering path stays an arm whatever is beside it and whatever the rows cover — what a body owes is a property of the body.

A call is not read through, and that limit is written into the specification: an inlined helper's arms are already here, and reading a recursive one means deciding what a call does.

An arm without a probe keeps its place in the array the emitter indexes (`NO_SITE`). Compacting it would move every arm below down one and light the neighbour's probe — a hit recorded against an arm nothing ran, which no count can tell from a real one. An `if` whose arms both answer nothing draws no line for a row to be at, so it registers no guard: kept, its boundary would be reported unmet however the model is exercised.

## The other two measures

There was no rule at all — both derived what there was to cover from the declaration and never read the body. What the body rules out is read off it once (`Exclusions`), and the classes a row can be written at are asked for in one place (`Axis.eligible`), so the signature's cases, a position's partition, the pair space and the generator cannot come to disagree again.

Only what the body states without anything having to be decided is read: the first fork on the body's spine, where it matches on an input position. Nothing below a fork, because an arm holding a `match` on a second input says what that input cannot be *given the first one*. Nothing off a condition, because which values satisfy a predicate is a question with a solver behind it. So exclusions that hold are missed — a class left in asks the author for a row, a class wrongly taken out tells them nothing.

An excluded case stays *declared* and leaves what the rows are held to: `declared = coverable + excluded`. The type still has it, the report names it and says why, and neither E1915 nor E1918 asks for a row that cannot be written.

## Over the example corpus

Thirteen modules, before and after. Only the arms move, and only behind an `unreachable`:

| | |
|---|---|
| crm | `branch 5/7 → 5/6`, `9/10 → 9/9` |
| hr | `branch 7/37 → 7/35` |
| ordering | `branch 7/10 → 7/9` |

The ten modules with no `unreachable` are byte-identical. The signature and partition measures do not move there: every `unreachable` in the corpus is in a helper or under a fork, so nothing is read off a spine — which is the sound-only extraction doing what it says.

## Notes

No ADR. ADR-0089 already decided that a measure that could not be made is a third value beside met and unmet; ADR-0091 already decided how a state a build reads is added; ADR-0083 already decided that an `unreachable` is a statement and not a proof, which is why this does not travel as `provenInfeasible`. The specification gains the rules and this is an implementation catching up with them.

`schemaVersion` stays 1. The added keys are optional, so a document written before they existed is still a document of this version, and that is now a test.

The generator round-trip is checked by writing the offered rows out, answering them and compiling: a generator that offered nothing would pass an assertion about `Off` alone, so what it does offer is fixed too.


kawasima:
Fixed in f218d51e. The blocker was real and cost more than the review supposed.

**`NormalReturn(Block)`.** A `Core.Block` is only ever a function value — the javadoc says so, and the general expression emitter throws on one — so evaluating that position makes the function and the body runs when the call applies it. Reading the body there is `Core.Block _ -> true` now.

The visible symptom on the reviewed head was a ratio that cannot happen:

```
let pick (f, xs) = match f with
    | On  -> Answer(1)
    | Off -> Answer(List.fold((acc, x) -> Answer(unreachable "no element arrives").value, 0, xs))

example pick
    | "on"  : (On, [1]) -> Answer(1)
    | "off" : (Off, []) -> Answer(0)
```

```
in #1 specified 2/1   excluded 1      →    in #1 specified 2/2
  · `Off` is declared unreachable          partition   axes 1   single-axis 2/2
partition   single-axis 1/1  excluded 1    branch      2/2
branch      1/1
```

Two rows specified at a position with one class: the `Off` row is written, runs and holds, and it had been excluded anyway.

`Exclusions.spine` stepped over a block too, which would have read a lambda's `match` as if it were the behavior's. It now steps over `LetIn` only. `CoverageSites.walk` keeps descending into a block, which is right — a lambda's forks are arms when it runs.

Both new tests were checked against the unfixed predicate and both fail there: `NormalReturnTest.aFunctionPassedToACallIsMadeAndNotRun` and `ACaseTheModelRulesOutIsNotOwedARowTest.aStepThatAbortsDoesNotRuleOutTheCaseThatPassesItAlong` (`expected: <[]> but was: <[example.probe.Off]>`). The integration one goes through the measures, not the predicate, so it pins the downstream exclusion.

**Reasons.** Collection stops at the first strict child that does not answer, so `Boxed { a = unreachable "A", b = unreachable "B" }` reports `["A"]`. A fork still contributes every arm, since each is a way the arm answers nothing — both cases are asserted.

**Schema.** The contract now says which direction the version protects, in the schema description and in the test: a key added since is optional so an older document still validates, and every object is closed so an older copy of the schema refuses a newer document. A reader that must accept newer documents wants the newer schema, not a higher number.

The example corpus is byte-identical to the reviewed head — nothing there is higher-order behind an `unreachable`, so the earlier numbers stand.

kawasima:
Both fixed in 24bd8764.

**The specification contradicted itself.** `what-is-read-off-a-body` still sent the reader through a block while the paragraph above it said a function written where a call takes one is made there and run elsewhere. The spine rule now reads:

```
* the body is followed to the first fork through what a `let` binds;
* a function written where a call takes one is not entered. Evaluating that position
  makes the function, and what it matches on when the call applies it is about its own
  parameters rather than about this behavior's inputs;
```

The `Exclusions` class comment carried the same stale sentence and now says the same thing as `spine()` below it.

**Construction order.** Confirmed, and confirmed by running it rather than by reading the emitter:

```souther
| Off -> Boxed { b = unreachable "B, written first"
               , a = unreachable "A, declared first"
               }.a
```

```
This row reached a point the model declares cannot arise: A, declared first (15:26)
```

The runtime stops at `a`, and the report was naming `B`. A construction is read in declaration order now — the field list `TypeOps.fieldTypes` gives, which is the list `emitFieldValues` walks. `Core.forEachChild` stays where slot order and evaluation order agree, and the comment says that is why it is used there rather than as an account of evaluation.

`theReasonsAreTheOnesEvaluationReaches` now writes the initializers against the declaration, so it fails on the written-order reading (`expected: <[declared first, so this is where it stops]> but was: <[written first and evaluated second]>`). Checked by making `declaredFields` return nothing.

Reading the fields wants `Symbols`, which the reader had no reason to hold, so it moved to `souther.compiler.partition` beside `Exclusions` — its only caller, and where the symbols were already to hand. `NormalReturn` stays in `coverage`: it is asked by `CoverageSites` too, and whether an expression answers does not depend on which of its parts aborts first.

The example corpus is unchanged from f218d51e.

kawasima:
Fixed in 37ddc485. Reproduced first, and it reaches further than the branch count:

```
in #1 specified 1/1   excluded 1
  · `No` is declared unreachable
partition   single-axis 2/3   excluded 1
branch      1/3                              ← two arms behind the abort
  · no row goes through `case Yes` (16:13)
  · no row goes through `case No` (16:13)
```

The signature and the partition go green on `a` while the branch count cannot — #426 exactly, one level in, and reachable from outside because the exclusion and the arm rule now disagree about the same arm.

**The walk carries whether a row that stands can get this far.** A row cannot get past something that aborts, so what a node cannot do nothing inside it can do either:

```java
boolean inside = reachable && answers(e);
```

and an arm needs both — `armOf(..., inside)` — so a fork under a dead arm is numbered no more than the arm is. Applied at the body too, so `let f (…) = { let x: Int = unreachable "…"; if c then A else B }` has no arms and no guard rather than two arms and a line.

`branch 1/3 → 1/1` on the reproduction.

**Every fork stays in the plan.** As you say, dropping the subtree would leave `probesOf` with no entry for a fork the emitter still generates and stop the generation. A fork nothing reaches gets `{NO_SITE, NO_SITE}` — the structure without the obligation — and `GuardRef` is not registered there, so this does not move from the branch measure to the boundary one.

That is why one of the four new tests is an integration one: the unit test never runs the emitter over the dead body. `CompileExampleCoverageTest.aForkNothingReachesIsNeitherCountedNorLostFromThePlan` measures through a real compile.

All four fail on the unfixed walk (checked by pinning `inside = reachable`):

```
aForkNothingReachesIsNeitherCountedNorLostFromThePlan  expected: <1> but was: <3>
aForkBelowSomethingThatAbortsIsNotAForkToCover         expected: <[case Yes]> but was: <[case Yes, case Yes, case No]>
aForkNothingReachesIsPlannedWithNoArms                 expected: <-1> but was: <1>
aBodyThatCannotAnswerHasNoArms                         expected: <[]> but was: <[then, else]>
```

Whether a node answers is memoised per node now: the walk asks it of every node it passes and of every arm it makes, and each answer is read off a whole subtree.

`a-fork-nothing-reaches-has-no-arms` states the rule in the specification beside the arm rule it belongs to.

The example corpus is unchanged from 24bd8764.

Issue #440, opened 2026-08-07
Decide whether a behavior may carry `Option` across its boundary


kawasima:
Two shapes compile today, and neither can be driven. Observed on a build of `develop` at d5fac718.

```
behavior pick : (x: Option<Int>) -> Int
let pick (x) = Option.withDefault(0, x)
```
```
$ souther run pick.sou --input 3
input #1 has type `Int?`, which `run` cannot decode yet (only a data type, a primitive, or a collection of those).

$ souther compile pick.sou -d out && javap -cp out pick.Pick
public interface pick.Pick extends souther.runtime.Behavior<souther.runtime.Option<java.lang.Long>, java.lang.Long>
```

```
behavior peak : (xs: List<Int>) -> List<Option<Int>>
let peak (xs) = [List.max(xs)]
```
```
$ souther run peak.sou --input '[1,5,3]'
the output has type `Int?`, which `run` cannot encode yet (only a data type, a primitive, or a collection of those).
```

## This is not an oversight

#202 measured six positions the name `Option` reached and listed a behavior's input among them.
#204 declined to guard the name: five of the six are the reading ADR-0011 already permits, and two
of those cannot be written any other way, because the annotation is mandatory there — a recursive
helper's return, and a `List.filterMap`-shaped step's function parameter. What #204 did decide is
that making an optional is a `?` field being given its value and nothing else, and that a behavior
does not *answer* an optional, read off the resolved output type so both spellings are refused as
one thing.

A behavior's input was one of the five kept. `OptionAsAWrittenTypeTest.aBehaviorInputMayNameTheType`
fixes it as the specification, and souther-bench's corpus carries a comment saying the same:
"`Option<T>` may be named wherever a model reads an optional, and this helper reads one."

## What the decision did not separate

Reading an optional inside a model, and transporting one across a boundary. From the type's side
they are the same permission. Only the second becomes an external representation and a Java-facing
signature.

The javap line above is one of the two grounds #204 gave for banning the output — ADR-0008, with
`JvmTypes:185` mapping an optional to `souther.runtime.Option`. A parameter produces exactly that,
and was never measured against it.

The output rule has the matching gap on the other side. It is asked of the whole output type, so
`-> List<Option<Int>>` is not "answering an optional" by its wording, and the optional travels
inside the shape instead of as it.

## Two questions, and the one statement that settles both

Neither follows from what is already decided:

1. May a behavior's input name an optional? Answering no changes #202's decision.
[...]

PR #441, opened 2026-08-07
Make `run` use the boundary Map-key contract in both directions

### LINKED ISSUE #425: `run` writes one of the five key kinds a boundary Map is allowed
A Map that crosses the boundary MUST be keyed by `String`, a String-backed newtype, `Date`,
`DateTime`, or an enumeration ([#collections], E1314). The compiler admits all five as a behavior's
output. `run` writes one of them.

```
behavior strKey     : (n: Int) -> Map<String, Int>       → {"k":7}
behavior newtypeKey : (n: Int) -> Map<Code, Int>         → refused
behavior dateKey    : (n: Int) -> Map<Date, Int>         → refused
behavior enumKey    : (n: Int) -> Map<Outcome, Int>      → refused
behavior tallyOut   : (n: Int) -> Tally                  → {"k":7}    // data Tally = Map<Code, Int>
behavior optOut     : (n: Int) -> Map<String, Code>      → {"k":"v"}
```

The last two are the point. Wrapping the identical map in a newtype prints byte-identical JSON, so
the encoder exists and is reached — just not from here. And a newtype in the *value* position is
written without trouble; it is the key position alone that refuses.

## Where

`Runner.encode` tests the key against the primitive by identity:

```java
if (out instanceof Type.MapOf map && map.key() == Type.STRING) {
```

A newtype key, a temporal key and an enumeration key are all `Type.Ref`, so they fall past this
branch to the catch-all. `Type.Ref` two lines below would have encoded them through the generated
codec, which is exactly what happens once the map is wrapped.

## The message argues against the diagnosis

```
the output has type `Map<Code, Int>`, which `run` cannot encode yet (only a data type, a primitive,
or a collection of those).
```

`Map<Code, Int>` is a collection of a data type. The message states a rule the reader satisfies, so
it points nowhere. The agent that hit this searched `run cannot encode` and `souther run output type
limitation`, found nothing, and recovered only by writing a four-case probe file to reverse-engineer
the real condition — which is the behavior this tooling exists to remove. It then changed its model
to route around a tool bug it had no way to identify as one.

`cli/run` says the result "is encoded back to JSON and printed", with no caveat.

kawasima:
Closes #425.

The reported symptom is a behavior answering with `Map<Code, Int>`, refused by `run` while the same
map wrapped in a newtype printed byte-identical JSON. The defect is not that output: it is that the
runner decides for itself what may key a boundary Map, in a version that admits one of the five
kinds.

## What owns the rule

`TypeOps.isBoundaryMapKey` says a Map that crosses is keyed by `String`, a String-backed newtype,
`Date`, `DateTime`, or an enumeration. `SpecChecker` and `DataChecker` ask it for E1314, and
`Deriver.mapKeyEnc`/`mapKeyDec` follow it by sending every named key through that type's own codec —
which is why wrapping the map worked.

`Runner` wrote the rule again, as `map.key() == Type.STRING`, in the decoder it composes and again
in the encoder. By the time either runs, E1314 has already passed, so the key is known to be one of
the five. The runner re-asked a question that was answered and got a narrower answer.

## Both directions

Fixing only the encode side would leave `run`'s input admitting a `String` key alone, leave the
runner and `TypeOps` disagreeing, and leave the next reader to repeat this investigation. So the
unit of the fix is the contract, not the output.

Encoding sends the key through the recursion the value already takes. Decoding rekeys the decoded
object's strings through the key type's own decoder, with a bad key reported at that key's own path
and every key read before the map is given up on. Neither direction names the newtype and the
enumeration apart: a `Type.Ref` goes to its class's codec, and the three primitives are the string
leaf.

The key decoder is the neutral-source one rather than the JSON one, because a key arrives already
read out of the object that carried it — the same choice `CodecGen.emitKeyDecoder` makes. A `String`
key is rekeyed with the rest: the keys of a decoded object do not pass the string leaf, so leaving
them alone is the one place a boundary would hand over text it had not made canonical.

## Tests

`DecoderPathAgreementTest` exists so the derived codec, the fixture builder and `run` cannot drift
apart. Two of its rows asserted that `run` refuses a key the other two read, so the table that
detects drift was holding this one open; both now state that all three read it, and `DateTime`,
enumeration, bad-key reporting and canonical `String` keys are stated beside them.

It also covered reading only, which is why an output-side defect met no test at all.
`EncoderPathAgreementTest` states the writing side as the same kind of table: a value written bare
and the same value written as a data's sole field must come out as the same JSON, since wrapping is
what routes it through the derived codec.

Each change was checked against a mutant. Restoring the encode guard fails six encoder rows,
restoring the decode guard fails five decoder rows, and skipping the rekey for a `String` key fails
the canonical-key row alone.

## Two refusals with nothing left to refuse

With the key read from one place, no type a behavior can declare reaches `run`'s `cannot decode yet`
or `cannot encode yet`: a tuple is E1311, an optional E1402, an anonymous union E1312, an
inadmissible key E1314. `RunnerTest.everyTypeThatCrossesIsOneRunDrives` states that as a table.

`RunnerTest.namesAnUndecodableInputTypeInSurfaceSyntax` used a newtype-keyed map only as a vehicle
for such a type and had none left, so it gives way to that table.
`aRunFailureIsRenderedInTheSelectedLocale` keeps its subject on `run.behavior.noimpl`.

What to do with the two refusals is a design question rather than a wording one — drop them, or keep
them as the internal error they describe — so it is #440 and not this PR.

### DIFF
--- a/souther-compiler/src/main/java/souther/compiler/FixtureReader.java
+++ b/souther-compiler/src/main/java/souther/compiler/FixtureReader.java
@@ -18,6 +18,7 @@ import net.unit8.raoh.Ok;
 import net.unit8.raoh.Result;
 import net.unit8.raoh.decode.Decoder;
 import net.unit8.raoh.decode.ObjectDecoders;
+import net.unit8.raoh.decode.builtin.StringDecoder;
 
 import java.math.BigDecimal;
 import java.util.ArrayDeque;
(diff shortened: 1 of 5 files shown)


kawasima:
All four points addressed. 1, 2 and 3 are in `160fcc80`; 4 is #442.

## 1. The collision was real, and it was reachable

Confirmed before fixing. The derived codec refuses the pair; the runner took the second over the
first and answered with a map holding one entry where the input wrote two.

`Runner.rekey` now checks `containsKey` before putting and reports `duplicate_key` /
`two keys are the same key once decoded` at `path.append(rawKey)` — the code, the wording and the
path the generated helper uses, so the two refuse the same input the same way.

## 2. The weak probe was hiding a live bug, not just a mutant

You were right that `Map.size(v)` answers whichever thing the reader hands over. The map rows now
look their key up, written as the key type. That immediately failed — on the fixture path.

`FixtureReader.decoderFor` rekeyed a `Type.Ref` key and returned the values decoder untouched for
anything else, so a `Map<Date, Int>` fixture reached the behavior still keyed by the strings the
decoded map carried. `Map.size` was 1 and `Map.get(Date("2026-01-01"), v)` was `None`, and no test
had ever asked. So the third reader had the same defect, and only the stronger probe could see it.

It now rekeys every kind through a key decoder of the same shape as the other two, and refuses a
collision for the same reason.

The enumeration row reads its keys rather than looking one up. `Map.get(Won, v)` types the map as
`Map<Won, Int>` (E1317), and a `let theOutcome : Outcome = Won` does not help — a `let` value is
substituted where it is named, so it types as the case again. The row filters `Map.keys(v)` on
`k == Won`, which is a sum-vs-case comparison, and a leaked String key fails it at run time.

## 3. Split

`everyTypeThatCrossesIsOneRunDrives` is gone. `aParameterTypeTheRunnerCannotDecodeIsRefusedBeforeItIsAsked`
and `anOutputTypeTheRunnerCannotEncodeIsRefusedBeforeItIsAsked` state the two claims over the two
lists, which are not the same list — an anonymous union is a parameter the checker refuses (E1312)
and an output it admits, and the output row now pins that it is driven, discriminator and all.

## 4. #442

Filed rather than fixed, as you suggested. The boolean is the part that cannot be made exhaustive:
`Runner.keyDecoder` and `FixtureReader.keyDecoderFor` each write the three primitives out, and a
sixth kind added to `isBoundaryMapKey` would leave both agreement tables green because their rows
are hand-written too. A closed `BoundaryMapKey` the readers switch on is what would fail to compile
instead. I have not claimed in the tests that the table removes the cause — its javadoc says the
rows are a claim about the readers, not about the boundary.

## Mutants

Your three, plus one for the fixture-path fix:

| mutant | killed by |
|---|---|
| skip the rekey entirely | `Map<Date>`, `Map<DateTime>`, `Map<Key>` typed lookups, the enumeration row (ClassCastException), canonical String, bad key, collision — 7 |
| skip it for a `String` key only | canonical String, collision — 2 |
| drop the collision check | collision — 1 |
| fixture builder rekeys named keys only | `Map<Date>`, `Map<DateTime>` via the fixture path — 2 |

`mvn clean install` on the rebased branch: BUILD SUCCESS, 3678 tests, 0 failures.

Issue #455, opened 2026-08-07
A comment keeps its owner and its relation to it


kawasima:
Part of #451.

A comment written at the end of a line comes back as the description of whatever is declared next:

```souther
data Inbound = WebForm | PhoneInquiry      // three units; nothing to attribute
data LeadSource = Inbound
```

becomes

```souther
data Inbound = WebForm | PhoneInquiry

// three units; nothing to attribute
data LeadSource = Inbound
```

The comment described `Inbound`. It now describes `LeadSource`, and a reader has no way to tell it once said something else. `souther-bench`'s corpus has thirteen comments written this way, seven of them in `crm.sou`.

## The provenance is in the tree

Whitespace is kept as `WHITESPACE` tokens, and the one immediately before a comment says which kind it is:

```
      WHITESPACE @30 "\n    "
      LINE_COMMENT @35 "// full line above b"
...
      WHITESPACE @68 "   "
      LINE_COMMENT @71 "// trailing on b"
```

Nothing in the grammar or the parser has to change. What is missing is a classifier that reads it and a model with somewhere to put the answer.

Three collectors convert a position into "the leading comments of the next node" today, and each has to be replaced by a lookup: `leadingDeep` (`:1023`) via `file()` (`:161`), which is the path the example above takes; `scanCommentsBefore` (`:1050`), which is the path a record field takes; and the `pending` list in `dataDef`'s `SUM_BODY` loop (`:350`).

## What is preserved is the owner and the relation

```
(owner, LEADING | TRAILING)
```

and not line or column. The formatter re-derives the layout, so a trailing comment's position in the output is a derived value: the owner of

```souther
data D = { a: Int, b: Int }   // c
```

ends on the `}` line after formatting, wherever it ended before. The canonical rule is that a trailing comment goes at the end of its owner's last content line — content line rather than last line, so that an owner whose document ends with a break still takes the line its text ends on.

## A trailing comment constrains the layout

A group holding a trailing comment cannot be printed flat: a `//` on a collapsed line swallows everything after it. So the decision is not width alone — whether a group may flatten has to account for the comments inside it. An attachment rule that does not say what it forbids the layout to do leaves the output underdetermined, which is the half of the canonical form #444 would otherwise have to infer from whatever the implementation happens to do.

## Testing

[...]

Issue #454, opened 2026-08-07
Comment consumption is complete, and checked


kawasima:
Part of #451.

Formatting `souther-bench`'s corpus drops comments: `crm.sou` 232 → 231 comment lines, `quoting.sou` 114 → 105, `pipeline.sou` 238 → 236. Nothing reports it.

Two ways a comment goes missing. `scanCommentsBefore` (`Formatter.java:1050`) flushes its `pending` run when it meets the next `SyntaxNode`, so a run after a parent's last member is never flushed — the last field of a product body, the last arm of a match, the last entry of an `exposing` list, the last statement of a block. `trailingComments` (`:1066`) covers that case for the file root and for nothing below it. Separately, an `invariant` clause (`:334`) and a `constructs` / `depends on` clause (`:411`) never ask for comments at all; eight of the nine lines `quoting.sou` loses are one comment block above an `invariant`.

`CommentSurvivalTest` lists the places the formatter is known to handle and passes. That list is the formatter's own model written down a second time, so it cannot find a place the model omits.

## Discovery in one pass, completeness by postcondition

Deciding every comment's owner in one walk over the CST removes the three collectors that disagree today. It does not make discovery complete: a construct that never looks the result up still drops its comments, which is how the two clauses above lose theirs. Restructuring reduces the number of places a comment can be lost; a postcondition is what closes it.

`Formatter` already keeps `written`, the offsets it has emitted. Comparing that set against every `LINE_COMMENT` offset in the tree at the end of `format` turns a silent loss into a failure, and comparing the sets rather than their sizes also catches a comment consumed twice or an offset registered that no comment has. `LINE_COMMENT` is the only comment kind in the grammar, so the check covers all of them.

Two emit sites bypass `written` and have to carry their tokens before the check can be turned on: `trailingComments` (`:1066`) returns `List<String>` and discards the offsets, and `dataDef`'s `SUM_BODY` loop (`:350`) emits its own `pending` list without registering anything.

The set records what was consumed, not what reached the output — a site could register an offset and forget to emit the text. So it is the inner half of a pair, and the outer half is a survival test over the source: insert a probe comment at each whitespace boundary of a fixture, keep the variants that still parse, and require `// probe` to appear exactly once in the formatted output. That needs no oracle and replaces the enumerated list. `written` should be named for what it holds — the comments consumed — so that "registered but never emitted" stays a describable bug.

kawasima:
Done except the postcondition, which moves to #455.

[...]

Issue #451, opened 2026-08-07
Formatter comment handling is neither lossless nor provenance-preserving


kawasima:
This was filed as one defect: a comment written at the end of a line comes back attached to whatever is declared next. Reproducing it turned up two more things, and together they say the formatter's model of a comment is short in three separate ways rather than wrong in one.

A comment is trivia to the language. It is not trivia to a pretty-printer, which owes it three things: it must survive the transformation, it must keep the construct it was written about, and it constrains the layout the printer is free to choose. The formatter has no model for any of the three.

## Comments are lost, not only moved

Formatting `souther-bench`'s corpus drops comments:

| file | comment lines before | after |
| --- | --- | --- |
| `crm.sou` | 232 | 231 |
| `quoting.sou` | 114 | 105 |
| `pipeline.sou` | 238 | 236 |

Two causes, neither of them about trailing comments.

`scanCommentsBefore` (`Formatter.java:1050`) collects `LINE_COMMENT` tokens into `pending` and flushes them when it meets the next `SyntaxNode`. A run of comments after a parent's last member never meets one, so it is dropped. This is what takes `, reach: ContactPoint // not a pair of optionals` out of `crm.sou`, and it is not specific to that construct: the last field of a product body, the last arm of a match, the last entry of an `exposing` list and the last statement of a block all lose it. `trailingComments` (`:1066`) is the only place that handles a run with no following node, and it is written for the file root alone.

An `invariant` clause (`:334`) and a `constructs` / `depends on` clause (`:411`) never ask for comments at all. An ordinary full-line comment written above either disappears; eight of the nine comment lines `quoting.sou` loses are one such block above an `invariant`.

`CommentSurvivalTest` enumerates "the nine places a comment can go" and passes, because those two clauses and the end-of-parent position are not among the nine. A test that lists where the formatter is known to handle comments cannot find a place where it does not.

## The comment is misattached in three collectors, not one

The original report named `scanCommentsBefore` as the origin. That is where the record-field example comes from, but not the first one. For

```souther
data Inbound = WebForm | PhoneInquiry      // three units; nothing to attribute
data LeadSource = Inbound
```

the parser puts the comment inside the *following* declaration:

```
DATA_DEF @39                    <- data Inbound = WebForm | PhoneInquiry
  ...
DATA_DEF @77                    <- data LeadSource
  WHITESPACE @77 "      "
  LINE_COMMENT @83 "// three units; nothing to attribute"
  WHITESPACE @119 "\n"
[...]

Issue #444, opened 2026-08-07
A canonical form is computed, and a deviation from it is not


kawasima:
`souther fmt --check` shows the difference between a file and its canonical form (#430, PR #443). It still cannot say why any line of that difference is there.

Two earlier framings of this were wrong in the same direction. The first put it down to lost provenance — `Doc` carries no node reference, `Formatter` keeps no record of a `Mode` it picked, nothing survives `render`. The second, after the boundary sweep (#468) and the spacing rule (#476), put it down to a missing rule label: give each boundary the row that answered it and a witness reads it off.

Both treat the canonicalization rules as things that exist and are merely unrecorded. They do not exist. The formatter has mechanisms that produce canonical text and no first-class model of the decisions those mechanisms implement, and that is what a witness needs — the decisions, at their own units, applied to arbitrary source forms.

## What the repository actually deviates by

Every `.sou` file in the repository, source tokens against canonical tokens, one comparison per adjacency of the two: 16 files, 15 of them not in canonical form, 17727 adjacencies, 2162 differences.

Split by what the document says about the boundary the difference sits at:

| policy at the boundary | differences | distinct nodes | what is in hand |
| --- | --- | --- | --- |
| `MAY` | 1228 | 494 | neither the verdict nor the measurement; `Doc.render` computes both into locals and returns a string |
| `NEVER` | 517 | 372 | the answer; the rule for the 336 that are spacing |
| `ALWAYS` | 417 | 178 | the answer; not which of the four structural obligations |

`TokenDoc.Break` is chosen at 41 places in `Formatter` — 26 `HARD_GAP`, 14 `SOFT_GAP`, one `MUST_BREAK`, with `GAP` the default at 126 more. Three values are a policy, not a rule. #468 established that the forced sites answer to four structural obligations and two comment ones, and that mapping exists as four test method names in `AForcedBreakIsWrittenWhateverTheWidthTest`. Nothing at runtime can say which obligation a `HARD_GAP` discharges.

## A rule written to generate does not answer as a predicate

`Spacing.between(construct, left, right)` is the one rule that became a function. Its domain is the boundaries the canonical form writes on one line, because those are the boundaries the formatter asks about. A witness asks about the adjacencies a source can write on one line, and that set is strictly larger.

[...]

PR #457, opened 2026-08-07
Formatter comment handling is lossless and provenance-preserving

### LINKED ISSUE #453: A prefix-bearing item and its comment are one layout unit
Part of #451.

A comment written above a member of a leading-comma block comes back below the comma, and the member drops to the nesting indent:

```souther
data D =
    { a: Int
    // the comment
    , b: Int
    }
```

becomes

```souther
data D =
    { a: Int
    , // the comment
    b: Int
    }
```

This is not about trailing comments — the input above has the comment on its own line, where it already belongs to `b`, and the output still puts it after the comma. A sum body does the same with its connector: `data S = A` / `// c` / `| B` prints `| // c` and then `B`.

`withComments` (`Formatter.java:1000`) expands a member into `comment + HARDLINE + member`. `productBody` (`:390`) writes the leading comma before that document and `chained` (`:135`) writes its connector before the segment, so `separator(member)` is really `separator(comment + member)` and the `HARDLINE` puts what follows at the nesting indent rather than after the separator.

The layout's unit is a line — `, b: Int`, `| Alternative` — and a comment decorates that whole unit. Decorating the member while the parent supplies the prefix cannot express it. A rendering item that carries its own prefix, and is decorated as a whole, can:

```
item(prefix = ",", body = field)
```

The prefix is a layout property of the item, not a claim that the comma belongs to the field grammatically. Both `productBody` and `chained` write prefix-bearing items, so both are served by one shape, and a third construct written the same way later gets it for free.

Independent of the rest of #451 and worth doing first: the other parts change which comment lands where, and their output cannot be judged while the layout around a comment is still wrong.

kawasima:
Closes #453, #454, #455, and with them #451.

A comment is trivia to the language and not to a pretty-printer, which owes it three things: it has to survive the transformation, it has to keep the construct it was written about, and it constrains the layout the printer is otherwise free to choose. #451 was filed as a defect in the second. Reproducing it found the formatter had no model for any of the three.

## What was wrong

Formatting `souther-bench`'s corpus dropped comments — `crm.sou` one, `quoting.sou` nine, `pipeline.sou` two — and nothing reported it. That was two separate holes: a run of comments after a parent's last member was handed to nobody, and an `invariant` clause, a `constructs` / `depends on` clause and a `fake` row never asked for comments at all.

A comment written at the end of a line came back describing whatever was declared next. Three collectors turned every position into "the leading comments of the next node", and the one the issue named turned out to handle the record-field example but not the headline one.

A comment above a member of a leading-comma block came back below the comma, with the member itself at the nesting indent. That one had nothing to do with trailing comments: an ordinary full-line comment above a field did it too.

Every symptom was a fixed point. Formatting the output reproduced it, so `fmt -w` wrote the wrong form and `fmt --check` demanded it, and the idempotence the formatter is tested for held throughout.

## What it does now

**A comment decorates a whole line, opener included** (#453). A member of a leading-comma block or a union is written on a line the block opens for it, so the opener is part of what the comments decorate. `separated` is the only place that has to know which side the punctuation goes on.

**Discovery is exhaustive, and counted** (#454). Comments are found in one pass and consumed in one place. The run after a parent's last member is kept apart from the runs a member is under, and the clauses that never asked now ask.

**A comment keeps which construct it is about** (#455). Whether a comment was written above a line or at the end of one is already in the tree — the whitespace before a full-line comment carries the newline that ended the line above it, and the whitespace before a trailing one does not. The parser was not losing the provenance; comment interpretation was discarding it. Nothing in the grammar or the parser changed.

What is preserved is `(owner, relation)` and not line and column. Where a trailing comment goes is derived: it is written at the end of the line its construct ends on *after* formatting, which is not the line it ended on before. Its width is measured against nothing, since nothing can share the line after it — which is also why a group holding one is never laid out flat. `Doc.Trailing` carries both facts.

The owner is the outermost construct ending where the code before the comment ends: `data D = A | B // c` is about the declaration and `, f: T // c` about the field, which is one rule. A comma is nobody's, so a comment after one is put against the member the comma closed. Where the owner has no line of its own — a comment after the condition of an `if` — it moves up to the nearest construct that does, and is written at the end of that line. It travels further than it was written. Writing it where it was written would put the rest of that line inside it.

## Measurements

A probe comment is written into every boundary between two of a fixture's tokens — not only where whitespace already was — and required back from every variant that still parses. Not every position the grammar admits one: what is quantified over is the boundaries of the fixtures, and a form no fixture is written in is a form nothing here asks about. A qualified name in a `constructs` clause was such a form, and the formatter refused it, so it is a fixture now.

Four things are required of each variant, and each was added because something got past the ones before it.

- the comments come back, as comments and not as text: one written into another one's line is still in the output and is no longer a comment
- the tokens that are not trivia come back unchanged: a comment that swallowed the code after it looked like a comment that survived, and that is how a pipeline's declared output was being deleted
- the output parses
- what each comment is next to, as an index into those tokens, is either unchanged or one of the boundaries that has no line to stay on

Of 404 boundaries, none loses a comment, changes the code or fails to parse, and none of the comments that had a line to stay on left it.

Which of them had one is read from the canonical form rather than from the formatter: a token that already ends a line is a place a comment can stay, and one in the middle of a line is a place it cannot, since the rest of that line would be written inside it. A comment written there goes to the nearest construct the layout gives a line to — further than it was written, and the alternative is writing code into a comment. That check replaced a count of how many travelled, which said nothing about which: a comment that started staying and one that stopped cancel out. It found thirteen reassignments the count had been hiding, all of them the same thing — a construct written over several lines has more lines than its last, and only the last was a line anything could be attached to.

Against `develop` the corpus loses twelve comment lines and now loses none: `crm.sou` 232 in and 232 out, `quoting.sou` 114, `pipeline.sou` 238. That number is the small one — the corpus writes trailing comments only where a construct already fits on one line.

`format` also compares the comments the tree holds against the ones some construct consumed, and refuses rather than returning a file that came back shorter. That is the inner half and the sweep is the outer: a construct can take a comment and write nothing, which is how an empty `exposing` clause lost one with the count green.

Five rounds of review found defects, and each round is as much a change to the sweep as to the formatter. What survival cannot ask is which member a comment is about — moved from one member to the next it is one comment in and one out with the code unchanged — which is why that question is now counted rather than enumerated, and why the cases written out by hand are written by how a member is rendered rather than by which construct holds it.

## Relation to #444

The canonical form changed, so #444 can be written from this output rather than from the previous one. Two of its rules did not exist before in any form: a trailing comment goes at the end of its owner's last content line, and a construct holding one is never laid out flat.

### DIFF
--- a/souther-compiler/src/test/java/souther/compiler/fmt/ACommentAtTheEndOfALineStaysOnItTest.java
+++ b/souther-compiler/src/test/java/souther/compiler/fmt/ACommentAtTheEndOfALineStaysOnItTest.java
@@ -81,6 +81,106 @@ class ACommentAtTheEndOfALineStaysOnItTest {
                 """, below);
     }
 
+    /** A sum's cases are identifiers rather than nodes, so a case cannot be named the way a field
+     * can. The comment is held against where the identifier is instead. The last case is the
+     * exception: the declaration ends where it does, and what ends on a line is what the line was
+     * about. */
+    @Test
+    void onTheSumCaseItWasWrittenAfter() {
+        String formatted = Formatter.format("""
+                module m
+                data A
+                data B
+                data C
+                data S = A   // about A
+                    | B
+                    | C
+                """);
+
+        assertEquals("""
+                module m
+
+                data A
+
+                data B
+
+                data C
+
+                data S = A // about A
+                    | B
+                    | C
+                """, formatted);
+    }
+
+    /** A comment above an attempt's departure, and above a pipeline's stage. Both are written one to
+     * a line, and neither was asked for its comments. */
+    @Test
+    void aboveADepartureAndAboveAStage() {
+        String departures = Formatter.format("""
+                module m
+                data Amount = Int
+                    invariant value > 0
+                data O = { n: Int }
+                behavior f : (n: Int) -> O | Amount constructs O, Amount
+                let f (n) = {
+                    guard Amount(n) as a else
+                        // why this arm
+                        | value -> Amount(1)
+                    O { n = a.value }
+                }
+                """);
+        String stages = Formatter.format("""
+                module m exposing ( p )
+                data A = { n: Int }
+                data B = { n: Int }
+                behavior one : (a: A) -> B constructs B
+                behavior p =
+                    one
+                    // why this stage
+                    >-> one
+                """);
+
+        assertEquals("""
+                module m
+
+                data Amount = Int
+                    invariant value > 0
+
+                data O =
+                    { n: Int
+                    }
+
+                behavior f : (n: Int) -> O | Amount
+                    constructs O, Amount
+
+                let f (n) = {
+                    guard Amount(n) as a else
+                        // why this arm
+                        | value -> Amount(1)
+                    O { n = a.value }
+                }
+                """, departures);
+        assertEquals("""
+                module m exposing ( p )
+
+                data A =
+                    { n: Int
+                    }
+
+                data B =
+                    { n: Int
+                    }
+
+                behavior one : (a: A) -> B
+                    constructs B
+
+                behavior p =
+                    one
+                    // why this stage
+                    >-> one
+                """, stages);
+    }
+
     @Test
     void onTheFieldItWasWrittenAfter() {
         String formatted = Formatter.format("""
--- a/souther-compiler/src/test/java/souther/compiler/fmt/ACommentAtTheEndOfALineStaysOnItTest.java
+++ b/souther-compiler/src/test/java/souther/compiler/fmt/ACommentAtTheEndOfALineStaysOnItTest.java
@@ -38,6 +38,49 @@ class ACommentAtTheEndOfALineStaysOnItTest {
                 """, formatted);
     }
 
+    /**
+     * The same words in the same place in the child list, told apart by the whitespace in front of
+     * them: the one below carries the newline that ended the line above it and the one beside does
+     * not. Written as a pair because either half alone is satisfied by a formatter that treats every
+     * comment the same way.
+     */
+    @Test
+    void whichOfTheTwoItIsComesFromTheWhitespaceBeforeIt() {
+        String beside = Formatter.format("""
+                module m
+                data D =
+                    { a: Int   // the comment
+                    , b: Int
+                    }
+                """);
+        String below = Formatter.format("""
+                module m
+                data D =
+                    { a: Int
+                    // the comment
+                    , b: Int
+                    }
+                """);
+
+        assertEquals("""
+                module m
+
+                data D =
+                    { a: Int // the comment
+                    , b: Int
+                    }
+                """, beside);
+        assertEquals("""
+                module m
+
+                data D =
+                    { a: Int
+                    // the comment
+                    , b: Int
+                    }
+                """, below);
+    }
+
     @Test
     void onTheFieldItWasWrittenAfter() {
         String formatted = Formatter.format("""
--- a/souther-compiler/src/test/java/souther/compiler/fmt/ACommentAtTheEndOfALineStaysOnItTest.java
+++ b/souther-compiler/src/test/java/souther/compiler/fmt/ACommentAtTheEndOfALineStaysOnItTest.java
@@ -38,6 +38,109 @@ class ACommentAtTheEndOfALineStaysOnItTest {
                 """, formatted);
     }
 
+    @Test
+    void onTheFieldItWasWrittenAfter() {
+        String formatted = Formatter.format("""
+                module m
+                data AccountId = String
+                data DomainName = String
+                data Industry
+                data Account =
+                    { id: AccountId
+                    , domain: DomainName?         // an honest optional
+                    , industry: Industry
+                    }
+                """);
+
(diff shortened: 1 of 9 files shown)


kawasima:
Reviewed through `fcbd1e5`. The fifteenth commit fixes the over-generalisation that made `ARROW`/`COLON` separators unconditionally, so that one is settled. #453 and #454 are in good shape. #455 is not: the contract this PR states is not the one the implementation holds, and the gap between them is still losing comments.

## The owner is not kept; it is projected onto a writable slot

The body defines what is preserved as `(owner, relation)`. `place` computes an owner and then stores the projection:

```java
private static void place(Attachments out, SyntaxNode owner, SyntaxToken comment, boolean above) {
    SyntaxNode slot = slotOf(owner);
    if (slot != null) {
        add(above ? out.above() : out.after(), slot, comment);
    }
}
```

Every map in `Attachments` is keyed by a slot node or by a token offset. Nothing holds the owner, so after this call it is gone.

That is not only an internal-representation point — a test fixes the counterexample. For

```souther
let value (n: Int) = if n > 0   // about the test
    then One else Two
```

the expected output is

```souther
let value (n: Int) = if n > 0 then One else Two // about the test
```

In the source the comment's owner is `n > 0`. Re-parse the output and its owner is the whole declaration. So

```
owner(c, format(x)) ≠ owner(c, x)
```

The "the owner did not change, only where there was room to write it" reading does not hold for a source-to-source formatter: the output has no channel other than placement in which to record an owner, so placement *is* the provenance the next parse reads.

Two ways out. Weaken #455 — a comment keeps its owner where it can and is reattached to the nearest writable construct where it cannot. Or keep #455 and let a comment demand a break, giving its owner a line of its own:

```souther
let value (n: Int) =
    if n > 0 // about the test
        then One
        else Two
```

This PR already accepts that a trailing comment stops its group going flat. Carrying that one step further — attachment creates a break opportunity, rather than attachment being redirected to wherever a break already is — is what agrees with `(owner, relation)`.

## The projection is many-to-one, and that loses comments

Distinct owners can share a slot, and `afterOf` emits everything that lands on one with nothing between:

```java
for (SyntaxToken c : comments.after().getOrDefault(n, List.of())) {
    if (consumedComments.add(c.start())) {
        parts.add(Doc.trailing(c.text().stripTrailing()));
    }
}
```

`Trailing` renders as `' ' + text`, so two of them run together on one line. Measured on this branch:

```souther
module m
data One
data Two
let value (n: Int) = if n > 0 // c1
    then One // c2
    else Two
```

```souther
let value (n: Int) = if n > 0 then One else Two // c1 // c2
```

```
source comments    = [// c1, // c2]
formatted comments = [// c1 // c2]
idempotent         = true
```

Two comment tokens in, one out: `// c2` is not misplaced, it is inside `// c1`'s text. And the result is a fixed point, so `fmt -w` writes it once, `fmt --check` never mentions it again, and the loss is silent. This is a direct counterexample to the losslessness this PR claims.

Neither guard catches it. `consumedComments` is checked before rendering and both comments are consumed by `afterOf`, so it passes. The sweep inserts one probe per variant:

```java
out.add(source.substring(0, at) + written + source.substring(at));
```

so no two-comment interaction is in its search space. Formatting the whole `souther-bench` and stdlib corpus loses nothing (16 files, comment counts equal before and after) — the corpus writes its comments where each has a line of its own, which is why five rounds of review did not reach this.

Adding one test closes this instance, but the shape is the cause: with

```
owner₁ ─┐
        ├─> slot
owner₂ ─┘
```

a collision policy is required, and there is none.

## What the sweep proves is not what #455 states

The sweep is much stronger than a comment count — tokens survive, non-trivia tokens are unchanged, the output parses, and a neighbour is unchanged where the canonical form already has a line. But its oracle is

```java
if (was != now && ... && mustStay.get(was)) {
```

which permits an owner with no canonical line to travel, and `aPartOfAnExpressionWithNoLineOfItsOwn` writes that permission down as an expected value. So what the tests establish is

```
a comment survives, and stays with its owner when that owner already has a writable line,
and may otherwise travel to a writable ancestor
```

not `(owner, relation) is preserved`. That is a contract mismatch rather than missing coverage.

## What is good

`Member(doc, trailing)` concentrating the separator/trailing order in one place is right, and `Segment`/`head` now asking the constructor what a segment stands for closes the path where a bare `Doc` was passed and its comment forgotten. Deriving `holdsLines` and `takesALineOf` from the single `linesOf` table matters more than it looks: the same layout fact living in both the attachment side and the rendering side had already produced a real bug.

## Order to fix in

1. Hold `owner` and `placement slot` as separate fields rather than one projected value.
2. Decide, for an owner with no line: reattach to an ancestor, or break the layout so the owner gets one.
3. If the first, weaken #455 and this PR's title and body to match. If the second, change the layout builder.
4. Add the regression for two comments with different owners reaching one slot.
5. Add a round-trip test: `(owner, relation)` is the same after `format(parse(format(x)))`.

My preference is to keep #455 and make the layout follow the attachment. Nearest-writable-ancestor is the cheaper change, but it discards, for the convenience of placement, exactly the provenance this PR set out to keep.

PR #459, opened 2026-08-07
A boundary is a value the record can hold

### LINKED ISSUE #427: A boundary read off a field alone asks for a value the record refuses
A boundary is derived from the invariant on the field's own type and from nothing else, so it can
name a value the record the field sits in refuses to hold.

```
data MinuteOfDay = Int
    invariant withinDay = value >= 0 && value <= 1440

data WorkInterval =
    { date: Date
    , startsAt: MinuteOfDay
    , endsAt: MinuteOfDay
    }
    invariant endsAfterStart = startsAt < endsAt
```

```
$ souther examples timesheet.sou
    boundary    4/6
      · no row is at classifyInterval/interval.startsAt = 1440 (invariant MinuteOfDay (max))
      · no row is at classifyInterval/interval.endsAt = 0 (invariant MinuteOfDay (min))
```

`startsAt = 1440` needs an `endsAt` above 1440, which `withinDay` refuses. `endsAt = 0` needs a
`startsAt` below zero, same. Neither row can be written, and the boundary that would make them
writable is three lines below the field the measure read.

The construction agrees:

```
$ souther run timesheet.sou --behavior classifyDay --input '[…,[{"date":"…","startsAt":1440,"endsAt":1440}]]'
input #2 could not be decoded — /0: invariant violated on timesheet.WorkInterval: endsAfterStart
```

[#example-partition] already reasons about intersecting a threshold with what the type admits. What
it does not do is carry a sibling field's invariant into that intersection, so a rule relating two
fields is invisible to the derivation that reads either one.

## Related

Same consequence as #426 from a different cause: the report asks for rows no model can write, and
nothing in the output separates those from the gaps that are real. In the model above, four of the
fourteen reported gaps were unsatisfiable — two from here, two from #426 — and the other ten were
worth closing. Four of the ten found something the author had not tested, including an entire
untested half of a late-night rule.

kawasima:
Closes #427.

A boundary was derived from the invariant on the field's own type and from
nothing else, so it could name a value the record the field sits in refuses to
hold. Both ends of a work interval are minutes of a day and both were offered
1440, though a `startsAt` of 1440 wants an `endsAt` past the end of the day.
Neither row can be written, the decoder says so, and nothing in the report
separated those gaps from the ones worth closing.

## What a position admits

A field's type says which values exist of that type. What the position admits is
that read under the rules of the record holding it, which is the record's own
invariant projected onto this field. `startsAt` stops at 1439 and `endsAt`
starts at 1; both are values a row can carry, and both are edges worth carrying
one at.

Only taken in, never put there: a position its own type left open stays one the
model draws no line through, because a rule between two fields is not a
partition of one of them.

## No second reasoner

The projection is a question the compiler could already answer. `NumericDomain`
holds per-atom intervals and difference-bound constraints and closes the
differences transitively, which is exactly the projection of an ordering onto
one of its ends. It sat inside the invariant-discharge check as that check's
own; it is now a facility of its own, and the seeding a parameter of a record
type already gets is read for its numbers instead of for what it discharges.

Three things had to be true first, and each is its own commit:

- the domain had to know whether an atom's values are whole numbers, because
  `a < b` is `a <= b - 1` over the integers and nothing tighter than `a <= b`
  over decimals — it took the weaker reading for both;
- the discharge check is fail-open, so an analysis that fell over and one that
  found nothing arrive the same way; made to throw on every atom it looks up,
  the analysis took two tests in the whole suite down with it, so it now says
  whether it finished and a test can hold it to that;
- the reasoner had no tests of its own. It has them now, and they were written
  before the arithmetic changed, so the commit that sharpens it shows exactly
  which answers moved.

## What changed for a compiling program

No diagnostic moves. Sharpening a bound on a whole number makes the discharge check prove more, so
E2011 could fall and E2010 could rise; on `souther-examples` the sharpening fires 110 times over that
corpus and every site comes out where it was.

```
E2011  19 -> 19
E2010   0 ->  0
```

Reading an imported declaration's invariant at all is what moved E2011 from 13 to 19, and that is
#466, which is in the base this now sits on. What is kept here is the boundary-side cover for it: a
relation over an imported type narrows the same either side of a module boundary, including where the
reader declares that spelling itself.

What `souther examples` reports does move, and none of it changes a verdict:

```
boundaries counted   181 -> 138
said and not counted  82      (hr and crm — records carrying a rule outside the fragment)
adequacy verdicts     every module unchanged
```

## Also in

An edge the record moved names the record beside the rule that put it there —
`invariant MinuteOfDay (max) within WorkInterval` — as one obligation and not
two, since neither rule wants a row of its own.

The generator fills the rest of a record from what is left of it once the fixed
position is settled. Beside a `startsAt` of 1439 the only `endsAt` the record
admits is 1440; it used to take the bottom of the type's range and report the
boundary as one every value tried was refused at. On the model the issue
reports, all four boundaries now have a row offered at them, where before two
were unwritable and three could not be filled.

ADR-0090 is revised in place. It stated that a threshold is intersected with
what the type admits, which is the sentence this corrects; a second ADR would
leave two statements of one rule.


kawasima:
Reproduced all four on `6b7c949`. The Int direct-sibling case is fixed; the contract this states is not.

**A strict rule over decimals.** The same shape as #427, one type away.

```souther
data Ratio = Decimal
    invariant withinOne = value >= 0.0m && value <= 1.0m
data Band = { low: Ratio, high: Ratio }
    invariant ordered = low < high
```

```
    boundary    0/4
      · no row is at classify/band.low = 1 (invariant Ratio (max))
      · no row is at classify/band.high = 0 (invariant Ratio (min))
```

`low = 1` needs a `high` above 1 and `high = 0` needs a `low` below 0. Both are refused, and neither
end says `within Band`, because a strict difference over dense atoms is recorded as the non-strict one
and nothing narrows. The true projection is `[0, 1)` and `(0, 1]`, which a closed `Bounds` cannot hold.

**A rule that skips a value.** `assume` drops `Rel.NE` at the door and records nothing.

```souther
data R = { a: N } invariant nonzero = a.value /= 0
```

```
      · no row is at classify/r.a = 0 (invariant N (min))
```

**A record's rule one level up.** The context is rebuilt per declaration, so a clause on the outer
record is gone by the time the walk reaches the inner one.

```souther
data Interval = { startsAt: Minute, endsAt: Minute }
data Input = { interval: Interval, cap: Minute }
    invariant capped = interval.startsAt < cap
```

```
      · no row is at classify/input.interval.startsAt = 1440 (invariant Minute (max))
      · no row is at classify/input.cap = 1 (invariant Minute (min) within Input)
```

`cap` is a direct field and is narrowed; `interval.startsAt` is one level down and keeps 1440, though
`cap <= 1440` puts it at 1439. `Seeded.atoms()` names direct fields only, and `Partitions.walk` builds
a fresh `FieldDomains` for `Interval` that never saw `Input`'s clause.

**A record no value satisfies.** `boundsOf` answers a bottom domain with `Bounds(null, null)`,
`FieldDomains` reads that as *nothing known* and drops the field, and the derivation falls back to the
type's own edges — so an uninhabited record is asked for rows at every one of them.

---

`exact()` is the thread through all of it and nothing in production reads it. Worse, it would not
catch the first two: `kept.isEmpty()` is true for a weakened strict bound and for a dropped `NE`, so
the answer is `EXACT` where the projection has already lost something.

Moving to draft. What has to change, in the order it has to change:

1. what a projection lost is recorded where it is lost — a strict bound weakened on a dense atom, an
   `NE` dropped, a form kept as written — rather than inferred from `kept` afterwards;
2. the domain is seeded once per parameter and queried by full path, so a clause on an enclosing
   record still reaches a field two levels down, and `bottom` is an answer of its own rather than the
   absence of one;
3. a consumer acts on it. A boundary derived from a projection that lost something is not known to be
   writable, and asking for a row at it is the whole of what #427 reports. The shape ADR-0091 already
   gives combinations fits: reported, not counted, and not something `--strict` refuses over.

The first two are mechanical. The third is where the third state finally has to exist, and it is worth
saying that the alternative — falling back to the type's own edge when the projection is inexact — is
not sound either: the clause that could not be read can refuse that edge just as well.

kawasima:
All three are in, plus the two the automated review found. Ready again at `588c331`.

**What a projection lost is recorded where it is lost.** Three losses against the atoms the rule
named: a strict bound weakened on a dense atom, a disequality dropped, a form kept as written.
`kept.isEmpty()` saw none of the first two, so `low < high` over decimals and `a /= 0` both came back
as the whole story.

**One reading per parameter, asked by path.** A clause on an outer record now reaches a field of a
field — `interval.startsAt` under `interval.startsAt < cap` stops at 1439 — and the generator fills it
beside a `cap` of 1440. `bottom` is a third answer: a value whose rules contradict has no positions to
cover rather than positions nothing bounds.

**An edge nothing promises is said and not counted.** Reported, out of the denominator, and not
something `--strict` refuses over.

Two more, from the review:

- a `guard` drawn outside what the record admits was dropped from the intervals and kept in the cuts,
  so a line nothing can be at was still asked for a row. One filter now, read by both;
- the class filler was never told what the boundary filler had settled, so it picked the rest of a
  record from each field's own type — #427 in the other of the two row-generating paths.

## What the criterion cost, measured

Per *value* it was unusable: 159 of 181 corpus boundaries demoted, including `a.employees = 1`, which
is writable and was a good obligation. Per *position* with the loss attributed to atoms it is 81 — and
81 was still wrong. Every one of those was an imported newtype: a second classifier was reading the
discharge form of the type, which for another module's declaration has already been settled into
folds, so every imported bound looked like a rule nothing could read. Classified by the reader that
produced the bound, it is 11.

```
counted   181 -> 172
demoted   11
verdicts  every module unchanged
```

The 11 are the six `Quote` positions under
`Decimal.compare(total.value, subtotal.value) <= 0`, four more Amounts under the same clause, and a
`TaxRate` whose `value < 1.0m` has no last value. Two of them are conservative rather than forced.

## Still open, and stated rather than fixed

The seeding takes nothing from a type another module declares, so an imported position keeps the
bounds its own type gives it and no clause relates it to a sibling across that boundary. A narrowing
missed, not an edge invented. Pinned by a test.

`#461` is the other one: a bound stops at the first layer a newtype wraps, so `data StartMinute =
Minute` carries no bound to either reading.

kawasima:
All three reproduced on `588c331`. Back to draft.

**A loss does not travel the closure it was derived through.**

```souther
data N = Int invariant within = value >= 0 && value <= 10
data R = { a: N, b: N }
    invariant same = a == b
    invariant notTen = b.value /= 10
```

```
    boundary    0/2
      · no row is at f/r.a = 10 (invariant N (max))
      · not known to be writable: f/r.b = 10 (invariant N (max))
// no row for `r.a = 10`: every value tried was refused at construction
```

`b` carries the disequality and is demoted; `a` reaches its bound *through* `a == b` and is not. The
bound travelled and the doubt about it did not.

**An imported type puts the relation and the range in different places.**

```
imported Amount, a < b     · no row is at f/p.a = 10 (invariant Amount (max))
                           · no row is at f/p.b = 0  (invariant Amount (min))
same module                · no row is at f/p.a = 9  (... within Pair)
                           · no row is at f/p.b = 1  (... within Pair)
```

The relation reaches the domain and the type's range does not, so nothing narrows and `Partitions`
puts back the type's own edges — 10 and 0, which the generator then refuses. Not a narrowing missed;
an edge nobody can write, which is the issue this branch is named after.

**A row sitting on a boundary is not read as the proof it is.**

```souther
data R = { a: N } invariant hole = a.value /= 5
example f | "zero" : (R { a = N(0) }) -> Ok
```

```
      · not known to be writable: f/r.a = 0 (invariant N (min))
```

There is a row at it. `met == YES` is a value that went through the decoder, and the report says the
value is not known to be writable anyway.

---

The three are one mistake, as you put it: whether a constraint reached this field's bound, and whether
this field's edge extends to a whole value, are being answered at the same locality. The first is
per-atom and travels the closure. The second is existential over the parameter and cannot be settled
by looking at which names a clause mentions.

So they become two answers:

- **the bounds are exact here** — per atom, and a loss on any atom the bound was derived through is a
  loss here. That is the closure the bound came from, so it is the closure the doubt takes;
- **the edge is writable** — parameter-wide and conservative, promoted only by a witness. A row at the
  value is one. A row the generator built at it is another, and that is what ADR-0091 already says
  construction is for.

Conservative-plus-witness is also what keeps the report useful: the demotions that matter are the ones
nothing has ever been written at, which is exactly the set worth being honest about.

The imported case is a third thing and not precision: an atom whose type bound never reached the
domain cannot take part in a projection at all, so it is `partial` rather than narrowed-by-nothing.

ADR-0090 and the spec still say the opposite of what this now does — "a rule of another shape leaves
the position where its type left it ... an edge never moved is one they can [write]". That sentence
was written before the third state existed and has to go with it.

kawasima:
All three fixed, held to the models that showed them, rebased on `develop`. Ready at `d2cc1f0`.

**A doubt travels the path its bound came along.** A loss was asked about the atom the rule named,
while `boundsOf` reads a bound through the differences to every other atom. `a == b` with `b /= 10`
now demotes both — `a`'s edge *is* `b`'s edge, carried by the equality. The whole component the atom
sits in, since a difference is closed transitively.

**A relation is a projection only where both ends brought their ranges.** A position something is
compared with, whose own range never reached the domain, is `partial`. So the imported `Pair` case
stops asking for 10 and 0.

**A row at the value is the proof the projection could not give.** `met == YES` is a value that went
through the decoder. `boundary 1/1` where a row sits on one edge and the other is unpromised, instead
of `0/0` with both called unknown.

## The root cause, run down

The imported case is not a limitation of the projection. `Clauses.typed` cannot re-elaborate an
imported declaration's invariant at all:

```
E1023: unknown identifier `value`
  | fieldTypes=[value] | fieldBindings=[value] | newtype=true
```

The scope has the field and its binding. What fails is resolution: a `Var` of a published declaration
carries the `denotes`/`reachedAs` its own module settled, and looking that up in a reader's scope
finds nothing. Every caller reads the resulting `null` as *a clause outside the fragment*, which is
the answer for a rule the language cannot express rather than for one that was not looked up.

Filed as #464 with the diagnosis. It is a defect in reading published declarations and it reaches
further than boundaries — the discharge check reads clauses the same way, so a construction of an
imported type is presumably discharged by nothing and reported by nothing. Not measured.

What is in this PR is not a workaround for it. *Not known to be writable* is the true answer while
#464 stands, it is the answer dense strict bounds and disequalities need anyway, and it stops being
reached the moment #464 is fixed — so it is not in the way.

## Normative text

ADR-0090 and the spec said an edge a rule of another shape left in place is one an author can write.
That is the sentence this PR disproves; both now say what settles such an edge is a row rather than an
argument, and how far a doubt reaches.

## Corpus, against the rebased base

```
counted   181 -> 172
demoted   11
refused   111 -> 109
verdicts  every module unchanged
```

Unchanged by this round: the corpus has no equality between bounded fields, no relational record over
an imported type, and no rows at the 11.

kawasima:
Blocker 2 is gone at the root rather than detected better. `60133d0`.

`TypeOps.fieldBindings` named the declaring type `symbols.own(data.name())` — this module's own,
whichever module actually declared it. A declaration's fields are bound once, where it is written, and
the clauses that read them carry those bindings in; a reader computing them under its own name gets a
different binding for the same field.

```
E1023: unknown identifier `value`
  | fieldTypes=[value] | fieldBindings=[value] | newtype=true
```

`Clauses.typed` answered null and every caller read null as *a clause outside the fragment*, which is
the answer for a rule the language cannot express, handed to one that was looked up in the wrong
place. The registry knows which module declares a bare name; it is asked now instead of assumed. One
line.

The imported `Pair` answers exactly as the same-module one:

```
· no row is at f/p.a = 9 (invariant Amount (max) within Pair)
· no row is at f/p.b = 1 (invariant Amount (min) within Pair)
```

So the `isRelated && boundsOf.isEmpty()` guard is deleted. You were right that it could not be made
complete — a missing range on the far side of a relation is invisible to it — and a guard that
outlives its cause is where the next defect hides.

**It reaches past boundaries.** The discharge check reads clauses the same way, so it has been silent
about every imported type's rules. On `souther-examples`:

```
E2011  13 -> 19   payroll.sou:195, socialinsurance.sou:461,
                  forecasting.sou:233, :321, :375, :492
E2010   0 ->  0
```

Six constructions of imported types that may abort at run time and had nothing said about them.
Warnings only; nothing stops compiling. Two tests that pinned the broken behaviour now assert the
fixed one, and they were the only two in the suite that noticed.

Boundaries on the corpus are unmoved: 172 counted, the same 11 demoted, every verdict the same.

Blocker 1 is still open and I have not touched it. `exact(path)` still answers the local question and
is still being read as the existential one; the split you described — `projectionExactAt` local,
`edgeKnownWritable` parameter-wide and promoted only by a witness — is next. The row witness is
already in; what it needs beside it is the generated row, since a parameter-wide condition without a
second witness source puts most of the corpus back into *not known*.

kawasima:
Both blockers and the ADR. `08be0f9`.

**A rule reaches as far as the construction it can refuse.** `allRulesRead()` was borrowing
`FIELDS_SEEDED`, which is a limit on what is worth measuring, for a question about whether a value can
be built at all. It walks its own way now: down the fields a value must have, at any depth, stopping
on a type already met; not into an optional or a collection, since a rule inside one is about a value
the construction need not make.

```souther
data Leaf = { x: N, tag: String } invariant tagged = String.matches("[a-z]+", tag)
data L2 = { leaf: Leaf }   data L1 = { l2: L2 }
data Root = { n: N, l1: L1 }
```

```
      · not known to be writable: f/root.n = 0 (invariant N (min))
      · not known to be writable: f/root.n = 10 (invariant N (max))
```

Three records down, past the old cutoff. Pinned, with its opposite beside it — the same rule behind a
`Note?` and a `List<Note>` leaves `n`'s edges promised, because a `Root` with neither is a `Root`.

**The seeding's own verdict is read.** `seedFields` answers whether it got through every clause and
`FieldDomains` was keeping the numbers and dropping that, so a projection that was never computed
could be classified clean by a second reading that walks the declarations instead of the same path.
Same shape as the fail-open silence earlier in the branch. Forcing that answer to false now fails five
assertions in the class, so it is being read.

The attribution of an unread rule to the fields it names is deleted with the locality it existed for —
one question, one answer, nothing to attribute.

**ADR-0090** said the question is asked of the position and that a pattern on a label leaves the
minutes beside it as they were. Both were true of the implementation two commits ago and are the
opposite of what it does now. It states the value-wide rule and how far it reaches.

## Corpus

```
counted   181 -> 153
demoted   67          (hr 38, crm 28, ordering 1)
verdicts  every module unchanged
E2011     13 -> 19    unchanged from the #464 commit
E2010      0 ->  0
```

The 67 are records carrying a rule outside the fragment — a call over two dates, `Decimal.compare`, a
pattern — and every edge in them is now honest about what nothing has shown.

## One thing deliberately not done

A row the generator builds at an edge is a witness as good as one already written, and would take most
of the 67 back. It is not in: `Findings` would have to ask `Generated`, so every `--adequacy` run
would build and decode values, and `--generate` is opt-in today. Worth doing as its own change with
that cost stated, not folded in here.

kawasima:
## State at `06b88dc0`

Draft, 22 commits, tree clean and pushed, 3910 tests green locally.

Two things are in the way of a merge and neither is a code change:

- **no CI run since `08be0f95`.** `.github/workflows/ci.yml` triggers on `pull_request` with no
  filter and ran on every head up to `d02420ec`. Closing and reopening the PR did not start one. An
  empty commit is the remaining nudge; a force-push would rewrite history, so it has not been done.
- **21 commits behind `develop`.** Wants a rebase before merge.

## Blockers 1 and 2 of the last round, done

**The reading a boundary comes from now reaches as far as the rules do.** Confirming a rule exists
and having read it into the domain the boundary came from were two things and only the first had the
construction's reach. `n < l1.l2.leaf.x` with `x` stopping at 10 three records down left `n` at 10;
it is 9 now. Both halves walk the required chain at any depth, stopping on a type already entered on
that path, and not into an optional or a collection. The walk over a body keeps `FIELDS_SEEDED`,
which is a cost bound rather than a claim.

**A newtype taken straight as a parameter is held to its own rules.** It answered the question at the
door and always said yes. Having no siblings is why it hands out no per-field bounds; it is not a
reason to call an edge writable, and `value /= 0` makes the bottom of `[0, 10]` a value the decoder
refuses. Writability no longer depends on whether the type is a parameter or a field of one.

Both pinned.

```
counted   181 -> 139
demoted   81         (hr and crm — records carrying a rule outside the fragment)
verdicts  every module unchanged
E2011     13 -> 19   E2010 0 -> 0
```

## Blocker 3, reproduced and not fixed

```souther
module probe.report
import probe.money as M
data Amount = String                                    // shadows M.Amount
data Pair = { a: M.Amount, b: M.Amount } invariant ordered = a < b
```

```
· no row is at f/p.a = 10 (invariant Amount (max))      // unshadowed: 9 within Pair
```

`TypeOps.fieldBindings` still guesses the declaring type from the bare name, so a local declaration of
the same name wins and #464 comes back. Your framing is the fix: do not go back to a bare name when
the caller already holds the resolved one. `fieldBindings(TypeName declared, Ast.Data, Symbols)`
threaded through `Clauses.bindingsOf`, `Clauses.typed`, `DataChecker.fieldScope`, `capabilityOf` and
three sites inside `InvariantChecker`. `Resolve` and `ValueClassGen` only ever see this module's own
declarations and stay on the bare name.

## Also outstanding

- The PR body's compatibility section still says the diagnostics do not move. Since the #464 fix it is
  E2011 13 → 19, E2010 unchanged.
- A row the generator builds at an edge would be as good a witness as one already written and would
  take most of the 81 back. Left out: `Findings` would have to ask `Generated`, so every `--adequacy`
  run would build and decode values, and `--generate` is opt-in. Worth its own change.
- #461 waits on this merging — it touches `Partitions`, `Terms` and `InvariantChecker`.
- `gh issue close 427` and `gh issue close 464` by hand after merge.

kawasima:
The contradictory newtype is closed. `54668c4d`, rebased on `develop`, CI green.

Taking a newtype straight as a parameter skipped the reading entirely — bounds from the type,
feasibility from nowhere. `value >= 10 && value <= 0` describes no value and both ends were still
counted as measurements waiting for a row that cannot exist, since `Bounds` calls a range empty only
when both ends are absent and `[10, 0]` is not that.

As you say, the domain is already the one authority for a contradiction, so the reading now runs for a
newtype like any other declaration and only the per-field bounds are dropped — its value is the same
position it is and there are no siblings to relate. Feasibility and whether every rule was read come
back with it, and nothing else needed a second notion of empty.

```
data Impossible = Int
    invariant low = value >= 10
    invariant high = value <= 0
```

```
    boundary    0/0          (was: 0/0 (2 not measured))
```

Held on the count rather than the findings: with no value of the type there is no row either, so
nothing was ever reported at those edges — what was wrong is that they were waiting for one.

One thing moves on the corpus, and it is the same defect: `Quota` is `value > 0.0m` taken straight as
a parameter, a strict bound on a decimal has no least value, and the edge the range reader names is
not one anything promises. Said rather than counted now.

```
counted   181 -> 138
demoted   82
verdicts  every module unchanged
E2011     19        E2010  0
tests     3972
```

Agreed on the generator witness being its own change. Marking ready.

Issue #446, opened 2026-08-07
Boundary validation is the construction of the signature downstream is allowed to see


kawasima:
> Boundary validation is not a check performed on a signature. It is the construction of the
> signature that downstream phases are allowed to see.

## Why the pair of options this was opened with is the wrong pair

Drop the two `unsupported` branches so the runner's switch is total, or keep them as the internal
error they describe. Both leave `Runner` switching over `Type` — the whole type language — and asking
the boundary's question of it.

`Runner.decoderFor` handles `Prim`, `Ref`, `List`, `Set` and `Map`; `Runner.encode` adds `Union`.
Everything else falls to a fallback arm, and showing that none of it arrives takes six arguments
spread across five rules: E1313 for an optional, E1311 for a tuple and a function, E1312 for a union
in a parameter, E1402 with E1502 for an open type, and "not writable in a signature" for the rest.
The runner knows none of them. That its accepted set coincides with the checker's admitted set is a
non-local fact that nothing checks, and the fallback reports a compiler defect in a reader's
vocabulary — *which `run` cannot decode yet*.

`CodecGen.flatMember` raised on a union member no module declares (#456) and `FixtureReader` asks the
same questions over again. Three consumers, three assumptions about one `Type`, three ways of being
wrong.

## What is not enough

Classifying the type just before the reader. A `Type -> BoundaryShape` function called from
`Compilation.boundary` produces a classified tree and changes nothing: the checker still validates
independently, the classifier re-derives, and a rule the checker relaxes without the classifier
following makes an ordinary compile succeed and `souther run` raise. The map-key classification would
be asked twice, which is the *ask twice* #447 removed, moved one level up.

Draft #460 is exactly that, and is kept as the base for the parts of it that stand: the two ADTs, the
input/output split, and the runner as their consumer.

## The structure

```
written behavior signature
        │
        ▼
PipelineSigs.signatures — resolve and validate, exactly once
        │
        ├─ failure → E1311 / E1312 / E1313 / E1314 / E1325
        │
        └─ success → BoundarySig
                      /        \
              BoundaryInput   BoundaryOutput
                      │
                      ├── module metadata
                      ├── pipeline composition
                      ├── Runner
                      └── FixtureReader / CodecGen
```

`BoundarySig` is canonical, and the types are derived from it rather than stored beside it:

```java
record BoundarySig(List<BoundaryInput> ins, BoundaryOutput out) {
[...]

PR #460, opened 2026-08-07
Make the signature, and admit what it carries in the making

kawasima:
Part of #446, which stays open: the rest of it is #462 and #463, and the scope is revised on the
issue itself to say why one of its items is decided the other way.

## What was wrong

The rules about what a behavior's boundary may carry were six passes over a written declaration —
an anonymous union in a parameter, a tuple, a function, an optional, a `Map` key, a name the language
declares of its own operations — and every phase below them held the behavior's `Type` and asked the
boundary's question of it again.

`Runner.decoderFor` handled `Prim`, `Ref`, `List`, `Set` and `Map`; `Runner.encode` added `Union`.
Everything else fell to a fallback arm reporting a compiler defect in a reader's vocabulary — *which
`run` cannot decode yet*. Showing that none of it arrives took six arguments spread across five
rules: E1313 for an optional, E1311 for a tuple and a function, E1312 for a union in a parameter,
E1402 with E1502 for an open type, and "not writable in a signature" for the rest. The runner knew
none of them. That its accepted set coincided with the checker's admitted set was a non-local fact
nothing checked.

Classifying the type just before the reader does not answer it, which is what the first attempt on
this branch did: the checker still validated independently, the classifier still re-derived, and the
map key was classified twice — the *ask twice* #447 removed, moved one level up.

## What this does

**Boundary validation is not a check performed on a signature. It is the construction of the
signature that downstream phases are allowed to see.**

`Sig` holds the shapes rather than the types:

```java
public final class Sig {
    Sig(List<BoundaryInput> ins, BoundaryOutput out) { … }   // package-private

    public List<BoundaryInput> ins();
    public BoundaryOutput out();
    public List<Type> inputTypes();
    public Type outputType();
}
```

Both are built by one walk inside `PipelineSigs.signatures`, which refuses what the boundary does not
admit as it goes. Holding a `Type` beside the witness would make a new thing to keep true — that they
agree — so the crossing is one-way: a signature always yields its types, and a type becomes a
signature only by going through the walk. `BoundaryShapes` does not survive; a public `Type -> shape`
utility is a standing invitation to re-derive.

**The constructor is closed, not merely unused.** A record publishes a canonical constructor, and
with one, anything below the check could assemble a `Sig` out of shapes nothing admitted — a
`Nominal` naming a type the language declares of its own operations, say — and every reader below
would take it for one the compiler stands behind. That would leave "the check happens to build its
signatures through the walk", which is a convention, in place of "a signature cannot be made any
other way", which is the claim. So `Sig` is a final class with a package-private constructor,
reachable from `SignatureBoundary` and the composition that calls it and nowhere else. It keeps value
equality by hand, because the query layer compares answers to decide whether recomputing one changed
anything. The shapes themselves stay open: what is closed is raising one to a signature.

**And one place makes them.** A closed constructor stops the wrong kind of signature from being made;
it does not stop a second walk, and there were three: `TypeChecker.checkRecovering` built its own,
and `Backend` built two more. Each was correct, so nothing observable was wrong — the walks are the
same walk, and they would have gone on agreeing until the trees they are given stopped being the same
tree. The check read one answer and `Runner`, through `Compilation.signatures`, read another.

`Bodies.Signatures` is now the only caller:

```
Bodies.Signatures ── PipelineSigs.signatures ── SignatureBoundary
        │
        ├── ModuleCheck / TypeChecker.checkModule
        ├── Output.Classes / Backend.generate
        └── Compilation.signatures / Runner, metadata
```

`TypeChecker` no longer takes `importedSigs` at all — it took them only to build signatures with —
and `Backend.generate` takes the signatures beside the imported ones. Identity through the query is
what "the same signature" now means.

A second caller is invisible to every other kind of test, so `ASignatureIsMadeInOnePlaceTest` reads
the sources for one. It is a tripwire and not a proof; it was confirmed to fail on a call put back
into `TypeChecker`, naming it.

**Each structural position owns admissibility of the subtree under it**, so there is no ordering
between rules to know:

```
Prim              → a scalar, or E1325 for `Raw`
Ref               → a name a model declared, or E1325
List / Set        → the element, recursively
Map               → the key as a whole, as a boundary map key; the value, recursively
Union in an input → E1312
Option            → E1313
Tuple / Fn        → E1311
```

So `Map<Option<Int>, String>` is E1314 and not E1313, reversing the precedence #458 pinned between
what were then independent passes. It is also the better report: fixing `Option<Int>` to `Int` on
E1313's advice earns E1314 next, because E1313 was never the reason *there*.

**Where a signature is wrong in more than one place, what is reported is the position the walk
reaches first** — the parameters as written, then the answer, and within a type from the outside in.
This is wider than the key position, and it is the same change. The passes were rule-major: every
parameter and the answer were asked about tuples before any of them was asked about optionals, so

```souther
behavior f : (x: Option<Int>) -> (Int, Int)
```

was reported for the tuple in the answer and is now reported for the optional in `x`. There is no
order between rules left to state, because there are no passes left to order.

**A scalar is a closed set with no `Raw` in it.** `Scalar(Type.Prim)` could hold the reserved type,
which left `Runner.leafDecoder` with a `RAW` arm and the catalog with `run.decode.raw` and
`run.encode.raw`. `BoundaryScalar` has six cases and one way in, and both arms and both messages are
gone with it — as `run.decode.unsupported` and `run.encode.unsupported` already were.

**Every signature is one, whatever its origin.** A composition's answer is a type nobody wrote and is
admitted where it is made; a declaration read back from a jar travels as source and is admitted by
the same walk when the importing compilation builds its signatures.

The issue asked instead for the validated representation in the jar. That is decided the other way
here, and the issue now says so: a module's declarations already travel as source — a data, an
invariant, a published helper, all read back by the same front end — so a serialized witness would be
a *second* representation of what the declaration already says, and the two agreeing would be a new
thing to keep true. `Backend.BOUNDARY_VERSION` moves 8 → 9, which is what keeps an import from
re-admitting a declaration written under rules this compiler does not share: a jar whose number is
not this one is refused before its declarations are read.

## What goes

The six `SpecChecker.reject*` passes, and the two whole-type walks that backed two of them
(`TypeOps.optionalInBoundaryShape`, `TypeOps.foreignNameInBoundaryShape`). Every diagnostic keeps its
code, its message and the position it is reported at. What changed is who raises them, and — for a
declaration wrong in more than one place — which of them arrives.

## The order the checks run in

`PipelineSigs.signatures` moved to just before `checkRequiresAreInjectionTargets`, with
`checkStagesAreSingleInput` ahead of it. Making the signatures is a prerequisite of the module check
rather than a step in the middle of it, and it fails fast: a signature that does not build leaves no
behavior to require, to compose or to expose. Left where it was, `ASignatureThatDoesNotBuildTest`
reported twice — E1607 was collected before the signature failed. Put ahead of
`checkStagesAreSingleInput`, E1702 would arrive first and the friendlier diagnostic would never be
seen.

## Verification

`mvn clean test` green — compiler 3572, runtime 109, lsp 184, and the rest.

Collapsing the three walks into one changed no test: the check is handed the signatures and abandons
the module at the point where it used to build them, so a module whose signatures do not build still
reports what its declarations, its `exposing` line and its stage arities had to say, and nothing
twice.

What a reader sees first, where a module is wrong in a way signatures decide and wrong in a way they
do not, was measured against `develop` rather than assumed. A module whose parameter carries an
optional and whose composition has a two-input stage reports `[E1313, E1702]` on `develop` and
`[E1313, E1702]` here — the same two, in the same order. That is pinned by a test through
`Compiler.compile` and `allReports`, and it is pinned rather than stated: which of a module's phases
reports first is decided by when each answer is worked out, which is not what this change is about.

`OneBoundaryAnswerReachesTheReaderTest` holds what is new. A written signature classifies to the
shapes its codecs are built for; a composition carries the same answer, with the retired case in it;
and a type the boundary does not admit builds no signature, each row naming the obligation that
refuses it — including the three that say which position owns the diagnostic, and the cross-position
rows that fix the first-error semantics above.

## What this does not do, and why it is not this change

`FixtureReader` (#462) and `CodecGen.flatMember` (#463) still hold a `Type`, so *three consumers, one
answer* is one-and-a-bit rather than one. Each needs something this change does not settle.

`FixtureReader.decoderFor` serves helper-parameter positions as well as boundary ones — a helper may
take an optional, a tuple or a function, which is exactly what a boundary may not — so the two have
to be told apart at every caller first, including the stand-in path, which rebuilds a dependency's
parameter types from the AST instead of reading its signature. Until then it still asks
`classifyConcreteMapKey` of a `Type`.

`CodecGen` reads a union's *leaf* members, and the witness does not carry them:
`BoundaryOutput.Cases` holds the members as written, so that `outputType()` reconstructs the type the
signature was built from. Expanding a named sum inside the witness would change what the signature's
type is, so that reader needs leaf expansion carried beside the members rather than instead of them.

`BoundaryNominal` waits on the second of those: a name carrying its resolved codec descriptor has
exactly one reader that would branch on it, and adding it first would be a witness nobody reads.

One widening, in the one rule that was asked of the written form rather than of the type: a parameter
written `A | A` has the type `A` and is now taken. Reading the type at every position is what lets a
composed and an imported signature be subject to the rule at all.

ADR-0100 records the decision and both of those.


kawasima:
Draft. What this says it does — "the boundary's answer now travels" — is not true of it: `Compilation.boundary` re-derives the classification after the fact instead of receiving what validation produced, and the map key is classified a second time. The design that replaces it is in #446: the signature *is* the witness, built once where it is resolved, with each structural position owning the diagnostic for the subtree under it. The ADTs, the input/output split and the runner as their consumer stay; the rest is rewritten on top.

kawasima:
On the ordering: measured rather than assumed, and the premise turns out to be the intermediate commit rather than `develop`.

The program in the review reports, through `Compiler.compile` and `allReports`:

| | first error | all reports |
| --- | --- | --- |
| `develop` | `3:17 E1313` | `[E1313, E1702]` |
| `9f93805` (intermediate) | `9:1 E1702` | — |
| `c3c6644` (head) | `3:17 E1313` | `[E1313, E1702]` |

So head matches `develop`. What `9f93805` did was the deviation: I had moved signature construction to sit after `checkStagesAreSingleInput` so that `ASignatureThatDoesNotBuildTest` would report once, and on `develop` the boundary passes ran early — before the `exposing` checks and before that one — so `E1313` came first there too. Hoisting `Signatures` into the query put it back where `develop` had it.

I did try demanding the signatures lazily, at the point in `checkRecovering` that needs them, on the theory that the demand decides when the report is published. It changes nothing: a child query completes before its parent either way, so `Signatures` publishes before `ModuleCheck` regardless of when it is asked. That is worth saying because it rules out the cheap fix — making the publication order follow the semantic order really would need the pre-signature checks to be a query of their own, as you suggest.

I have not done that, on the grounds that there is now nothing to repair: the order is the one `develop` had. What is decided by this change is the order *within* one signature, and that is the walk, which `whereTwoPositionsAreWrongTheOneWalkedFirstIsReported` fixes. Across a module’s phases the order is when each answer is worked out, which this does not touch — `aBoundaryRefusalIsShownAndTheOtherMistakeIsReportedToo` pins it through `Compiler.compile` and `allReports` and says in its own words that it is pinned rather than stated.

Happy to take the `PreSignatureChecks` split as its own issue if you want the publication order to be a decision rather than a consequence — it is a real gap, just not one this PR opens.

PR #465, opened 2026-08-08
Say what a fixture may supply

### LINKED ISSUE #462: FixtureReader has no answer for what a fixture may build
`FixtureReader` decides whether it can supply a value to a position by borrowing three unrelated
mechanisms: the boundary's map-key classification, reflective discovery of a generated `decoder()`,
and a handwritten list of the type forms it knows how to walk. Those are three different facts, and
what a reader is told when one of them says no exposes the mismatch.

Measured, with a helper called from an `example` row — a position that crosses nothing:

| written | what happens |
| --- | --- |
| `(xs: List<Int>)`, `(s: Set<String>)`, `(m: Map<UserId, Int>)` | builds |
| `(m: Map<Int, Int>)` | `` `Int` cannot key a Map that crosses `` |
| `(o: Option<Int>)`, `(p: (Int, Int))` | `is not supported as an example value yet` |
| `(m: RoundingMode)` | ``no decoder for `RoundingMode` `` |
| `(r: Raw)` | ``no decoder for `Raw` `` |

A `Map<Int, Int>` a helper takes reports a restriction on crossing a boundary, and it crosses none.
`RoundingMode` reports a missing decoder, which is an implementation property of a type the runtime
provides rather than anything about supplying a fixture. An optional and a tuple fall through the
reader's own vocabulary. Each message is accurate about the mechanism that produced it and wrong
about what the author has to change.

So the defect is not the wording. **There is no policy for the position to have a diagnostic about.**

| position | who decides what it admits |
| --- | --- |
| a behavior's boundary | `SignatureBoundary`, which leaves the admitted shape behind |
| a data field | `DataChecker` |
| a position a fixture supplies | nobody |

The three mechanisms are standing in for the missing one, and what is admitted today is the
intersection of three accidental predicates. That is also why a helper parameter looks like the odd
case: it is not. It is the ordinary case of a position nothing decided.

## The `Raw` arm is evidence, not cleanup

`decoderFor` carries

```java
case RAW -> throw new FixtureException("the Raw type cannot be an example input");
```

and nothing reaches it. Written `Raw` denotes a `Type.Ref`, so it takes the nominal branch and fails
by reflection as ``no decoder for `Raw` ``. The arm neither states a rule nor stops an input.

The same arm went out of `Runner` in #460, and for a different reason: there it had become
unreachable because the boundary stopped admitting `Raw` at all. Here it was never implementing the
policy it appears to implement. `decoderFor` looks like it says what the type system allows, and it
does not.

## Three things, in this order

**1. Make fixture admissibility an answer of its own.** Take "may a fixture supply a value of this
type here" out of `FixtureReader`. For a position the boundary did not establish, its structural rule
is the data field's — what a derived codec can read — which `DataChecker` already states
(`TypeOps.withoutExternalForm`, `TypeOps.nonBoundaryMapKey`). That leaves `Map<Int, Int>` no way to
reach the boundary's rule, and refuses `RoundingMode` for being a type a fixture cannot construct
rather than for a reflection call that came back empty.

Not as a `boolean fixtureable(Type)`. Checking and then handing the `Type` back to the reader to
re-derive from is the shape #460 removed: the admitted classification is what travels.

```
helper parameter ──fixture admission──> FixtureShape
a behavior's boundary ────────────────> BoundaryInput / BoundaryOutput
```

**2. Then a boundary position stops being asked.** It has already been through
`SignatureBoundary`, so there is nothing for `FixtureReader` to classify: no second
`classifyConcreteMapKey`, no reflection standing in for a decision. This is what this issue used to
be about, and it is the consequence rather than the cause — the reason the `Type` is being asked
again is that nothing else answers for the other positions.

**3. What an expression may denote stays the fixture language's question.** Whether an `Ast.Expr` can
express a value of an admitted shape is not the same question as whether the shape is admitted, and
the two have to stay apart or `is not supported as an example value yet` comes back under a new name.
`Option` is where the difference is already visible: a fixture can write `None` for an optional field
and a present value through the enclosing data's decoder, both of which are pinned by
`CompileOptionalFixtureTest`. What has no notation is a top-level `Option<T>` argument. Tuples and
functions are the same kind of question and not this one.

## Not this issue

This issue does not decide which admitted shapes the fixture expression language should be able to
denote. In particular it adds no syntax and no construction rules for a top-level `Option<T>`, for
tuples, or for function values.

The reason to say so out loud is that step 1 makes a shape admissible, and admitting one reads as an
instruction to support writing it:

```
Option<Int> is admitted now
  → decoderFor should learn Option
    → and then some syntax has to construct Some / None
```

Nothing in that chain follows. A helper may take a function parameter and it does not follow that an
`example` row should be able to write a function value — a `fake` and an injected behavior are
already how a row supplies behavior, and a fixture language with no function literal may well be the
coherent answer. Deciding it here would settle it by implementation convenience, which is the shape
of the defect this issue is about.

## What success looks like

Not that everything builds. That each refusal comes from the layer that owns it:

| written in a position a fixture supplies | after this |
| --- | --- |
| `Map<Int, Int>` | admissible; built |
| `RoundingMode`, `Raw` | inadmissible as a fixture target, said by the fixture policy |
| `Option<Int>`, `(Int, Int)`, a function | target admissibility and expression denotability reported apart |

`Option` and a tuple need not start building. They need to stop being refused from the wrong layer.

The separate issue — *define which admitted values the fixture language can denote* — is worth
opening when one of these has happened, and not before: the set of "admissible but unwritable" types
is known because this issue settled the first half; a real model wants a top-level `Option<T>` or a
tuple in a row; or the fixture language's completeness is being written down as specification.

## Why this is not just the rest of #446

#446's rule is not "carry the boundary's witness". It is:

> admissibility is decided where a position is established, and what is downstream receives the
> admitted representation rather than reconstructing the decision from a `Type`.

#460 applied that to a behavior's boundary. This applies it to a position a fixture supplies, which
has no such decision at all yet. Where the position came from a boundary, the answer #460 already
built is the one to use.

Part of #446.

kawasima:
Part of #446. Closes #462.

## What was wrong

Nothing said what could stand at a position a fixture supplies. `FixtureReader` worked it out as it
went, from three things that are not that question: the boundary's rule about what may key a `Map`
that crosses, whether reflection found a generated `decoder()`, and which constructors of `Type` it
happened to have a branch for. What was admitted was the intersection of three accidental predicates.

Measured, with a helper called from an `example` row — a position that crosses nothing:

| written | before | after |
| --- | --- | --- |
| `(m: Map<Int, Int>)` | `` `Int` cannot key a Map that crosses `` | builds |
| `(m: RoundingMode)` | ``no decoder for `RoundingMode` `` | `implemented by the runtime rather than derived, so it has no decoder` |
| `(r: Raw)` | ``no decoder for `Raw` `` | `` `Raw` is the reserved type and has no decoder `` |
| `(p: (Int, Int))` | `is not supported as an example value yet` | `has no external representation, so no decoder reads one` |
| `(o: Option<Int>)` | `is not supported as an example value yet` | `absence belongs to the data that holds it, on a `?` field` |
| `List`, `Set`, a newtype-keyed `Map` | builds | builds |

Each message was accurate about the mechanism that produced it and wrong about what the author has
to change. The defect was not the wording: there was no policy for the position to have a diagnostic
about.

## What this does

**`FixtureShape` answers it on its own.** A fixture is written text that becomes a value by being
decoded, so what may stand at such a position is what a decoder reads: a scalar, a type whose codec
was derived, or a collection of them. A classification and not a yes — the reader is handed the
shape rather than the type it was admitted from.

Which names have a derived codec is one question the symbol table already answers.
`Symbols.declaredByCompilation` says whether a module of this compilation declared a name —
including one read off the path, whose own build derived its codec — and the language's own
vocabulary is the other side of that line. Measured before switching, since it decides
admissibility: the two differ exactly on the runtime-backed data, so reading the runtime namespace
and a failed lookup was computing that same set out of a spelling convention and an absence.

**A map a fixture writes gets its own decoder.** The neutral-source map decoder reads an object of
strings, which is the form a map crosses a boundary as; wrapping it in a rekey was the last of the
boundary's assumptions standing in a position that crosses nothing. A fixture writes a list of pairs
and hands over a map of the keys themselves, so both sides go through their own decoder and a key's
invariant runs where a key is written. That is what makes `Map<Int, Int>` buildable.

**A boundary position is read rather than classified again.** `FixtureShape.of(BoundaryInput)` is a
projection: every concrete type a boundary admits is one a fixture admits — the same scalars, a name
a model declared where this wants one whose codec was derived, a key that renders as text where this
takes any key a decoder reads — so the crossing has no refusal in it. A test holds that, of witnesses
a compile produced rather than of ones written by hand.

An answer is `ofWholeAnswer`, and it is partial on purpose: a union nobody named is not one fixture
shape, since which of its members a value is is what the row writes rather than what the position
answered. It says so with nothing rather than by raising, and carries that outward, so a shape
holding nothing cannot be built either. A second test reaches a `Cases` a compile produced.

**One fact, one type.** `BoundaryScalar` is now `LeafScalar` — a primitive a derived codec reads and
writes as itself — held by `BoundaryInput.Scalar`, `BoundaryOutput.Scalar` and `FixtureShape.Scalar`
alike. What a boundary writes and what a fixture reads list the same six because both are asking
which primitives a codec exists for; two spellings of that agreeing today would have been the whole
of the guarantee.

**An admitted name with no decoder is not a fixture that cannot be read.** Admitting a `Nominal` is
the compiler saying a codec was derived for it, so a missing class or a missing `decoder()` raises
with its cause rather than being reported in the reader's vocabulary about a program its author would
find nothing wrong with.

**A map key is shaped like anything else at that position.** The classification made a key a position
of the same kind and the neutral form had not followed, so an inner collection reached the key
decoder as the list of pairs it was written as: `Map<Map<Int, Int>, Int>` was admitted and then
failed as `not_a_map`. Both sides are shaped now, which closes the nested cases with it.

A dependency's stand-in reads that dependency's own signature: `resolveFake` rebuilt its output type
from what it declared and `tableProxy` rebuilt its parameter types, in the one place that had a `Sig`
to hand all along.

`FixtureReader` no longer asks `classifyConcreteMapKey`. The two callers left are a data's derived
codec and the signature boundary, which are both boundaries.

## The `Raw` arm

`decoderFor` carried `case RAW -> throw …("the Raw type cannot be an example input")` and nothing
reached it: written `Raw` denotes a `Type.Ref`, so the name fell through to reflection and failed
there. The same arm went out of `Runner` in #460 for a different reason — there the boundary had
stopped admitting `Raw` at all. Here it was never implementing the policy it appeared to.

## What still starts from a `Type`, and why

The position nothing established — a helper's parameter — and the case a row names, which is what
the row wrote rather than what the position admitted: a row may name a case the behavior does not
answer with, and that is a disagreement to report rather than a shape to read off the answer. `raw`
keeps its `Type` throughout, since which sum a bare case belongs to and what a written list holds are
questions about the expression.

## Not in this

Which admitted shapes the fixture expression language should be able to denote. No syntax and no
construction rules for a top-level `Option<T>`, for tuples, or for function values — the issue says
why, and the success condition here is that each refusal comes from the layer that owns it rather
than that everything builds.

## Verification

`mvn clean test` green — compiler 3583, runtime 109, lsp 184, and the rest. No commit here changed an
existing test; what is new is `WhatAFixtureMaySupplyIsItsOwnQuestionTest`, which holds the measured
table, the nested map key, an imported type, and both halves of the answer projection.


kawasima:
All four, plus the two smaller ones. Head is `9fe1b499`; `mvn clean test` green — compiler 3583.

**1. Reflection failure is no longer a fixture refusal.** Admitting a `Nominal` is the compiler saying a codec was derived for it, so a missing class or missing `decoder()` raises with the cause attached instead of being reported as a fixture that cannot be read. You are right that moving the reflection from deciding to executing left the old meaning on the failure.

**2. Confirmed as a real bug, and fixed.** Measured first:

```
Map<Map<Int, Int>, Int>  →  [[1, 2]]: a `Map` fixture is a list of (key, value) pairs
Map<List<Int>, Int>      →  builds        (a list needs no reshaping, which is why it hid)
```

`NeutralForm.shaped` now shapes both sides, and `Map<List<Map<Int, Int>>, Int>` closes with it as you predicted. Both are regression tests.

**3. Taken as you framed it.** `BoundaryScalar` is now `LeafScalar` — "a primitive a derived codec reads and writes as itself" — held by `BoundaryInput.Scalar`, `BoundaryOutput.Scalar` and `FixtureShape.Scalar` alike. One fact read by two questions that are not the same question, rather than one borrowing the other's vocabulary because the sets coincide today.

**4. The projection is partial and now says so.** `ofWholeAnswer` returns nothing for a union nobody named instead of raising. The test is split: `everyConcreteShapeABoundaryAdmitsProjectsWithoutRefusing`, and `anAnswerOfSeveralTypesIsNotOneShape`, which reaches `Cases` — the old one never did.

**On `declaredByCompilation`:** measured before switching, since it decides admissibility.

```
UserId (local)            get=true   declaredByCompilation=true
RoundingMode              get=true   declaredByCompilation=false
```

The two differ exactly on the runtime-backed data, so my `TypeName.RUNTIME` + null-lookup pair was computing that same set out of a spelling convention and an absence. It is one call now. A jar-imported type still builds — `declaredByCompilation` counts what was read off the path, whose own build derived its codec — and that is pinned by a test, since it is the side of the line the wording does not name.

PR #468, opened 2026-08-08
Sweep the canonical form's layout rules, and say what decides each of them

kawasima:
Groundwork for #444, which asks how `fmt --check` could say *which rule* a deviation from the
canonical form breaks. Before that can be answered the rules have to exist as statements rather than
as formatter output, and this sweeps them: what decides each piece of the canonical layout, and what
that decision is allowed to read.

Mostly tests. Six changes to `Formatter` are listed at the end, four of which change canonical
output — each one a defect the sweep found, and none of them observed by any test that existed.

**Scope.** This sweeps the rules that decide *layout*. The formatter also canonicalizes code
tokens — it removes a trailing comma, adds the `|` in front of a match's first arm, and rewrites a
definition whose right-hand side is a lambda so its parameters sit left of the `=`. Those are left
alone deliberately and raised as #480, because whether canonical form should change a token the
author wrote is a different question from where a line breaks — and because they invalidate the
positional token correspondence #444 was relying on to make a deviation witness cheap. What the
formatter does today is characterized in `ACanonicalFormCanRewriteCodeTokensTest` rather than left
implicit.

No rule is added to `specification.adoc` and no diagnostic code is added. Whether canonical surface
syntax is normative is still open, and that ordering is deliberate.

## The six rule families

| family | what decides it |
| --- | --- |
| conditional layout | a group, and whether the line it would take fits the canonical width |
| forced layout | a structural obligation — four of them, over twenty-four sites |
| top-level separation | the kinds of the previous and the next item |
| indentation | the nesting level, four spaces per level |
| inline spacing | the construct that joins two tokens, and the kind on each side |
| comment placement | the owner and the relation (#451) |

Each is stated at the same granularity: the inputs the decision reads, the values it produces, and a
corpus that covers the domain.

## What the sweep found

**The conditional break decides at a column, and three of the nine `group` sites had no test.** Twenty
constructs written padded to exactly 99, 100 and 101 columns: 100 or fewer stays on one line, one
more breaks it, with no exceptions. Removing the group from an `if` expression, from an intrinsic
body, or from an example row's `with` bindings previously left the whole formatter suite green.

**Twenty-four forced breaks answer to four structural rules and two comment rules.** `HARDLINE` was
being counted as if each site were a rule. Removing one break at a time and reading which obligation
fails says otherwise: eight constructs write their members one to a line and that is one rule with
nine sites, a product body and a block take a line for each bracket, the blank line between top-level
items is a third, a file ending with one newline is a fourth.

**What goes between two top-level items is a function of both kinds** — forty-three cells, one for
every ordered pair the grammar admits. Two of them cannot be reached by a fixture holding a module
header and an import, and one is an asymmetry worth a second opinion: an examples file header is not
a module header, so its imports are written a blank line below it.

**One level of nesting is one indent further in**, asserted as the difference between levels rather
than as a four observed somewhere, and in spaces rather than in columns of whatever whitespace.

**Between two tokens on a line there is nothing or one space, and the tokens do not decide which.**
The key that does is the construct that joins them — their lowest common ancestor — with the kind on
each side. Over the corpus that key never takes both answers. It is load-bearing in only nine of the
164 token-kind pairs; the other 155 are transcribed, because the rule does not compress: grouping the
kinds into words, brackets, commas, dots and operators leaves eight of twenty-four class pairs taking
both answers.

**Cardinality is a formatting decision, and it is neither a kind of node nor a property of one.**
Every node kind can be built by a corpus that never writes a bracketed construct empty, and the empty
one is where the layout of the brackets themselves is settled. The unit is the bracket *site* — a
node kind and the bracket that opens the run, because a match arm holds a constructor's parentheses
and a record's braces and they are not one decision. Sweeping none/one/many per site found three
defects.

**A row's comment is the row's**, and one branch read comments above an example row's description
that nothing ever puts there.

## How the rules were told apart

The existing fixtures assert whole formatted files, so removing any forced break at all fails between
fifteen and fifty test methods. That reports that something changed, not which rule it was, and it is
why the count of rules was previously the count of `HARDLINE` sites.

Each obligation is one test method, quantified over the constructs it covers, with assertions phrased
so a site touches exactly one of them. Two things only combined mutations show: each of the nine
sites of the members-one-to-a-line rule fails exactly its own construct alone and all nine together
fail eight of nine rows; and the ninth, the file itself, is held by the line break and the blank line
together, so neither alone moves two declarations onto one line.

Coverage is asserted rather than assumed. Every node kind in `SyntaxKind` is built by the corpus,
every bracketed construct is written at every cardinality the grammar admits, and every kind the
file's walk treats as an item has a row. The cardinalities and pairs the grammar refuses are read
from `CstParser`, so a row missing from a table does not look like a case the language forbids.

## The changes to `Formatter`

Two that change nothing an author would see:

- `isTopLevel` becomes package-visible, so the table of what separates two items can be asked whether
  it covers every kind the file's walk writes.
- `head(SyntaxToken, Doc)` no longer reads comments above an example row's description or a match
  arm's pattern. Nothing attaches one there, and the whole suite passed with that branch replaced by
  a `throw`. The test saying where such a comment does go was added first.

Four that change canonical output, each a defect:

- **A match arm's pattern was spaced unlike every other pattern.** It is a run of tokens rather than a
  node, so the formatter joins them itself — the one place it does — and it joined every two with a
  space. `Some ( x )` in a match arm against `Some(x)` in the `let` that opens the same value.
- **A construct written open put two spaces between its brackets when empty.** `exposing ()` came out
  `exposing (  )`: the boundary just inside each bracket is a space when flat, and with no members
  both land in one gap. This is the counterexample to the range claim above, and it is why cardinality
  had to become a coverage unit.
- **A product body with no fields lost its opening brace and did not reparse.** The block writes that
  brace on its first member's line, and there was no member to write it on.
- **A match arm's record destructuring wrote `A { }`.** A node can hold more than one pair of
  brackets — an arm holds a constructor's parentheses and a record's braces — and keyed by the node
  they were one cell, so the braces were never swept empty. The pattern is built as a run of tokens
  and never reaches the `delimited` that answers this elsewhere; both paths say it now.

Each is a separate commit, preceded by the statement that would have caught it.

## Verification

`mvn clean test` green across all modules.

Every rule is checked by mutation: the width moved by a column and a line of exactly a hundred
columns broken; each `group` site and each forced break removed, alone and in the groups they were
claimed to share a rule with; one blank line between every pair of items, only a module header
keeping its imports tight, a kind added to the file walk; an indent of two, of eight, and of one tab
per level; no space before a definition's parameter list, a parameter spaced from its type, the
`exposing` list written closed, a match pattern spaced between every two tokens again; empty brackets
written by their boundary again, an empty product body written as a block again, a comments-only body
writing no brace. Each fails at the rule it belongs to.

The measurements behind each of these are on #444, which stays open: this is the sweep it asked for,
not an answer to its first question.


kawasima:
Addressed, with one addition that came out of P1 and changes the formatter.

## P1 — inline spacing is now a rule, not a dependency claim

The reviewer is right that the previous version stopped at "the tokens do not decide it" and never
exhibited `f`. It does now, and the key that closes it is the **construct that joins the two tokens**
— their lowest common ancestor in the tree.

```
f(joining construct, left kind, right kind) -> none | space
```

Over the corpus that key never takes both answers: 271 keys, zero conflicts. That is the function.

What makes it readable rather than a 271-row transcript is that the construct is load-bearing in
only nine of the 161 token-kind pairs. For the other 152 the pair alone decides. The nine are written
out with every construct's answer, and each differs for a reason that can be stated:

| pair | what the construct decides |
| --- | --- |
| `LPAREN IDENT`, `IDENT RPAREN` | a list of names is written open (`exposing ( a, b )`, `import m ( a, b )`); every other bracket is closed |
| `IDENT LPAREN` | a declaration's name is spaced from what follows it (`let f (a)`, `import m ( a )`); what is applied or opened is not |
| `IDENT COLON` | a type ascribed to a name is tight against it (`a: Int`); a signature is spaced from its name (`behavior b : ...`) |
| `IDENT LT`, `LT IDENT`, `IDENT GT`, `RPAREN GT` | `<` and `>` are a type's brackets or a comparison's operator |
| `MINUS IDENT` | `-` negates or subtracts |

The 152 pair-decided ones are deliberately not transcribed. Freezing them would pin the corpus rather
than state a rule, and they are already pinned by single-valuedness plus the coverage below.

**On the corpus being the wrong oracle.** Also right, and fixed at the root. The corpus is no longer
the standard library plus whatever it happens to contain: it is the standard library plus sources
written to reach what it does not, and **every node kind in `SyntaxKind` is asserted to be built by
it**. A kind added to the grammar with no source here fails the test rather than going unmeasured.
The kinds that never join two tokens on a line are listed with the reason each one cannot, so they
are accounted for rather than absent.

## The addition: a match arm's pattern was spaced unlike every other pattern

Enumerating the table turned up why `MATCH_CASE` appeared in five of the exceptional pairs. A match
arm's pattern is a run of tokens rather than a node — the one place the formatter joins two tokens
itself instead of a construct writing what goes between them — and it joined every two of them with a
space. So the canonical form was:

```
match o with
    | Some ( x ) -> x        // a match arm
...
let W(inner) = w             // the same syntax in a let
```

The whole suite passes either way, which is the point: nothing observed it. The run now answers as
the constructs writing the same syntax do, `aPatternReadsTheSameWhereverItIsWritten` says so, and the
exception table went from nine reasons with an accident in it to nine with none.

This is a canonical-form change and it is its own commit (`e362b4ad`), so it can be dropped if you
would rather it were a separate issue. I think it belongs here: the table is what found it, and
leaving it in would have written an accident into the rule.

## P2 — the separator test was stronger than the rule

Split as suggested. Per module it is now `{"", " "} ⊇ seen`; that both answers occur is a separate
assertion over the whole corpus. A module that stopped containing one of them no longer fails a rule
it does not break.

## P2 — indentation was measuring leading whitespace, not spaces

Also right, and it mattered more than it looks: four tabs per level would have passed. `indentsOf`
now asserts the leading prefix is spaces before counting it, so the rule is "four spaces" and not
"four columns of something". Checked with a mutant that indents with one tab per level — twelve
failures.

## Minor — the positive rows read the grammar too

`theGrammarAdmitsThisPair` asserts `errors().isEmpty()` for every row of the table, so the parser is
the oracle on both sides. Without it a row the grammar refused would have been formatted from a tree
with an error token in it and answered about something else — `Formatter.format` assumes a clean
parse and does not check.

## Verification

`mvn clean test` green across all modules.

New mutants, each failing at the rule it belongs to: a match pattern spaced between every two tokens
again (3 table rows, the pattern statement, and the pair set), the `exposing` list written closed,
a parameter spaced from its type, no space before a definition's parameter list, an indent of two,
an indent of one tab per level, and only a module header keeping its imports tight.

With this the six rule families close at the same granularity:

```
conditional layout    <- group + width
forced layout         <- structural obligation
top-level separation  <- (previous kind, next kind)
indentation           <- nesting level, in spaces
inline spacing        <- (joining construct, left kind, right kind)
comment placement     <- (owner, relation)
```

kawasima:
The blocker is fixed, and sweeping it found a second defect underneath.

## `exposing ()` — confirmed, and it was not the only one

`module m exposing ()` parsed fine and formatted to `exposing (  )`. The reviewer's reasoning is
exactly right: `delimited` writes the boundary just inside each bracket, its flat form is a space, and
with no members the two land in the same gap. Nothing said two spaces, and it falsified the range
claim.

Sweeping the same cardinality across every bracketed construct found a worse one:

```souther
data R = {}
```

formatted to

```
data R =
    
    }
```

The opening brace is gone and the output does not reparse. `productBody` writes that brace on the
first member's line — `text(first ? "{ " : ", ")` — so with no member there is no line to write it
on. This is a token-preservation failure, the sweep's own top-level invariant, and it was reachable
from three characters of valid source.

Both are fixed:

- `delimited` writes `open + close` when there are no members. That is the answer a record literal
  already gave (`R {}` was special-cased in `newData`), so the special case is gone and the rule is
  stated once.
- An empty product body is written `{}`; one holding only comments keeps its lines, with the brace on
  one of its own. (Writing that check cost a bug of its own first: the emptiness test called
  `endLines`, which *takes* the comments it reports, so the comment was consumed and never written.
  Worth noting that the formatter's `unconsumed` guard does not catch that — it detects comments that
  reach no construct, not comments a construct took and dropped.)

## Cardinality is now a coverage unit

`AllOfABracketedConstructsCardinalitiesAreSweptTest`. Every bracketed construct — detected from the
tree, as a node whose own children hold an opening bracket and a closing one — has a row saying which
of none/one/many the grammar admits, and the corpus is asserted to write it at each.

The cardinalities a construct cannot take are read from `CstParser`, the same way the top-level
table's forbidden pairs are: a block ends in a result expression, `let f ()` is refused, a constructor
opens one value. What keeps those rows short is the grammar and not a choice made here.

`everyBracketedConstructHasARow` compares the row set against what the corpus actually builds, so a
construct that starts holding brackets is one whose empty form has been looked at.

Two rows are honest about a limit rather than claiming coverage they do not have: a block's statements
and a product's fields are not comma-separated, so the member count cannot tell one from several
there. That they are written one to a line is what the forced-break test says, and the row is about
the brackets.

## The 152 pairs: it does not compress, so it is transcribed

Measured before deciding. Grouping token kinds into words, brackets, commas, dots and operators leaves
**eight of twenty-four class pairs taking both answers**:

| class pair | tight | spaced |
| --- | --- | --- |
| `CLOSE OP` | `):` `)>` | `) *` `) ->` `) =` |
| `OP OP` | `>>` | `= -` `> =` |
| `OP OPEN` | `-(` `<(` | `-> (` `-> [` `++ [` |
| `OP WORD` | `-n` `...P` `<'a` | `+ 1` `&& p` |
| `OPEN WORD` | `('a` `(0` | `( alpha` `{ a` |
| `WORD CLOSE` | `'a)` `0]` | `1 }` `T )` |
| `WORD OP` | `'a>` `'b?` | `0 -` `'a =` |
| `WORD OPEN` | `Amount(` | `P (` `Q {` |

To make classes work you would have to split angle brackets from operators, prefix from binary,
postfix from binary, and each bracket from the others — at which point the class is the kind and the
rule is the pair. So the pair table *is* the rule, and it is written out: 39 pairs tight, 116 spaced,
9 decided by the joining construct.

`everyPairIsWrittenAsItsRowSays` checks it in both directions — no pair the corpus writes is missing
from the lists, none is written against its row, and no listed row goes unreached. A new adjacency now
fails the test, which is the point: it is a decision, and it should be made in the table rather than
by whichever construct writes it first.

## PR description

Updated. It now says five changes to `Formatter`, three of which change canonical output, each named
with the defect it fixes.

## Verification

`mvn clean test` green across all modules. New mutants: empty brackets written by their boundary
again (8 failures across both tests), an empty product body written as a block again, a comments-only
body writing no brace, a comment written with a trailing space. Each fails at the rule it belongs to.

kawasima:
Fixed at the level the review points at, not at the three cells.

## `MATCH_CASE` was right, and the key was one level too coarse

`| A {} -> 1` parses and came out `| A { }`. Confirmed, and it is a counterexample to the rule the
previous commit added, so it had to be fixed rather than excluded.

The site key is now **(node kind, opening bracket)**, and the sites are found by matching each
opener with its closer among a node's own children rather than taking the first of each. That splits
`MATCH_CASE LPAREN` from `MATCH_CASE LBRACE`, and nested constructor parentheses fall under the first
where they belong. Twenty sites, up from nineteen node kinds.

The formatter fix is where you would expect. A pattern is built as a run of tokens rather than as a
document, so it never reaches the `delimited` that answers this everywhere else, and the run's
separator knew about parentheses, commas and dots but not about a bracket closing the one before it.
`spaceBetween` now says brackets with nothing between them are written together — the same sentence
`delimited` says — and the knowledge of which tokens are brackets moved into one predicate that both
the run and the comment attachment read.

## `ADMITTED` is gone

The sharper half of the review was that the table was a hand-written grammar model spot-checked
against the parser. It is not written down any more. Each cell carries a source that would produce
it, and what the cell admits is **what the parser and the tree say that source builds** — refused by
the grammar, or parsed as a different site, and it admits nothing. `whatTheCellClaimsIsWhatItsSourceBuilds`
checks the claim against the reachable set in both directions, so they cannot drift.

That immediately confirmed both other omissions: `PATTERN_TUPLE` and `TUPLE_TYPE` do admit empty —
`let () = r` and `(t: ())` both parse — and the old table said they did not.

Sixty cells, twenty sites, three cardinalities each. The ones that admit nothing now carry their
witness rather than a comment: `let f ()` is refused, `A()` is refused, `(a, a)` parses as a tuple and
not as a parenthesised expression, `()` as neither.

## The pair table earned its keep

Adding the arm's braces to the corpus introduced `RBRACE ARROW`, which no row held, and
`everyPairIsWrittenAsItsRowSays` refused it until the row was written. That is the behaviour the
table was for: a new adjacency is a decision, made in the table rather than by whichever construct
writes it first.

## P2

The class Javadoc said the remaining pairs were not transcribed and that listing them would freeze
the corpus. It now says the opposite, which is what the code does and why: the rule does not
compress, so the table is the rule.

## Verification

`mvn clean test` green across all modules; CI green on `1eac32f0`.

New mutants: a pattern run spacing its empty brackets again (3 failures across both tests), a brace
not counted as an opening bracket, a brace not counted as a closing one (17 failures each, including
13 cells whose claim stops matching what their source builds).

## On the underlying shape

Worth saying plainly, since three reviews have now each found one more instance of it: none of these
were independent bugs. The formatter has no function that decides what goes between two tokens — it
spells its separators as string literals in fifty-five methods, and the adjacency rule is the
by-product. That is why the same token pair takes different answers in different constructs, why the
pattern run was a third mechanism disagreeing with the other two, and why the empty case had to be
found separately on each path.

This PR makes that rule stateable and checked, which is what #444 needs before it can compute a
deviation witness. Making it *implemented* — one `spaceBetween(left, right, joining construct)` that
the document construction goes through — is a separate and much larger change, and I would rather
raise it as its own issue than grow this one further.

kawasima:
Both done. `5a0fb998`, CI green.

## The oracle no longer has the formatter in it

`builds()` reads the parse of the source. `sitesWritten(source)` is the grammar question;
`sitesInTheCanonicalForm(source)` is the formatter question, and only the corpus-coverage check asks
the second.

That immediately exposed what the review predicted. `held()` counted commas, so `g(a,)` measured as
two members. It had never shown because the source was normalised before it was measured — the
formatter writes no trailing comma, so the detector was agreeing with the answer. Members are counted
now, with the comma that separates a member from nothing left out:

```
(a)     one
(a,)    one
(a, b)  many
(a, b,) many
```

`aTrailingCommaSeparatesAMemberFromNothing` holds seven cells written with one, including two that
assert the old reading is *wrong* — `g(a,)` is not `many`.

## Trailing comma is written down and scoped out

`ATrailingCommaIsRemovedAndIsNotWhitespaceTest`: the grammar admits it, the canonical form drops it,
the code-token stream is one shorter, and it is the trailing comma that went. Five kinds of list.

`FormatterTest`'s contract said formatting changes only whitespace and the non-trivia token stream
survives. It now says that holds over that corpus and not of the language, and points here. That test
still passes for the same reason it always did — none of its sources writes one — but it no longer
claims more than it checks.

The PR description now scopes itself to layout and says why the trailing comma is out.

## And it is #444's premise, not a detail

Agreed, and this is the part worth the most. #444's second comment derived that the deviation witness
is cheap because

```
nonTriviaTokens(source) = nonTriviaTokens(canonical)
```

makes the correspondence positional — and said that ought to be a design constraint rather than
something the formatter happens to do. It is not something the formatter does. It holds over sixteen
sources that never write a trailing comma.

Raised as **#480**, with the three answers laid out: preserve code tokens and accept that canonical
form is no longer single; delete them and give `E` a token alignment; or delete them and name the
deletion as a rule of its own so the witness can cite it. Recorded on #444 as a correction to that
comment, together with the decomposition this implies:

```
code-token canonicalization     trailing comma -> absent     #480
layout canonicalization         the six families             this PR
comment canonicalization        (owner, relation)            #451
```

The first and the third may not get the same answer to question 1. Deleting a token an author wrote
is the only one of the three that changes what the compiler reads rather than how it looks.

Also raised **#476** for the shape underneath all four defects this sweep found: the formatter has no
function that decides what goes between two tokens, so the rule is the by-product of two hundred
string literals and has to be re-established on each path.

kawasima:
Right on both, and the second one is worse than "one more case". `993c07e7`, CI green.

## There are three kinds of code-token canonicalization, not one

Verified both, and they are not the same shape as the trailing comma:

| | what happens |
| --- | --- |
| a token is **removed** | a trailing comma, in five kinds of list |
| a token is **added** | the `\|` in front of a match's first arm, which the grammar makes optional |
| tokens are **rewritten** | `let f = (x) -> x` becomes `let f (x) = x` — the `->` goes and the `=` moves past the parameter list |

```
source:     LET IDENT ASSIGN LPAREN IDENT RPAREN ARROW IDENT
canonical:  LET IDENT        LPAREN IDENT RPAREN ASSIGN IDENT
```

That third one is the reason to care about the count. The correction I posted on #444 said the fix
for a broken positional correspondence is a token alignment. That is true of a deletion and true of
an insertion. It is not true of a token that moved past another — the lambda rewrite is a structural
rewrite, and describing it as an edit to a sequence loses what it is. So #444's premise fails in
three ways and only two of them are repairable the way I said.

`ACanonicalFormCanRewriteCodeTokensTest` characterizes all three, records which kind each row is, and
asserts that all three kinds occur — which is the assertion that would have failed while the file
claimed there was one. `FormatterTest`'s contract points at it and names all three.

## #480 and #476 were empty — my mistake

`gh issue create --body -` set the body to a literal `-`; it needed `--body-file -`. Both are filled
in now, and #480 is retitled to **"Canonicalization rewrites code tokens, and the deviation witness
assumed it would not"** with the three kinds and what each costs a witness. It also carries a third
option I had not written down: rewrite, and give each rewrite its own rule and code, so the witness
names the rule rather than aligning around it — which for the lambda case is the only one of the
three that describes what actually happened.

## PR description

Fixed: four changes that alter canonical output, not three; three defects from the cardinality sweep,
not two; the cardinality unit described as the bracket site; and the scope paragraph now covers all
three kinds of code-token canonicalization rather than the trailing comma alone.

## On the shape of these five rounds

Each round has found one more instance of the same thing: a rule that exists only as the by-product
of where it was written, so it is true of the corpus and not of the language. The layout families,
the empty bracket, the bracket site, token preservation. The sweep is what turns those from
invisible into stated, and I would rather each of them was found here than while drafting the
specification — but it does mean "is this the last one" is not a question I can answer from inside
the sweep. What I can say is that the six layout families now have their domain covered by
construction rather than by corpus, which is the property that made the last two rounds find things
in the *scope statement* rather than in the rules.

Issue #461, opened 2026-08-07
A nested numeric newtype is not a number to the analysis


kawasima:
`data StartMinute = Minute` where `Minute = Int` is a number the language compares, orders and bounds, and the analysis does not model as a number at all. No atom is made for it, so a `guard` over it enters no reasoning, its type's bound reaches no boundary, and the position is reported as one the model draws no line through.

```souther
data Minute = Int
    invariant withinDay = value >= 0 && value <= 1440

data StartMinute = Minute

data Span = { a: T, b: T }
    invariant ordered = a < b
```

```
T = Minute        boundary 0/4
                    · no row is at f/s.a = 1439 (invariant Minute (max) within Span)
                    · no row is at f/s.b = 1 (invariant Minute (min) within Span)

T = StartMinute   boundary 0/0
                    · not derivable: s.a
                    · not derivable: s.b
```

A `guard` goes the same way. `guard a < b` before constructing something the difference satisfies discharges its invariant where the fields are `Minute` and does not where they are `StartMinute` — the comparison type-checks, and by the time it reaches the analysis it is gone.

## It is not that a lookup stops after one layer

That distinction is the language's and it is deliberate. Arithmetic is direct-only:

> Only a newtype whose value is *directly* `Int` or `Decimal` has arithmetic — a newtype over another newtype does not. (spec §newtype-arithmetic)

Comparison reaches the recursive base:

> (`Manager = Level = Int`) reaches its base. (ADR-0047)

`TypeOps` holds both — `directNumericNewtypeBase` for the first, `base` for the second — and the names are right. What is wrong is that the analyzer reuses the first as its general notion of a numeric value:

```java
boolean numericNewtype(Type t) {
    return TypeOps.directNumericNewtypeBase(t, symbols) != null;
}
```

`Terms.isNumeric` is what decides whether a position becomes an atom, so atoms, bounds and guard reasoning all inherit arithmetic's cutoff. Making `newtypeInner` recursive would repair the symptom by deleting a distinction the language means, and would leave the reason it was reached for in place.

```
nested newtype
  -> direct arithmetic capability = false      as specified
  -> analysis numericity = false               the error
  -> no atom
  -> comparison and guard cannot enter the reasoning
  -> nothing discharged, no boundary derived
```

## Three capabilities, not two

Being a number to the analysis is its own question and should be named as one, rather than given the arithmetic answer under a name that reads like a general one.

| question | today | for a nested numeric newtype |
| --- | --- | --- |
[...]

PR #470, opened 2026-08-08
A boundary is settled by evidence, and one place gathers it

### LINKED ISSUE #467: A row the generator built at a boundary settles it, and is not read as one
A row the generator built at a boundary is a witness, and is not read as one.

An edge derived from a projection that could not read every rule of the value is reported and not
counted (#427): the bounds admit it, and a rule the domain could not hold can refuse it, so nothing
has shown a row can be written there. A row that already sits at the value settles it — `met == YES`
is a value that went through the decoder, and the edge counts from then on.

A row `souther examples --generate` built at that value went through the same decoder and settles it
just as well. It is not read.

## What it costs today

On `souther-examples`, of the 82 edges reported as not known to be writable:

```
39   the generator builds a row at
13   the generator tries and every value is refused
30   not reached (a guard's line the arms did not measure, or no row names the behavior)
```

So a little under half of what is said and not counted is already known to be writable by the only
authority there is for it, and is being said anyway. The 13 are the honest remainder — a refusal is
not a proof of impossibility (ADR-0091) and they stay where they are.

## Why it was left out of #459

`Adequacy.Findings` would have to ask `Adequacy.Generated`, and generating rows builds and decodes
values against the module's own classes. `--generate` is opt-in today, and `--adequacy` is not; making
every adequacy run pay for a row search is a change to what that flag means, not a detail of this
measure. That is the decision this issue is for.

Three shapes, roughly in order of cost:

- ask `Generated` from `Findings`, and pay for it on every adequacy run;
- promote only where the caller already asked for generation — the numbers above would then appear
  under `--generate` and not otherwise, which makes one measure depend on another flag;
- keep the boundary unpromised in the measure and say the generator settled it beside the row it
  offers, which costs nothing and leaves the count conservative.

The third is closest to what the report already does with an unfillable combination, and the first is
what would actually move the number.

## Related

#427 introduced the third state and the row witness. Nothing here is needed for that to be sound —
conservative-plus-observed-witness is already sound, and this is precision.

kawasima:
Closes #467.

## What the issue asked, and what turned out to be under it

#467 asks whether `Findings` should read `Generated`, and offers three
shapes for saying yes. None of them is taken here. The generator's witness
being unread is a symptom: coverage, the static projection and runtime
construction were each producing their own answer about the same boundary
obligation, which is the problem `Adequacy.Excluded` already names —
deriving the same fact several times is several chances to disagree.

Two of the issue's premises did not hold up:

- The cost is not "a row search on every adequacy run". `Generator.fill`
  searches the pair space; a boundary witness is one decode per unmet
  obligation. They were bundled in one query, not one computation.
- `Generated` was asking `Output.Linked`, generating a second uncounted
  set of classes, while the rows had already been run against
  `Output.EvaluationLinked`. Those give the same generated bytes, so the
  second codegen bought nothing. (`FixtureReader.constructing` still makes
  its own `MemoryClassLoader`; no `Class<?>` identity is shared with the
  evaluation.)

## What is here

`BoundaryAssessment(obligation, coverage, writability, attempt)`, made
once in the new `Adequacy.Boundaries` query and read by `Findings`, by
`Generated` and by the report.

Three answers, because they are about three things and are established by
three different means:

```
Coverage    — what the rows showed
Writability — what is proven about a value existing there
Attempt     — what the search tried, and what came of it
```

`Writability` is sealed: `ProvenByProjection`, `WitnessedByRow`,
`WitnessedByConstruction`, `Unknown`. There is no case for "cannot be
written". A decoder refusing every candidate it tried says nothing about
the ones it did not, so a refusal leaves the edge unknown — and the
mutation that turns a decode failure into proof of impossibility is not
expressible, not merely tested against.

`Attempt` is `Built | Refused | NotAttempted(reason)`. It is kept whether
or not it changed the verdict, because an edge the projection proves is
still one a search can fail to reach — folding it into the verdict left
`--generate` with nothing to say about exactly that case.

Gone: `Coverages.unmet`, `Generator.forBoundaries`,
`PartitionEvidence.BoundaryCoverage`, and `knownWritable || met == YES`
— which was two evidence sources joined with a boolean `or`.

### Evaluation order

A row at the value settles it first, and nothing is built for a value
that is already there. Otherwise the probe runs, and the projection
stands behind it rather than in front of it. That is one step away from
what was proposed on the issue thread: skipping the probe where the
projection already proves the edge would also skip building the row the
generator offers there, and running one attempt that both readers use is
the point of moving it here.

## Effect on souther-examples

```
not known to be writable   82 -> 32
no row is at              138 -> 166
```

50 edges leave the unknown set. 28 become rows an author is owed. The
other 22 are on behaviors no row names (`returnFromLeave`,
`loseCoverage`, `applyDecision`): writability now has an answer, coverage
still does not, and the report no longer prints the second as though it
were the first.

`sharedAttendees` stays unknown, correctly — nothing can be written at
its `related` position, so no row completes.

Generator output is unchanged from develop: same rows, same notes.

## Tests

`AnEdgeIsWritableBecauseSomethingSaidSoTest` pins the design at the
assessment rather than through the printed line, which cannot tell a
witness from a proof. Its main fixture is one range with `value /= 0` on
it: edge 0 is refused because the unread rule reaches it, edge 10 is
witnessed by construction because it does not. The pair is the point —
reading `allRulesRead == false` as writable would settle both, and
keeping it as unwritable settles neither. A second model covers a proven
edge whose search still came back with nothing, in the assessment and in
the block an author reads.

Three cases in `ABoundaryIsAValueTheRecordCanHoldTest` changed, and are
rewritten to assert both ends. `Band { low, high } invariant low < high`
now distinguishes `band.low = 1`, which stays unknown, from
`band.high = 1`, which is owed.

`AdequacyLensTest.aLineWhoseValueCouldNotBeReadIsNotInTheRatio` holds the
editor lens to counting only what came to an answer.

Full clean build: 3691 + 185 + 109 + 1 tests, no failures.

## Not in this change

- The JSON report keeps its shape. `hit` and `knownWritable` are derived
  from the assessment instead of stored beside it. Naming which evidence
  settled a line is worth emitting and needs a schema version, so it
  waits for one.
- The candidate search is asymmetric: probing `band.low = 0` refuses
  every candidate while probing `band.high = 1` builds `low = 0, high = 1`
  — the same pair, found from one side only. It predates this change and
  is sound here (unknown never means impossible), but it costs coverage.
  Worth its own issue.

### DIFF
--- a/souther-compiler/src/main/java/souther/compiler/partition/Generator.java
+++ b/souther-compiler/src/main/java/souther/compiler/partition/Generator.java
@@ -241,15 +241,18 @@ public final class Generator {
     /**
      * What building a row at one boundary came to: the row, or why there is none.
      *
-     * <p>Exactly one of the two. A caller measuring the boundary reads whether a row was built, which
-     * is a value that went through the decoder and so a witness that the edge can be written; a caller
-     * offering work to a person reads the row itself. The attempt is made once and both read it.
+     * <p>One or the other, and the type says so. A caller measuring the boundary reads whether a row
+     * was built, which is a value that went through the decoder and so a witness that the edge can be
+     * written; a caller offering work to a person reads the row itself. The attempt is made once and
+     * both read it.
      */
-    public record BoundaryAttempt(GeneratedRow row, UnresolvedCombination unresolved) {
+    public sealed interface BoundaryAttempt {
 
-        public boolean built() {
-            return row != null;
-        }
+        /** A value with the edge in it, built and accepted. */
+        record Built(GeneratedRow row) implements BoundaryAttempt {}
+
+        /** No row came of it, and why. Never a statement that none exists. */
+        record Unresolved(UnresolvedCombination why) implements BoundaryAttempt {}
     }
 
     /**
@@ -267,13 +270,13 @@ public final class Generator {
         Axis axis = subject.axes().stream().filter(a -> a.id().equals(obligation.axis())).findFirst()
                 .orElse(null);
         if (axis == null) {
-            return new BoundaryAttempt(null, new UnresolvedCombination(List.of(),
+            return new BoundaryAttempt.Unresolved(new UnresolvedCombination(List.of(),
                     UnresolvedCombination.Reason.NO_REPRESENTATIVE));
         }
         String label = axis.path() + " = " + written(obligation.value());
         FixtureTemplate at = valueOf(obligation.value(), axis.type(), subject.symbols());
         if (at == null) {
-            return new BoundaryAttempt(null, new UnresolvedCombination(List.of(label),
+            return new BoundaryAttempt.Unresolved(new UnresolvedCombination(List.of(label),
                     UnresolvedCombination.Reason.NO_REPRESENTATIVE));
         }
         List<FixtureTemplate> inputs = new ArrayList<>();
@@ -290,12 +293,12 @@ public final class Generator {
                             ? Map.of(axis.path().toString(), List.of(at)) : Map.of();
             Outcome tried = valueAt(subject, p, here, settled, check);
             if (tried.value() == null) {
-                return new BoundaryAttempt(null,
+                return new BoundaryAttempt.Unresolved(
                         new UnresolvedCombination(List.of(label), tried.reason(), tried.detail()));
             }
             inputs.add(tried.value());
         }
-        return new BoundaryAttempt(new GeneratedRow(List.of(label), inputs), null);
+        return new BoundaryAttempt.Built(new GeneratedRow(List.of(label), inputs));
     }
 
     // --- the pair space -------------------------------------------------------------------------
--- a/souther-compiler/src/main/java/souther/compiler/partition/Generator.java
+++ b/souther-compiler/src/main/java/souther/compiler/partition/Generator.java
@@ -239,55 +239,63 @@ public final class Generator {
     }
 
     /**
-     * Rows at the boundaries nothing has been written at.
+     * What building a row at one boundary came to: the row, or why there is none.
+     *
+     * <p>Exactly one of the two. A caller measuring the boundary reads whether a row was built, which
+     * is a value that went through the decoder and so a witness that the edge can be written; a caller
+     * offering work to a person reads the row itself. The attempt is made once and both read it.
+     */
+    public record BoundaryAttempt(GeneratedRow row, UnresolvedCombination unresolved) {
+
+        public boolean built() {
+            return row != null;
+        }
+    }
+
+    /**
+     * A row at one boundary, built through the module's own decoders.
      *
      * <p>One row per boundary rather than one row covering several, because a row is a question put to
      * a person and a row sitting on three edges at once is three answers they have to separate.
+     *
+     * <p>Nothing here decides that a boundary cannot be written at. A refusal is a refusal of the
+     * candidates that were tried, and another value of the same edge may build; what comes back says
+     * which of the two happened and leaves the reading to the caller.
      */
-    public static GenerationResult forBoundaries(Subject subject, List<BoundaryObligation> unmet,
-                                                 CandidateCheck check) {
-        List<GeneratedRow> rows = new ArrayList<>();
-        List<UnresolvedCombination> unresolved = new ArrayList<>();
-        for (BoundaryObligation each : unmet) {
-            Axis axis = subject.axes().stream().filter(a -> a.id().equals(each.axis())).findFirst()
-                    .orElse(null);
-            if (axis == null) {
-                continue;
-            }
-            String label = axis.path() + " = " + written(each.value());
-            FixtureTemplate at = valueOf(each.value(), axis.type(), subject.symbols());
-            if (at == null) {
-                unresolved.add(new UnresolvedCombination(List.of(label),
-                        UnresolvedCombination.Reason.NO_REPRESENTATIVE));
-                continue;
-            }
-            List<FixtureTemplate> inputs = new ArrayList<>();
-            UnresolvedCombination left = null;
-            // What the rest of the row has to sit beside. A field of a record is not chosen from its
-            // own type once another field of that record is fixed: the rule relating them says what
-            // is left, and taking the bottom of the type's range instead is how a boundary that can
-            // be written came back as one every value tried was refused at.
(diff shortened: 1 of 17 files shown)


kawasima:
Both blockers are fixed in `ce40bd46`. Thank you — the first one was the same defect this PR is about, and I had reproduced its symptom without recognising it.

## 1. The attempt is a third answer

You are right that recovering the generation result from `Writability` made `ProvenByProjection × failed-attempt` unrepresentable, and that this combination is not contradictory but expected. `BoundaryAssessment` now carries it:

```java
record BoundaryAssessment(BoundaryObligation obligation, Coverage coverage,
                          Writability writability, Attempt attempt)
```

`Attempt` is `Built(row) | Refused(UnresolvedCombination) | NotAttempted(Reason)`, with `Reason` distinguishing a row already being there, the line never having been measured, the classes being absent, and the runtime being absent. `Generated.offered` reads only this now. `Writability` keeps the verdict alone — `WitnessedByConstruction` lost its payload and `Unknown` lost its reason, because the row and the reason each have one home in the attempt.

`Generator.probe` returns a sealed `Built | Unresolved` as you suggested, so the "exactly one of the two" invariant is in the type rather than a comment. That fell out of the same change: with the attempt stored, its shape stops being a private detail of one call site.

The `LinkageError` path is restored — `NotAttempted(RUNTIME_ABSENT)` and `NotAttempted(NO_CLASSES)` both become the `Incompleteness.RUNTIME_ABSENT` the block used to print.

**How much this was costing.** Comparing `--generate --boundaries` against develop over souther-examples, seven notes had gone silent on crm:

```
no row for `floor = 0` in `openNegotiation`
no row for `money.amount = 0` in `convert`
no row for `proposed = 0` in `sendProposal`
no row for `lead.reach: ContactPoint` in `convertLead`
...
```

Three name the boundary; the other four name a position that had no value, reached through `UnresolvedCombination.subject()`. All seven are `ProvenByProjection` edges whose attempt failed. Generator output is now identical to develop's on both counts — 182/29/0 notes and 30/425/50 rows for crm/hr/ordering.

I found these while checking your claim and initially mis-attributed them to `Linked` → `EvaluationLinked`. Reverting that switch produced byte-identical output, which is the cleanest evidence I have that the counted classes decode the same; the cause was entirely the projection you identified.

## 2. The lens

Confirmed, and it was a plain transcription error: `status() == COMPLETE` is `Hit | Missed`, and `!(NotMeasured)` also admits `Undecided`. Fixed to the two cases.

`AdequacyLensTest.aLineWhoseValueCouldNotBeReadIsNotInTheRatio` uses the `budgetSpent` shape you pointed at — a row wide enough that the observation truncates the numeric position — and asserts the boundary figure is absent rather than `0/2`. I checked it fails against the version in `47112a13` and passes now.

## 3. Non-blockers

- `BoundaryAttempt` is sealed, as above.
- PR body corrected: `EvaluationLinked` gives the same generated bytes and avoids a second codegen; `FixtureReader.constructing` still makes its own `MemoryClassLoader`, so no `Class<?>` identity is shared with the evaluation.
- The two comments that said a report should not pay for building values (`Adequacy.Filling`, `GeneratedRows.of`) now say what is true: filling the combinations searches the pair space and stays behind the flag, while the rows at the edges are read from the attempt the measurement already made.

## Fixtures

`AnEdgeIsWritableBecauseSomethingSaidSoTest` covers the combinations you named. `Missed × ProvenByProjection × Refused` needed a model where every rule of one input is read while the row cannot be completed for another reason, which is:

```
data Amount = Int
    invariant value >= 0

data Code = String
    invariant String.matches("(?=.*a).*b", value)

behavior place : (amount: Amount, code: Code) -> Ok
```

`amount = 0` is proven, and no row is produced because nothing can be written for `code`. There is a second test on the same model asserting the block still says so, which is the user-visible half.

`NoRows × ProvenByProjection × Built` and the arms-not-asked case are covered too. `Missed × Unknown × RuntimeAbsent` is not: I could not arrange a missing runtime through `Compilation` without a fake loader, and I would rather leave it uncovered than pin it with a test that constructs the assessment by hand. Say the word if you want that one and I will add the seam.

Corpus figures are unchanged: 82 → 32 not known to be writable, 138 → 166 rows owed. Full clean build green (3691 + 185 + 109 + 1).

kawasima:
Both done in `3b952edc`.

## `Attempt.Refused` → `Attempt.Unresolved`

Agreed, and the reason you give is the one that matters: the payload carries `NO_REPRESENTATIVE` and `SEARCH_LIMIT` as well, and neither of those is the decoder having said anything. A reader who trusted the name would take a decision out of two cases the decoder was never asked about — the same mistake this type exists to prevent, one size down. It also now matches `Generator.BoundaryAttempt.Unresolved`, so the two layers use one word for one thing.

The javadoc opening says three answers, and says why the third is kept where it changes neither of the others.

## `NO_CLASSES`

I traced when it can happen rather than leave it as a guess. `constructing` returns null when `Output.EvaluationLinked` is absent, and that key is absent when a module *this compilation declares* failed to generate — it refuses to hand back a set with a hole in it. So `NO_CLASSES` means "this module did not compile", which is a different thing from `RUNTIME_ABSENT`'s "the runtime is not on this host's classpath".

Both still leave the block under `Incompleteness.Code.RUNTIME_ABSENT`, which as you say loses information. Separating them needs a new code and a message for it, so I have left the mapping alone and made the comment state plainly that this is as fine a distinction as the report currently has, rather than implying the two are the same. Happy to file it as a follow-up alongside the candidate-search asymmetry if you want them tracked.

Clean build green (3691 + 185 + 109 + 1). Nothing else changed — the corpus figures and the generator output are the same as in `ce40bd46`.

kawasima:
Rebased onto develop (#469 and #471 landed since this branched). No file overlap, but #461 touched the numeric-newtype analysis this reads, so I re-ran everything on the rebase: clean build green, and the corpus figures are unchanged — 82 → 32 not known to be writable, 138 → 166 rows owed, no report errored.

Follow-ups filed: #472 (the schema names a code the compiler does not emit — a fix to the file, no version bump), #473 (a bounded newtype offers only its low edge, so a pair is found from one side only), #474 (NO_CLASSES reported as a missing runtime — needs a schema version, and carries the `writability` field this PR deferred as a second passenger).

PR #471, opened 2026-08-08
The value envelope is what a sum adds, not what a newtype is

### LINKED ISSUE #463: A newtype's representation changes when an unrelated sum names it
## What the specification says

A case's contents are decided by the case's shape (`[#sum-discrimination]`): a product lays its
fields beside `"type"`, a unit writes `"type"` alone, a newtype puts its contents under `"value"`.
And a standalone newtype is not in a sum, so it stays bare — `[#sum-discrimination]` says it in as
many words:

> A standalone newtype (not a sum case) stays bare (`3`); the `"value"` wrapping appears only inside
> a sum.

## What happens

Adding a sum that names a type changes what that type writes somewhere else. Measured through
`souther run`, on a behavior that answers the newtype itself and has nothing to do with the sum:

```souther
data Code = String
behavior f : (n: Int) -> Code constructs Code
let f (n) = Code("x")
```
```
"x"
```

```souther
data Code = String
data Z = { z: Int }
data S = Code | Z            // ← the only difference
behavior f : (n: Int) -> Code constructs Code
let f (n) = Code("x")
```
```
{"value":"x"}
```

The same behavior, the same type, the same position. A declaration elsewhere in the module changed
what an existing boundary publishes, and the second form is the one the specification says appears
only inside a sum.

## Why

The `"value"` envelope has two producers, and which of them runs is decided by whether the type is a
case of some sum.

`Deriver.deriveEncoder` takes an `isCase` flag. `bareField` answers nothing when it is set, and the
`d.newtype() && !isCase` arm is skipped too, so a newtype that is a case falls through to the
`ObjectRaw` branch and the *derived encoder itself* writes `{"value": …}`. A newtype that is nobody's
case keeps the standalone derivation and writes its base's form.

`CodecGen` then has to put a discriminator on a member of a behavior's answer, and what it finds
depends on which of those two it was given:

```
Auth  (a case of S)   Deriver wrote {"value":…}   → CodecGen injects "type" into it
Plain (nobody's case) Deriver wrote {"n":1}       → CodecGen wraps it under "value"
```

Both reach `{type, value}`, by different routes — which is externally visible as a key order:

```
Auth  as a member of `Auth | Z`   →  {"value":{"n":1},"type":"Auth"}
Plain as a member of `Plain | Z`  →  {"type":"Plain","value":{"n":1}}
```

Semantically the same object, so this is a fingerprint rather than the defect: it is how you can see
from outside that one rule is being produced by two code paths.

## What `rawOutputType` is doing there

`flatMember` asks `encoderOutput(member)`, which projects the derived `Ast.RawExpr` onto a JVM
descriptor, and treats `CD_Map` as "the discriminator can go in here". That is not a lazy way of
asking about the case's shape. It is how `CodecGen` tells the two producers apart — an adapter for
the split, and it cannot simply be replaced by reading the shape.

It is also not sufficient for the rule it is used for. A newtype over a record produces an object and
is still wrapped:

```
data Rec     = { x: Int }   →  {"x":1,"type":"Rec"}
data OverRec = Rec          →  {"type":"OverRec","value":{"x":1}}
```

`rawOutputType` answers `CD_Map` for the record's `ObjectRaw` and `CD_Object` for the `EncodeRaw` a
newtype over a data gets, and *produces an object* does not separate them. So the projection is both
semantically insufficient and operationally necessary — which is what a split responsibility looks
like from the inside.

## The invariant to restore

The `"value"` envelope is not part of a newtype's representation. It is what membership in a sum
adds, which is why a standalone newtype is bare. So:

> `Deriver` derives the standalone representation of a type. A sum's encoding adds what membership in
> that sum requires.

Under it, a derived encoder does not depend on whether some other declaration names the type — which
is the property this issue is really about — and the sum side is a function of the case's shape:

```
product   → { "type": name, ...object }
unit      → { "type": name }
newtype   → { "type": name, "value": <standalone representation> }
primitive → { "type": name, "value": <standalone representation> }
```

Three cases, not two. `flatMember`'s boolean is the shape of the adapter rather than of the rule.

## Order

1. `primKind` — done, and independent of all of this.
2. The specification — done. It already says wrapped, and it already says standalone is bare.
3. Expose the two producers: a test that adding `data S = Code | Z` does not change what `Code`
   writes, which fails today.
4. Give the envelope one owner — the sum side, per the invariant above.
5. Make derived encoders independent of `isCase`.
6. Decide a union member's representation from the case's shape.
7. Drop `flatMember`'s dependence on `rawOutputType`.

4 and 5 are what makes 6 possible; doing 6 first would be writing the adapter's behaviour down as if
it were the rule.

## Why this belongs to #446

The decision is made where it is made, and what is downstream receives it rather than inferring it
from something weaker. Here the decision was made in two places at once, and `CodecGen` recovers
which one ran by looking at the shape of what came out.

Part of #446.

## Where this stands

`primKind` is done, on `feature/issue-463`, and it is independent of everything above: how a
primitive is written is one table now, read back through the inverse of the mint a primitive case
name comes from. A primitive standing as a member of a behavior's answer was uncovered and is now
pinned at the form it writes.

The rest is not a migration from one side to the other. It replaces a contract:

```
was:  being a case changes the derived codec of the type
is:   the derived codec is the standalone representation, and membership adds the envelope
```

So it goes in one step or not at all. Deriving without `isCase` and leaving the sum encoder on the
old contract breaks immediately; fixing the sum encoder without the decoder breaks the symmetry
between them; and stopping before the union path shares the rule leaves the duplication this issue is
about in another shape.

**Read the decoder side before changing anything.** Twice now the abstraction visible in the code has
been shallower than the contract a measurement showed, so the shared unit is worth deciding after
every path has been read, not before.

1. Enumerate what `deriveEncoder` and `deriveDecoder` do with `isCase`.
2. Confirm what `generateSumEncoder` requires of a case's encoder today (it `checkcast`s a `Map`).
3. Read `generateSumDecoder` against raoh's `discriminate` / `variant`: whether a case's decoder can
   be handed the value under `"value"` rather than the whole object.
4. Find the case rule `emitMemberEncode` and the sum path have in common.
5. Decide the shared case-shape operation. **No code changes before this.**
6. Switch `Deriver`, the sum encoder and the sum decoder together.
7. Regressions: a standalone newtype unchanged by another sum naming it; product, unit, newtype and
   primitive members of a sum and of a union answering the same rule; encode/decode round-trip.

The key order a newtype case writes moves with step 6 — `{"value":…,"type":…}` becomes
`{"type":…,"value":…}`, which is the form the specification's own examples show. Existing exact-string
fixtures are updated to it; nothing here makes object key order part of what the language guarantees.

kawasima:
Closes #463. Part of #446.

## The defect

Adding a sum that names a type changed what that type wrote somewhere else:

```souther
data Code = String
behavior f : (n: Int) -> Code constructs Code   →  "x"

data Code = String
data Z = { z: Int }
data S = Code | Z            // the only difference
behavior f : (n: Int) -> Code constructs Code   →  {"value":"x"}
```

The second form is the one the specification says appears only inside a sum.

## Why

The `"value"` envelope had two producers. `Deriver.deriveEncoder` took an `isCase` flag, and a
newtype that some sum listed fell through to the `ObjectRaw` branch, so the derived encoder wrote
`{"value": …}` itself; a newtype nobody's case kept the standalone derivation. `CodecGen` then told
the two apart by projecting the derived encoder onto a JVM descriptor and asking whether it was a
`Map`.

That projection is not the rule it was used for. A newtype over a record produces an object and is
wrapped all the same:

```
data Rec     = { x: Int }   →  {"x":1,"type":"Rec"}
data OverRec = Rec          →  {"type":"OverRec","value":{"x":1}}
```

So it was semantically insufficient and operationally necessary at once — an adapter for a split
responsibility rather than a reading of the rule.

## What this does

A derived codec is the standalone representation of its type. A sum's encoding adds what membership
in that sum requires. The question is asked of the declaration, in one place:

```java
TypeOps.caseShape(TypeName, Symbols) -> PRODUCT | UNIT | WRAPPED
```

- `Deriver` no longer asks whether some other declaration names the type.
- `CodecGen.emitTagged` writes a discriminated case, shared by a named sum's encoder and a
  behavior's anonymous answer — the same rule, previously written twice with different arms.
- The sum decoder is the mirror: a wrapped case is handed what is under `"value"` through
  `field(…).asDecoder()`, the pair the discriminator's own tag is already read with. Raoh's
  `discriminate` passes the input whole, so the projection belongs where the variant is built, and
  it composes the same way for all three sources.
- `flatMember` is gone. `rawOutputType` is back to answering what an encoder returns, which is what
  `encoderSig` needs it for.
- `recordCompatible` reads the case's shape rather than the derived decoder, so a newtype case of a
  sum keeps the jOOQ column it read.

## Two readers the issue did not list

**A fixture wrote the envelope from the case's sums, not from its position.** `NeutralForm` was a
third producer of the same rule and had the same defect for newtypes — only unit cases consulted the
position. A row written at the case's own type produced a form that type's decoder refuses. The
position now reaches both directions, including the written one, where `expected` already flowed as
far as `raw` and was dropped at `record`.

**A field typed by the case read `{"value":3}`.** It now reads `3`. `CompileNewtypeOrderingTest` had
that asymmetry written down as its expectation — it fed `{"type":…,"value":…}` and asserted
`{"value":…}` came back, with a comment noting the tag was not written out — and is updated to the
rule.

## Boundary version

`BOUNDARY_VERSION` moves 9 → 10. A jar built before this carries a `$Enc` that writes the envelope
itself, which this compiler's sum encoder would wrap a second time, and a `$Dec` that demands it,
which the sum decoder would hand the inner value to.

## Verification

`mvn clean test` is green across every module (3635 tests). Each of the three changed decision
points was mutated and killed by a test; `NeutralForm`'s was not covered when first written, which
is why `AFixtureWritesTheFormItsPositionReadsTest` writes both positions — a row that only ever
stands in a sum passes under either rule.

Measured through `souther run`, the table the issue states:

```
Rec      →  {"x":1,"type":"Rec"}
Absent   →  {"type":"Absent"}
Code     →  {"type":"Code","value":"x"}
OverRec  →  {"type":"OverRec","value":{"x":1}}
```

Every example row in `souther-examples` member / ordering / hr — the three that declare a newtype
case — still holds.

The key order a newtype case writes moves to `{"type":…,"value":…}`, the form the specification's
own examples show. No fixture pinned the old order, and nothing here makes key order part of what
the language guarantees.

## Specification

Two sentences in `[#sum-discrimination]` and `[#custom-codec]` described the old mechanism — the
discriminator handing the case's decoder the whole object for it to read `"value"` out of. The
format they state is unchanged.


kawasima:
Fixed in 676be574. Thanks — the jOOQ finding was right, and chasing it back turned up the rule it belongs to.

## The root cause

The question is not *which shape is this case*, it is **which decoder reads a value found under a key of a source** — and taking a value out from under a key leaves the source behind. A jOOQ row is flat, so what is under a key is a column and not a row.

That question was already answered, in the arm `emitDecoderObject` uses for a data-typed field, jOOQ comment and all. I wrote a second answer beside it in `generateSumDecoder`, which is the very thing this PR is about. So rather than adding a shape × source table, the two are now one operation:

```java
private void emitUnderAKeyDecoder(CodeBuilder code, Ast.Name typeName, Src src)
```

- **JOOQ** → `decoder()`. A column is read by the type's own `Object` decoder, and a type that is not a whole row has no `recordDecoder()` to reach for anyway.
- **NEUTRAL** → `decoder()`, bridged through `MapDecoders.nested` when the type reads a `Map`.
- **JSON** → `jsonDecoder()`. A field of a `JsonNode` is a `JsonNode`, so this source alone carries through.

Both `emitDecoderObject` and the sum decoder's wrapped branch call it.

## Regression tests

`ACaseIsReadByTheDecoderForWhatItIsHandedTest`, two rows:

- A record-compatible sum with a primitive-backed newtype case, read through `recordDecoder()` from a real jOOQ `Record` with `type` / `value` columns. It reproduced `NoSuchMethodError: demo.Code.recordDecoder()` before the fix, and reverting the fix kills it again.
- A newtype over a data as a case, handed a non-object under `"value"`, reports `type_mismatch` at `/value`.

The NEUTRAL half I checked before writing it down: the object guard is already there one level in, through the newtype's delegation to the inner type's decoder. So the second row is a guard rather than a second defect — only the jOOQ path was broken.

## The javadoc

You are right that the fixture does not read `caseShape`, and that it should not. Its question is which form a value takes at the *position* it is written in, which the declaration's shape does not answer on its own — the same newtype is bare at its own type and wrapped inside a sum. The claim is narrowed to what code generation does, with a paragraph saying why the fixture is a different question.

`mvn clean test` green across every module.

PR #475, opened 2026-08-08
A witness holds a name the boundary admitted, and nothing guards against one it did not

kawasima:
Part of #446 — the two items that were left after #462 and #463 landed.

Audited the issue's scope against the tree first. Scope 1, 2, 4, 5 and 6 are in place: `Sig` is a final class with a package-private constructor and the types are read off it; the six `SpecChecker.reject*` passes are gone and the position-ownership table (`Map<Option<Int>, String>` → E1314, `List<Option<Int>>` → E1313, `Map<String, Option<Int>>` → E1313) is pinned in `OneBoundaryAnswerReachesTheReaderTest`; an imported signature comes from the same `PipelineSigs.signatures` walk with no serialized witness; `SignatureBoundary` has two callers and there is no `BoundaryShapes` left. What remained was the half of scope 3 the issue names, and the guards it said would follow from it.

## What a witness is worth

Two readings were both being stated, and they cannot both hold. Either a signature is the capability — a shape may be assembled by anyone, and the closed `Sig` constructor keeps a forged one from ever being *raised* to a signature — or a witness is itself the evidence, and holding one means the name in it was admitted.

This takes the second, everywhere. #446 asks for it in as many words ("the witness cannot hold an invalid state … constructible without validation, which leaves the readers guarding against them"), `BoundaryScalar` has been it since #460, and a reader is handed a union's case, a list's element or a map's key at least as often as a whole signature — so a guarantee that held only of `Sig` is not the one the readers rest on.

**Closing the constructors.** A record cannot be closed: its canonical constructor is as accessible as the record. So `Nominal` and `Cases` are written out as final classes with package-private constructors, and `BoundaryInput` / `BoundaryOutput` move from `types` to `check`, beside the walk that admits a name — a `check` type held by a `types` type would be a package cycle, so the move is what makes it possible at all. The cases carrying no name stay records.

**Closing a constructor is not enough where a public function hands the value out.** `TypeOps.classifyConcreteMapKey` answered a `BoundaryMapKey`, and what it looked at was whether the type *has* a representation — which a name of the language's own has as much as a model's. `SignatureBoundary` proved the gap by re-asking `declaredByAModel` of the classified name before using it.

So the two questions become two types:

```
Type
 ↓ classification — a fact about the type, anything may work it out
MapKeyRepresentation          (types, public records, carried by the codec IR)
 ↓ admission — the position's own obligation
BoundaryMapKey                (check, one closed constructor, made by SignatureBoundary)
```

The readers follow the question they are asking: `Ast`, `Deriver` and `CodecGen` take the representation; `Runner` and `FixtureShape` take the witness and branch on what is inside it. This is #462's principle in the general form — a key position is established by a boundary, by a data field and by a fixture, and they do not admit alike, while the representation is what all three share. It also removes a package cycle: carrying the witness into `Ast.MapDecRef` / `Ast.MapEnc` had made `ast` depend on `check`, which depends on `ast`.

## The guards that were left

`run.decode.nodecoder` and `run.encode.noencoder` asked whether the class a name denotes has the factory being reached for. Measured over every shape a witness has a name in — a record, a newtype, a unit, an enumeration, a field-bearing sum, a union a behavior answers with, and both kinds of named map key — in both directions, which is every place the runner turns a name into a factory. Neither is reachable: the name came out of the witness, so a model declared it, and this compilation generated its class with both factories. `EveryNameAWitnessCarriesHasItsCodecTest` holds that; the two catalog entries and their messages are gone, and a reflective failure is now an `IllegalStateException` about the class this run itself emitted.

`CodecGen.jsonCompatible` goes with them, and is why the decode guard looked reachable. It was the one thing that could have withheld a `jsonDecoder()`, and it had stopped implementing the rule its own comment states — "JSON has no temporal leaf, so a type is JSON-decodable iff no `Date`/`DateTime` appears anywhere in its shape". `jsonOkType` answers yes to a temporal, so it answered yes to everything. Measured before removing it: a `Date`-bearing data already had a working `jsonDecoder()`.

```souther
data Booking = { on: Date, who: String }
behavior confirm : (b: Booking) -> Confirmed constructs Confirmed
```
```
$ souther run temporal.sou --behavior confirm --input '{"on":"2026-01-01","who":"a"}'
{"who":"a"}
```

## Tests

`AWitnessIsMadeWhereANameIsAdmittedTest` pins the closure of every witness — `Sig`, both `Nominal`s, `Cases`, `BoundaryMapKey`. javac is the real enforcement, but nothing stopped a constructor being widened back, and the closure is what this PR is. Checked it is not vacuous: widening one to public fails it by name.

`EveryNameAWitnessCarriesHasItsCodecTest` measures a field-bearing named sum as well as an enumeration — `ValueClassGen.generateSum` branches on `isUnitOnlySum`, so the two go down `generateEnumSumDecoder` against `generateSumDecoder` and answer a bare name against a discriminated object.

One thing is **not** tested, deliberately. Of the two questions a key position asks, only the first has an example in today's library: a name with no representation at all is refused as that (E1314, already pinned), and there is no type the language declares of its own that a key *could* be classified as — `DivisionByZero` is a unit data rather than a unit-only sum, so it never reaches the admission arm. That obligation is the position's whether or not anything can reach it now, which is why it is written where the position is, but no test here claims to exercise it.

`BOUNDARY_VERSION` does not move. Nothing changes about what is emitted, only about what could be asked for.

## Not in this change

`BoundaryOutput.Cases` carries members as written rather than leaf-expanded. That is about what `CodecGen` needs beside the members rather than about admission, and ADR-0100 keeps it as an open item.

ADR-0100 is revised in place rather than superseded: this is the same decision carried to the cases it had not reached, it now states the choice between the two readings as a choice, and its "what is still open" section is updated to what #462 and #463 actually settled — including that `BoundaryNominal`'s codec descriptor was refuted by measurement, so what was left of that case was the name alone.

`mvn clean test` green across every module.


kawasima:
Fixed in 8ea42dbb. All three, and 1 was the right thing to press on — the inconsistency was mine, introduced by this PR.

## 1. Which model, and the same one everywhere

Confirmed all three parts of the claim. `Sig`'s javadoc did state model A verbatim ("The shapes themselves stay open: what matters is what may be raised to a signature, not what may be described"), `BoundaryMapKey.StringNewtype` and `UnitEnum` were public records, and `BoundaryInput`'s javadoc — which I wrote in this PR — claimed a map is built out of witnesses already admitted, which was false of exactly that case.

Went to **B**, everywhere. The reasons it is not a coin flip:

- #446 asks for B in as many words: "the witness cannot hold an invalid state … constructible without validation, which leaves the readers guarding against them."
- `BoundaryScalar` has been B since #460 — six cases, no way to name `Raw` — so A would leave that half unexplained.
- A would make closing `Nominal` and `Cases` pointless, which is the work the issue asks for.
- A reader is handed a union's case, a list's element or a map's key at least as often as a whole signature. A guarantee that held only of `Sig` is not the one the readers rest on.

So `BoundaryMapKey` moves beside the walk that admits a name, and its two naming cases are closed like the others. The three that name none stay records — each stands for a representation the boundary always admits, so there is no refused state to assemble. Nothing in `types` referenced it, so the move is not a cycle; the only thing that ever constructed one is `TypeOps.classifyConcreteMapKey`.

`Sig`'s javadoc now says why the shapes are closed rather than why they need not be, and ADR-0100 states the choice as a choice — the two readings, why they cannot both hold, and which one this takes.

## 2. The sum that was only ever an enumeration

Right, and the reason matters: `ValueClassGen.generateSum` branches on `isUnitOnlySum`, so an enumeration and a field-bearing sum go down `generateEnumSumDecoder` against `generateSumDecoder` and answer a bare name against a discriminated object. The test claimed a sum among the shapes it measured and had only the first. `Decision = Approved | Rejected` with payloads is measured now, as a nominal input and output.

## 3. The change itself was not regression-tested

Also right — `EveryNameAWitnessCarriesHasItsCodecTest` passes with the constructors public, so it never held the thing this PR is. `AWitnessIsMadeWhereANameIsAdmittedTest` pins the visibility of every witness that names a type, plus `Sig`. Went with reflection rather than an in-process `JavaCompiler`: the test lives in a different package, so what it can honestly say is that the door is shut, and javac is still the enforcement.

Checked it is not vacuous — widening `StringNewtype`'s constructor to public fails it with the message naming that case, and reverting restores green.

`mvn clean test` green across every module.

## Still not in this change

`BoundaryOutput.Cases` carrying members as written rather than leaf-expanded. That one is unrelated to admission — it is about what `CodecGen` needs beside the members — and ADR-0100 keeps it as an open item.

kawasima:
Fixed in 9131a6f1. The blocker was real and the diagnosis was exact — `SignatureBoundary` re-asking `declaredByAModel` of the classified name was the proof, sitting right there in the code I had just read.

Took the type split rather than pushing `declaredByAModel` into the classifier, for the reason given: a key position is established by a behavior's boundary, by a data's field and by a fixture, and they do not admit alike, while the representation is what all three share. Folding admission into the classifier would put one position's policy in front of all of them.

```
Type
 ↓ classification — a fact about the type, anything may work it out
MapKeyRepresentation          (types, public records, carried by the codec IR)
 ↓ admission — the position's own obligation
BoundaryMapKey                (check, one closed constructor, made by SignatureBoundary)
```

`Ast`, `Deriver` and `CodecGen` take the representation; `Runner` and `FixtureShape` take the witness and branch on what is inside it. The `ast ↔ check` cycle goes with it, as you expected — the representation is a fact about a type and belongs in `types`, where `Ast` could already reach. The structural test follows the type: one witness, one closed constructor, rather than five cases.

**On testing the admission half.** I tried to pin it and could not, honestly. `Map<DivisionByZero, Int>` does not reach the admission arm — `DivisionByZero` is a unit data rather than a unit-only sum, so it has no key representation and the position's *first* question refuses it with E1314. For the second question to fire, the language would have to declare a String-backed newtype or a unit-only sum that a signature can name, and it declares none. So the `nominal()` call inside `mapKey` has no reachable example today. I removed the test I had written rather than let it assert E1314 under a name claiming it showed admission, and said so in ADR-0100 and the PR body instead. The obligation stays where the position is either way.

**PR body** is updated — the `Not in this change` line about `BoundaryMapKey` was contradicting the second commit, and is gone.

`mvn clean test` green across every module.

PR #477, opened 2026-08-08
The words the schema allows are the words the compiler writes

### LINKED ISSUE #472: The shipped adequacy schema names a code the compiler does not emit
`Incompleteness.Code` is a published vocabulary: `adequacy-schema-1.json`
constrains `incompleteness[].code` to a fixed enum, and a build reading
the report validates against it.

The enum and the compiler disagree.

```
schema : value_unreadable, value_truncated, row_timed_out,  runtime_absent, ...
java   : VALUE_UNREADABLE, VALUE_TRUNCATED, ROW_UNDECIDED,  RUNTIME_ABSENT, ...
```

`ExampleVerifier` builds `Code.ROW_UNDECIDED` for a row that spent its
budget or whose evaluation did not answer, and `AdequacyReport.json`
writes every code as `name().toLowerCase()`. So a compile with an
undecided row emits `"code": "row_undecided"`, which `schema-1` rejects.

The schema has never described this. `row_timed_out` was the name before
the code was widened to cover an evaluation that did not answer at all,
and only the Java side moved. Nothing is deprecated and no consumer has
ever seen `row_timed_out` in output, so correcting the schema does not
raise `schemaVersion` — it makes the file say what has always been
emitted.

## Why it was not caught

`MainExamplesSubcommandTest.theEmittedJsonMatchesTheShippedSchema` walks
`properties` and `required` and checks that what is emitted has the
fields the schema demands. It does not check that a value belongs to an
`enum`, and its fixture has no incompleteness in it, so neither half of
the disagreement is visible to it.

Two things are worth holding, and they are not the same:

- every value the compiler can emit for an enumerated field is in the
  schema's enum — checkable from the Java enum, without needing a fixture
  that produces each one;
- the fixture the test runs actually exercises the shapes it walks.

The first is what would have caught this, and it is the cheaper of the
two: `Incompleteness.Code`, `MeasurementStatus`, `BoundarySide` and the
per-measure `Reason` enums are all written into the schema by hand today.

kawasima:
Closes #472.

## The defect

`adequacy-schema-1.json` allows `row_timed_out` for
`incompleteness[].code`. The compiler writes `row_undecided`. A report
carrying an undecided row fails validation against the schema shipped
beside it.

Reproduced end to end, not read off the source: a compile whose row does
not come back emits

```
emitted code : row_undecided   allowed=false
```

`f1f42ecf` renamed `ROW_TIMED_OUT` to `ROW_UNDECIDED` when a row stopped
being held to a clock and the code widened to cover an evaluation that
did not answer at all. The rename was right. The schema was not part of
the change and has promised the old word since.

## The fix, and why it is not a version

The schema is corrected. `row_timed_out` has never been emitted, so no
consumer has read it and none needs telling: the file described something
that was never true.

## What let it happen

The vocabulary is spelled twice — a Java enum, and a hand-written list in
the schema — with nothing tying them. There are twelve such fields.

I checked all twelve against the compiled enums. Eleven agree. That is
eleven pieces of luck rather than eleven pieces of evidence: the same
rename in any of them would have gone the same way.

`EveryWordTheSchemaAllowsIsOneTheCompilerWritesTest` holds the
correspondence, one entry per field naming the enum its words come from,
plus a third case holding the table complete so a new enumerated field
arrives as a failure rather than an omission. `implementation` is named
as the exception — it is written from a boolean and has no enum behind
it.

Not generated from the enums. A contract generated from an internal type
changes whenever the type does, and widening what a consumer must handle
is a decision about the contract rather than about the compiler. Both
decisions stay deliberate; only the accidental one is now impossible.

This is the shape `Adequacy.Kind` already describes — "the agreement is
held by a test rather than by reading one off the other" — and that
`CstLexer.keywords()` → `souther.tmLanguage.json` and
`Main.knownOptions()` → the usage text already follow.

## The second hole

Only one of the two things that hid this is the missing correspondence.
The report the schema is checked against is a clean compile, so
`incompleteness` is an empty array in it: even a value-checking version
of the existing test would not have produced `row_undecided`. A report
that carries one is now emitted and read.

## The rename

`theEmittedJsonMatchesTheShippedSchema` →
`theEmittedJsonHasTheShippedSchemaShape`. It checks required keys and
declared keys and has never checked a value. A name claiming to match the
schema is how a reader concludes validation is already covered — which is
roughly how this defect survived.

No JSON Schema validator dependency is added. The defect mechanism is
caught by the correspondence, and the fixture gap by the emitted report.

## Where else this could be

I swept the repository with the two questions — must a human remember to
change a second file, and does the build fail if they disagree. Eight
boundaries: six already protected (keyword highlighting, the message
catalogs, diagnostic codes against the specification, message keys
against the catalog, CLI options against the usage text, the doc set
index), and two not. This schema is one. The other is the ADR index, 99
files and 99 entries, in step by hand.

Two more were raised and checked: the stdlib's intrinsic keys against the
backend table turn out to be held in both directions by
`EveryKernelIsDeclaredTest`, and the LSP's advertised capabilities
against its dispatch are not held by anything — filed separately.

### DIFF
--- a/souther-compiler/src/main/java/souther/compiler/report/AdequacyReport.java
+++ b/souther-compiler/src/main/java/souther/compiler/report/AdequacyReport.java
@@ -642,22 +642,46 @@ public record AdequacyReport(int schemaVersion, String compilerVersion, Adequacy
 
     private static final JsonMapper JSON = JsonMapper.builder().build();
 
+    /**
+     * How this report spells an enumerated value on the wire.
+     *
+     * <p>Here rather than at each field, so that the words a consumer reads have one origin. The
+     * shipped schema names the same words in its own file and is held against this — against what is
+     * written, not against a second reading of the enum, because those are different things and only
+     * one of them is what a consumer sees. A spelling rule applied at ten call sites is ten places for
+     * the schema to stop describing the output while every one of them still agrees with the enum.
+     */
+    public static String word(Enum<?> value) {
+        return value.name().toLowerCase(java.util.Locale.ROOT);
+    }
+
+    /**
+     * The same, for the one field with no enum behind it.
+     *
+     * <p>Whether a behavior has a {@code let} is a boolean, and these are the two words it is written
+     * as. Having no enum is why it needs this more than the others rather than less: nothing else in
+     * the compiler says what these words are.
+     */
+    public static String implementationWord(boolean injected) {
+        return injected ? "injected" : "implemented";
+    }
+
     public String json() {
         ObjectNode root = JSON.createObjectNode();
         root.put("schemaVersion", schemaVersion);
         root.put("compilerVersion", compilerVersion);
-        root.put("status", status.name().toLowerCase(java.util.Locale.ROOT));
-        root.put("adequacy", adequacy().name().toLowerCase(java.util.Locale.ROOT));
+        root.put("status", word(status));
+        root.put("adequacy", word(adequacy()));
         ArrayNode modulesOut = root.putArray("modules");
         for (ModuleReport module : modules) {
             ObjectNode m = modulesOut.addObject();
             m.put("module", module.module());
-            m.put("status", module.status().name().toLowerCase(java.util.Locale.ROOT));
+            m.put("status", word(module.status()));
             ArrayNode gaps = m.putArray("incompleteness");
             for (Incompleteness gap : module.incompleteness()) {
                 ObjectNode g = gaps.addObject();
-                g.put("code", gap.code().name().toLowerCase(java.util.Locale.ROOT));
-                g.put("scope", gap.scope().name().toLowerCase(java.util.Locale.ROOT));
+                g.put("code", word(gap.code()));
+                g.put("scope", word(gap.scope()));
                 g.put("subject", gap.subject());
                 gap.at().ifPresent(where -> {
                     ObjectNode at = g.putObject("at");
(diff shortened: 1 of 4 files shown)


kawasima:
Both fixed in `13d851a8`.

## 1. Held against the writer, not against a second reading of the enum

You are right, and the first version had the defect this PR is about inside the test for it. What was held was `schema == lowercase(enum names)`, computed in the test, while the report computed the same rule separately at ten call sites. Changing the writer alone left every new check green.

`AdequacyReport.word(Enum<?>)` is the spelling now, used at all ten sites in `json()`, and the test reads its words through it. The chain is the one you drew:

```
Java enum -> AdequacyReport.word -> emitted vocabulary
                                         |
                                     agreement
                                         |
                                  published schema
```

The schema keeps its own list, so widening what a consumer must handle stays a separate decision from renaming a constant.

`implementation` was the worse half: excused from the correspondence because it has no enum, then written out a second time in the test. `implementationWord(boolean)` now names those two words in production and the test reads them from there. Having no enum is why it needed this more than the others.

I checked the new agreement is load-bearing rather than assuming it. Changing the encoder to `.replace('_', '-')` and nothing else:

```
everyEnumeratedFieldSpellsTheEnumItComesFrom  expected: <[satisfied, not-satisfied, undetermined]>
                                              but was:  <[satisfied, not_satisfied, undetermined]>
anUndecidedRowIsReportedInAWordTheSchemaAllows expected: <[row_undecided]> but was: <[row-undecided]>
```

Both fail. Before this commit, both passed.

Note the human report keeps its own formatting — one line lowercases and replaces `_` with a space for prose. That is a different surface with a different rule, and it is deliberately not routed through `word`.

## 2. The regression names the word

`assertEquals(List.of("row_undecided"), written)`. As you say, the role here is to keep one reproduction alive rather than to check the schema generally, so the specific value belongs. A fixture that stopped producing an undecided row and produced some other legitimate gap instead would have gone on passing.

Full clean build green.

Issue #474, opened 2026-08-08
`runtime_absent` at source scope does not mean the runtime is absent


kawasima:
Follows #470. The first reading of this was that `Adequacy.Generated.offered`
writes two distinct `Attempt.Reason`s under one `Incompleteness.Code`:

```java
NO_CLASSES      // the module's classes were not there to build against
RUNTIME_ABSENT  // the runtime is not on this host's classpath
```

That is true of the code and it is not where the defect is. `NO_CLASSES`
never reaches a report, and the code it collapses into was already carrying
something else.

## `runtime_absent` at source scope

Two places turn "this source produced no observation" into a reason —
`AdequacyReport.moduleReport` and `Adequacy.rowsOf` — and both write
`Code.RUNTIME_ABSENT`. `Output.Examples.evaluate` goes absent for five
conditions:

```text
the frontend did not answer
the module did not type-check
the classes are not there
the requirements are not there
a value here is already declared by the module
```

None of them is the runtime being absent, and at that point none of them
can be: a real `LinkageError` is caught per example inside
`ExampleVerifier` and recorded at behavior scope, so it never makes a
source absent. The code names a cause it cannot have.

`CompilePartialAdequacyTest` already asserts this code, and the compilation
it asserts it on has one diagnostic — E1906, a `let` declared twice in an
attached file. The runtime is present, the module generated, and both
boundary rows built.

So the vocabulary reads by scope rather than by code:

```text
(behavior, runtime_absent)   the runtime could not be reached
(source,   runtime_absent)   nothing was observed here, for any of five reasons
```

The consumer is not the problem. `Observed.someRowsUnseen` asks whether
rows exist that nothing read, which is a question about the state and not
about the cause, and the second meaning is the one it wants. A cause was
named first, then reused as the state, and each new producer widened it —
`from either place it arises` in `Adequacy.someRowsUnseen` is where that
was written down.

What the report should distinguish is the two states, not the five causes.
Why a compilation failed is already the diagnostics' answer, and a second
taxonomy of it in the adequacy schema would be two files to keep in step.
`observation_absent` says what the source-scope entry means;
`evaluation_absent` follows the shape of `Output.Examples.evaluate` too
closely, and `source_unobserved` repeats what `scope` already says.

## `NO_CLASSES` is not a candidate for the schema

Adding it would publish a state no artifact can carry.

`attemptAt` reaches it only where `worthBuilding(coverage)` holds and the
[...]

Issue #484, opened 2026-08-08
A selector that matches nothing reports `complete`, and `--strict` passes


kawasima:
A selector that matches nothing reports that everything it measured was
complete, and `--strict` passes.

```text
$ souther examples --format json --module example.nosuch --strict model.sou
{
  "schemaVersion" : 1,
  "compilerVersion" : "unreleased",
  "status" : "complete",
  "adequacy" : "undetermined",
  "modules" : [ ]
}
$ echo $status
0
```

The model compiles, has a behavior, and has a row that leaves a gap — run
without the selector the same command answers `not_satisfied`. Misspell the
module name and the gap is gone, the report says `complete`, and the build
gate that exists to refuse a gap lets it through.

`--behavior nosuch` does the same thing one level down: the module is kept,
its behaviors are all filtered out, and what is left says `complete`.

## Where it comes from

`AdequacyReport.only` starts at `COMPLETE` and folds the shown behaviors
into it, and `AdequacyReport.of` does the same over the modules. Over an
empty selection the fold returns what it started with. As a predicate that
is right — everything selected was completely assessed, and nothing was
selected — but `status` is not a predicate. It is the word a person and a
build read to find out whether the measurement happened.

So the two are different statements and only one of them is `only`'s to
make:

```text
a name that matches nothing leaves an empty report     selection
an empty report is complete                            assessment
```

`only` is documented for the first and is not wrong about it. The second is
the status aggregation's, and it has no case for a selection that resolved
to nothing.

## What it is not

Not partiality. `partial` means the subject exists and part of it could not
be measured, and there is no subject here at all — an empty selection and a
module whose rows went half-read should not answer with the same word.

`adequacy` already answers `undetermined`, which is the right shape: it has
a case for "no measure was asked that could find a gap". What it does not
have is a reason to refuse when the selection itself came to nothing, so
`--strict` reads `undetermined` and exits 0.

## Worth deciding together

Whether `status` needs a word for this, or whether an empty selection should
be refused before a report is made of it. A selector naming something the
compilation does not have is arguably a usage error — `souther examples
--module` with a name no source declares is closer to a misspelled path than
to a measurement result, and answering it with a document at all is what
makes the rest of this possible.

Either way `--strict` must not exit 0 over a selection that assessed nothing.
That part does not wait on the vocabulary question.

Independent of #482 — this reproduces on a compilation that raised no
diagnostic at all, and does not touch `incompleteness[]`.

## Acceptance criteria

[...]

PR #486, opened 2026-08-08
The report is what was observed, not whether it compiled

### LINKED ISSUE #482: `souther examples` suppresses its own report for every failure the report exists to describe
The adequacy report carries `incompleteness[]` — one entry per reason a
measure could not read what it needed. Nothing a person reads through
`souther examples` ever has one in it.

`Main` renders the report from `Compiler.compiled`, which throws on the
first error in the compilation. Every producer of an incompleteness sits
behind a condition that raises one:

```text
duplicate declaration in an attached file   E1906
a module that did not type-check            E1321 and its neighbours
a composition that reaches itself           E1608
a row that did not come back                E1923
classes that could not be generated         the backend's own CompileException
```

E1923 is built without `.warning()`, so it takes the builder's default of
`Severity.ERROR` like the rest. Run through the CLI each of these exits 1
with the diagnostic and prints no report; a clean compile prints
`"incompleteness": []`.

The one remaining producer is `ExampleVerifier`'s `LinkageError`, which is
the only place the name `runtime_absent` is accurate. It needs the runtime
off the classpath while the compiler runs, and the path where that happens
is the annotation processor — which builds no report. `AdequacyReport.of`
has one caller outside the tests.

So the condition that fills the field and the condition that publishes it
exclude each other.

`status: partial` is not in the same position, and that is what makes this
a defect rather than a design. Axis omission, pair truncation and a value
read past its limit all reach the CLI report, because none of them raises
a diagnostic. The schema publishes partiality and cannot publish
incompleteness, and nothing decided that — it is what the pipeline came to.

## What the gate should be

Not the compilation succeeding. A parse failure leaves nothing to say
which model the rows were meant to cover, and a report over it would be a
report about nothing. What adequacy needs is that the semantic model
formed; a failure in a later phase then costs particular evidence rather
than the answer.

`AdequacyReport.of` already behaves that way when it is asked. Asked
directly, past the throw:

```text
parse failure             modules=0  status=complete  adequacy=undetermined
name resolution failure   modules=1  status=partial   incompleteness=[runtime_absent/source]
type formation failure    modules=1  status=partial   incompleteness=[runtime_absent/source]
```

Nothing throws, and the gate falls out of the evidence rather than being
written: a source that did not parse declares no module, so there is
nothing to report on. The suppression is entirely in the throw ahead of it.

The empty report is wrong about itself, though. With no module to be
partial about it answers `complete`, which reads as everything having been
measured. Whatever this issue settles, that status has to come from
somewhere other than the absence of modules.

Emission and exit status are separable and the exit contract does not have
to move. A compile error is already 1 and an adequacy gap is 1 only under
`--strict`; only the report is tied to the throw. Read the three apart:
what was observed goes to the artifact, whether the command succeeded goes
to the exit code, and what is wrong with the source goes to the
diagnostics.

## What this does not yet buy

Changing the gate alone publishes reports with almost nothing in them.

Evidence is recovered per module. `Bodies.Checked` is absent for the whole
module, so one behavior that does not type-check costs every row
observation in it — including a behavior beside it that checks and whose
rows are fine. A module with two behaviors, one of them broken, reports
both names, a source-scope reason, and no rows at all. Honest, and close to
useless.

Making the field carry its weight is the same direction the frontend
already took with its error type: keep going past the failure and record
what was lost. That is separate work and wants its own issue. The contract
belongs first, because it says what the recovery is for.

## What is blocked on this

#474 corrects `runtime_absent`, which at source scope means "no
observation" and cannot mean what it says. #472 corrects a code the schema
names and the compiler does not emit. Both are about a field whose
contents no artifact currently carries, so what they are worth follows from
this decision — and if the answer is that a compilation with an error has
no adequacy report, then `incompleteness[]` describes states the artifact
cannot reach and the question becomes whether to keep it at all rather than
how to name its members.

kawasima:
Fixes #482.

`souther examples` took its compilation from `Compiler.compiled` /
`compiledModules`, which raise the first error they find. Those entry points
are `Compilation.answerEverything()` plus three throws, so the command got
the work it needed and a contract it did not: the specification says
"Nothing here changes what compiles; it reports", and the converse was
never implemented.

What that cost is the `incompleteness` field. It carries what stopped the
rows from being observed, and run through the CLI each of those exits 1
with the diagnostic and prints no report:

```text
duplicate declaration in an attached file   E1906
a module that did not type-check            E1321 and its neighbours
a composition that reaches itself           E1608
a row that did not come back                E1923
classes that could not be generated         the backend's own CompileException
```

One producer carries no diagnostic — the runtime not being reachable,
which no source is at fault for — and it needs the runtime off the
classpath while the compiler runs, which is the annotation processor's
situation, and that path builds no report. So the field a report could
carry and the report a compilation could produce excluded each other, and
`AdequacyReport.adequacy` — which refuses `satisfied` where the field is
not empty — has never had the chance to fire in a published report.

What stopped a *measure* from reading what was observed is a separate half,
said by that measure: an axis past the limit, a combination count that
stopped at its own, a value too large to read back. Those reach the report
today, carry no `incompleteness` entry, and are why `partial` was never in
the same position. An empty `incompleteness` has therefore never meant that
everything was read, and the specification now says which half is which.

## What changed

`Compiler.analyzed` / `analyzedModules` drive the same compilation and
raise nothing. `Compilation.errors(...)` sits beside `warnings(...)` and
answers with all of them where `firstError` answers with one.
`examplesSubcommand` routes through those and decides its three results
apart: the report to stdout wherever the compilation formed something to
assess, the diagnostics to stderr as before, the exit code unchanged — an
error is still 1, and `--strict` still decides the exit code of a
compilation that raised none.

Whether there is anything to report on and what the report shows are asked
of different things — the first of everything that formed, the second of
what the selection named. Reading the first off the filtered report would
put them back under one condition, which is the shape of the defect this
change is about, and would leave the emission gate depending on the
selector having been resolved somewhere else first.

`Compiler.linked` and `compiled` keep their fail-fast contract. `souther
compile` is untouched.

The specification gains `[#example-report-emission]` and a normative
account of `incompleteness`, which it had never described at all — its
description of the JSON form covered `status` and `reason` only.

## Behaviour that changes

- A failing invocation now writes a report to stdout alongside a non-zero
  exit status. A consumer treating empty stdout as "no report" reads the
  exit status instead. This is a change to the process I/O contract, not to
  the schema: `incompleteness[]` has never been emitted, so no document
  shape a consumer has seen is altered.
- `souther examples` now prints every error rather than the first. A
  command that goes on to describe the whole compilation should not show
  one error beside an account of everything else.
- A source that declares no module gets no report. It has no subject for a
  measure to be about.

## Not in this change

A selector that matches nothing is refused ahead of the report, with its own
exit code. That landed in #487 while this was open, and this rebases onto it:
the two rules meet in a test here, where a selector that resolves gets its
report from a compilation that did not.

The test asserts that the entry exists and is about the source, not which
word it carries. `runtime_absent` at source scope does not mean the runtime
is absent, and that is #474's to fix now that the field is reachable.

Evidence is still recovered per module: `Bodies.Checked` is absent for the
whole module, so one behavior that does not type-check costs every row
observation in it. A report that says `partial` and carries little is
honest and not yet worth much. That is the next piece of work and wants its
own issue.

## Verification

`mvn clean test` — BUILD SUCCESS across all eight modules.

Six tests in `ACompilationThatFailedStillReportsWhatItsRowsCoverTest`, and
`ASelectorThatResolvesToNothingIsRefusedTest` still green beside them. The
emission and `incompleteness` tests were watched failing against the old
routing before this was written.

### DIFF
--- a/souther-compiler/src/main/java/souther/compiler/Compiler.java
+++ b/souther-compiler/src/main/java/souther/compiler/Compiler.java
@@ -219,6 +219,44 @@ public final class Compiler {
         return compilation;
     }
 
+    /**
+     * The compilation of one source, driven as far as it goes, with nothing raised.
+     *
+     * <p>Beside {@link #compiled} rather than in place of it. A caller about to write classes out has
+     * to stop at the first error, there being nothing to write; a caller about to say what the rows
+     * cover has an answer either way. The specification puts the two apart — an example report
+     * "changes what compiles" not at all — and this is the half of that the entry points did not
+     * have. What refused the compilation is refused still: it is in the reports, and the caller reads
+     * it there rather than catching it.
+     */
+    public static Compilation analyzed(String source, String defaultModuleName,
+                                       List<Located> warningsOut, Adequacy.Asked measure) {
+        return answered(Compilation.ofSource(source, defaultModuleName), warningsOut, measure);
+    }
+
+    /** As {@link #analyzed}, for a module set resolved against {@code path} — what
+     *  {@link #compiledModules} is to {@link #compiled}. */
+    public static Compilation analyzedModules(List<String> sources, ModulePath path,
+                                              List<Located> warningsOut, Adequacy.Asked measure) {
+        return answered(Compilation.ofSources(sources, path), warningsOut, measure);
+    }
+
+    /**
+     * Asks the whole compilation and collects its warnings.
+     *
+     * <p>The warnings are collected whatever the compilation came to, which is where this parts from
+     * the raising entry points: they reach the collection only by getting past every error, so a
+     * compilation with one reports no warnings at all. Nothing about a warning depends on the errors
+     * beside it.
+     */
+    private static Compilation answered(Compilation compilation, List<Located> warningsOut,
+                                        Adequacy.Asked measure) {
+        compilation.measure(measure);
+        compilation.answerEverything();
+        warningsOut.addAll(compilation.warnings(compilation.db().allReports()));
+        return compilation;
+    }
+
     private static Map<String, byte[]> classesOf(Compilation compilation) {
         return new LinkedHashMap<>(compilation.classes());
     }
@@ -272,8 +310,8 @@ public final class Compiler {
      * The compilation of a module set, driven to completion, with the first error raised — what
      * {@link #linking} returns the classes of, for a caller that wants more than the classes.
      *
-     * <p>{@code souther examples} is such a caller: it asks how well the rows cover the model, which
-     * is read off the same compile that would have written the classes out.
+     * <p>For a caller that has to have a compilation there is nothing to go on with. A caller that
+     * can say something either way takes {@link #analyzedModules} instead.
      */
     public static Compilation compiledModules(List<String> sources, ModulePath path,
                                               List<Located> warningsOut) {
--- a/souther-compiler/src/main/java/souther/compiler/Main.java
+++ b/souther-compiler/src/main/java/souther/compiler/Main.java
@@ -316,9 +316,17 @@ public final class Main {
      * {@code souther examples <file.sou>...}: how well the model's {@code example}s cover it.
      *
      * <p>Its own command rather than a flag on {@code compile}, because the two answer different
-     * questions. {@code compile} writes classes out and stops at the first error; this asks a model
-     * that already compiles how much of it the rows have pinned down. Mixing them would put a report
-     * on stdout next to {@code wrote <path>} and make a failing build the only way to see it.
+     * questions. {@code compile} writes classes out and stops at the first error; this asks how much
+     * of a model the rows have pinned down, which is a question about what was observed and not
+     * about whether anything can be written out. Mixing them would put a report on stdout next to
+     * {@code wrote <path>} and make a failing build the only way to see it.
+     *
+     * <p>So the three things this answers with are decided apart. The report goes to stdout wherever
+     * there is a subject to report on; the diagnostics go to stderr as they always did; the exit code
+     * says whether the command succeeded, and an error still means it did not. Reading the report off
+     * a compile that had to survive every error is what left the whole {@code incompleteness} channel
+     * unreachable — all but one of the reasons that field carries arrive with a diagnostic beside
+     * them, so the fail-fast path hid the whole of it.
      */
     private static int examplesSubcommand(String[] rawArgs) {
         RenderOptions render = new RenderOptions();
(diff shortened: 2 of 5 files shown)


kawasima:
Both blockers addressed; the branch was amended and force-pushed (`aa72ebc8` → `ca4e6425`), since the first commit message carried the same overstatement as the specification text.

**1. Emission decided before the selection.** `assessable` is now read off `AdequacyReport.of(compilation)` and the selector is applied after it, so a compilation that formed no module and a selector that matched nothing are no longer one condition. `aSelectorThatMatchesNothingIsUntouchedByThis` pins #484's repro as unchanged by this PR — report of no modules, exit 0 — and it was watched failing against the gate the PR first had.

**2. The specification no longer overstates.** Two corrections, both of which the code confirms:

- `statusOf` makes a behavior `PARTIAL` from `partition.omitted()`, `pairs().truncated()` and the per-axis statuses, none of which reach `moduleReport`'s list. So `incompleteness: []` with `status: partial` is a legitimate state, and "read everything it asked for" is gone. The section now divides the two halves — what stopped the rows being observed, which is this field, and what stopped a measure reading what was observed, which is that measure's own `status`, `reason`, omissions and truncation.
- `ExampleVerifier` adds a `RUNTIME_ABSENT` incompleteness on `LinkageError` and adds no diagnostic, so "every one of them arrives with a diagnostic" is now "all but one", with the exception named.

**Non-blocker.** `bothProblemsAreSaidRatherThanTheFirstOfThem` gives two behaviors with independent bad field reads and asserts both reach stderr.

`mvn clean test` is green across all eight modules.

kawasima:
Rebased onto `develop` with #487 in it, and force-pushed (`ca4e6425` → `ce1a4f8c`).

**`Main.java` conflict.** Both contracts kept, in this order: diagnostics are said first (so a usage refusal never hides what is wrong with the source), then the selector is resolved and refused with exit 2, then the report is assessed, emitted and the exit status decided. `Selection.unresolved` resolves against `Compilation.declaredBehaviors`, which is a frontend question, so a type error does not make a good selector unresolvable — that is what makes the composition work.

**Obsolete test removed, composition test added.** `aSelectorThatMatchesNothingIsUntouchedByThis` asserted the pre-#487 contract and is gone, along with its fixture. In its place `aSelectorThatResolvesStillGetsItsReportFromACompilationThatDidNot` fixes the meeting point: `--module example.partly --behavior take` on a module with an E1321 in another behavior gives the selector resolving, a partial report on stdout narrowed to the one behavior, E1321 on stderr, exit 1.

**Comments corrected.** The `every reason … arrives with a diagnostic` generalisation is gone from the `examplesSubcommand` Javadoc and from the test class; both now say "all but one", with the runtime-not-reachable case named. The commit message carried the same claim and was rewritten with the amend.

Issue numbers have also been taken out of the source comments, which this repository does not carry.

`mvn clean test` green across all eight modules, with `ASelectorThatResolvesToNothingIsRefusedTest` (7) and this PR's tests (6) both passing.

PR #489, opened 2026-08-08
A reason is named for what happened, not for one of its causes

### LINKED ISSUE #474: `runtime_absent` at source scope does not mean the runtime is absent

kawasima:
Fixes #474.

`souther examples` told an author with a type error in front of them that
their runtime was missing:

```text
$ souther examples model.sou            # one behavior reads a field its input has not got
example.partly                                           measurement: partial
  ...
    · not measured: 0 (runtime_absent)
```

Where a source produced no observation, `AdequacyReport.moduleReport` and
`Adequacy.rowsOf` both wrote `RUNTIME_ABSENT`. At that point the runtime
cannot be what happened: `Output.Examples.evaluate` asks the evaluation only
of a module that checked, so a source with no observation failed something
earlier, and a real `LinkageError` is caught per example inside
`ExampleVerifier` and recorded against the behavior it was thrown for. The
code was named after one of its causes, then reused for the state, and every
producer since widened it — which nobody had to notice while the field it is
written in reached no artifact. #482 changed that a few hours ago, and this
is the first thing the field says out loud.

## The vocabulary

`OBSERVATION_ABSENT` is a source none of whose rows were observed, whatever
stopped them.

The other half of the split had the same defect one level down.
`RUNTIME_ABSENT` was written wherever a `LinkageError` was caught, and the
JVM raises one for a class that will not verify, one whose format it does
not accept, and one compiled against a different version of what it calls —
not only for a runtime that is off the classpath. Narrowing the catch would
mean reading an error message for a class name; naming what is caught costs
nothing and claims nothing. It is `LINKAGE_FAILED`, and the case it was
written for is said where it belongs, as the case rather than as the rule.

What the report distinguishes is the two states, not the five conditions
that produce the first — the frontend not answering, the module not
type-checking, the classes not being there, the requirements not being
there, a value already declared. Why a compilation failed is the
diagnostics' answer, and a second taxonomy of it in this schema would be two
files to keep in step.

`Observed.someRowsUnseen` now reads both codes. It asks whether rows exist
that nothing read, which is about the state and not the cause, so a single
code standing for both was the only reason it could read one.

**`schemaVersion` does not move.** Nothing is removed and nothing is renamed
that any document carries: this field has never reached an artifact, so no
document names either word. The shipped enum gains one and loses one that
was never written. The rule that field carries — raised when something is
removed or renamed, not when something is added — is now stated for a word
as it was for a key, in the specification and in the schema's own
description. Say so if you would rather have the version, it is a one-line
change.

## What a person reads

The report line and the note beside the generated rows both printed the
enum's own name and the subject. They are read side by side in one terminal,
so the wording lives in one place and both read it:

```text
    · no rows were read from `0`, so what they cover is unknown
// generation stopped for `take`: no rows were read from `0`, so what they cover is unknown
```

The subject stays the source id, which is what the JSON carries and what
tells a reader which of several files it was. In a single-source compile it
is a bare `0`, which is not much; giving the report file names is a bigger
change than this and not one #474 asks for.

**Only three codes are written out in words**, and the rule is that a
sentence keyed on the code may say only what is true at every place that
writes it. The code is not what happened; the producer is, and two producers
writing one code mean two different things by it.

```text
ROW_UNDECIDED        1 producer
OBSERVATION_ABSENT   2 producers, meaning the same thing
LINKAGE_FAILED       3 producers — an example that would not run, a fill
                     that could not build its candidates, a boundary that
                     could not build one
```

The three agree that the classes would not link and disagree about what did
not happen next, so that is where the sentence stops — and so do the code's
own javadoc and the schema's description of it, both of which said `nothing
ran`, which is true at one of the three.

The other five keep `subject (code)`, which asserts nothing the code does
not already say. That walk turns up more than it looks like it will: the
subject of `SEARCH_LIMIT` is `<n> combinations past the row limit` rather
than a name, and `PROBE_MAPPING_LOST` is written wherever the instrumented
classes could not be made, at module scope, which is wider than a probe
mapping that was lost — its javadoc is corrected here for the same reason,
though the name is left for the audit. Doing the rest is #491.

## `--generate` said nothing where it had something to say

It printed no block at all when the only thing to say was that it could not
look. `GenerationResult.isEmpty` counted rows and unresolved combinations,
so a result that was nothing but a reason was empty, and
`GeneratedRows.of` dropped the whole block — leaving an author who asked
what to write unable to tell "nothing left to write" from "could not look".

`isEmpty` now counts the reason too. The heading is about work there is to
do, so it is printed where there is some; the reason is printed whenever
there is one.

Three tests asserted the old silence by comparing against `""`. They now
assert what they were protecting — that no row is offered, a row being a
specific piece of work that may already be sitting in the file that could
not be read — and that the output says why.

## `NO_CLASSES`

Stays where it is, and stays out of the schema. Reaching it takes
`worthBuilding` holding while the probe is null, and the probe is null only
where `EvaluationLinked` is absent — which needs the backend to fail on a
module that checked, since anything that fails earlier leaves no rows and a
boundary with no rows behind it is undecided rather than missed. That is a
defect in the backend rather than a state of a source. The comment at the
mapping says that now instead of inviting a distinction the report has no
occasion to make.

## Verification

`mvn clean test` — BUILD SUCCESS across all eight modules.

Five new tests in `ASourceThatProducedNoObservationSaysSoTest`, each watched
failing first.


kawasima:
Both blockers addressed; amended and force-pushed (`f4f4dc60` → `b4c98b46`), since the commit message carried the same over-specification as the code.

**1. `RUNTIME_ABSENT` narrowed by renaming, not by narrowing the catch.** You are right that the producer catches any `LinkageError`. Of the two options I took the second: narrowing the catch would mean matching an error message for a runtime class name, which is fragile in exactly the way this issue is about, and letting other linkage errors escape would be worse than reporting them. It is `LINKAGE_FAILED` now — the enum member, the schema word, `BoundaryAssessment.Attempt.Reason`, and the prose at each producer, which said "the runtime is not on this host's classpath" in five places. The runtime being `provided` is still recorded as the case it was written for, as a case rather than as the rule.

**2. `Reasons.said` no longer asserts for un-audited codes.** Auditing the producers found two bad sentences, not one:

- `PROBE_MAPPING_LOST` — one producer, `Output.Examples.evaluate`, written whenever the instrumented classes could not be made (a backend `CompileException`, an `IllegalStateException`, absent inputs), at MODULE scope. "the arms of X could not be tied back to its body" claims more than that, and the subject is a module.
- `SEARCH_LIMIT` — its subject is not a name at all. `Generator` writes `<n> combinations past the row limit`, so quoting it as an identifier garbles the line.

Only `OBSERVATION_ABSENT`, `LINKAGE_FAILED` and `ROW_UNDECIDED` are written out in words; each has one producer, one scope, and a subject that is a name. The other five keep `subject (code)`. `AReasonSaysOnlyWhatWasEstablishedTest` pins that, including that a linkage failure does not claim the runtime is missing and that `PROBE_MAPPING_LOST` is not turned into a claim about arms. The full audit is #491.

**On the non-blocker:** the additive policy now says it covers a word added to an enumerated field, in the specification and in the schema's own description.

`mvn clean test` green across all eight modules.

kawasima:
This review is against `f4f4dc6`, which was superseded by `b4c98b46` — both blockers were addressed there, and the branch is now at `963bb2af`. Current state:

**1.** `RUNTIME_ABSENT` no longer exists in the tree. It is `LINKAGE_FAILED`, taking your second option: narrowing the catch would mean matching an error message for a runtime class name, which is fragile in the way this issue is about, and letting other linkage errors escape would be worse than reporting them. The enum member, the schema word, `BoundaryAssessment.Attempt.Reason` and the prose at all five producers were renamed with it.

**2.** `Reasons.said` sentences three codes — `OBSERVATION_ABSENT`, `LINKAGE_FAILED`, `ROW_UNDECIDED` — each read against its producers. `PROBE_MAPPING_LOST`, `SEARCH_LIMIT` and the three value codes fall through to `subject (code)`. `AReasonSaysOnlyWhatWasEstablishedTest` pins that `PROBE_MAPPING_LOST` is not turned into a claim about arms. The audit of the rest is #491.

Re-reading against your points did turn up two leftovers from that rename, now fixed in `963bb2af`:

- the schema's `code` description still explained `runtime_absent` — a published file describing a word it no longer allows. The `sed` had replaced the enum value and not the prose beside it.
- `theHumanLineIsASentenceAndNotACodeName` asserted the output does not contain `runtime_absent`, which no longer appears anywhere, so it had gone vacuous. It now asserts on the two words that exist.

`mvn clean test` green across all eight modules. If you were reviewing the current head and still see either blocker, say which file and I will look again — I may be reading my own change too kindly.

kawasima:
Fixed, and you are right that it is the same defect again — the third time in this PR. `34657d37`.

`LINKAGE_FAILED` has three producers, not one:

```text
ExampleVerifier:109   an example whose classes would not load    — its rows did not run
Adequacy:795          a fill that could not build its candidates — the rows were already read
Adequacy:837          a boundary that could not build one        — likewise
```

The sentence is now `the classes for \`submit\` would not link` and stops there.

Your point also lands on a claim I made that you did not quote. The `Reasons` javadoc said "each has one place that writes it", which is false for two of the three: `OBSERVATION_ABSENT` has two producers as well. They happen to mean the same thing, so its sentence survives — but the justification I wrote for it was wrong, and I would not have caught that by checking only the code you named.

So the class now carries the rule rather than a list of what I happened to check: **a sentence keyed on the code may say only what is true at every place that writes that code**. The javadoc names the producer count for each of the three and says where the agreement stops. #491 carries the same rule, plus the structural note that a consequence only some producers have belongs to the producer and cannot be recovered from the code.

Also fixed: the class javadoc of `ASourceThatProducedNoObservationSaysSoTest` said "both wrote \`LINKAGE_FAILED\`", which the rename made false — historically it was the other word.

`mvn clean test` green across all eight modules.

kawasima:
Fixed in `5636e0da`, and the point that the definitional sites outrank the display layer is the right way round — I had fixed the renderer and left the two places that say what the code means.

`nothing ran` is gone from both:

- `Incompleteness.Code.LINKAGE_FAILED` now stops at "The generated classes would not link", and names the three producers underneath so the next reader does not have to find them again.
- the schema's `code` description says "and no more than that", with the same note about what did not happen next.

I checked the rest of the definitional sites against their producers rather than only the one you named, and found a third with the same defect. `PROBE_MAPPING_LOST` was documented as "A branch probe could not be tied back to the node it belongs to", but its one producer writes it wherever the instrumented classes could not be made — a backend that failed on them, inputs that were not there. Its javadoc now says that. The name is left alone for #491, since renaming it is the audit's call and not this PR's.

Two that I checked and left as they are: `Attempt.Reason.LINKAGE_FAILED` has one producer, where the decoders genuinely could not be reached, so "the decoders could not be reached" holds; `OBSERVATION_ABSENT` has two producers that mean the same thing.

Non-blockers both done — the PR body carries the producer counts (1 / 2 / 3) instead of "one producer each", and the `--generate` test javadoc is in the past tense for behaviour this PR changed.

`mvn clean test` green across all eight modules.

PR #496, opened 2026-08-08
The distinctions Incompleteness.Code draws now reach a reader, or leave

### LINKED ISSUE #491: Half of `Incompleteness.Code` is a distinction nothing reads
The audit this issue asked for has been done. It does not end in a sentence or a
rename for the codes it was about, and what it turned up instead is this:

> Nothing in the type says which distinctions are meant to survive to a reader and
> where information is allowed to collapse. The two places that collapse one do so
> by reading a shape rather than by deciding to.

`Incompleteness.Code` is not one vocabulary. It holds the cause an observation could
not be read (`VALUE_TRUNCATED`), the state a measurement ended in
(`OBSERVATION_ABSENT`), and a limit a process ran into (`SEARCH_LIMIT`), and
`Reasons.said` being total over the enum is what makes all eight look like words a
report is written in. Half of them are never read as words at all. For those, the
question this issue opened with — what is true at every place that writes this code
— cannot be put yet, because a code nothing reads as vocabulary has no reader to be
honest to.

## What writes each code, and what reads it

| code | written at | scope / subject | read as a word by |
|---|---|---|---|
| `OBSERVATION_ABSENT` | `AdequacyReport:161`, `Adequacy:651` | `SOURCE` / source id | `AdequacyReport.human` |
| `LINKAGE_FAILED` | `ExampleVerifier:109`, `Adequacy:795`, `Adequacy:837` | `BEHAVIOR` / behavior name | `AdequacyReport.human`, `GeneratedRows.notes` |
| `ROW_UNDECIDED` | `ExampleVerifier:118` | `BEHAVIOR` / behavior name | `AdequacyReport.human` |
| `PROBE_MAPPING_LOST` | `Output:877` | `MODULE` / module name | `AdequacyReport.human` |
| `SEARCH_LIMIT` | `Generator:211` | `MODULE` / `"N combinations past the row limit"` | `GeneratedRows.notes` |
| `VALUE_UNREADABLE` | `Generator:189` | `POSITION` / `axis.path()` | `GeneratedRows.notes` |
| `VALUE_UNREADABLE` | `RowClasses:52,57,61,75` | `POSITION` / `axis.id()` | nothing |
| `VALUE_TRUNCATED` | `ObservedValues:50,53,74,84,94` | field of `ObservedValue.Truncated` | nothing |
| `VALUE_TRUNCATED` | `RowClasses:65` | `POSITION` / `axis.id()` | nothing |
| `AXIS_OMITTED` | `Partitions:121` | `POSITION` / `axis.id()` | nothing |

The unit here is a producer and not an enum constant. `VALUE_UNREADABLE` is read
where the generator writes it and unread where `RowClasses` does, and the two are
not the same value — more on that below.

Two channels carry a code to a reader and they do not meet. The module's list —
what `Output.Examples.Of` returns, plus the `OBSERVATION_ABSENT` added beside it —
is printed by `AdequacyReport.human`, counted by `countsAgainst`, and written to
JSON. `GenerationResult.incompleteness` is read by `GeneratedRows.notes` and by
nothing else.

## The fact is read. The distinction is not

The last column says nothing reads the code. It does not say nothing reads what
happened, and for three of those rows the fact is read through another route.

`RowClasses` returns `Classification.Unclassified` and every consumer of a
classification matches on `Classified` and treats anything else as a position the
row did not place. `Coverages.Readings.classIn` returns null for it,
`couldNotSay` counts the row, and the count becomes
`AxisCoverage.unclassifiedRows`, which is a required field of
`adequacy-schema-1.json` and is written per axis. The same count makes the axis
`PARTIAL`. So a truncated value does reach a reader — as a number and a status,
through a channel that has never carried the code.

```text
RowClasses records Truncated apart from Unreadable
    |
    +-- Coverages   reads the shape only  -> unclassifiedRows, PARTIAL  (JSON)
    |
    +-- Generator   reads the shape only  -> a fresh VALUE_UNREADABLE   (notes)
```

Neither of those is obviously wrong. Coverage may have no use for the difference
between a value that was too large to keep and one that could not be read at all,
and a generator that withholds work at a position it cannot see may not either. What
is wrong is that neither of them says so. `Coverages.Readings.classIn` and
`Generator.readEverywhere` both decide by `instanceof Classified`, so the collapse
is a side effect of a shape test and not a decision anything states or can be held
to. There is no boundary in the model where the distinction is declared to end.

`Generator.fill` in particular writes `VALUE_UNREADABLE` whatever was recorded. Put
a classification of `VALUE_TRUNCATED` in and `VALUE_UNREADABLE` comes out. The code
the generator writes is not the value `RowClasses` wrote; it is a second reason
built after the first was dropped, which is why the two rows of the table sharing
that name have different readers.

The renderer defect this collapse feeds into — a note reading `generation stopped`
over a block that generated a row — is #495. It is a separate failure of the same
kind, one layer further out, and does not need this issue's decision to be fixed.

## `SEARCH_LIMIT` is what a three-field record looks like when it does not fit

Its subject is `"N combinations past the row limit"`, which is a phrase and not a
name, and its scope is `MODULE` while `Generator.fill` runs per behavior.
`identity()` is `(code, scope, subject)`, so two behaviors that stopped at the same
count are one reason. Nothing observes this yet: the dedup by identity runs only
over the module's list at `AdequacyReport:168`, and this value never enters it.

Correcting the scope and the subject together removes the count, which is the part
a reader wants:

```text
code    = SEARCH_LIMIT
scope   = BEHAVIOR
subject = behavior name
detail  = N combinations past the row limit
```

`Incompleteness(code, scope, subject)` has nowhere to put the last line, because
`subject` is doing identity and presentation at once. That is a second axis of the
same problem — one shape holding codes that need different things — and it is here
as evidence. What to do about it falls out of the decision below rather than
being a fix of its own.

## The order the questions have to be put in

The question this issue opened with is the last one.

```text
1. Does anything read this code as a word?
   |
   +- Yes -> is its meaning preserved from every producer to that reader?
   |          +- Yes -> a sentence, or a name that says what happened
   |          +- No  -> the propagation is the defect, not the name
   |
   +- No  -> does the fact it classifies reach a reader another way?
              +- No  -> the value itself may not be needed
              +- Yes -> does that reader need the classification?
                         +- No  -> the collapse is part of the model. Say so in the
                         |         type and take the discriminator out
                         +- Yes -> the collapse is a propagation defect. Carry it
```

What is being decided at the bottom is not whether to keep an enum constant. It is
whether the collapse is the specification or a defect, and that is a question about
what the reader is owed rather than about the word.

- `VALUE_TRUNCATED` — the fact is read, the distinction is not, and whether a reader
  is owed it is open. Removing it is not dead-code cleanup; it is deciding that the
  cause of an unplaced value is not part of the model.
- `VALUE_UNREADABLE` at `RowClasses` — the same, and the two travel together.
- `VALUE_UNREADABLE` at `Generator` — a consequence rebuilt after the cause was
  dropped, wearing the name of the cause. If the collapse is intended, this wants a
  type and a name of its own rather than a reused one; if it is not, the recorded
  reason should reach it. Either way the name comes after that.
- `AXIS_OMITTED` — `AdequacyReport:778` writes `o.reason().subject()` into the JSON
  `omitted` array and `AdequacyReport:334` reads `carriedAnObligation`. The code
  changes no output and no behavior, and what the consumers want is
  `OmittedAxis(axis, carriedAnObligation)`. Not "no renderer, so delete it" — the
  discriminant the consumers need is not this one.
- `SEARCH_LIMIT` — read as a word, and wrong in scope, subject, and the field it has
  no room for.
- `PROBE_MAPPING_LOST` — one producer, at `MODULE` scope, written wherever the
  classes an arm-measuring evaluation needs are absent. Wider than the name, as this
  issue said. This one is ready for the original question.

## What was decided

The distinction is carried. The criterion is whether it changes what the author does
next: a truncated value goes away if the fixture is written smaller and an unreadable
one does not, so collapsing the two loses the information a remediation is chosen
from. A distinction that would not change an author's next step is one this could
drop, and that is the test to put to the next one.

It goes through the module's incompleteness channel rather than into `AxisCoverage`.
`unclassifiedRows` is the measurement — how many rows could not be placed at this
position — and an `Incompleteness` is why that happened. Adding `truncatedRows` and
`unreadableRows` beside the count would make a diagnostic taxonomy a dimension of the
measurement schema, and the schema would then grow sideways with every cause added
later. A count that is stable and a vocabulary that is not stay apart:

```text
axis:
  unclassifiedRows: 3          <- the measurement

incompleteness:
  request.kind: the value was truncated    <- why it came out that way
```

This is not the judgement #484 made. There the selector's failure to resolve was not
a measurement at all and so belonged on neither scale. Here the fact happened during
the measurement, and what is being settled is which of the two layers holds it.

Where one axis has rows failing both ways, both reasons are written — one per kind,
not one per row. `identity()` is `(code, scope, subject)`, which makes the two
distinct and the repeats within a kind one reason. The count stays in
`AxisCoverage.unclassifiedRows`, so nothing carries it twice.

## The work

1. Carry the classification cause from `RowClasses` to the module's incompleteness
   channel, under the aggregation contract above, and give the two codes sentences in
   `Reasons.said`. `RowClasses` runs inside `Coverages` and does not reach the list
   `ExampleVerifier` builds, so this is where the wiring is.
2. Delete `ObservedValue.Truncated.reason`. Five producers pass the same constant and
   the record's own name already says it. Which limit truncated the value —
   `node-limit`, `depth-limit:X`, `elements`, `entries`, text — currently rides in
   `digest`, whose job is telling two truncated values apart. That is a resolution
   below the one being decided here and is not part of this.
3. `AXIS_OMITTED` becomes `OmittedAxis(AxisId, boolean)`. Not a cleanup: it makes the
   type say what the consumers actually distinguish.
4. `PROBE_MAPPING_LOST` goes to the original question — one producer, take what it
   guarantees, rename where the name is narrower. The shape `RUNTIME_ABSENT` →
   `LINKAGE_FAILED` had.
5. Separate the generator's reasons from the classification causes. Once the cause
   reaches a reader, what `Generator:189` writes is not an unreadable value; it is a
   position not offered as work, which is a different taxonomy and wants a different
   type rather than another constant in this enum. `SEARCH_LIMIT` belongs with it, and
   moving it there dissolves the scope, subject, and detail problem into a product
   type — a behavior and a count — instead of a fourth field on `Incompleteness`.
   #495 is where this is first needed, since telling a generation that stopped from a
   position that was left out is exactly the distinction its note is missing.

## Where this leaves the sentence the issue opened with

> A code that cannot be given an honest sentence is a code whose name is doing work
> it cannot support, and the fix is the name.

That holds for a code something reads as a word, and `PROBE_MAPPING_LOST` can be
taken straight to it. It does not reach the rest, and applying it there would rename
values on the strength of a reader they do not have. Narrowed:

> A code used as report vocabulary has to support an honest sentence across all of
> its producers. A code nothing reads as vocabulary has to justify the distinction
> before it is worth naming at all.

### LINKED ISSUE #495: A row is generated, and the note under it says generation stopped
`GeneratedRows:184` prints a note over every element of a generation's
incompleteness:

```java
"// generation stopped for `%s`: %s"
```

which asserts `an incompleteness exists => generation stopped`. The implication is
false, and the counterexample is printed in the same block. Where some positions
were read from the existing rows and others were not, `Generator.fill` leaves the
unread ones out, generates from the rest, and reports both:

```text
// generated by `souther examples --generate`: 1 row to fill what nothing covers.
// example submit
//     | "request.urgent=false" : (Request { kind = Domestic, urgent = false }) -> <?>
// generation stopped for `submit`: request.kind (value_unreadable)
```

The row two lines above the note is the refutation. Generation did not stop; one
position was left out of it, and the rows offered are the ones the remaining
positions could still be filled at.

`Generator.fill` returns early only when every axis is unread — `axes.isEmpty()` at
`Generator:193`. Otherwise the unread positions go into `undecided`, the list is
copied into `incompleteness` at `Generator:207`, and the loop runs. So the note and
the rows come out of the same call by construction, not by some unlikely
interleaving.

## Why the test does not catch it

`AdequacyNeverAssertsFromPartOfTheRowsTest.nothingIsGeneratedFromRowsThatWereNotRead`
asserts `written.contains("generation stopped")` and passes. Both of its models
leave every axis unread — one holds a row that never comes back, the other spends
the observation budget on a value written before the positions a measure would
read — so `Generator.fill` returns before generating anything and the sentence is
true. A model where one position is unread and another is not has never been put
through it.

## Reproducing it

`GeneratorTest.modelOf` builds a subject from a source. Hand `Generator.fill` one
recorded row where one axis is `Classification.unreadable(...)` and another is
`Classification.in(...)`:

```java
Map<AxisId, Classification> row = new LinkedHashMap<>();
row.put(first.id(), Classification.unreadable(Incompleteness.Code.VALUE_TRUNCATED,
        first.id().toString()));
row.put(second.id(), Classification.in(second.classes().get(0).id()));

Generator.GenerationResult filled =
        Generator.fill(subject, List.of(row), Generator.CandidateCheck.ANY);
```

which comes back with one row and one incompleteness, and renders as the block
above.

## What a fix has to hold to

`Reasons.said` is held to a rule — a sentence keyed on the code may say only what is
true at every producer — and this is the same failure one layer above it, at a
renderer that is not keyed on the code at all. The frame is applied to whatever
turns up in `result.incompleteness()`, and it states a consequence the value does
not carry.

Two of the three things that reach this list are consistent with the current frame
and one is not. `SEARCH_LIMIT` from `Generator:211` and `LINKAGE_FAILED` from
`Adequacy:795` are both a generation that stopped. `VALUE_UNREADABLE` from
`Generator:189` is a position left out of one that went on. So the note has to come
from something that distinguishes them rather than from the presence of a reason.

That may not be another constant in `Incompleteness.Code`. What the generator has to
say — a generation that stopped, a position left out of one that went on, a search
that ran past its limit — is a taxonomy of its own, and #491 settles that it should
be typed apart from the causes an observation could not be read for. A generation
reason as a product type would carry the behavior and the count that
`Incompleteness(code, scope, subject)` currently has no room for, and `SEARCH_LIMIT`
would move with it. Worth looking at here before adding a code to tell the two apart.

The vocabulary question underneath — whether `VALUE_UNREADABLE` is even the right
value at `Generator:189`, given that it is rebuilt after the recorded reason was
dropped — is #491. This issue is only that the note states more than its value
carries, and that nothing tests the case where it is wrong.

kawasima:
Closes #491. Closes #495.

The audit #491 asked for did not end in a sentence or a rename. What it found is
that `Incompleteness.Code` held three kinds of thing at once, that nothing in the
type said which distinctions were meant to reach a reader, and that the two
places dropping one did so by reading a shape rather than by deciding to. These
five commits settle each of those and leave the enum as the vocabulary a report
is written in.

## What is in each commit

**A record that says what it is does not repeat it in a field.**
`ObservedValue.Truncated` carried a code every one of its five producers set to
the same constant. `AXIS_OMITTED` was held in an `OmittedAxis` whose two
consumers read the subject and the flag beside it and never the code, so
`OmittedAxis` now takes the `AxisId` it is about. The schema listed
`axis_omitted` among the codes a report can write, and nothing could ever write
it.

**A reason is named for the absence, not for one way of arriving at it.**
`PROBE_MAPPING_LOST` → `INSTRUMENTATION_ABSENT`. Reading its one producer also
established more than the old name allowed: the branch is taken only where arm
coverage was asked for and returns no rows, so the sentence may say both.

**A generation that went on does not say it stopped.** The note under a
generated block was written from the presence of a reason. A position left out
is a reason and is not a stop, so a block came out offering rows above a line
saying it had offered none. What the generator has to report about its own run
is now `GenerationReason`, which is a different taxonomy from what a measurement
could not read — and `SearchLimit` carries the behaviour and the count as
fields, which `Incompleteness(code, scope, subject)` had nowhere to put.

**A position is inside one behaviour, and now it says which.** `countsAgainst`
answered `POSITION` with the fallback for anything larger than a behaviour.
Nothing had ever asked: no producer wrote a position into the list a report
reads. The subject is now held as what it names, so the scope is derived from it
and the two cannot disagree; readers are untouched. `Scope.isOneBehavior()` is
gone — two readers used it and then compared the subject, which is the same
question asked a second and third way.

**Why a row could not be placed reaches the reader.** The cause `RowClasses`
records now travels to the module's reasons, one per kind per position. It goes
there rather than into `AxisCoverage` because the count is the measurement and
the reason is why it came out that way; putting `truncatedRows` beside the count
would make a diagnostic taxonomy a dimension of the measurement schema.

## What decided it

Whether a distinction is carried is decided by whether it changes what the
author does next. A value too large to keep goes away if the fixture is written
smaller; one that could not be read does not. That is the test to put to the
next one.

## Evidence

`generation stopped` was refuted by running it, not by reading it: with one
position readable and another not, the generator fills the rest and the block
printed a row two lines above the line saying it had stopped. The test that
guarded the wording asserted it over two models where every axis was unread, so
the generator returned before offering anything and the sentence was true.

The reasons now print under the behaviour they name. That one only showed up in
the CLI — the module's reasons are printed after every behaviour, so a reason
about `take` sat under `cancel` at the same indent.

Measurement is unchanged throughout. The one existing test that failed under the
propagation was the assertion that nothing is recorded as a reason; its three
status assertions passed untouched.

Full reactor `clean test` green.


kawasima:
The blocking point is right, and reading the producers sharpens it: the answer
was correct and the reason given for it was not.

Both places that write a `SOURCE` write `OBSERVATION_ABSENT` for a source that
was not evaluated at all, and `Adequacy` says why beside one of them — *which
behaviors it wrote rows for is exactly what cannot be read, so it counts against
every one of them*. So `true` is a reading of the producers. What `Target` gave
instead was the shape: *anything larger, which is about every behavior inside
it*. That is the reasoning this branch removed from `POSITION`, put back one
case along, and the test asserted the yes without the reason — which fixes the
assumption as specification exactly as you say.

Taken the way you recommend. `Target` now answers only what it names —
`onlyBehavior`, which is intrinsic — and `countsAgainst` moves to
`Incompleteness`, where the reading can be written down. It also says what would
have to change if a `SOURCE` whose contents were known ever turned up: that one
needs the compilation to answer which behaviors those were, and does not belong
on the value. Nothing writes one today.

The `OfSource` javadoc no longer claims anything about what a source holds. A
source id is a file; which behaviors wrote rows in it is the compilation's
answer.

The test is now three, and the two that answer yes say which yes it is: a module
holds every behavior in it by what a module is, and a source does because of what
is written there.

`f759da2e`

## GenerationReason.stopped()

Taken further than requested, and it is fair to push back on that. The predicate
restated in a boolean what the variants already say, and its only reader was a
test. What stops a variant added later from going unsaid is the exhaustive
switch in the renderer, not this. Deleting it is the first commit on this branch
applied to what the branch itself wrote — a field whose producers all agree is a
field carrying nothing — and it comes back with a second reader if one turns up.

`0ea6e590`

Full reactor `clean test` green on both.

kawasima:
All four are in. `e3b86ca6` and `0e48cc57`, full reactor `clean test` green.

## Blocker 1 — you are right and my reading of the precedent was wrong

I took the two earlier renames as a licence to rename an emitted word in v1. They
are not. `4f921a5b` says so in the commit message — *`row_timed_out` has never
appeared in output* … *the schema is corrected rather than versioned* — and the
same condition held for `runtime_absent`, which was renamed before the
incompleteness list had reached a published report at all. Every word taken out
until now was one no document could be holding. `probe_mapping_lost` is the first
that is not.

So it goes back, and the change becomes additive rather than a removal, which is
what the preamble already allows without raising the number.

`search_limit` and `axis_omitted` stay out. Neither could reach the report's
incompleteness list: the first went to the generator's own block, and the second
no further than an `OmittedAxis`. They are the `row_timed_out` case — words this
version promised that nothing could ever write — so taking them out is the same
correction.

`EveryWordTheSchemaAllows...` is extended the way you describe. A vocabulary
carries what it retired as well, named beside it, so the expected set is what the
compiler writes now together with what it wrote then. A word kept on purpose is
one somebody wrote down; a word that quietly stops being emitted still fails.
Checked by taking `probe_mapping_lost` back out of the schema and watching it
fail.

The equality was right for as long as it held, and what changed is not the
contract but that a rename finally happened after the field had been published.
That is in the test's javadoc now, since it is the reason the shape is what it
is.

## The other three

`VALUE_TRUNCATED` — the budget is one for the whole walk, and the fixture in
those very tests is the counterexample: the position named holds the number the
invariant's lower bound names, and what spent the budget was the collection ahead
of it. The sentence says the observation was stopped by a limit. The javadoc on
`ObservedValue.Truncated` had the same error and is fixed with it. The
remediation the two codes are told apart for survives — a fixture written smaller
still frees the budget.

`ClassesWouldNotLink` — taken to the intersection: `NothingToBuildAgainst`. Of
your three options this is the third. The difference between classes that would
not link and classes that were not there changes nothing an author would do,
since neither leaves a row to offer, so one word at what both establish is what
this branch's own rule asks for.

`Reasons` producer audit — re-walked rather than only corrected.
`LINKAGE_FAILED` has one producer now, an example whose evaluation raised one and
observed nothing, so the sentence may say the rows did not run. It could not
while three producers disagreed about what did not happen next, and two of those
three are the generator's now.

kawasima:
Both in, `c658d690`. Full reactor `clean test` green.

## `incompleteness.scope`

Rewritten around what each kind answers rather than around size. A module counts
against every behaviour for the reason the old sentence gave, and now gives it on
its own; a source counts against every behaviour for the other reason, which is
that one is written only where the source was not evaluated and which behaviours
had rows in it is exactly what could not be read. A behaviour and a position
count against the behaviour they are in.

That was the shape argument taken out of the code in `f759da2e`, still standing
in the published contract — the worse of the two places for it to be, since a
consumer is written against this file and not against the javadoc.

## `code`

`linkage_failed` no longer says what did not happen next is not part of it. It
was true while three producers disagreed about it; two of those are the
generator's now and the one left observed nothing, so the schema says the rows
did not run, which is what `Reasons.said` says.

## The test name

Taken as well, since it is a nit that stopped being true at a known moment:
`EverySchemaWordIsAccountedForTest.everyEnumeratedFieldNamesCurrentOrRetiredWords`.
The old name was right until a word outlived the constant that spelled it.

Issue #479, opened 2026-08-08
Numeric candidate generation cannot exploit relational constraints between positions


kawasima:
Candidate generation fixes what each position may take before it knows what any
other position took. `Generator.choicesOf` builds one candidate list per position,
once, and `Generator.walk` searches the product of those lists. So a rule relating
two positions is satisfied only when both lists happen to already contain values
that satisfy it.

```
C(r.a) × C(r.b) × ...
```

For

```
data R =
    { a: Int
    , b: Int
    }
    invariant ordered = a < b
```

`C(r.a) = {0}` and `C(r.b) = {0}`, so the search has one assignment to try and it
is refused. Nothing here involves a newtype, a decimal or a boundary — it is
ordinary partition fill, and the report says

```
// no row for `flag=false` in `f`: every value tried was refused at construction
```

Widening the lists is not the answer either. With `C(r.a) = C(r.b) = {0, 1}` the
model above builds, and

```
    invariant ordered = a + 100 < b
```

still does not: no pair drawn from two lists chosen in ignorance of each other
satisfies it. Both models are measured below.

### A1. A projected range exists and representative generation does not read it

Every primitive branch of `Partitions.representativesOf` reads the type and
returns before `within` is looked at:

```java
static List<FixtureTemplate> representativesOf(Type type, Symbols symbols,
                                               NumericDomain.Bounds within) {
    if (type == Type.INT) {
        return List.of(FixtureTemplate.integer(0));
    }
    ...
}
```

This is reachable. Give one of the fields a bound of its own and the relation
projects onto the other:

```
data R =
    { a: Int
    , b: Int
    }
    invariant lower = a >= 0
    invariant ordered = a < b
```

`b` is left `[1, ∞)`, and the branch above discards it and offers `0`. Letting the
primitive branches honour `within` closes this model and nothing else.

A newtype and a bare `Int` cannot be compared (E1319), so a projection reaches a
primitive branch only where a record clause gave one of the fields its own
interval. That is why this defect is narrow.

### A2. No projected range is derived, because there is none to derive

For `a < b` alone the domain holds `a - b <= -1` and no interval on either atom,
so `bestLo`/`bestHi` return null for both and `FieldDomains` has no row for
either field. That is correct: `∃a. a < b` bounds `b` nowhere.

`b >= 1` is not a projection that was derived and dropped. It is what conditioning
on a choice yields:

```
projection:    ∃a. a < b          ⇒  no bound on b
[...]

PR #497, opened 2026-08-08
An end says whether the value it stops at is one of the range's own

### LINKED ISSUE #483: Where a rule stops is recorded as a value it admits, and each reader invents a different one where there is none
Where a rule stops is recorded as a value the rule admits. Over whole numbers
there is always such a value — `x > 0` stops at 1 and 1 is one of its own — so
the representation costs nothing and reads as exact. Over decimals there is no
value beside the line, and every reader that needs one invents it somewhere
else, differently.

## What is measured

One record, seven positions, one row written against every edge:

```
data Positive    = Decimal invariant positive    = value > 0m
data AboveFive   = Decimal invariant above       = value > 5.0m
data BelowOne    = Decimal invariant below       = value < 1.0m
data AtLeastZero = Decimal invariant atLeast     = value >= 0m
data AtLeastFive = Decimal invariant atLeastFive = value >= 5.0m
data AtMostOne   = Decimal invariant atMostOne   = value <= 1.0m
data IntPositive = Int     invariant positive    = value > 0
```

| declaration | where the rule stops | `boundsOf` | the cut | reported |
| --- | --- | --- | --- | --- |
| `value > 0m` | 0, not its own | `min = 0` | **1** | a row is asked for at 1 |
| `value > 5.0m` | 5, not its own | `min = 5.0` | **none** | `not derivable` |
| `value < 1.0m` | 1, not its own | `max = 1.0` | **none** | `not derivable` |
| `value >= 0m` | 0 | `min = 0` | 0 | a row is asked for at 0 |
| `value >= 5.0m` | 5 | `min = 5.0` | 5.0 | a row is asked for at 5 |
| `value <= 1.0m` | 1 | `max = 1.0` | 1.0 | a row is asked for at 1 |
| `value > 0` (Int) | 1 | `min = 1` | 1 | a row is asked for at 1 |

Three wrong answers to one missing distinction. The first names a value no rule
in the model mentions and reports it as the line `positive` drew. The next two
say the model divides the position nowhere, of a model that divides it at five
and at one. The projection admits the one value the rule refuses.

They compound. `AboveFive` offers `5m` and `0m`, and its own rule refuses both,
so no assignment of the record completes and **every** edge of it — including the
four that were read correctly — comes back as `every value tried was refused at
construction`.

## Where each of them comes from

The cuts are read from `Partitions.boundsOf`, which reads
`codegen.InvariantConstraints`. That is the list of checks the runtime has a word
for, and it answers a different question from where a rule stops:

```java
case InvariantConstraints.Positive _, InvariantConstraints.DecimalPositive _ ->
        BigDecimal.ONE;
```

`Positive` is an `Int` above zero, and 1 is exact for it. `DecimalPositive` was
added to the same arm. And a decimal above five has no annotation at all:

```java
case GT -> zero ? Optional.of(new DecimalPositive()) : Optional.empty();
```

`Optional.empty()` is right for codegen — emit a general predicate — and is read
here as "this rule states no bound". `Partitions.boundsOf` has one absence for
both "the type draws no line" and "the type draws one I have no case for".

`NumericDomain.addLe` repairs the same gap in the opposite direction, widening
to the non-strict bound and tagging `WEAKENED_STRICT`, which is sound for proving
and admits the refused value.

`BoundaryDomain` already holds the correct statement — "a neighbour needs a
smallest step, and the type does not carry one, so there is none to give" — and
none of the three consults it.

## What an end has to say

```
>  5.0  ⇒  End(5.0, exclusive), derivable, no row owed at it
>= 5.0  ⇒  End(5.0, inclusive), derivable, a row owed at it
<  1.0  ⇒  End(1.0, exclusive), derivable, no row owed at it
<= 1.0  ⇒  End(1.0, inclusive), derivable, a row owed at it
```

An open end is derived, not underived. `not derivable` is for a line whose value
could not be worked out; this is a value that is known and is not one of the
range's own. Keeping the two apart is what stops a report asking for a row at a
value construction will refuse, which is how one such position takes every edge
of the record down with it.

Composing ends is an algebra rather than a flag beside a number. A lower bound
takes the greater value and the exclusive end at equal values, an upper bound the
lesser and the same tie; a range is empty where its ends meet and either leaves
the value out. An end derived through `closedDiff` is reached only where every
end on the path is, so exclusiveness propagates along a derivation. Both carriers
— the domain, which computes with the numbers, and a partition's end, which also
says which rule put it there — have to answer that the same way or the
distinction is gone one call later.

## Choosing a value to write is a different question

Over a dense order `5 < x` has no least solution, so a candidate generator is not
being asked for the value beside the line. It is being asked for one value from
inside the range, decided the same way every time: between the ends where both
are known, and a step in from the only one where there is one. Nothing about the
step is the nearest admitted value, and it does not need to be.

Stepping by the literal's scale — `5.0` to `5.1`, `5.00` to `5.01` — is not that.
`5.0m` and `5.00m` are one constraint, and letting the spelling decide the value
puts the representation back into the semantics, which is the same mistake as the
1 above. An end reached through a difference has no literal to take a scale from
at all.

What this issue owes is that an open boundary is never mistaken for a value the
rules admit. A range with a hole in it — `x > 5` beside `x /= 6` — is a witness
search, and belongs to whatever answers that.

## Consumers

Two places compare a guard against what a position holds, and both read the
domain's ends as closed: `Partitions.holds`, which decides whether a value
invented one step off a line is one the position can hold, and
`GuardEdge.provenDisjoint`. The second is not "read the domain's flag too" — two
half-lines meeting at one number share it only where both hold it, and either
side can be the one that does not.

## Not a fix for #479

#479 is that candidate generation fixes each position's values before knowing
what any other position took. With openness recorded exactly,
`invariant ordered = a + 100 < b` still has no row; with #479's partial
assignment implemented, a strict decimal end is still approximated here. #479
made the second of these visible: its projection now reaches the candidate
generator, so `AboveFive` is offered `5m` rather than `0m`, and the one thing
left between that and a row that builds is the end's openness.

kawasima:
Closes #483.

Where a rule stops was recorded as a value the rule admits. Over whole numbers
there is always such a value, so the representation cost nothing and read as
exact. Over decimals there is no value beside the line, and each reader that
needed one invented it somewhere else:

| declaration | where the rule stops | what was reported |
| --- | --- | --- |
| `value > 0m` | 0, not its own | a row asked for at 1, a value no rule mentions |
| `value > 5.0m` | 5, not its own | `not derivable` |
| `value < 1.0m` | 1, not its own | `not derivable` |

They compounded. A position whose only candidates were `5m` and `0m` — one
refused by its own rule and one outside it — left the record with no assignment
at all, so **every** edge of it came back as `every value tried was refused at
construction`, including the ones read correctly.

## What is here

An end is a value and whether the range holds it. The algebra is written once, in
`numeric.Endpoint`, because two carriers derive it: the domain, which computes
with the numbers, and a partition's end, which also says which rule put it there.
A lower bound takes the greater value and, at one value, the end that leaves it
out; a range is empty where its ends meet and either of them does.

`NumericDomain` keeps strictness where it records it instead of widening and
tagging the widening as a loss, so `WEAKENED_STRICT` has no producer left and is
gone. An end reached through a difference is reached only where every end on the
way is, which `bestLo`/`bestHi`, the transitive closure and `upperBound` carry
along. `proveLe` gets the precision back and `feasible` sees a contradiction the
numbers alone do not show.

The bounds a position is measured at are read off the comparison rather than off
`codegen.InvariantConstraints`, which is the list of checks the runtime has a word
for. That list has a case for a decimal above zero and none for a decimal above
five, and `Partitions` had one absence for a type that draws no line and a type
that draws one it could not spell. A literal's scale is not part of what it says,
so `5.0m` and `5.00m` reach a range as one number.

A cut is where a rule drew its line; whether a row is owed there is whether the
position holds the value. The two places that compare a guard against a position
ask it as one question — two half-lines meeting at a number share it only where
both hold it, and either side can be the one that does not.

A position offers a value from inside its range where the end is not one of its
own: between the ends where both are known, and a step in from the only one where
there is one. Deliberately not the literal's scale — `5.0m` to `5.1m` would let
how a rule was written decide the value proposed for it, and an end reached
through a difference has no literal to take a scale from.

## Evidence

One record with seven numeric positions and one row written against its edges:

```
before  boundary 0/0 (5 not measured)
          · not known to be writable: take/h.p = 1 (invariant Positive (min))
          · not known to be writable: take/h.z = 0 / h.f = 5 / h.m = 1 / h.i = 1
          · not derivable: h.a        (value > 5.0m)
          · not derivable: h.b        (value < 1.0m)
        // no row for ... every value tried was refused at construction  (x5)

after   boundary 4/4
        adequacy: satisfied
```

`souther examples` over every model in `souther-examples`, before and after,
differs in one place:

```
-     boundary    0/2
-       · no row is at quotaGap/quota = 1 (invariant Quota (min))
+     boundary    0/1
```

`crm`'s `data Quota = Decimal invariant value > 0.0m` was being asked for a row at
1. The real edge of that rule is a value nothing can be constructed at, and the
other obligation on that behavior — an inclusive edge — is untouched.

Each of the five commits builds and its tests pass on their own.

### DIFF
--- a/souther-compiler/src/main/java/souther/compiler/check/InvariantBound.java
+++ b/souther-compiler/src/main/java/souther/compiler/check/InvariantBound.java
@@ -13,8 +13,10 @@ import java.util.Optional;
  * <p>Read off the comparison. What a report measures a position at is where the rules about it stop,
  * and that is not the same question as which runtime check the rule becomes: the runtime states a
  * decimal above zero directly and has no word for a decimal above five, so a reader written for it
- * answers "no bound" to a rule that plainly draws one. Asked here of {@code value OP literal} and of
- * nothing else, which is the whole of what a range can hold anyway.
+ * answers "no bound" to a rule that plainly draws one. Asked here of an ordering of
+ * {@code value} against a literal, and of one end at a time. An equality states both ends at once
+ * and is not read: it was not read before either, and giving a position one value is a different
+ * answer from bounding it, which the report has nowhere to put yet.
  *
  * <p>Over whole numbers a strict comparison names the adjacent value, and the end lands on a value
  * the rule admits. Over decimals there is no adjacent value, so the end stays on the literal and says
--- a/souther-compiler/src/main/java/souther/compiler/check/InvariantBound.java
+++ b/souther-compiler/src/main/java/souther/compiler/check/InvariantBound.java
@@ -1,0 +1,117 @@
+package souther.compiler.check;
+
+import souther.compiler.ast.Ast;
+import souther.compiler.numeric.Endpoint;
+import souther.compiler.types.Type;
+
+import java.math.BigDecimal;
+import java.util.Optional;
+
+/**
+ * Where one conjunct of a numeric newtype's invariant leaves its value able to stop.
+ *
+ * <p>Read off the comparison. What a report measures a position at is where the rules about it stop,
+ * and that is not the same question as which runtime check the rule becomes: the runtime states a
+ * decimal above zero directly and has no word for a decimal above five, so a reader written for it
+ * answers "no bound" to a rule that plainly draws one. Asked here of {@code value OP literal} and of
+ * nothing else, which is the whole of what a range can hold anyway.
+ *
+ * <p>Over whole numbers a strict comparison names the adjacent value, and the end lands on a value
+ * the rule admits. Over decimals there is no adjacent value, so the end stays on the literal and says
+ * that the literal is not one of its own.
+ *
+ * @param lower whether this bounds the value below; otherwise above
+ */
+public record InvariantBound(boolean lower, Endpoint end) {
+
+    private static final String VALUE = "value";
+    private static final BigDecimal LONG_MIN = BigDecimal.valueOf(Long.MIN_VALUE);
+    private static final BigDecimal LONG_MAX = BigDecimal.valueOf(Long.MAX_VALUE);
+
+    /** What {@code clause} says about a value carried as {@code base}, or empty where it says
+     * nothing a range can hold. */
+    public static Optional<InvariantBound> of(Ast.Expr clause, Type base) {
+        if (base != Type.INT && base != Type.DECIMAL) {
+            return Optional.empty();
+        }
+        if (!(clause instanceof Ast.Binary bin)) {
+            return Optional.empty();
+        }
+        // `0 <= value` says what `value >= 0` says: read the value-bearing side as the left one.
+        Ast.Expr left = bin.left();
+        Ast.Expr right = bin.right();
+        Ast.BinOp op = bin.op();
+        if (!isValue(left) && isValue(right)) {
+            Ast.Expr swap = left;
+            left = right;
+            right = swap;
+            op = mirrored(op);
+        }
+        if (!isValue(left)) {
+            return Optional.empty();
+        }
+        BigDecimal bound = literal(right);
+        if (bound == null || (base == Type.INT && bound.stripTrailingZeros().scale() > 0)) {
+            return Optional.empty();
+        }
+        boolean whole = base == Type.INT;
+        return switch (op) {
+            case GE -> Optional.of(new InvariantBound(true, Endpoint.inclusive(bound)));
+            case LE -> Optional.of(new InvariantBound(false, Endpoint.inclusive(bound)));
+            case GT -> whole ? stepped(true, bound.add(BigDecimal.ONE))
+                    : Optional.of(new InvariantBound(true, Endpoint.exclusive(bound)));
+            case LT -> whole ? stepped(false, bound.subtract(BigDecimal.ONE))
+                    : Optional.of(new InvariantBound(false, Endpoint.exclusive(bound)));
+            default -> Optional.empty();
+        };
+    }
+
+    /** A whole number's strict bound moved onto the value beside it — where the type has one. At the
+     * ends of what an {@code Int} holds there is nothing to move onto, and nothing is claimed. */
+    private static Optional<InvariantBound> stepped(boolean lower, BigDecimal onto) {
+        return onto.compareTo(LONG_MIN) < 0 || onto.compareTo(LONG_MAX) > 0
+                ? Optional.empty() : Optional.of(new InvariantBound(lower, Endpoint.inclusive(onto)));
+    }
+
+    private static boolean isValue(Ast.Expr e) {
+        return e instanceof Ast.Var v && v.name().equals(VALUE);
+    }
+
+    private static Ast.BinOp mirrored(Ast.BinOp op) {
+        return switch (op) {
+            case LT -> Ast.BinOp.GT;
+            case LE -> Ast.BinOp.GE;
+            case GT -> Ast.BinOp.LT;
+            case GE -> Ast.BinOp.LE;
+            default -> op;
+        };
+    }
+
+    /** A numeric literal, negation included. A bare integer counts against a decimal, since a literal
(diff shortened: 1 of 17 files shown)


kawasima:
Both P1s were real, and the second is worse than it looks from the interval alone.
Pushed as three commits.

## [P1] the candidate paths

`admissible` read the end's number whatever the end said, and it was the path every
position that is not a numeric newtype took. So the fix is not another
endpoint-aware sampler beside `inside` — it is that a pair of ends gives up a number
in exactly one place, `Endpoint.valueBetween`, and nothing else reads an end for
that. `inside`, `admissible` and `Intervals.representative` were three answers to
one question; there is one now, and it is where the whole/dense difference is
decided too.

It can return nothing, which is the honest answer for a range that holds no value
of the type — `(5, 6)` over `Int`. `choicesUnder` already had the shape for that: a
position of a type nothing stands for is not a value that was written and refused.

    data Pair = { low: Decimal, high: Decimal }
        invariant lower = low > 0m

    before  // no row for `flag=true`: every value tried was refused at construction
    after   | "flag=true" : (Pair { low = 1m, high = 0m }, true) -> <?>

`inReserve` was reading `bounds.max()` as the far value to hold back. It now holds
back nothing where the range stops short of that end — there is no value beside it
to hold back instead, and what is inside is what the first tier already offers.

## [P1] the intervals

The ends are handed over as ends now, so a range cut out of an open one is open
there.

What `representative` did was worse than picking the excluded value: for a decimal
range open below it asked `BoundaryDomain.DECIMAL.successor()`, which is
permanently empty, so it returned null and the class was reported as
`no value of this range can be written without a smallest step`. Every dense class
open at its lower end was ungeneratable. Through the shared selector it gets a
midpoint.

    data AboveFive = Decimal invariant above = value > 5m
    guard r.value < 10m else No

    before  · no row is in `r/5 <= x < 10`
            // no row for `r=5 <= x < 10`: every value tried was refused at construction
    after   · no row is in `r/5 < x < 10`
            | "r=5 < x < 10" : (AboveFive(7.5m)) -> <?>

## [P2] the narrowed provenance

Changed to compare the ends. But I could not construct a case where it changes
anything printed, and I think it cannot: an obligation is owed only where the
position holds the cut's value, and an end whose openness the record took away
never holds it. So the two cases are "the record moved the number", which was
already `moved`, and "the record took the value away", which now has no obligation
to carry the origin. The predicate says what it means; nothing reads the difference
today.

## The rule

`NumericDomain.Bounds` hands out `Endpoint`s and the three readers that turned one
into a value are gone, so the remaining `.value()` calls on an end are two in
`inReserve`, both after `max().inclusive()` has been checked.

## The two regression tests

Both are in `AnEndIsWhereItsRuleStopsTest`:

- `aBareDecimalIsOfferedAValueItsRangeHolds` — a position of no type but `Decimal`,
  left above zero by the record, is offered a value the record accepts.
- `aRangeCutOutOfAnOpenEndIsOpenThere` — a guard on a newtype open at its end
  leaves a class whose label and whose representative both keep the value out.

The first is written against a record clause rather than `low < high`, because
`low < high` bounds neither field until the search feeds an assignment back into
the projection, which is #479's remaining step and not something this can show.

## Also

`displacedRepresentativesOf`'s javadoc no longer describes the widening this PR
removes. `InvariantBound`'s claim is narrowed to an ordering against a literal, one
end at a time; `value == 5` is a range and is still not read, as before this PR.

`souther examples` over `souther-examples` still differs from `develop` in exactly
one place, the `quotaGap/quota = 1` obligation that no rule drew.

Issue #499, opened 2026-08-09
A classifier that could not read a value says only that it does not match


kawasima:
A truncation one wrapper in is reported as a value that could not be read. Both
halves of that are worth stating, and the second is where the defect is.

## Measured

The same failure, at the node a path names and one wrapper inside it:

```text
plain = Truncated                -> VALUE_TRUNCATED
plain = Unknown                  -> VALUE_UNREADABLE
cost  = Amount{value=Truncated}  -> VALUE_UNREADABLE     wrong
cost  = Amount{value=Unknown}    -> VALUE_UNREADABLE     right, by luck
```

One wrapper in, the kind is gone, and one of the two kinds comes out right
anyway. That is why nothing noticed.

## The equation that is false

Three things are layered here and the middle one is invisible to the third:

```text
path navigation          RowClasses.walk follows a record's fields. A newtype is
                         never a step: the path names `request.cost`, not
                         `request.cost.value`, and that is deliberate.

interpretation           Intervals.numberOf recurses through a Constructed with a
                         `value` field. The number at `request.cost` is not the
                         node `request.cost` names.

failure reporting        RowClasses.classify tests that node for Unknown and for
                         Truncated, hands it to `boolean matches(..)`, and names
                         the failure from what is left.
```

So the code assumes

> the `ObservedValue` a path names is the value a classifier classifies

and that is false wherever a classifier owns a projection. Nothing about this is
particular to newtypes: `PartitionClass` separates what a class means from how a
value is told to be in it, so a classifier is entitled to decode. What it cannot
do is say that its decoding failed.

## Two losses, not one

```text
ObservedValue.Truncated
        |  numberOf
       null
        |  matches
      false
        |  the fall-through
 VALUE_UNREADABLE
```

`numberOf` answers `null` for a value it could not read and for a value that is
not a number at all, and `matches` folds that into the same `false` as a number
outside the range. By the time `classify` has to name something, both the reason
and the fact that there was one are gone, so it names the failure that is right
for the case its comment is about:

```java
// The classes are exhaustive over the position, so a value in none of them is one this could
// not read rather than one outside the partition.
```

That argument is sound. It is being applied to values it was not written about.

Two of the three readers that need the value at a position already agree with the
classifier: `Coverages.readable` is `numberOf(at) != null`. `RowClasses` is the
one that reads the node instead.

## What is decided

The interpreter returns why it could not interpret. Whoever holds the knowledge
to derive an answer holds the reason it could not be derived — the same rule as
[...]

PR #501, opened 2026-08-09
A truncated observation says that it was, and nothing else

### LINKED ISSUE #498: A truncated observation carries data no measure can ask for
`ObservedValue.Truncated` carries two things and nothing but a test reads either:

```java
record Truncated(int observedSize, String digest) implements ObservedValue {}
```

That is the symptom. Reading what they were for turns it into something else:

> `Truncated` carries payload for semantics the model does not have.

Both were written for a reader named at the time. The size wants a dimension the
model does not divide positions by; the digest wants an operation the model does
not perform. Different halves of the same sentence, and neither is a reader
somebody forgot to write.

## What they were for

The commit that introduced the type says what the size is for:

> what it dropped becomes `Truncated`, **carrying the size a rule over that value
> would have read**

and the record's javadoc says what the digest is for:

> enough to tell two of these apart without holding either

## The size is for a dimension the model does not have

A position is classified only where it has classes, and there are two ways to get
them: `Partitions.caseClass`, which matches a sum case by its type name, and
`Intervals`, which matches a number against a range. Neither reads a size.
`Axis.derivable()` is `!classes.isEmpty()`, and `RowClasses.of` skips a position
that is not derivable before looking at any value.

A collection has no cases and no rules drawing a line through it, so it has no
classes. Nor does a string, or a newtype over one:

```text
· not derivable: request.memo      (String)
· not derivable: request.who       (a newtype over String)
· not derivable: request.items     (List<Item>)
```

The two positions in that model that were classified were a sum and a number an
invariant divides.

So the problem is not that a reader was forgotten. There is no classification
dimension that could read a size, and adding one is a change to what an axis is.
Which produces an inversion worth stating on its own:

> `observedSize` is informative exactly where the model cannot look, and `-1`
> everywhere it can.

The three producers that record a real size — a string past `maxText`, a
collection past `maxElements`, a map past the same — all write at positions
nothing classifies. The two that reach a classified position are the shared node
budget and the depth limit, and both pass `-1`.

## The digest is for an operation that does not exist

Two observed values are compared in one place, where a boundary asks whether a
row put its value there, and a truncated value is dropped before the comparison:

```java
private static boolean readable(ObservedValue at) {
    return numberOf(at) != null;
}
```

`numberOf` answers nothing for a `Truncated`, so the fingerprint is never
reached. Nothing in the compiler asks whether two observations are the same.

And the field would not serve that if something did. Only one producer puts a
fingerprint there:

```text
text past maxText     ->  hash of the string
node budget spent     ->  "node-limit"
past maxDepth         ->  "depth-limit:" + the type
past maxElements      ->  "elements"
past maxEntries       ->  "entries"
```

Four of the five write which limit was reached. So the field holds an identity at
one producer and a cause at four, which are different things, and no reader could
take it as either.

## Three things, two fields, one of them a union

What a truncation could say separates cleanly:

```text
what stopped it     node budget / depth / text length / element count / entry count
how much there was  123 characters / 42 elements / 17 entries
which value it was  a fingerprint
```

The record has a field for the second, a field that is the first at four
producers and the third at one, and no field for the first as such. So *should
the report say which limit was reached* is not a reason to keep `digest`; it is a
reason to have a cause, which is not what either field is.

## Not the same failure as #491

There a distinction was recorded and a reader that existed dropped it by testing
a shape. Here the representation arrived first and the operation that would give
it meaning was never built. The first is information lost in a projection; this
is a representation with nothing to project onto.

## What is decided

Under the model as it stands, the only thing about a truncation that anything
consumes is that it was one — `RowClasses` turns the case into
`VALUE_TRUNCATED`, and the report says so in words. So:

```java
record Truncated() implements ObservedValue {}
```

Not because the fields are unused. Because no classification dimension can
consume a size, and no operation consumes the identity the digest promises.

One thing to hold the change to. A record with no components makes every
`Truncated` equal to every other under Java's own `equals`, and nothing may be
allowed to read that as two observations being the same. Structural equality of
an `ObservedValue` is not used for anything today, which is what makes the change
safe; it is not a licence to start.

Two real questions are folded into the current payload and would be answered
separately if either came back:

- whether the cardinality of a collection or a string is something the model
  divides a position by, which is a change to what an axis is;
- whether an observation needs an identity that survives being dropped, which is
  about comparing observations rather than about partitioning them.

Either would arrive with the data it needs. Neither is served by keeping this.

kawasima:
Closes #498.

`ObservedValue.Truncated` becomes a case with no payload. The two fields it
carried were each written for a reader named at the time, and the model has no
way to build either — which is what the investigation on #498 turned up, and why
this is not a dead-field cleanup.

## What the fields were for, and why nothing could read them

The size was for *a rule over that value*, in the words of the commit that
introduced the type. A position gets classes from the cases of its type or from
where a rule cuts its range, and `RowClasses` skips a position with no classes
before looking at any value. A collection, a string, and a newtype over one have
neither:

```text
· not derivable: request.memo      (String)
· not derivable: request.who       (a newtype over String)
· not derivable: request.items     (List<Item>)
```

So the size was informative exactly where nothing looks: the three producers that
knew a real one wrote at those positions, and the two that reach a classified
position — the shared node budget and the depth limit — both passed `-1`.

The fingerprint was for telling two truncated values apart. Two observations are
compared in one place, and `readable()` drops a truncated one before the
comparison because it is not a number. Nothing asks whether two observations are
the same. And only one of the five producers put a fingerprint there; the other
four put the name of the limit they hit, so the field was an identity at one and
a cause at four.

## What is not in this

Nothing about how a truncation is classified. A truncation one wrapper in is
still reported as a value that could not be read — that is #499, and mixing it in
here would join the payload question back to the classification vocabulary the
investigation separated it from.

No `equals` of its own. Two of these now compare equal, which is Java saying they
carry the same nothing. Structural equality of an `ObservedValue` is not read
anywhere; writing an identity back in would restore by another route exactly the
semantics this removes.

## Checks

```text
git grep -nE 'observedSize|new Truncated\('   nothing
git grep -n digest                            only CoverageSites and Continuation, other concepts
git diff --check                              clean
mvn clean test                                full reactor green
```

Four files: the record, its five producers, and the two tests that asserted
payload values rather than the case.

### DIFF
--- a/souther-compiler/src/main/java/souther/compiler/ObservedValues.java
+++ b/souther-compiler/src/main/java/souther/compiler/ObservedValues.java
@@ -75,9 +75,10 @@ final class ObservedValues {
     }
 
     /** A collection past the element limit is dropped whole rather than kept as a prefix. A prefix
-     * read back is a collection nobody wrote, and it arrives looking like one that was: a measure
-     * would classify it, and be answering about a value that was never there. Saying the
-     * observation was stopped is the one thing that stays true. */
+     * is not the value that was written, and returned as a {@link ObservedValue.Sequence} it would
+     * be the same thing as a complete observation of a shorter collection — nothing about the two
+     * would differ. {@link ObservedValue.Truncated} keeps the one fact that stays true: the
+     * observation stopped. */
     private ObservedValue sequence(Iterable<?> it, int depth) {
         List<ObservedValue> out = new ArrayList<>();
         for (Object e : it) {
--- a/souther-compiler/src/main/java/souther/compiler/ObservedValues.java
+++ b/souther-compiler/src/main/java/souther/compiler/ObservedValues.java
@@ -74,8 +74,10 @@ final class ObservedValues {
                 : new ObservedValue.Truncated();
     }
 
-    /** A collection past the element limit is dropped whole rather than kept as a prefix: its size is
-     * what a rule over it reads, and a prefix would answer a length question with the wrong number. */
+    /** A collection past the element limit is dropped whole rather than kept as a prefix. A prefix
+     * read back is a collection nobody wrote, and it arrives looking like one that was: a measure
+     * would classify it, and be answering about a value that was never there. Saying the
+     * observation was stopped is the one thing that stays true. */
     private ObservedValue sequence(Iterable<?> it, int depth) {
         List<ObservedValue> out = new ArrayList<>();
         for (Object e : it) {
--- a/souther-compiler/src/main/java/souther/compiler/ObservedValues.java
+++ b/souther-compiler/src/main/java/souther/compiler/ObservedValues.java
@@ -20,11 +20,11 @@ import java.util.Map;
  * one {@link NeutralForm#of} makes in the other direction: that one produces the form a decoder reads,
  * this one produces the form a measure reads.
  *
- * <p>The walk never fails. A value it cannot read becomes {@link ObservedValue.Unknown} and an
- * observation {@link Limits} stopped becomes {@link ObservedValue.Truncated}, both carrying why,
- * because a measure that cannot classify a row has to say so rather than count the row as covering
- * nothing. The second is not the same as a large value: the node budget is one for the whole walk,
- * so a small value observed after a large sibling is truncated as well.
+ * <p>The walk never fails. A value it cannot read becomes {@link ObservedValue.Unknown}, which says
+ * why, and an observation {@link Limits} stopped becomes {@link ObservedValue.Truncated}, which says
+ * only that — a measure that cannot classify a row has to say so rather than count the row as
+ * covering nothing. The second is not the same as a large value: the node budget is one for the
+ * whole walk, so a small value observed after a large sibling is truncated as well.
  */
 final class ObservedValues {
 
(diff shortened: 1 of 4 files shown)


kawasima:
Fixed, `8eace517`. Full reactor `clean test` green.

You are right that this is not a stale comment. `sequence` justified dropping a
collection by *the size a rule over it reads*, which is the concept this branch
removes — so the change would have taken the field out and left its reason in
place, with the record's javadoc and the method's javadoc asserting opposite
models of what a position can be divided by. That is grounds for putting the
field back, written into the file that removed it.

Going whole is still right, and now says why in terms that survive: a prefix read
back is a collection nobody wrote, and it arrives looking like one that was, so a
measure would classify it and be answering about a value that was never there.
The test's javadoc says the same in one line.

I checked whether the same rationale had been written anywhere else. It had not —
the other matches for `prefix` in the tree are the formatter, the vector, and
string handling, none of them about observation limits.

kawasima:
Fixed, `2d0076b6`. Full reactor `clean test` green.

Right again, and it is the same mistake one step smaller: I reached for a
classifier to justify something that does not need one, in a branch whose finding
is that nothing classifies a collection. `RowClasses.of` skips a position with no
classes before it looks at any value, so the prefix would never reach a measure
at all.

The reason is representational and stops there: a prefix returned as a `Sequence`
is indistinguishable from a complete observation of a shorter collection — the two
are the same value, and nothing anywhere could tell them apart. `Truncated` keeps
what stays true.

The test's javadoc already said it that way and is unchanged.

PR #502, opened 2026-08-09
A construct the width could keep on one line still owns its comment

### LINKED ISSUE #500: A comment above a definition's body is given to the declaration
A comment written above a definition's body comes back above the declaration.

```souther
module m exposing ( g )

let g (alpha: Int, beta: Int, gamma: Int): Int =
    // about the body
    alpha + beta + gamma + alpha + beta + gamma + alpha + beta + gamma + alpha + beta
```

`souther fmt` writes

```souther
module m exposing ( g )

// about the body
let g (alpha: Int, beta: Int, gamma: Int): Int =
    alpha + beta + gamma + alpha + beta + gamma + alpha + beta + gamma + alpha + beta
```

The body is on a line of its own in both. The comment was about the body and now describes the declaration, and a reader has no way to tell it once said something else. This is #455 again in another place; there it was a trailing comment handed to the next declaration, here it is a leading comment handed to the enclosing one.

The same comment written above a statement of a block body stays where it was, and one written above the branch of an `if` stays too.

## Why

`fnDef` writes a boundary before the body that the layout may break (`Formatter.java:729`). Comment attachment does not ask the document about that. It asks `takesALineOf(parent, child)` (`:1837`), reading the table `linesOf` (`:1856`), and `linesOf` has no row for `FN_DEF`. So `place` (`:1492`) files the comment against the nearest construct above the body that the table says has a line, `slotOf` (`:1610`) walks past the body, and the declaration takes it.

Nothing catches it. `TheFormatterAccountsForEveryCommentTest` counts the comments consumed against the comments the tree holds, and this is one comment in and one comment out. `ACommentKeepsItsMemberHoweverItIsWrittenTest` says so of itself:

> A comment moved from one member to the next is one comment in and one comment out, and the code around it is unchanged; what changed is the only thing a comment is for.

## Adding a row would keep the cause

Which children of a construct can own a line is written down twice.

`linesOf` is one of the two, twenty-one parent kinds, read only by comment attachment. The other is the placement of gaps while the document is built. Of the twenty-three uses of `HARD_GAP` in `Formatter`, twelve write this obligation and eleven are other ones:

| what the site writes | uses |
| --- | --- |
| a construct's members, one to a line | 12 |
| a comment on a line of its own | 7 |
| a bracket of a construct written down the page | 2 |
| the newline a file ends with | 1 |
| the blank line between two top-level items | 1 |

and the conditional form of the same obligation is written as `SOFT_GAP` by `separated` (`:176`) and `chained` (`:251`), with the bracket obligation's own conditional form in `delimited` (`:207`).

So the two spellings are not the same shape, and adding `FN_DEF` to one of them fixes this source and leaves the next construct to be added exposed the same way.

## What the two of them share is a slot, not a line

`FN_DEF` is what makes that precise. The boundary before a definition's body may break or may not, so whether the body has a line of its own is a fact about the width. The comment still belongs to the body. What comment attachment needs to know is that the construct has somewhere to put a line, not that it has one:

```
                                   NONE  |  CONDITIONAL  |  ALWAYS
    a comment may be owned here      no  |      yes      |   yes
    the document writes             GAP  |   SOFT_GAP    |  HARD_GAP
```

The order also runs the other way from what a name like "line structure" suggests. Attachment is decided before any of the document is built (`Formatter.java:279`), it reads the slot, and its answer then changes the layout: a comment given to a construct is written with a `HARD_GAP` after it, so a group that would have been flat breaks. Ownership reads the slot and drives the line, rather than reading the line.

## Acceptance

- The relation is one thing, saying whether a child can own a line and whether the break is forced or conditional. Its unit is a child's place in its construct rather than the pair of kinds: an `if` writes three expressions, the first of them between `if` and `then` with no boundary the layout can break, and `if flag then p else q` makes all three a `VAR_EXPR`.
- Comment attachment reads it. `linesOf` is gone.
- Only the sites writing that obligation read it — not the bracket, the top-level separation, the file's newline or the comment placement, which are four other obligations that also spell `HARD_GAP`.
- Its domain is swept from candidates built on the `(parent kind, child kind)` axes and admitted by the parser, rather than from the constructs the formatter happens to write.
- The existing comment fixtures and the forced and conditional layout fixtures pass.
- Turning one cell of the relation off fails a comment fixture **and** a layout fixture. Two readers that do not both fail have not been reading one thing.

The pair to hold it with is this defect and the `if` branch beside it: the same comment, one construct where the body's line is conditional and one where it is not, with the owner required to be the same in both.

Found while writing out the rule inventory for #444, which is where the two spellings were counted. The defect is the formatter's own and does not depend on anything that issue decides.

kawasima:
A comment written above a definition's body came back above the declaration, and so did one above either branch of an `if`.

```diff
 let g (x: Int): Int =
-    // about the body
     x
```

The comment was about the body. The reader is left with one describing the declaration and no way to see it once said something else. This is #455 in another place — there a trailing comment handed to the next declaration, here a leading comment handed to the enclosing one.

## Why

Which construct a comment is about is decided before any of the document is built, so it cannot be read off the layout. It is read from `linesOf` (`Formatter.java`), a table of which children a construct writes on lines of their own, and that table had no row for a definition or for an `if` — while `fnDef` and `ifExpr` both write a boundary before those children that the layout may break. `slotOf` walked past them to the enclosing construct.

Nothing caught it. The count of comments consumed against comments held sees one comment in and one comment out, and `ACommentKeepsItsMemberHoweverItIsWrittenTest` says so of itself:

> A comment moved from one member to the next is one comment in and one comment out, and the code around it is unchanged; what changed is the only thing a comment is for.

## What is here

The table is asked of the two nodes rather than of their kinds, because for these two constructs the kinds do not answer it.

An `if` writes three expressions and only the two after the condition open a line: written `if flag then p else q` all three are a `VAR_EXPR`, so which of them has a line is a fact about where it stands. The condition keeps its comment travelling, and there is a fixture saying so — holding children is not what makes a construct able to carry one.

A definition written as a lambda has its parameters moved to the left of the `=`, so what the canonical form writes after the `=` is that lambda's body rather than the child the source has there. The comment is about the body, the source gives it to the lambda, and the two are not the same construct. **That case keeps the behaviour it had**, with a fixture stating it as a limit rather than leaving it as a case nobody tried.

## What is deliberately not here

An earlier version of this branch tried to make the layout and the comment attachment read one relation — the same relation, since both need to know which children a construct writes down the page. Measuring it found that they do not want the same answer:

- `DATA_DEF` writes `data Positive = Int` with no boundary before the representation, while the representation owns the trailing comment of that header line. `BEHAVIOR_SIG` and its return type are the same shape. So *what goes before a place* and *whether a place can own a comment* are different questions, and the existing code already knew: both of those are excluded from `opensALine`.
- A slot is a position in the canonical structure, not in the source. Where a definition's lambda is lifted, two different source elements name one canonical place and the canonical body is a descendant rather than a child, so no function of the source tree identifies it.

Merging one relation over those would move the duplication one level down rather than remove it, so this PR does not claim it. The findings belong to #444, which is where the model that has to account for them lives.

## What holds it

`ADefinitionsBodyOwnsItsCommentTest`: a definition's body, both branches of an `if`, the condition that owns nothing, and the lifted lambda that is left as it was. 4353 tests pass.

Closes #500

### DIFF
--- a/souther-compiler/src/test/java/souther/compiler/fmt/ADefinitionsBodyOwnsItsCommentTest.java
+++ b/souther-compiler/src/test/java/souther/compiler/fmt/ADefinitionsBodyOwnsItsCommentTest.java
@@ -1,0 +1,129 @@
+package souther.compiler.fmt;
+
+import org.junit.jupiter.api.Test;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+
+/**
+ * A comment written above what a definition or an {@code if} writes after its keyword stays there.
+ * The construct is given a line by the layout — the boundary before it breaks where the body does
+ * not fit — and a comment is about what it was written above whether or not the width would have
+ * kept that on one line. Owning one is then what puts it on a line of its own, since a comment
+ * cannot share the line after it.
+ *
+ * <p>{@link ACommentKeepsItsMemberHoweverItIsWrittenTest} sweeps the ways a member can be written,
+ * and the survival sweep cannot ask either of them: a comment handed to the enclosing construct is
+ * one comment in and one comment out, and only which construct it is about has changed.
+ */
+class ADefinitionsBodyOwnsItsCommentTest {
+
+    @Test
+    void aDefinitionsBody() {
+        assertEquals("""
+                module m
+
+                let g (x: Int): Int =
+                    // about the body
+                    x
+                """, Formatter.format("""
+                module m
+                let g (x: Int): Int =
+                    // about the body
+                    x
+                """));
+    }
+
+    @Test
+    void aBranchOfAnIf() {
+        assertEquals("""
+                module m
+
+                let k (x: Int): Int =
+                    if x > 0 then
+                        // about the then branch
+                        1
+                    else
+                        0
+                """, Formatter.format("""
+                module m
+                let k (x: Int): Int =
+                    if x > 0 then
+                        // about the then branch
+                        1
+                    else
+                        0
+                """));
+    }
+
+    @Test
+    void andItsOtherBranch() {
+        assertEquals("""
+                module m
+
+                let k (x: Int): Int =
+                    if x > 0 then
+                        1
+                    else
+                        // about the else branch
+                        0
+                """, Formatter.format("""
+                module m
+                let k (x: Int): Int =
+                    if x > 0 then
+                        1
+                    else
+                        // about the else branch
+                        0
+                """));
+    }
+
+    /**
+     * And what an {@code if} writes on its header line owns nothing. The condition stands between
+     * {@code if} and {@code then} with no boundary the layout can break, so a comment above it is a
+     * comment about the {@code if} — which is why holding children is not what makes a construct
+     * able to carry one.
+     */
+    @Test
+    void butNotWhatTheHeaderLineHolds() {
+        assertEquals("""
+                module m
+
+                let k (x: Int): Int =
+                    // about the test
+                    if x > 0 then 1 else 0
+                """, Formatter.format("""
+                module m
+                let k (x: Int): Int =
+                    if
+                        // about the test
+                        x > 0
+                    then 1 else 0
+                """));
+    }
+
+    /**
+     * A definition written as a lambda is left as it was. Its parameters move to the left of the
+     * {@code =}, so what the canonical form writes after the {@code =} is not the child the source
+     * has there — it is that child's body, a level further down. The comment is about that body and
+     * the construct the source gives it to is the lambda, and until the two are the same question
+     * this is one construct where the comment still travels to the declaration.
+     *
+     * <p>Held here so that it is a stated limit rather than a case nobody tried. Making it right
+     * needs the canonical structure to be something a comment can be filed against, which is
+     * issue #444.
+     */
+    @Test
+    void butADefinitionWrittenAsALambdaIsLeftAsItWas() {
+        assertEquals("""
+                module m
+
+                // about the body
+                let f (x) = x
+                """, Formatter.format("""
+                module m
+                let f = (x) ->
+                    // about the body
+                    x
+                """));
+    }
+}
--- a/souther-fmt/src/main/java/souther/compiler/fmt/Formatter.java
+++ b/souther-fmt/src/main/java/souther/compiler/fmt/Formatter.java
@@ -717,8 +717,10 @@ public final class Formatter {
         if (intrinsic.isPresent()) {
             SyntaxToken body = intrinsic.get().token(SyntaxKind.STRING_LIT).orElseThrow();
             return TokenDoc.node(n.kind(), concat(head, GAP, ASSIGN,
-                    group(nest(INDENT, concat(SOFT_GAP, TokenDoc.node(intrinsic.get().kind(),
-                            concat(ident("intrinsic"), GAP, token(body))))))));
+                    group(nest(INDENT, concat(SOFT_GAP, aboveOf(intrinsic.get()),
+                            TokenDoc.node(intrinsic.get().kind(),
+                                    concat(ident("intrinsic"), GAP, token(body))),
+                            afterOf(intrinsic.get()))))));
         }
         var block = n.child(SyntaxKind.BLOCK_EXPR);
         if (block.isPresent()) {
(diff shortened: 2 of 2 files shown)


kawasima:
Both points taken, and both are now held by a mutation rather than by argument.

## The unit is the place

`memberBoundaryOf` took a construct and the positional fact was written twice — once by where `ifExpr` put its boundaries and once by a predicate the attachment kept for itself. Dropping that predicate failed comment tests and no layout test, which is the measurement the objection asked for.

A `MemberSlot` is now the construct together with which of its places this is, and `slotFor` is the one thing that decides the place. Both readers go through it. `Member` and `Segment` carry their slot, so `separated` and `chained` take no construct at all — they ask the member in front of them. `ifExpr` asks about its condition too, which is where the answer that a condition opens no line lives.

Two things this surfaced, both of them the divergence the PR is about:

- a data declaration's product body is written on a line of its own and the attachment did not know, so a comment above it travelled;
- the break before a sum body is the comment's own doing rather than a member boundary, so it stays a forced break and the attachment stays out of it.

## The sweep is per place

`answerFor(SyntaxKind)` proved a function total and nothing about the rule's domain, exactly as you say. It now walks every parent occurrence the corpus builds and asks about every child it holds — 5000-odd places — and a second method pins that an `if` answers its three places differently, so a relation that answered per construct fails it.

## What the mutation says now

Turning the condition into a branch in `slotFor` fails, by name:

```
layout      AMemberWhoseLineTheWidthDecidesKeepsItsCommentTest.andTheLayoutKeepsTheConditionOnTheHeaderLine
attachment  ACommentKeepsItsMemberHoweverItIsWrittenTest.aPartOfAnExpressionWithNoLineOfItsOwn
            AMemberWhoseLineTheWidthDecidesKeepsItsCommentTest.butNotWhatTheHeaderLineHolds
            ...aBranchOfAnIf, ...andItsOtherBranch
sweep       EveryConstructSaysWhatItWritesBetweenItsMembersTest.andAConstructWithSeveralPlacesAnswersEachOfThem
```

The layout row needed a fixture that did not exist. An `if` short enough to fit writes the same text whether its condition's boundary can break or not, because an unbroken boundary holds the spacing rule's answer either way — so every layout test stayed green under the mutation until an `if` over the width pinned where the condition goes. That fixture is its own commit.

## The `if` history

Here I think the record is the other way round. Against `develop` at `5d5cf2c6`, the fixture in `aBranchOfAnIf` gives

```
module m

// about the then branch
let k (x: Int): Int = if x > 0 then 1 else 0
```

so the comment left the branch before this PR. What I said earlier in the session — that an `if`'s branch kept its comment — came from a source I had written without `then`, which does not parse (`expected THEN_KW but found IDENT`); I read a filtered slice of the output and took it for a formatted result. The PR text is right and that remark was wrong.

4356 tests pass.

kawasima:
Right on both counts, and the correction to my fallback was the important one — I checked it rather than taking it.

Built at `0e6c026c` and run:

```
let f = (x) ->
    // about the body
    x
```

```
CRASH 1 comment(s) in this source reached no construct and would have been dropped
```

So the regression is in the first commit, not in `MemberSlot`, and "shrink to the four commits" would have shipped it. `MemberSlot.owner` made the mismatch legible; it did not create it.

The branch is reset to a single commit against `develop`. `memberBoundaryOf` is gone with it — an abstraction with two known counterexamples is worse in `develop` than two questions kept apart, and the counterexamples are already written down:

- `data Positive = Int` writes no boundary before the representation while the representation owns that header line's trailing comment, and `BEHAVIOR_SIG` with its return type is the same shape. `boundary-before` and `can-own-a-comment` are not one decision, and `opensALine` already excluded exactly those two.
- The lifted lambda: two source elements name one canonical place and the canonical body is a descendant, so no function of the source tree identifies the place.

What is left is the fix and its limits. The node-keyed `takesALineOf` stays, because the `if` condition and the lifted definition are both positional; the lifted case keeps `develop`'s behaviour and has a fixture that says so is a stated limit rather than an untried case. 4353 tests pass.

The reading of the sequence is one I would keep. Four rounds, one shape:

```
#476       an interval had no representation      -> the generator's boundaries were not the source's
#500       line ownership had two spellings       -> boundary and ownership were not one question
review 1   the answer was per construct           -> the unit was the place
review 2   the place was a source child           -> the place belongs to the canonical structure
the sweep  the domain was what the walk reached   -> not what the grammar admits
```

Each time a point in the generator's traversal was given a name and asked to be a decision. That is going to #444 as its own comment, with the measurements attached, since it is that issue's model these were testing.

PR #503, opened 2026-08-09
A classifier says why it could not read the value

### LINKED ISSUE #499: A classifier that could not read a value says only that it does not match

kawasima:
Closes #499.

`Classifier` answers a `Membership` instead of a `boolean`, and the numeric
reader stops writing two kinds of nothing as one.

## What was wrong

A newtype is not a step in a path, so the value at `request.cost` is the
construction and the number is one node inside it. `Intervals` reads through it;
`classify` tested the node the walk found. A limit reached inside left a
construction with nothing visibly wrong, and the last line named a failure it had
not observed:

```text
plain = Truncated                -> VALUE_TRUNCATED
plain = Unknown                  -> VALUE_UNREADABLE
cost  = Amount{value=Truncated}  -> VALUE_UNREADABLE     wrong
cost  = Amount{value=Unknown}    -> VALUE_UNREADABLE     right, by accident
```

Two losses, not one: `numberOf` answered `null` both for a value it could not
read and for a value that is not a number, and `matches` folded that into the
same `false` as a number outside the range.

## What it is now

```text
any Match                 ->  that class
else Incomplete, agreeing ->  that incompleteness
else                      ->  internal failure
```

The shape tests in `classify` are gone. They were the same question asked in a
second place and answered from a different node, which is what let the two come
apart.

`Incomplete` answers are not returned on as they arrive: a class may read less of
a value than the one after it, so one saying it could not read says nothing about
the rest.

## The last line

`Axis` says its classes are *exclusive and exhaustive over the position's
values*, so every class reading a value and none holding it is that contract
broken, not a third thing a row can be. Answering it with `VALUE_UNREADABLE`
would report a defect in the partition as a measurement failure and name the row
for it.

Walked, and nothing produces it: a parameter is a single named type so no
anonymous union reaches an input axis; a fixture that would not build never
reaches `inputs`; the case classes are built from every leaf; the decoder refuses
a number outside its newtype's invariant; and `Intervals` divides a domain end to
end, thresholds adding boundaries rather than holes. The whole existing suite
runs without reaching it.

Two classes disagreeing about why they could not read one value is held to the
same line, rather than given a precedence nobody has asked for.

## Checks

Eight cases, including the two contract ones as unit tests rather than
reproductions. Dropping the reason on the way back out of the numeric reader
fails two of them, so they are not passing on the shape alone.

Full reactor `clean test` green.

## Noted, not done here

`numberOf` is written four times — `Intervals`, `Coverages`, `BoundaryAssessment`,
`Generator` — each its own copy of the same walk through a newtype. Only the
first is touched here, and `numberOf` is kept on top of the new reader so this
does not add a fifth.

### DIFF
--- a/souther-compiler/src/main/java/souther/compiler/partition/Classifier.java
+++ b/souther-compiler/src/main/java/souther/compiler/partition/Classifier.java
@@ -2,6 +2,8 @@ package souther.compiler.partition;
 
 import souther.compiler.observe.ObservedValue;
 
+import java.util.function.Predicate;
+
 /**
  * Whether a value an {@code example} row was given belongs to one equivalence class.
  *
@@ -9,14 +11,33 @@ import souther.compiler.observe.ObservedValue;
  * author already wrote while nothing can generate a representative for it — a type whose invariant is
  * a pattern, a record whose fields constrain each other — and treating those as unmeasurable would
  * throw away coverage that is already there.
+ *
+ * <p>The answer is a {@link Membership} and not a {@code boolean} because a classifier may have to
+ * read the value before it can say anything about it, and reading can fail. What did the reading
+ * says why it failed; nothing downstream is in a position to work it out.
  */
 @FunctionalInterface
 public interface Classifier {
 
-    boolean matches(ObservedValue value);
+    Membership membershipOf(ObservedValue value);
+
+    /**
+     * One that answers by looking at the value it is given.
+     *
+     * <p>For the classes told apart by shape — a case of a sum, a {@code Bool}, whether an optional
+     * holds anything — where the value at the position is the value to look at. An observation that
+     * did not arrive is answered before the predicate is asked, because that is a fact about the
+     * observation rather than a thing the class declines.
+     */
+    static Classifier byShape(Predicate<ObservedValue> holds) {
+        return value -> {
+            Membership.Incomplete unread = Membership.unread(value);
+            return unread != null ? unread : Membership.of(holds.test(value));
+        };
+    }
 
     /** Recognises nothing: a class that exists but cannot be told from another by looking. */
     static Classifier none() {
-        return _ -> false;
+        return _ -> Membership.NO_MATCH;
     }
 }
--- a/souther-compiler/src/main/java/souther/compiler/partition/Intervals.java
+++ b/souther-compiler/src/main/java/souther/compiler/partition/Intervals.java
@@ -122,9 +122,9 @@ final class Intervals {
         for (Interval range : intervals) {
             String id = path + "/" + range.label();
             BigDecimal inside = representative(range, decimal);
-            Classifier is = v -> {
-                BigDecimal number = numberOf(v);
-                return number != null && range.holds(number);
+            Classifier is = v -> switch (read(v)) {
+                case Read.Number number -> Membership.of(range.holds(number.value()));
+                case Read.Missing missing -> new Membership.Incomplete(missing.code());
             };
             classes.add(inside == null
                     ? PartitionClass.ungeneratable(id, range.label(), is,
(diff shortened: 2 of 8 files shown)


kawasima:
Both fixed, `546a9838`. Full reactor `clean test` green. #504 filed for the third.

Worth saying what the two have in common, because it is the same thing this
branch is about, committed inside the fix for it.

## 1. The disagreement was judged from a prefix

Right, and the reason is in the PR text I wrote: *a class saying it could not read
says nothing about the rest*. It says nothing about whether one of them holds the
value either, so nothing about an incompleteness can be settled until every class
has answered. It is what is left once no class has claimed the value — a residue,
not an event — and I treated it as something that happens.

Noted and judged after the loop. The ordering test is in, and reverting the fix
fails it.

## 2. `read` still collapsed two nothings into one

Also right, and this one is the collapse the branch removes, rebuilt one layer
down. `VALUE_UNREADABLE` is a value that could not be read back into an observed
form at all; a `Text` was read. What does not fit is the partition.

So `Read` has the third answer too — a number, an observation that did not
arrive, or a value that arrived and is not one — and not-a-number is a class
declining, which lands on the exhaustiveness contract where it belongs.

## What both of them are

I widened one interface from two answers to three and left the reader underneath
it at two. The rule is not about `Classifier`: it is that every reader in the
chain needs somewhere to put *I read it and it is not mine* apart from *there was
nothing to read*. Applied at one seam, the defect reappears at the next one down.

Which is also what #504 is, at the seam above: `walk` answers `null` both for a
path that leads nowhere and for a record a limit stopped, and a position below a
truncated record reports as a value that could not be read. The chain is
`walk -> read -> classifier -> classify`, and this branch fixed exactly one link
of it.

## Tests

Three added. The shape-classifier one you asked for is the direct regression test
for moving that check out of `classify`, and there was nothing covering it.
Reverting either fix fails the new tests; reverting the reason inside `read`
fails the two truncation tests.

Issue #524, opened 2026-08-09
Compiler transformations can build IR trees deeper than recursive walks can safely traverse


kawasima:
The compiler bounds CST depth on the assumption that every recursive walk therefore has bounded depth. `AstBuilder`, `HelperInliner` and `Elaborator` can each build an IR of unbounded depth from a shallow CST, so any later recursive walk can exhaust the JVM stack.

## What happens

None of these sources nests as written. Measured through `Compiler.compiled` on a 256 KB stack — the size `CompileDeepExpressionTest` pins, so that the boundary is the same everywhere.

| source | |
|---|---|
| a block of 100 sequential `let`s | compiles |
| a block of 110 sequential `let`s | exhausts the stack |
| a block of 110 sequential `guard`s | exhausts the stack |
| a block of 800 sequential `let`s | exhausts it inside `AstBuilder`, while building |
| 40 chained values (`let vN = vN-1 + 1`) | compiles |
| 60 chained values | exhausts the stack |
| 3 chained values, each body 60 operators long | exhausts the stack |
| 10 chained values, each body 10 operators long | exhausts it walking `Core` |
| 100 chained helpers | exhausts the stack |

Which walk gives out is not a property of the source. It moves between runs and between builds — `AstBuilder.foldStatements`, `Resolve.expr`, `Lower.desugar`, `Elaborator.elaborate`, `BinaryElaborator.operand`, `Terms.affine`, `HelperInliner.inline` / `substituted` / `expandCall`, and `InvariantChecker.walk` over `Core` have all been the one — because what runs out is whichever walk is deepest in the stack at that moment. What is stable is which side of the boundary a source falls on.

Only a pinned stack has a boundary at all. On a default stack the same source compiles or does not depending on what the JVM did before it: a chain of 1000 values compiled three times out of three in a fresh JVM and exhausted the stack twice out of twice after other compiles in the same one.

## Why it matters

`CstParser.MAX_DEPTH` is 64, and `CstParser` publishes what the bound is for:

> Published because it is what a caller inherits: `Green#depth()` of a tree from `parse` never exceeds it. A walk that descends this tree by recursion — which is every walk there is — needs no limit of its own.

That is true of the CST. Not one of the walks above descends a CST. They descend `Ast` and `Core`, and three passes build those with a depth that has no relation to the CST's:

[...]

Issue #523, opened 2026-08-09
A chain of value declarations costs the square of its length


kawasima:
A chain of value declarations costs the square of its length.

## What happens

A value is substituted where it is named (ADR-0072), and the substitution copies the body. `v100`'s body contains a copy of `v99`'s, which contains a copy of `v98`'s, and every value is also checked on its own — so a module of N values holds O(N²) expression nodes and elaborates that many.

```
module chainN exposing (f)

let v0 = 1
let v1 = v0 + 1
...
let vN = vN-1 + 1

behavior f : (x: Int) -> Int
let f (x) = x + vN
```

Counting the calls to `Elaborator.elaborate`, with the doubling of #507 already fixed:

| N | elaborate calls | `compile` |
|---|---|---|
| 10 | 167 | 0.44 s |
| 22 | 623 | 0.45 s |
| 100 | 10,607 | 0.46 s |
| 1000 | 1,006,007 | 3.88 s |

The differences grow linearly, which is the square.

## Why it matters

This is not what #507 was about — #507 was the doubling in `BinaryElaborator`, which made the same chain cost 2^N, and it is fixed. What is left is the substitution itself, and it is the reason the original issue asked for the cost to be "linear in the number of names". A thousand names is a large module but not an unreasonable one, and four seconds is spent on structure the source shares and the compiler does not.

## Expected

A module costs time proportional to what it declares. The question is a representation one rather than a tuning one: the source has one `v99` and elaboration has a hundred copies of it. Sharing the expansion, answering each value once and reusing the answer, or keeping the name and binding it rather than copying the tree are the shapes an answer could take.

## Where

`HelperInliner.valueOf`/`substituted` (the copying) and every phase that walks what they produced.

Split out of #507, where the measurements were taken.

kawasima:
Measuring this turned up a second term, larger than the one described here and independent of it:
the walk that marks what a value carried ran once per level of substitution, which is N cubed. It
is #531, fixed in #532, and it is not this issue — with it gone, a chain of eight hundred values
still costs 644,807 elaborations, the same number as before.

What that leaves is worth writing down, because it moves where the fix has to go.

The quadratic is not the reference. Held against a chain whose behavior names nothing:

| shape, N=800 | `Elaborator.elaborate` |
|---|---|
| chain, the behavior names `v799` | 644,807 |
| chain, the behavior names no value | 641,603 |
| 800 values naming none of each other | 2,413 |

[...]

PR #526, opened 2026-08-09
An operator reads each operand once

### LINKED ISSUE #507: A chain of value declarations costs the compiler time that doubles with every link
A chain of value declarations costs the compiler time that doubles with every link.

## What happens

A module whose values are written one from the last takes time exponential in the length of the chain.

```
module chainN exposing (f)

let v0 = 1
let v1 = v0 + 1
...
let vN = vN-1 + 1

behavior f : (x: Int) -> Int
let f (x) = x + vN
```

| N | `compile` |
|---|---|
| 20 | 0.71 s |
| 22 | 1.17 s |
| 24 | 2.85 s |
| 26 | 9.70 s |
| 100 | did not finish; killed after 120 s |

About 1.85x per link. No example rows are involved — the file above has none, so the example evaluation budget (`souther.example.recursion.depth`, E1910) does not cover it.

## Where the time goes

`BinaryElaborator.elaborateBinary` reads each operand twice: once in `erroneousOperand`, to ask whether the operand's type is one the compiler could not work out, and again in the rule for the operator. An operand of an operand is read twice again, so an expression whose operators nest N deep costs 2^N. Counting the calls to `Elaborator.elaborate`:

| N | elaborate calls |
|---|---|
| 10 | 24,533 |
| 14 | 393,161 |
| 18 | 6,291,389 |
| 22 | 100,663,217 |

Exactly twice per link. Removing only the second reading takes N=22 to 623 calls and N=100 to 0.46 s.

A value is substituted where it is named (ADR-0072), which is what gives the chain its depth: `vN`'s body nests N operators deep. A chain that does not nest anything is not affected — `let vi = vi-1` compiles at N=26 in 0.44 s, and so does a chain through a helper call, whose argument becomes a binding rather than another level of tree.

## The nesting ceiling is not what sends an author here

This was first read as E2104 sending authors down a path the compiler then hangs on. It does not. E2104 bounds the parser's frame stack, not the depth of the tree the parser builds, and a left-associated operator chain does not grow that stack. `let f (x) = x + 1 + 1 + ...` written 30 times takes 96 s with no diagnostic and no names involved at all. The doubling is reachable by writing the expression literally, which is what E2104 was supposed to be refusing.

## Expected

Elaborating a binary expression must not read either operand more than once. Observed from outside: the cost of elaborating an already-expanded binary expression is proportional to the size of that expression, rather than exponential in its nesting depth.

The operand reading order is a rule rather than a side effect of reading twice. Left to right: an operand that says what is wrong with it is reported, and the operand beside it does not take that back. Today a right operand whose type could not be worked out covers a left operand's own diagnostic, because `erroneous` swallows it on the assumption that the second reading will raise it again.

## Not this issue

Two other things the investigation turned up, tracked separately because they have different causes and different answers:

- #523 — with the doubling gone, the same chain still costs O(N²), because substitution copies each body rather than sharing it. That is what the original "linear in the number of names" wording was asking for.
- #524 — a chain of 3000 ends in a `StackOverflowError` from the substitution walk, with no diagnostic. E2104 does not reach it: the depth is created after parsing.

Found in round-5 MCP dogfooding (bill-of-materials model).

kawasima:
Closes #507.

## What was wrong

`BinaryElaborator.elaborateBinary` read each operand twice: once in `erroneousOperand`, to ask whether the operand's type was one the compiler could not work out, and again in the rule for the operator. An operand of an operand was read twice again, so an expression whose operators nest N deep cost 2^N.

Counting the calls to `Elaborator.elaborate`:

| N | before | after |
|---|---|---|
| 10 | 24,533 | 167 |
| 14 | 393,161 | 287 |
| 18 | 6,291,389 | 439 |
| 22 | 100,663,217 | 623 |

Exactly twice per link. A value is substituted where it is named (ADR-0072), so a chain of values written one from the last nests that deep — which is the shape an author is told to write a deep expression in.

The nesting ceiling never covered this. E2104 bounds the parser's frame stack, not the depth of the tree it builds, and a left-associated operator chain does not grow that stack: `let f (x) = x + 1 + 1 + ...` written thirty times took 96 s with no names involved at all.

| | before | after |
|---|---|---|
| chain N=26 | 9.70 s | 0.45 s |
| chain N=100 | killed after 120 s | 0.46 s |
| 30 literal `+` | 96.3 s | 0.43 s |

## What changed

The operands are read once, and the Core each reading answers with is what the rule reads.

Reading twice also decided which operand was reported, and decided it wrongly: `erroneous` swallowed an operand's own `CompileException` on the assumption that the second reading would raise it again, but where the operand beside it had no type the definition was abandoned first and that diagnostic was never raised at all — `("x" ++ 1) + a.n` said nothing about `++`. Left to right is a rule now: an operand that says what is wrong with it is reported, and what stands beside it is not read where this one left nothing to check against.

`&&` and `||` are the operators that ask something of an operand on its own, so that is asked where the operand is read, before the one beside it is read. Everything else is about the pair and is answered once both have been read. That keeps `1 && ("x" ++ 1)` reporting that `1` is not a Bool, as it did before.

## The diagnostics that change

Two, both of them the rule above rather than a consequence of how many times an operand is read:

- `("x" ++ 1) + a.n`, where `a.n` has no type — the append is reported. Before, `erroneous` swallowed it and the definition was then abandoned for `a.n`, so nothing about `++` was said at all.
- `1 && a.n` — that `1` is not a Bool is reported. Before, the definition was abandoned for `a.n`. Giving up on a definition is for what follows from a name that denotes nothing, and an operand refused before that name is reached does not follow from it.

## Tests

`AnOperatorReadsEachOperandOnceTest` measures a chain of 100 named values and a 26-deep chain of each operator that nests into itself (`+`, `++`, `&&`, `==`), against a 10 s budget on a thread whose wait is bounded. Before this change all five failed — three at 27–32 s and two by not coming back within 60 s, 206.6 s for the class. After it, 0.229 s.

The reading is one place ahead of the switch, so every arm goes through it; the comparisons cannot nest into themselves and are covered by that rather than by a chain of their own.

`TheOperandThatFailedFirstIsTheOneReportedTest` fixes the order: the operand beside one that failed does not cover it, an operand `&&` refuses on its own comes before what stands beside it (whether that is an error of its own or an operand with no type), an operand with no type stops the one beside it from being read, and an operand with no type is still not reported as the wrong type for the operator.

`mvn clean install` is green (4,406 + 315 tests).

## Left out

Two things this investigation turned up, split out because they have different causes:

- #523 — with the doubling gone the same chain still costs O(N²), because substitution copies each body rather than sharing it.
- #524 — a chain of 3000 ends in a `StackOverflowError` from the substitution walk, with no diagnostic.

### DIFF
--- a/souther-compiler/src/main/java/souther/compiler/check/BinaryElaborator.java
+++ b/souther-compiler/src/main/java/souther/compiler/check/BinaryElaborator.java
@@ -19,35 +19,41 @@ public final class BinaryElaborator {
 
     private BinaryElaborator() {}
 
-    /** Whether either side of {@code bin} has a type the compiler could not work out. */
-    private static boolean erroneousOperand(Ast.Binary bin, Scope env,
-                                            CheckContext ctx) {
-        return erroneous(bin.left(), env, ctx) || erroneous(bin.right(), env, ctx);
-    }
-
-    private static boolean erroneous(Ast.Expr operand, Scope env, CheckContext ctx) {
-        try {
-            return Elaborator.elaborate(operand, env, ctx).type() instanceof Type.Erroneous;
-        } catch (CompileException _) {
-            return false;   // it has its own error; the operator's check will raise it as before
+    /**
+     * Reads one operand, and gives up on the definition where its type is one the compiler could not
+     * work out.
+     *
+     * <p>An operator wants a particular shape of type — Int or Decimal to add, two lists or two
+     * strings to append — and an operand the compiler could not work out has no shape. Absorbing is
+     * for a comparison, which can answer "no disagreement"; there is no answer to give here, so the
+     * definition is abandoned and the name that denoted nothing stands as what was wrong. Left alone,
+     * the operand's type reaches the message as `?`, which names nothing the author could go looking
+     * for.
+     *
+     * <p>What this answers with is what the operator's rule then reads. Asking the question of a
+     * reading of its own cost the expression twice its own subtree, and an operand of an operand
+     * twice again, so a chain of operators nested one inside the last cost two to the power of its
+     * length: twenty-six links took ten seconds and a hundred did not finish.
+     */
+    private static Core operand(Ast.Expr e, Ast.Binary bin, Scope env, CheckContext ctx) {
+        Core read = Elaborator.elaborate(e, env, ctx);
+        if (read.type() instanceof Type.Erroneous) {
+            throw new Unanswerable(bin.pos());
         }
+        return read;
     }
 
     static Core elaborateBinary(Ast.Binary bin, Scope env, CheckContext ctx) {
-        // An operator wants a particular shape of type — Int or Decimal to add, two lists or two
-        // strings to append — and an operand the compiler could not work out has no shape. Absorbing
-        // is for a comparison, which can answer "no disagreement"; there is no answer to give here, so
-        // the definition is abandoned and the name that denoted nothing stands as what was wrong.
-        // Left alone, the operand's type reaches the message as `?`, which names nothing the author
-        // could go looking for.
-        if (erroneousOperand(bin, env, ctx)) {
-            throw new Unanswerable(bin.pos());
-        }
+        // Left to right, and the first operand that failed is the one reported: an operand that says
+        // what is wrong with it has said it, and what stands beside it is not read at all where this
+        // one left nothing to check against.
+        Core left = operand(bin.left(), bin, env, ctx);
+        Core right = operand(bin.right(), bin, env, ctx);
         return switch (bin.op()) {
             case AND, OR -> {
-                Core left = Elaborator.requireTyped(bin.left(), Type.BOOL, env, ctx,
+                Elaborator.requireType(bin.left(), left.type(), Type.BOOL, ctx.symbols(),
                         "operand of logical operator");
-                Core right = Elaborator.requireTyped(bin.right(), Type.BOOL, env, ctx,
+                Elaborator.requireType(bin.right(), right.type(), Type.BOOL, ctx.symbols(),
                         "operand of logical operator");
                 yield new Core.Binary(bin.op(), left, right, Type.BOOL, bin.pos());
             }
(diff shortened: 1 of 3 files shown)


kawasima:
Both addressed in c8355f8b.

**P1 — the `&&`/`||` ordering.** You are right, and it contradicted the rule this branch states: `1` failed first and the append on the right was reported. Reproduced before changing anything — the keys came back as `[check.unknown.type.msg, check.concat.msg]`.

Taking the first of your two options rather than the deferred-rethrow. Reading both operands and then judging them is not what "the first operand that failed is the one reported" says; the operand is finished where it stands. `&&` and `||` are the operators that ask something of an operand on its own, so that is asked in `operand()`, where the operand is read. Everything else is about the pair and is answered once both have been read, which is where those arms already were.

The old order was itself the artifact: `erroneousOperand` probed both operands and swallowed the right one's exception, so the left check reached its mismatch first only because the probe had thrown that exception away.

So two diagnostics change, and both are the rule rather than a side effect of the reading count:

- `("x" ++ 1) + a.n` — the append is reported, where before nothing about `++` was said.
- `1 && a.n` — that `1` is not a Bool is reported, where before the definition was abandoned. Giving up on a definition is for what follows from a name that denotes nothing, and an operand refused before that name is reached does not follow from it.

Both are regression tests now (`anOperandTheOperatorRefusesOnItsOwnComesBeforeWhatStandsBesideIt`, `anOperandTheOperatorRefusesIsReportedThoughTheOneBesideItHasNoType`), and the PR body lists them under "The diagnostics that change".

**P2 — the worker.** Daemon now, and interrupted where the wait ran out. The comment says what actually bounds it: the compiler does not read the interruption, so it is the JVM not waiting for a daemon that does.

`mvn clean install` green (4,406 + 315). Timings unchanged: chain N=26 0.47 s, N=100 0.47 s, 30 literal `+` 0.43 s, a 26-deep `&&` chain 0.44 s.

PR #529, opened 2026-08-09
A comment is filed against a place of the canonical form

### LINKED ISSUE #506: The places a comment can be carried by are made during construction and not kept
The formatter builds the canonical form's places while it writes them, knows which source elements produced them, and keeps neither. Comment ownership then asks the source tree the same question and, wherever the two structures differ, gets a different answer.

`souther fmt` refuses a legal source today because of it — or would, if the ownership rules were extended to a definition's body without this. #502 could not fix

```souther
let f = (x) ->
    // about the body
    x
```

because the canonical body is the lambda's last expression child and the source's child is the lambda; the comment is filed against a construct the canonical form does not have there. That case is pinned as a limit rather than fixed, and this is the issue that lifts it.

The reading is in #444: `K` is not a tree but a skeleton with places and a correspondence, `TokenDoc` is already most of the skeleton, and the places and correspondence are what fall out of the construction unrecorded.

## What this is, and what it is not

**In scope.** The minimum of `K` that comment resolution needs, produced by the construction rather than reconstructed from the source: places that can be told apart, the relation from source elements to them, and a resolve point for comment carriers between construction and layout.

**Not in scope.** Completing `K`. The other thirteen rules keep answering the way they do now. In particular the deviation witness of #444 is not built here — what is built is the substrate that lets a witness name a place without inferring it from the source tree again.

### Places exist

- Two siblings of one `SyntaxKind` are different places. An `if` written `if flag then p else q` has three children that are all a `VAR_EXPR`, and its condition, its then branch and its else branch are three places.
- A place's parent and which of that parent's places it is can both be read. Whether that is a property of the place or a label on the edge from its parent is an implementation choice and not a requirement here.
- A place's identity is not another name for a source `SyntaxNode`'s. A construction that returns the source node as the place has not made one.

### The correspondence is a relation

- Nothing assumes it is one-to-one. A fixture holds the lambda lift: the canonical parameter-list place comes from the source `LAMBDA_EXPR` and the canonical body place from that lambda's last expression child, so one source element stands behind one place and its descendant behind another.
- The API does not offer a single-valued `origin()`. A place with one source element and a place with none are both ordinary.

### Comment resolution sits between construction and layout

- No carrier is settled during construction by looking a source node up — the hundred and seven `aboveOf` / `afterOf` / `endOf` / `withEndComments` calls do not decide placement where they stand.
- Carriers are resolved once the places and the correspondence are complete, and what goes to the layout is a document with the comments already in it.
- A fixture holds that a comment makes the group carrying it break, so that deferring the resolution past the layout fails rather than passing quietly.

### The known counterexample turns over

- `butADefinitionWrittenAsALambdaIsLeftAsItWas` stops stating a limit and starts requiring the body to own its comment. A nested lift (`let g = (x) -> (y) -> x`) is a row beside it.
- The assertion reaches the carrier and not only the rendered position. Two carriers can print the same and a test that reads the text alone lets a misattribution back in — which is what the comment count already cannot see.

## One fact, two readers

The criterion is not that two tables agree. It is that there is one table.

> A canonical fact about places and lines is represented once. Mutating that one representative fails at least one layout fixture and at least one comment-ownership fixture, and the two fixtures are independent of each other.

Written the other way — two representations kept honest by a mutation test — the defect #444 found would be closed off by verification while staying in the design, and the next reader added would need a third. What the mutation has to show is that the two readers took the same value, not that two values were kept level.

#500 is the worked example of getting this wrong: `linesOf` and the `HARD_GAP` / `SOFT_GAP` placements said the same thing in two places, nothing held them together, and a construct entered in one and not the other lost its comments silently.

## What stays as it is

A comment has two decisions on it and they do not merge (#444, rules 4 and 5):

```
source comment
    |  rule 4 — what it was written about
source construct
    |  the correspondence
canonical place
    |  rule 5 — which place can carry it
carrier
    |  rule 13, layout
written comment
```

Only the lower half moves. What a comment was written about is a fact about the source and is read from the token stream, as it is now; deriving that from the canonical structure would be the same mistake pointing the other way.

## Done when

```
before                                after

construction                          construction
  |- canonical document                 `- skeleton + places + correspondence
  `- comments: the source tree                    |
     searched again for a placement          resolve carriers
                                                  |
                                            resolved document
                                                  |
                                                layout
```

Refs #444. Lifts the limit stated in #502.

kawasima:
The formatter built the canonical form's places while it wrote them, knew which source elements produced them, and kept neither. Comment ownership asked the source tree the same question afterwards, and wherever the two structures differ it got a different answer.

The construction now records the places it writes and what the source had at each, and carriers are answered between the construction and the layout: the construction leaves a slot and says nothing about what goes in it, one pass fills every slot once every place exists, and what reaches the layout is a document with the comments already in it.

## Places exist

A place is told apart from every other by being it — an `if` written `if flag then p else q` has three children that are all a `VAR_EXPR` and three places — and says which of its parent's places it is. It is not another name for a source node: the parameter list of `let f (x) = x` is a place the source tree has no node at.

## The correspondence is a relation

A place stands for none, one or several source elements, and no API offers a single-valued `origin()`. The lambda supplies the parameter list place and its last expression child supplies the body place beside it, so a single-valued answer would have to pick one — and picking is what asking the source tree again already did.

## The counterexample turns over

```souther
let f = (x) ->
    // about the body
    x
```

The body owns the comment now. `butADefinitionWrittenAsALambdaIsLeftAsItWas` stated this as a limit; it requires the ownership instead, with a nested lift beside it, and reads the carrier rather than the rendered position — a comment carried by the declaration prints on the same line, and text alone cannot tell the two apart.

## Four answers did not survive being read back

Three were the classification reading the canonical form it had itself produced: a `with` clause is written on the input's line, so its place is that segment's and not the row's; a connector opens the line what follows is written on, so a comment above the connector is above all of that line and not above whatever it opens with; and a comment written inside a name is written after the whole name. The fourth was the walk that widens a carrier, which compared against the construct the comment is inside rather than against where the comment was written.

A comment held against a token also had no carrier at all in half the boundaries below. It was looked up in an index of places by offset, which only names and header lines are in; it reaches its place through the run the token is part of now, and where the canonical form writes no place for it, through the construct the token is inside.

## One table

The line table the old ownership walked is gone. `linesOf` and the placements said the same thing in two places, which is what #500 found. Nothing reaches it now, and the one representative left is a place's `Opening`: writing `Break.NEVER` where a member's line may break fails the layout fixtures and the ownership fixtures independently of each other.

## Checks

A probe comment written into every boundary of the two sweep fixtures, on a line of its own and at the end of a line, comes back once and formats to itself: 404 variants each way, none moved, refused, or rewritten. `mvn clean install` is green.

Closes #506. Lifts the limit stated in #502. Refs #444.

### DIFF
--- a/souther-compiler/src/test/java/souther/compiler/fmt/ADefinitionsBodyOwnsItsCommentTest.java
+++ b/souther-compiler/src/test/java/souther/compiler/fmt/ADefinitionsBodyOwnsItsCommentTest.java
@@ -1,8 +1,16 @@
 package souther.compiler.fmt;
 
 import org.junit.jupiter.api.Test;
+import souther.compiler.cst.CstParser;
+import souther.compiler.cst.SyntaxKind;
+import souther.compiler.cst.SyntaxNode;
+
+import java.util.ArrayList;
+import java.util.List;
 
 import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertNotSame;
+import static org.junit.jupiter.api.Assertions.assertTrue;
 
 /**
  * A comment written above what a definition or an {@code if} writes after its keyword stays there.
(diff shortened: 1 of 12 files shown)


kawasima:
All three hold. Addressed in 3d4dd71a.

## The correspondence was still being worked out from the source tree

Measured before changing anything: of 45,417 source nodes across the examples and the bench corpora, 3,678 had no place — 33 of 65 `PIPE_EXPR` and 54 of 779 `BINARY_EXPR` among them, so for a flattened run this was the ordinary path and not a fallback, as you said.

The construction records where it writes every construct now. One it writes a place for is written at that place; one it writes inside another construct's line — a type in a field's, a name in an import's — is written at that construct's place; and a flattened run opens at the head place and ends at the segment its own right operand is written at, recorded in `collectChain` and `collectPipe` where both ends are known. The count is 0 of 45,417, and `andEveryConstructOfTheSourceIsWrittenSomewhere` holds it over one of each form.

`entryPlaces` is one lookup with no walk, and `entryPlaces` on a construct nothing was recorded for throws rather than falling back. `endsTheLine` widens over the places — it asks a place what it was written from and how far that reaches, and moves to the outermost place ending where the line does. The only `parent()` left in rule 5 is `Place.parent()`. Rule 4 now hands over the construct a token is written in, since finding it is a question about the source and belongs on that side.

## ByOpener

Both cases where a place claimed a line while a `HARD_GAP` elsewhere made it — a union that moves down a line, and the first member of a product block — say what opens their line now. The one `ByOpener` left is the first line of a file, which nothing opens, so the variant is `FileBegins` and carries no opener.

Writing `Break.NEVER` where a product's first member opens its line now fails `ACanonicalFormCanRewriteCodeTokensTest` and `ACommentKeepsItsMemberHoweverItIsWrittenTest` — layout and ownership, independently, which is what the mutation showed for a member's break and did not show for this.

## ordinal

Read off the document now. `andItIsCountedInTheOrderTheDocumentWritesThem` holds the case you named: a declaration builds its `invariant` clauses before its body and writes the body first, so counted as they are made the two come back reversed.

## Unchanged

The sweep: 404 variants each way, none moved, unstable, lost, rewritten or refused. `mvn clean install` green. Formatting the examples and this repository's `.sou` files round-trips with nothing refused or unstable, and the only file whose output differs from the old formatter is a hand-written bench fixture that neither formatter leaves alone.

kawasima:
Both hold. Addressed in 9ac27d25 (the branch is rebased on the current `develop`).

## Unrecorded constructs

`IMPORT_ALIAS` throws exactly as you described. So does a comment written after a `private`:

```
module m

private // c
let f (x: Int) = x
```

Both are recorded now, at the line of the construct that holds them, and both sources come back with the comment kept.

The fixture was the real defect, and it found a third: asked of `WhatGoesBetweenTwoTokensOnALineTest.corpus()`, the invariant reported `[IMPORT_ALIAS, PARTIAL_MODIFIER, PRIVATE_MODIFIER, PRODUCT_BODY]` — the empty product body was the same hole with nothing yet reaching it. `andEveryConstructOfTheSourceIsWrittenSomewhere` runs over that corpus now.

One departure from what you asked: I did not add a node-kind coverage table here. `thereIsASourceForEveryKindOfNode` already holds that corpus against `SyntaxKind`'s node kinds, so a kind the grammar gains has to reach a fixture before either test passes, and a second copy of that list here would be the same list kept in two places. The javadoc names it instead. If you would rather this file assert it directly, say so and I will move it.

## Places with no marker

Measured over the corpora: 308 of 47,773 places had no entry in `order()`, all of them `BEHAVIOR_SIG` and `PIPE_BEHAVIOR` under `BEHAVIOR_DEF` — your two. I took the stronger option: both are written at their place, and `andEveryPlaceTheConstructionMakesIsWrittenInTheDocument` holds it over the corpus, so a structural place made without a marker fails rather than being numbered by a walk that invents one. `Correspondence.made()` is what lets the test enumerate them. Count is 0 of 47,773.

## placesOf

The javadoc promised the order the places were written and delivered the order the formatter made them — the same trace `ordinal` was taken off the construction to stop being. It promises no order now, and the two tests that read a single-element list assert the whole list rather than indexing into it.

## Unchanged

Sweep: 404 variants each way, none moved, unstable, lost, rewritten or refused. `mvn clean install` green on the rebased base. Formatting the examples and this repository's `.sou` files round-trips with nothing refused or unstable.

PR #532, opened 2026-08-09
What a value carried is written once, by the substitution nothing else is inside

### LINKED ISSUE #523: A chain of value declarations costs the square of its length

### LINKED ISSUE #531: A value's expansion is marked once per level of substitution
A value's expansion is marked once per level of substitution, and the mark is idempotent.

## What happens

`HelperInliner.substituted` expands a value's body and walks the result with
`HelperNames.carriedByValue`, which sets `ConstructionOrigin.viaValueReference` on every
construction under it. A value that names another value substitutes inside that expansion, so the
walk runs again at every level: a chain of K values is walked at sizes 1, 2, ..., K, which is K²
steps for one value and N³ for a module of N.

The walk is idempotent — the mark is a boolean, and the outermost walk covers every subtree an
inner one already covered. `Ast.atSlots` returns the same object where nothing changed, so the
repeated walks allocate nothing. What they cost is the traversal.

Measured on the chain of #523, counting calls:

| N | `compile` | `Elaborator.elaborate` | `carriedByValue` node visits |
|---|---|---|---|
| 100 | 0.03 s | 10,607 | 358,752 |
| 200 | 0.09 s | 41,207 | 2,767,502 |
| 400 | 0.27 s | 162,407 | 21,735,002 |
| 800 | 1.80 s | 644,807 | 172,270,002 |

The elaborations quadruple per doubling and the visits octuple: N² against N³ (the visits are
N³/3). At N=800 there are 267 visits for every elaboration, and a sampling profile of that compile
is `Ast.atSlots`, `HelperNames.carriedByValue` and `Ast.mapChildren` before anything else.

## Why it matters

This is not #523. #523 is that a module of N values holds O(N²) expression nodes, because the
source has one `v99` and elaboration has a hundred copies of it. This sits on top of that and is
larger, so it is the term that answers for the time: marking at the outermost substitution alone
takes N=800 from 1.80 s to 0.18 s, and what is left is the O(N²) #523 asks about — 644,807
elaborations, unchanged.

Split out of #523 so that the two are measured apart. Fixing this does not close #523.

## Expected

A value's expansion is marked once. The outermost substitution is the one that has the whole tree,
and it reaches every slot an inner one would have: `Ast.Expansion.given` is not a slot of
`Ast.atSlots`, and `HelperNames.alsoInGiven` is what puts it back.

## Where

`HelperInliner.substituted`.

kawasima:
A value's expansion is walked to mark what it built as carried, and the walk ran at the end of
every substitution. A value that names another substitutes inside that expansion, so a subtree the
inner substitution produced was walked again by the one above it, and again by the one above that.
A chain of K values was walked at sizes 1, 2, ..., K, which is N cubed over a module of N.

The mark is a flag set from what a node denotes, so the walk says the same thing however many
times it runs, and the outermost substitution holds every subtree the ones under it produced.
Writing the mark there alone says of each node what writing it at every level said. Which
substitution that is, the path this pass already keeps answers: it is the one the path holds
nothing else beside.

## What it moves

`values` joins the benchmarks: a chain of values against as many values naming none of them,
reported per value so a term above linear reads as a rising figure rather than a bigger total.

Before:

```
VALUES n=100   chain    19.6 ms (0.196 ms/value)   flat     5.6 ms (0.056 ms/value)
VALUES n=200   chain    62.3 ms (0.312 ms/value)   flat     4.8 ms (0.024 ms/value)
VALUES n=400   chain   439.3 ms (1.098 ms/value)   flat     5.8 ms (0.015 ms/value)
VALUES n=800   chain  3471.9 ms (4.340 ms/value)   flat     8.6 ms (0.011 ms/value)
```

After:

```
VALUES n=100   chain     8.5 ms (0.085 ms/value)   flat     6.5 ms (0.065 ms/value)
VALUES n=200   chain    16.7 ms (0.084 ms/value)   flat     4.6 ms (0.023 ms/value)
VALUES n=400   chain    49.1 ms (0.123 ms/value)   flat     5.9 ms (0.015 ms/value)
VALUES n=800   chain   165.7 ms (0.207 ms/value)   flat     8.9 ms (0.011 ms/value)
```

The per-value figure quadrupled per doubling before and now under doubles: cubic against quadratic.

## What it does not move

This does not close #523, and #523 is left open. A module of N values still holds N squared
expression nodes — the source has one `v99` and elaboration has a hundred copies of it — and the
count of elaborations is unchanged at 644,807 for eight hundred values. What this removes is the
cubic term that sat on top of the quadratic one, which had to go before the quadratic one could be
measured at all.

## Tests

Two for the shape the change turns on: a value reached through values of the same module, and one
reached through a value that expands a helper — a helper call becomes an expansion, whose `given`
is not a slot of the walk that rebuilds an expression and is put back by the pass that marks. Both
fail if the mark is dropped rather than moved, along with the eight that already covered it.

Closes #531.

### DIFF
--- a/souther-bench/src/main/java/souther/bench/Bench.java
+++ b/souther-bench/src/main/java/souther/bench/Bench.java
@@ -6,9 +6,9 @@ import java.util.List;
 /**
  * The benchmarks, run from the command line: {@code java -jar souther-bench.jar [measurement...]}.
  *
- * <p>With no argument it runs all four. Naming one runs only that, which is how a change is checked
- * against the thing it was meant to move without waiting for the rest — {@code scale} in particular
- * compiles two hundred modules five times over.
+ * <p>With no argument it runs all of them. Naming one runs only that, which is how a change is
+ * checked against the thing it was meant to move without waiting for the rest — {@code scale} in
+ * particular compiles two hundred modules five times over.
  *
  * <pre>
  *   cold    one compile in a JVM that has not compiled before
@@ -17,6 +17,7 @@ import java.util.List;
  *   edit    what an edit costs a store that already holds the answers
  *   run     what the generated code costs to run, per element
  *   scale   how a whole-workspace compile grows with the number of modules
+ *   values  how one module's compile grows with the number of values it declares
  * </pre>
  *
  * <p>Every corpus is compiled once and checked before anything is timed. A language change that
@@ -30,7 +31,7 @@ public final class Bench {
     public static void main(String[] args) {
         Report report = new Report(System.out);
         List<String> wanted = args.length == 0
-                ? List.of("cold", "warm", "phase", "edit", "run", "scale")
+                ? List.of("cold", "warm", "phase", "edit", "run", "scale", "values")
                 : new ArrayList<>(List.of(args));
 
         List<Corpus> corpora = Corpus.all();
@@ -73,6 +74,10 @@ public final class Bench {
         }
         if (wanted.contains("scale")) {
             Scale.measure(report);
+            report.blank();
+        }
+        if (wanted.contains("values")) {
+            Values.measure(report);
         }
     }
 }
--- a/souther-bench/src/main/java/souther/bench/Values.java
+++ b/souther-bench/src/main/java/souther/bench/Values.java
@@ -1,0 +1,73 @@
+package souther.bench;
+
+import souther.compiler.meta.ModulePath;
+import souther.compiler.query.Compilation;
+
+import java.util.List;
+
+/**
+ * How a compile grows with the number of values one module declares.
+ *
+ * <p>{@link Scale} asks the same question of a workspace and this asks it of a module, because the
+ * two have different answers: a value is substituted where it is named, so what a module declares
+ * can cost more than the sum of its declarations while what a workspace holds does not.
+ *
+ * <p>Two shapes, held the same apart from what a value names. Each value of the chain names the one
+ * before it, so the last of them stands for the whole chain and the module's source shares what its
+ * elaboration copies. The flat shape declares as many values naming none of them, which is the same
+ * number of declarations with nothing to share — so the distance between the two lines is what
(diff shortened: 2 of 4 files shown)


kawasima:
Folded into #536, which carries this commit unchanged and the rest of #523 with it. Closing so there is one branch to review and one to merge.

PR #536, opened 2026-08-09
A module of values costs what it declares

### LINKED ISSUE #523: A chain of value declarations costs the square of its length

### LINKED ISSUE #531: A value's expansion is marked once per level of substitution

kawasima:
A module of N values cost more than N, in three separate ways. All three are here, and #531 is
folded in rather than left to merge first — PR #532 is closed in favour of this one.

## What was costing what

Measured on the chain of the issue, one module, a whole compile:

| N | before | after |
|---|---|---|
| 200 | 23.8 ms | 11.7 ms |
| 400 | 52.2 ms | 17.0 ms |
| 800 | 130.2 ms | 28.8 ms |
| 1600 | 449.6 ms | 50.0 ms |

The cost per value now falls with the module (0.058 → 0.031 ms) where it rose with it
(0.119 → 0.281 ms).

**The mark on what a value carried was written once per level of substitution** (#531). The walk is
idempotent, so a chain of K values was walked at sizes 1, 2, ..., K — N cubed, and the term that
answered for the time. At eight hundred values, 172,270,002 node visits against 644,807
elaborations.

**Two checks searched the whole graph once per declaration**, and both were at their most expensive
when they found nothing. `ValueCycles` asked each value whether it reaches itself; which names lie
on a cycle is now worked out once, by Tarjan's. `PartialReachability` searched forward from every
declaration for a `partial` one; which names can reach one is now worked out once, backwards from
them. Both keep the search that a report's path comes from, and run it only where the answer is
yes, so nothing reported changes.

**A value's own check read a copy of every value it names.** That is the issue's own term: the
source has one `v99` and the check had a hundred copies of it, so N answers cost N squared nodes.

## What a value's check now reads

What it needs of a value it names is what that value's own check settled, and there are two such
things.

The type it stands under. A reference is kept standing and read as that, which is what a
representation keeping a name standing already means, so it is said through `Preserved` — the
capability belongs to that representation and not to elaboration at large.

The constant it is one of, where a fold reaches one. That is written out at the reference as the
literal it folded to, because everything that asks whether an expression is known at compile time
folds a tree rather than resolves a name (`ConstEval`, and it says so). A `String.matches` pattern
composed out of values goes on reading a literal.

Values are therefore checked each after the ones it names, which is not the order the module wrote
them. The order decides what is worked out twice and nothing else: a value not yet settled is
copied exactly as it was before.

Only while checking a value. A helper is checked as though its body had been written inline, and
what it names is expanded into it as before; so is every tree the backend emits from, because a
value is substituted at each of its references and that is what runs.

## What does not change

No program that compiled stops compiling and none that was refused is accepted. A value whose body
is an empty collection still takes its element type from each place that names it — the behavior
that names it substitutes, as it always did — so this settles nothing about that and leaves it
exactly where it was.

Diagnostics are the one thing that can differ in count: checking `v100` used to re-report what was
already reported at `v99`, and not doing that twice is what the change is. Nothing changes in what
is reported first, or in how anything is worded.

## Tests

Nine, over the two changes that could have altered an answer.

For the mark: a value reached through values of the same module, and one reached through a value
that expands a helper — a helper call becomes an expansion, whose `given` is not a slot of the walk
that rebuilds an expression. Both fail if the mark is dropped rather than moved.

For the settled answer: a chain still computing what it says once emitted, a value written above
the one it names, a value the fold does not reach standing under its type, a mis-typed value
reported at itself, and a compile-time pattern composed through a chain — inside a value's own
body, which is the check that reads a settled answer, and refused as a regex rather than as an
expression nothing could evaluate. The last two fail if the constant is not written out.

The benchmark grows a `values` measurement: a chain of values, the same chain written bottom to
top, and as many values naming none of them. Per value, so a term above linear reads as a rising
figure rather than a bigger total, and an order lost shows up in the reversed column alone.

Closes #523. Closes #531.

### DIFF
--- a/souther-bench/src/main/java/souther/bench/Bench.java
+++ b/souther-bench/src/main/java/souther/bench/Bench.java
@@ -6,9 +6,9 @@ import java.util.List;
 /**
  * The benchmarks, run from the command line: {@code java -jar souther-bench.jar [measurement...]}.
  *
- * <p>With no argument it runs all four. Naming one runs only that, which is how a change is checked
- * against the thing it was meant to move without waiting for the rest — {@code scale} in particular
- * compiles two hundred modules five times over.
+ * <p>With no argument it runs all of them. Naming one runs only that, which is how a change is
+ * checked against the thing it was meant to move without waiting for the rest — {@code scale} in
+ * particular compiles two hundred modules five times over.
  *
  * <pre>
  *   cold    one compile in a JVM that has not compiled before
@@ -17,6 +17,7 @@ import java.util.List;
  *   edit    what an edit costs a store that already holds the answers
  *   run     what the generated code costs to run, per element
  *   scale   how a whole-workspace compile grows with the number of modules
+ *   values  how one module's compile grows with the number of values it declares
  * </pre>
  *
  * <p>Every corpus is compiled once and checked before anything is timed. A language change that
@@ -30,7 +31,7 @@ public final class Bench {
     public static void main(String[] args) {
         Report report = new Report(System.out);
         List<String> wanted = args.length == 0
-                ? List.of("cold", "warm", "phase", "edit", "run", "scale")
+                ? List.of("cold", "warm", "phase", "edit", "run", "scale", "values")
                 : new ArrayList<>(List.of(args));
 
         List<Corpus> corpora = Corpus.all();
@@ -73,6 +74,10 @@ public final class Bench {
         }
         if (wanted.contains("scale")) {
             Scale.measure(report);
+            report.blank();
+        }
+        if (wanted.contains("values")) {
+            Values.measure(report);
         }
     }
 }
--- a/souther-bench/src/main/java/souther/bench/Values.java
+++ b/souther-bench/src/main/java/souther/bench/Values.java
@@ -14,14 +14,14 @@ import java.util.List;
  * two have different answers: a value is substituted where it is named, so what a module declares
  * can cost more than the sum of its declarations while what a workspace holds does not.
  *
- * <p>Three shapes, held the same apart from what a value names and where it is written. Each value
- * of the chain names the one before it, so the last of them stands for the whole chain and the
- * module's source shares what its elaboration copies. The reversed chain is that module written
- * bottom to top, where no value is settled by the time the one naming it is read unless the check
- * puts them in an order of its own. The flat shape declares as many values naming none of them,
- * which is the same number of declarations with nothing to share — so the distance from it is what
- * substitution costs, and a chain's own per-value figure is what says whether that cost is
- * proportional to what the module declares.
+ * <p>The shapes differ in what a value reaches and in where the thing it reaches is written, which
+ * is what the answer turns on. A chain has each value name the one before it. A chain through
+ * helpers has each value call a helper that names the one before it, so the same reaching is there
+ * and no value writes another's name. Either written bottom to top is the module where nothing a
+ * value reaches has been read yet by the time the value is. The fan-out has one value reach every
+ * other, which is the shape a walk that re-reads a name's edges once per edge is quadratic in and a
+ * chain says nothing about. The flat shape declares as many values reaching none of them: the same
+ * number of declarations with nothing to share, and so the floor the rest are read against.
  *
  * <p>The per-value figure is the one to read. A total that grows is only the module growing; a
  * per-value figure that grows with the module is a term above linear, and it is reported at sizes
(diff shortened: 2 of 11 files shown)


kawasima:
Both blocking findings are fixed in e7d1053d. The third is left alone, on your reading of it.

## 1 — the Tarjan frame reads its edges once

Right, and the chain hid it exactly as you say. The frame holds the iterator now. The benchmark
grows a fan-out where one value reaches every other, which is the shape that shows it: at eight
hundred values, 8.3 ms and 0.010 ms/value, falling.

## 2 — reaching through a helper

Right, and worse than the fallback you describe. The order was the smaller half of it: a helper is
not settled either, so `step_i` copied the whole chain below it whether or not the values were in a
good order. Both are fixed — the order follows the call edge as well as the read edge, over every
declaration rather than over the values alone, and a helper reads a settled value as a value does.
What settles one stays narrower: a value of this module. Neither a tree the backend emits nor a
recursive helper's own body reads any of it.

At eight hundred values, a chain written through helpers:

| | develop | first two commits | now |
|---|---|---|---|
| top to bottom | 26.7 s | 3.4 s | 0.11 s |
| bottom to top | 25.8 s | 3.4 s | 0.11 s |

## Your acceptance test does not pass yet, and here is what is left

You asked for the reversed helper-chain's ms/value to be flat. It is not:

| n | direct chain | via helpers |
|---|---|---|
| 200 | 0.056 | 0.103 |
| 400 | 0.031 | 0.101 |
| 800 | 0.037 | 0.128 |
| 1600 | 0.031 | 0.213 |

The direct chain is flat and the helper one rises from 1600. It is not the re-derivation this PR is
about, and it is not new — it is on develop, under terms large enough to hide it.

A profile of the helper chain at sixteen hundred is `Scope.with`, `Denotations.with`,
`BindingId.hashCode` and `InvariantChecker.conditionalIn`. `Scope.with` and `Denotations.with` copy
the whole binding map per insert, so a body holding N bindings costs N squared to build a scope
for. The direct chain has no bindings at all — `v = v' + 1` expands to arithmetic — and the helper
chain has one per link, which is why only one of them shows it. The body in question is the
behavior's, where the chain is substituted because it is what runs, so the N bindings are real and
what is quadratic is only the inserting.

That is a different mechanism in a different place: the checker's own scope structure, not a check
re-deriving an answer. Changing it means making `Scope` and `Denotations` persistent, which is the
core of the checker and not something to fold into this PR at review. Say the word if you want it
here anyway and I will do it; otherwise it wants its own change, and I have not opened an issue for
it.

## 3 — `PreservedCall`

Taking your call: left as it is. The javadoc says "a name" and the record still says "call", which
is a real mismatch, and the rename is eleven files of churn against this diff.

Full build green (souther-compiler 4452 and the rest), rebased on develop.

Issue #510, opened 2026-08-09
The boundary measure does not read an invariant written with `String.length`


kawasima:
The boundary measure does not read an invariant written with `String.length`.

## What happens

```
module nd exposing (Name, Size, T, look)

data Name = String
    invariant nonEmpty = String.length(value) > 0

data Size = Int
    invariant atLeastOne = value >= 1

data T = { name: Name , size: Size }

behavior look : (t: T) -> Int
let look (t) = t.size.value
```

```
  look                     implemented   rows 1    pending 0
    boundary    1/1
      · not derivable: t.name
```

`Size`'s numeric bound draws an edge. `Name`'s length bound does not, and the report calls it *not derivable* — a word `example-report-vocabulary` defines as "a fact about the model and not a gap in the rows".

## Why it is wrong

It is not a fact about the model. The discharge procedure names exactly this term:

> A term it compares is the numeric content of a location — a parameter, a field chain, a newtype's wrapped value — or the size of a container or a string: `List.length`, `String.length`, `Set.size`, `Map.size` applied to such a location.
> — `specification.adoc:1137`

So the boundary derivation is weaker than the documented fragment, and where it falls short it states something false rather than saying it could not measure.

The consequence is the wrong one for the author: a length bound is the most common invariant in the specification's own examples, and the shortest value the bound admits is exactly the row a reader wants written. The report steers them away from writing it.

## The same happens to a guard, which says what the defect actually is

A body that compares a length draws no line either, and the position is still called not derivable:

```
behavior guarded : (t: T) -> Int
let guarded (t) = {
    guard String.length(t.name.value) > 3 else 0
    t.size.value
}
```

```
  guarded                  implemented   rows 0    pending 0
    partition   axes 0   single-axis 0/0
    boundary    0/0   (1 not measured: no row names this behavior)
      · not derivable: t.name
```

The numeric guard beside it gives `axes 1` and three obligations. So this is not an invariant reader that is weaker than the discharge procedure's. Every part of the partition machinery identifies *the numeric term a boundary is drawn on* with *the location it is drawn at*, and a length is a number taken of a location rather than the number in it. `String.length` is not a missing case; it is the first term that shows the identification was wrong.

Saying "the model draws no line through this position" about a position the body compares in the line above is a false sentence, not a coverage gap.

## Where it stops

[...]

PR #544, opened 2026-08-09
A candidate is proposed from the rule and not the carrier alone

### LINKED ISSUE #528: A newtype whose invariant refuses the default fixture generates no rows at all
A newtype whose invariant refuses the default fixture leaves the generator with nothing to offer, and every row of the behavior is lost with it.

## What happens

```
module nd.gen

data Domestic
data Overseas
data Kind = Domestic | Overseas

data Tag = List<String>
    invariant someTags = List.length(value) >= 1

data T = { kind: Kind, tags: Tag }

behavior look : (t: T) -> Int
let look (t) = 1

example look
    | (T { kind = Domestic, tags = Tag(["t"]) }) -> 1
```

`t.kind` is divided into two classes and one of them has a row, so the generator is asked for the other. It comes back with nothing:

```
unresolved: UnresolvedCombination[classes=[t.kind=Overseas], reason=ALL_CANDIDATES_REJECTED, detail=null]
```

The same with a string:

```
data Name = String
    invariant longEnough = String.length(value) >= 5
```

## Why

What a position offers for a collection is the empty one, and only that (`Partitions.java:633`). It is the value that always builds — except against an invariant that says the collection is not empty, which is the most common rule anyone writes about one, and then it is the one value that never builds.

A string newtype is offered the bound's own value where the bound is numeric, a shortest string accepted by a `Pattern` where there is one, and otherwise the fixed `"x"` (`Partitions.java:744-774`). A length bound is none of those, so `"x"` is all there is, and a minimum of five refuses it.

Nothing else is tried. `inReserve` holds back the far edge of a numeric range and has nothing for a length (`Partitions.java:799-823`), and `displacedRepresentativesOf` steps a number (`Partitions.java:666-696`). So a rule about a size or a length leaves the position with candidates chosen as though the rule were not there.

The loss is not confined to the position. A row is the product of what its positions offer, so one field of such a type takes every row of the behavior with it — including rows for classes at other positions that have nothing to do with the rule.

## What it is not

It is not #510. That one is that the boundary a length bound draws is never derived, so nothing is asked for at the edge. This is that a value satisfying the bound is never constructed, which is already true of ordinary partition rows that ask for no edge at all. Neither causes the other and the two can be fixed in either order. They meet only afterwards, where a derived length boundary needs a witness at it.

## Expected

A position whose type is inhabited should offer at least one candidate its own rules admit. The two above are the smallest cases:

```
List.length(value) >= 1     ->  a list of one element
String.length(value) >= 5   ->  a string of five characters
```

`Set.size` and `Map.size` have the same shape.

The rules are already read for the decoder — `MinLength`, `MinSize`, `NonEmpty` and the rest are derived in `InvariantConstraints` (`InvariantConstraints.java:157-244`) and handed to Raoh — so what a candidate has to satisfy is known at the point the candidate is chosen. It is the choosing that does not consult it.

A candidate remains a proposal the decoder answers. Offering a list of one element is not a claim that the position admits it; it is having something to be refused at other than the one value the rule names as inadmissible.

kawasima:
Closes #528.

## What was wrong

`Partitions.representativesOf` chose a value from the carrier type and only the carrier — the empty one for every collection, the fixed `"x"` for a string — and the rules written on the newtype were consulted afterwards, by the decoder that refuses what was built. A rule saying a collection is not empty is the most common thing anyone writes about one, and it is then the one rule that refuses the one value on offer.

A row is the product of its positions, so the loss was not confined to the position. With `look : (t: T, urgent: Bool)` and a single `List.length(value) >= 1` under `t`, all eight combinations came back `ALL_CANDIDATES_REJECTED` — including the two the second parameter is about on its own.

The string case is narrower than the issue states: `"x"` is a string of one, so `String.length(value) >= 1` was met by accident and only `>= 2` and up were lost. The collection cases were lost from `>= 1`.

## The reading already existed

The concern with fixing this in `Partitions` was a second semantic reading of an invariant beside the codec's. There is none: `check/NumericMeasures` holds the one list of the operations that count a value, and `check/InvariantBound.ofSize` reads a bound on any of them. That is how a boundary on `List.length(t.tags)` is drawn, and `Partitions.measure` already asks it. The chooser did not.

Asking it there answers a set by the same call as a list. Reading the decoder's constraints instead would not have: `InvariantConstraints` has no entry for `Set.size` — a set crosses the boundary as a list, so a size chained after the mapping that drops duplicates would count the wrong things — and that absence is a fact about the decoder rather than about what the rule says. `NumericMeasures` says so in its own comment.

New code is `Partitions.leastHeld`, `partition/Witnesses`, and the two template forms in the commit before this one.

## What is held to

A minimum this recognises is a minimum the candidate is built at — not a witness for whatever anyone writes. Nothing here proves an invariant; a candidate remains a proposal the decoder answers, which is what lets one be built from the part of a rule this can read while the rest of the rule refuses it.

- `String.length(value) >= 5` gives a string of five, `List.length(value) >= 3` a list of three.
- A set gets elements no two of which are equal, a map keys no two of which are equal.
- An element's own rules are read where the element is built: `List<Word>` under `Word: String.length(value) >= 4` gives `[Word("xxxx")]`. The descent is bounded, so a type written in terms of itself gives up rather than recursing forever.
- A format and a minimum are two proposals rather than a precedence.
- A rule of a shape nothing reads leaves the position with what it had, and the combination is still reported as refused rather than as one nobody can write a row for.

Not covered: `List.length(value) == 3`. `InvariantBound` does not read an equality — a range has nowhere to put one — and changing that would move where boundaries are drawn.

Independent of #510, which is that the boundary a length bound draws is never derived. Nothing here touches boundary derivation.

## Evidence

Twelve tests in `ACandidateIsProposedFromTheRuleAndNotTheCarrierAloneTest`. With the witness disabled, ten of them fail; the two that pass either way are the ones that must not change — that `String.length(value) >= 1` still gives `"x"`, and that an unreadable rule is still reported as refused.

One test takes the generated row's text, puts it back into the model as an `example` and compiles it, so the text is held against the reader rather than against a spelling. Spelling a map entry wrongly on purpose makes it fail, so it is not a vacuous check.

On the crm corpus the lines saying a boundary is not known to be writable fall from 68 to 57. All eleven recovered are numeric edges of an `Amount`, at positions the collection rules have nothing to do with: the record could not be built at all, so the boundary row for a field beside them went with it.

`mvn clean test` is green across all modules.

### DIFF
--- a/souther-compiler/src/main/java/souther/compiler/partition/FixtureTemplate.java
+++ b/souther-compiler/src/main/java/souther/compiler/partition/FixtureTemplate.java
@@ -113,7 +113,31 @@ public record FixtureTemplate(String text, Ast.Expr value) {
     /** No elements. A list, a set and a map are all written this way in a fixture: what the position
      * is decides what the empty brackets become. */
     public static FixtureTemplate emptyCollection() {
-        return new FixtureTemplate("[]", new Ast.ListLit(List.of(), NOWHERE));
+        return collection(List.of());
+    }
+
+    /**
+     * A collection of what it holds, in the order the elements were chosen.
+     *
+     * <p>The same brackets for all three, for the same reason the empty one has them: a fixture writes
+     * a set as its elements and a map as its entry pairs, and which of them the brackets are is what
+     * the position declares rather than anything spelled here.
+     */
+    public static FixtureTemplate collection(List<FixtureTemplate> elements) {
+        List<Ast.Expr> values = new ArrayList<>();
+        List<String> written = new ArrayList<>();
+        for (FixtureTemplate each : elements) {
+            values.add(each.value());
+            written.add(each.text());
+        }
+        return new FixtureTemplate("[" + String.join(", ", written) + "]",
+                new Ast.ListLit(List.copyOf(values), NOWHERE));
+    }
+
+    /** One entry of a map: the pair a fixture writes a key and its value as. */
+    public static FixtureTemplate entry(FixtureTemplate key, FixtureTemplate value) {
+        return new FixtureTemplate("(" + key.text() + ", " + value.text() + ")",
+                new Ast.Tuple(List.of(key.value(), value.value()), NOWHERE));
     }
 
     /** A record, field by field, in the order the fields were declared. */
--- a/souther-compiler/src/main/java/souther/compiler/partition/Generator.java
+++ b/souther-compiler/src/main/java/souther/compiler/partition/Generator.java
@@ -652,7 +652,33 @@ public final class Generator {
                 || conditioned.reason() == UnresolvedCombination.Reason.SEARCH_LIMIT) {
             return new Outcome(null, UnresolvedCombination.Reason.SEARCH_LIMIT, null);
         }
-        return product;
+        // Every value that was offered was refused, which is only the whole story where every value
+        // the rules allow was offered. A position that read a count past what a row is built to carry,
+        // or that has more pairings than are built at once, held something back, and saying so is the
+        // difference between a fact about the model and a fact about this.
+        UnresolvedCombination.Reason held = heldBack(subject, p, decided);
+        return held == null ? product : new Outcome(null, held, null);
+    }
+
+    /** Why a position of this parameter offered less than its rules allow, or null where none did. */
+    private static UnresolvedCombination.Reason heldBack(Subject subject, int p,
+                                                         Map<String, List<FixtureTemplate>> decided) {
+        List<Position> found = new ArrayList<>();
+        positionsUnder(subject.types().get(p), TermPath.of(subject.parameters().get(p)),
+                subject.symbols(), 0, found, decided.keySet());
+        UnresolvedCombination.Reason held = null;
+        for (Position each : found) {
+            UnresolvedCombination.Reason here = Partitions.notBuilt(each.type(), subject.symbols());
+            // Nothing of the shape having been built outranks some of it having been: the first says
+            // the search never had what the rule asks for, and a reader owed one sentence is owed that.
+            if (here == UnresolvedCombination.Reason.NOTHING_COMPOSES_ONE) {
+                return here;
+            }
+            if (here != null) {
+                held = here;
+            }
+        }
+        return held;
     }
 
     /**
--- a/souther-compiler/src/main/java/souther/compiler/partition/Partitions.java
+++ b/souther-compiler/src/main/java/souther/compiler/partition/Partitions.java
@@ -820,12 +820,7 @@ public final class Partitions {
      * row that was written has no reason to pay for that.
      */
     static Generator.UnresolvedCombination.Reason notBuilt(Type type, Symbols symbols) {
-        Type carrier = TypeOps.base(type, symbols);
-        if (Witnesses.pastWhatIsBuilt(carrier, leastHeld(type, symbols))) {
-            return Generator.UnresolvedCombination.Reason.NOTHING_COMPOSES_ONE;
-        }
-        return Witnesses.moreThanIsPaired(carrier, symbols)
-                ? Generator.UnresolvedCombination.Reason.SEARCH_LIMIT : null;
+        return Witnesses.heldBackFor(TypeOps.base(type, symbols), leastHeld(type, symbols), symbols);
     }
 
     /**
--- a/souther-compiler/src/main/java/souther/compiler/partition/Partitions.java
+++ b/souther-compiler/src/main/java/souther/compiler/partition/Partitions.java
@@ -720,19 +720,21 @@ public final class Partitions {
      */
     static List<FixtureTemplate> representativesOf(Type type, Symbols symbols,
                                                    NumericDomain.Bounds within) {
-        return representativesOf(type, symbols, within, 0);
+        return representativesOf(type, symbols, within, java.util.Set.of());
     }
 
     /**
-     * The same, {@code depth} values inside the one a position was asked for.
+     * The same, with the newtypes this is already inside the value of.
      *
-     * <p>Counted because what stands for a collection is built from what stands for its element, which
-     * is this question again. A type written in terms of itself would ask it forever, and a value below
-     * the bound is given up on rather than invented.
+     * <p>Carried because what stands for a collection is built from what stands for its element, which
+     * is this question again. A name met while its own value is being built is a type written in terms
+     * of itself and is given up on — the names and not a count of them, since how many names a value
+     * wears on the way down is not what has to be stopped.
      */
     static List<FixtureTemplate> representativesOf(Type type, Symbols symbols,
-                                                   NumericDomain.Bounds within, int depth) {
-        if (type == null || depth > Witnesses.MAX_DEPTH) {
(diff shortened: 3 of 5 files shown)


kawasima:
Both P1s reproduced and are fixed in 1ac5e1d5. P2 reproduced as well and is fixed the way the review suggested.

**The element's proposals are kept.** `holding` returns one proposal per value the element offers, bounded at three, instead of building from `at(element, 0)`. `List<Word>` under a `Word` carrying both a format and a minimum now offers a list per candidate and the decoder settles which — the same thing the position itself is owed, one level down. A map pairs its key and value proposals off by index rather than crossing them: the search around a position is already a product, and taking another one inside a position would spend it on assignments the position is not what they are about.

**Distinctness comes from inside the element's rules.** `Set<Positive>` under `Positive: value >= 10` was offered `Positive(0)` and refused for its elements while saying nothing about its size. The second value is now `Partitions.numberInside`, which steps through the range the invariant leaves, or a string one character past the length its minimum asks for.

**The descent is bounded by the names being expanded, not by a count.** The review is right that a count cannot tell a value inside itself from a value wearing several names on the way down — five newtypes over a `String` inside a list had no witness. `representativesOf` carries the set of newtypes whose value is being built and gives up on a name met twice, which is the answer there rather than a limit: no value of such a type exists. The depth guard the previous commit put on ordinary representative generation is gone with it.

One case from the review is not expressible: `Map<Positive, String>` is refused by E1314, since a map that crosses the boundary is a JSON object and its keys are strings. The key case uses a `Code` bounded by its length instead, which asks the same question.

Six tests added, five for the defects above. The crm corpus is unchanged at 57 boundaries not known to be writable — the earlier 68 to 57 stands.

Test time, compiler module, best of three runs: 1:06 with the witness disabled, 1:00 with it. The change does not cost test time.

kawasima:
Both reproduced and are fixed in 899e9c52. The framing in the review is the one I should have used — the budget was sitting on what was read rather than on what is invented.

**A map's key and value meet each other.** The justification for taking them in step was wrong about which product it was appealing to: the search outside is over the behavior's positions, and a key beside a value is inside one position, so nothing out there can arrive at `Key[1] + Val[0]`. Every pair now, enumerated nearest first (by `i + j`) so that a search which stops has walked the low proposals of both rather than every one of the key's against the value's first.

**`MAX_PROPOSALS` is gone.** The reviewer's point is the deciding one: those entries are not variations on one value, and the minimum's is appended after every format's, so any budget over that list takes away precisely the candidate this change exists to add — and takes it away only inside a collection, so the same newtype was writable at a position of its own and not as an element. Nothing prunes a candidate a rule was read to produce. What consumes them is `Generator`'s own bounded walk, which reports `SEARCH_LIMIT` when it stops, and that is where a limit can be read for what it is.

**A bound went on the invented side instead**, and it turned out to be needed. The count comes from the model, so a rule can ask for a million: without a guard the generator produced two rows of 5,000,036 characters each. Past sixty-four elements nothing is proposed.

**One thing the review did not raise, found while testing the above.** Two tests wrote a fixture that breaks the rule it is written under — `nest = Nest([])` under `List.length(value) >= 1` — so the model did not compile and the empty answer agreed with what the test expected of one. The helper the tests generate through now holds the model to compiling, and both cases are rewritten without a written row, which is the only way to state a rule no fixture satisfies. The million-element test was written the same way and was passing vacuously until this; with the guard removed it now fails with the row it should never have built.

Twenty-one tests. The crm corpus is unchanged at 57 boundaries not known to be writable. Compiler tests, best of three: 58s.

kawasima:
Both reproduced and are fixed in 24cd5d56.

**The count guard was above the carrier check**, so it bounded a string's length by a number that is about collections. `String.length(value) >= 65` lost its witness for no reason the type suggests — a string of sixty-five is one literal where a collection of sixty-five is sixty-five values each built in turn. Two budgets now, each checked inside the branch that knows its carrier: 4096 characters, 64 elements.

**The reason was wrong and there is already a right one.** `NOTHING_COMPOSES_ONE` says the search has no way to write a value of the shape asked for and carries no opinion about the model; `probe` uses it for exactly this, and its note names a string of a given length as the case. A rule asking for a million elements now reports that instead of `ALL_CANDIDATES_REJECTED`, which was inviting a reader to look for the rule that refuses a value nothing refuses.

**The map product is bounded and reported as a search that stopped.** You are right that `Generator`'s walk cannot cover it — every pair is built before any of them is tried, so what the product costs is allocation, not the walk. Bounded at 64 pairings in the diagonal order that was already there, which is what makes the bound cheap: it drops the pairs furthest from what either side proposed first, so `Key[1] + Val[0]` is never among them. `SEARCH_LIMIT` is what it reports, since values of the shape were built and refused and more were never reached.

Both questions are asked only where nothing was written — once, at the end, of the positions of the parameter that failed. Re-reading what a position could have offered is work a row that exists has no reason to pay for.

Twenty-three tests. The crm and issuetracker corpora are unchanged byte for byte, still 57 boundaries not known to be writable. Compiler tests, best of three: 1:00.

On the principle: I agree with where you put the line, and I had it in the wrong place twice — first pruning what was read, then classifying what I declined to build as what the model refuses. What is bounded now is only what this invents, and each bound says which of the two it is.

kawasima:
Both reproduced. `Set<Bool>` came back holding one element, and the `/= 0` map reported `SEARCH_LIMIT` for a search nothing ran. Fixed in ccba87f2 — but not where they showed, because they are not two defects.

Across four rounds this file produced five findings, and they are two mistakes made repeatedly:

| round | finding | mistake |
|---|---|---|
| 2 | `holding` collapsed the element's proposals | a carrier chain, incomplete |
| 3 | the size guard sat above the carrier switch | the same chain |
| 4 | `varied` steps only strings and numbers | the same chain |
| 3 | the map product is eager | `notBuilt` re-derives what the builder decided |
| 4 | `moreThanIsPaired` ignores the minimum | the same re-derivation |

There were four independent carrier chains — what to build, how to vary it, what is too large, how many pairings — so adding a carrier meant remembering all four, and `notBuilt` was a second copy of the builder's conditions that had already drifted twice. Patching the two findings leaves both causes alive and buys a fifth round.

**What a type's values are is asked once now.** A type divides into its cases first — a `Bool` into two, a sum into its own — because each of those is a value of the type rather than a proposal about it. Then values made by stepping the carrier, which stay inside the rules it carries; that ordering is what keeps `Set<Positive>` off the carrier's `Positive(0)`. Then the type's own proposals last, which is all there is where the carrier neither divides nor steps.

`Set<Color>` over a three-case sum is in the tests as the check that this is coverage rather than a list of carriers I thought of: nothing in the file is written for sums, and it builds `Palette([Red, Green])`.

**The builder answers what it did not build.** `holding` and the held-back reason are two halves of one decision now, so there is nothing left for a second reading to disagree with. Your suggested `least > 0` gate falls out of it rather than being another condition to keep in sync — which matters, because that is the third time this pair drifted.

Twenty-six tests. The crm and issuetracker corpora are unchanged byte for byte. Compiler tests, best of three: 55s.

Thank you for the four rounds — the last two findings are what made the shape visible.

PR #548, opened 2026-08-09
The layout is a value, and its text is that value's projection

### LINKED ISSUE #545: The layout is realized and only its text is kept
`Doc.render` decides how the canonical form is laid out and returns only its text. Which groups were written flat, at what column each of them was measured, what indent each break was written at, and where each place ended up are all computed there and dropped.

That is the last canonical decision in the formatter that is not a value. #506 made the construction's places and their correspondence values; the boundaries became values with `TokenDoc` and `Gaps.resolve`; what is left is the layout's own.

## What is missing, verified

- `Gaps.lower` drops the places on the way down: `case TokenDoc.At a -> lower(a.doc(), answers, next)` (`Gaps.java:263`). A `Doc` knows nothing about them.
- `Doc.Group(Doc doc)` and `Doc.Nest(int indent, Doc doc)` carry no identity. Two groups under one place are indistinguishable, and the unit of the conditional-layout rule is the group.
- `Doc.render` computes `Mode` per group into a local and returns a `String` (`Doc.java:84`).

The consequence is measured: of 2162 differences between this repository's sources and their canonical forms, 1228 sit at boundaries the layout may break, and nothing in the document says whether it did.

## What this is

```
Doc     the layout program
Layout  the realized canonical layout
text()  its serialization
```

`Doc -> String` becomes `Doc -> Layout -> String`. The formatter is not rewritten: `Doc` is 197 lines, `render` is called from one place (`Formatter.java:123`), and nothing outside `souther-fmt` names the type.

## What `Layout` holds

Places are not enough. The rules ask about three kinds of unit and each needs its own identity:

- **places**, carried down from `TokenDoc.At` rather than stripped, each with the span it was written at;
- **groups**, with an identity of their own — not a record's structural equality, which would make two identical groups one;
- **nesting**, likewise, so that a break can say which nesting it was written under. The indentation rule's unit is a pair of consecutive levels, so a realized indent of `8` on its own leaves the pair to be reconstructed from the text.

A group's outcome has to say more than flat-or-broken:

```
Flat
BrokenByWidth
BrokenBy(forced)     the Hard, MustBreak or Trailing that refused the flat layout
```

`fits` returns false for a `Hard`, a `Trailing` and a `MustBreak` as well as for a line too long (`Doc.java:176`). A decision recording only the column would leave a reader unable to tell "this did not fit at column 73" from "this could never be flat", and working that out afterwards means running `fits` again — which is the reconstruction this issue exists to stop.

Naming which of the four structural obligations a forced break discharges is **not** part of this. What is recorded here is the evaluation fact — this group broke because of that `Hard`. Which rule that `Hard` answers to is a separate piece of work, and it can be given later without touching this.

Breaks and their indents are the primary record; lines are derived from them. Making a line the first-class object would need blank lines, place spans and token boundaries each related to it separately.

## Blank lines come back for free

`Gaps.boundaries` shows one boundary per adjacency, so a blank line — two forced breaks where the tokens have one adjacency (`Gaps.java:84`) — is a canonical fact the view cannot show. A `Layout` that records the breaks it realized keeps both of them, and the top-level separation rule stops needing a view of its own.

## Not a renderer trace

The thing to keep is not that `fits` was called, or that a `Nest` was pushed. That is the mechanism and it changes when the code is rearranged. What is kept is the layout that was realized: which groups are flat, what each break wrote, where each place is. #506 drew the same line for places — they name a position in what was written rather than the order the formatter built it in.

## Done when

1. **The projection is unchanged.** `layout(doc, WIDTH).text()` is byte-identical to today's `doc.render(WIDTH)` over every formatter fixture and every `.sou` in the repository.
2. **Every decision is there.** One `Group` in the document gives one decision in the layout, including where several sit under one place.
3. **Speculation is not recorded.** `fits` reads ahead over groups and nests it is only measuring. A decision becomes a value at the one point the outer loop evaluates that group, and a fixture holds that a group visited only by a look-ahead produces none.
4. **The same text keeps different decisions.** A fixture where a width break and a forced break write the same characters, asserting that the text matches and the decisions differ. Without this the layout could be re-deriving its answers from its own output and nothing would say so.

## Order

This comes before the deviation witness of #444, and that is a dependency rather than a preference. Written the other way round, the witness would re-run `fits` for the conditional rule, read nesting back out of the rendered text for indentation, work around the boundary view for blank lines, and search the canonical document again for places — which is the reconstruction #444 records failing five times.

Refs #444.

kawasima:
`Doc.render` decided how the canonical form was laid out and returned only its text. Which groups were written flat, at what column each was measured, what indent each break was written at, and where each place ended up were all computed there and dropped.

That was the last canonical decision in the formatter that was not a value. `Doc -> String` is now `Doc -> Layout -> String`, and `render` is the projection.

## What the layout holds

**Group decisions.** A group is named by an identity of its own, because two groups written the same way are two decisions. The outcome says why, not only whether: `fits` returned false both for a line over the width and for a group holding something that cannot be laid out flat at all, and told apart only by the text those two are the same characters. Which obligation a forced break answers to is deliberately **not** here — that is the forced-layout rule's and can be given without changing this.

**Place spans.** `Gaps.lower` dropped the places #506 made on the way to the layout program, so a canonical form knew which places it had and not where any of them ended up. Finding one in the output meant matching text, which is the search places were made to replace.

**Breaks.** Every newline the layout wrote, with the indent the line after it starts at and the nestings it was written under. A blank line is two forced breaks at one adjacency and `Gaps.boundaries` shows one boundary there, so the separation rule's answer had nowhere to be read from; both breaks are here because both were written. The nestings are named rather than measured — the indentation rule answers about a pair of consecutive levels, and an indent of eight says only what the second came to.

## A defect it found

`Place.parent` says "the place this one is written inside", and a function type did not. Its place named the bracketed run — `('k, 'a)` of `('k, 'a) -> Bool` — while the result was written at a place under it and after it. Seven of the standard library's types are like that.

Nothing could have said so before: the two are the same characters whichever place they belong to, and the parent relation was only ever read as a chain to walk up. Measuring where each place was written is what shows it. The fix moves one `at(...)` and changes no text.

## What holds it

The existing fixtures hold the projection: 4487 tests pass and the golden corpus is unchanged, which is what says the layout writes what it always wrote.

Four mutations, each required to fail:

- a decision naming a fresh group rather than the one laid out — caught by the identities, and **not** by counting them, which is what an earlier version of the sweep did;
- a shared identity for every nesting — caught by reading the layout's two chains against each other. The first version read its expectation off the document it was checking and held;
- dropping the second break of a blank line — caught by name;
- taking a decision while measuring rather than while laying out — this cannot arise: `fits` has no way to record, and a fixture holds that a group visited only by a look-ahead produces one decision and not two.

## Not a renderer trace

Nothing here says that `fits` was called or that a nesting was entered. Those are the mechanism and change when the code is rearranged. What is kept is what the layout realized, which is what the canonicalization rules are about — the same line #506 drew for places.

Closes #545

### DIFF
--- a/souther-compiler/src/test/java/souther/compiler/fmt/ALayoutSaysWhatEachBreakWroteTest.java
+++ b/souther-compiler/src/test/java/souther/compiler/fmt/ALayoutSaysWhatEachBreakWroteTest.java
@@ -1,0 +1,72 @@
+package souther.compiler.fmt;
+
+import org.junit.jupiter.api.Test;
+
+import souther.compiler.cst.CstParser;
+
+import java.util.List;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+
+/**
+ * A break is a decision and the layout keeps it: where it was written, how far in, and which
+ * nesting it was written under.
+ *
+ * <p>Two of the fourteen rules ask about this and neither can be answered from the document. The
+ * indentation rule asks about a pair of consecutive nesting levels, so a written indent of eight on
+ * its own leaves the pair to be worked out from the text. And a blank line is two breaks at one
+ * adjacency — {@link Gaps#boundaries} shows one boundary there, so what the top-level separation
+ * rule decided is a canonical fact the document's view cannot show.
+ */
+class ALayoutSaysWhatEachBreakWroteTest {
+
+    /** A blank line is two breaks and comes back as two. */
+    @Test
+    void aBlankLineIsTwoBreaks() {
+        Layout layout = Formatter.document(CstParser.parse("""
+                module m
+
+                data A
+
+                data B
+                """).root()).resolve().layout(100);
+
+        List<Integer> offsets = layout.breaks().stream().map(Newline::offset).toList();
+        String text = layout.text();
+        assertEquals("module m\n\ndata A\n\ndata B\n", text);
+        assertEquals(List.of(8, 9, 16, 17, 24), offsets,
+                "each newline the layout wrote, and the two of a blank line are both there");
+    }
+
+    /** A break says how far in it was written and which nesting wrote it. */
+    @Test
+    void andABreakSaysWhichNestingItWasWrittenUnder() {
+        Doc inner = Doc.nest(4, Doc.concat(Doc.text("a"), Doc.HARDLINE, Doc.text("b")));
+        Layout layout = Doc.group(inner).layout(100);
+
+        assertEquals("a\n    b", layout.text());
+        assertEquals(1, layout.breaks().size());
+        Newline wrote = layout.breaks().get(0);
+        assertEquals(4, wrote.indent());
+        assertEquals(List.of(((Doc.Nest) inner).ref()), wrote.under(),
+                "the nesting it was written under, by that nesting's own identity");
+    }
+
+    /** Two nestings written the same way are two, so the pair of levels a break sits between is
+     * readable rather than guessed from the amounts. */
+    @Test
+    void andConsecutiveLevelsAreTellableApart() {
+        Doc innermost = Doc.nest(4, Doc.concat(Doc.text("b"), Doc.HARDLINE, Doc.text("c")));
+        Doc outer = Doc.nest(4, Doc.concat(Doc.text("a"), Doc.HARDLINE, innermost));
+        Layout layout = Doc.group(outer).layout(100);
+
+        assertEquals("a\n    b\n        c", layout.text());
+        assertEquals(List.of(4, 8), layout.breaks().stream().map(Newline::indent).toList());
+        assertEquals(
+                List.of(List.of(((Doc.Nest) outer).ref()),
+                        List.of(((Doc.Nest) outer).ref(), ((Doc.Nest) innermost).ref())),
+                layout.breaks().stream().map(Newline::under).toList(),
+                "the second break is one level further in than the first, and says so by naming"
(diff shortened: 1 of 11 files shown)


kawasima:
Both taken, and both were measurable. The second one first, since it is the sharper of the two:

```
Doc.group(concat(text("way-too-wide"), hardline(), text("tail")))

width 4     BrokenByWidth
width 1000  BrokenByForcedLayout
```

The same group, written down the page at both, answering with whichever failure the walk met first. That is where the measuring stopped and not what decided the layout, which is the reading this PR claims not to be.

Refusing wins now, and the walk goes on past an overflow over the group's own flat content to see whether one is there. Both widths answer `BrokenByForcedLayout` with the same break.

And the decision names it. `Hard`, `MustBreak` and `Trailing` each have an identity of their own — the shared `Doc.HARDLINE` and `Doc.MUST_BREAK` constants are gone, since two written the same way have to be two — and `BrokenByForcedLayout(Doc refusing)` carries the one that refused. Which obligation it answers to is still not here; naming the break is exactly what lets that be given later without the group being measured again to find out which of its hardlines it was, which was the reconstruction the whole change is for.

Two fixtures hold it: the same group answering the same way at 4 and at 1000, and a group with a hardline inside it and another after it, where the decision names the one inside.

4489 tests pass.

Issue #520, opened 2026-08-09
There is no way to look up the syntax of a form


kawasima:
Round-5 dogfooding reported five syntactic questions the tooling could not answer. Investigation confirms three of them, shows two did not reproduce, and puts the root somewhere other than where the original report put it.

The root is not a missing grammar page. The specification has no mechanism holding its account of concrete syntax to either completeness or discoverability. A `grammar` topic is one projection of that.

## Three layers

1. *Missing specification.* The information is absent, so no search reaches it.
2. *Missing reachability.* The information exists but no name reaches it, or a diagnostic lands on a sentence that names no destination.
3. *Missing machine-checkable coverage.* The compiler holds a closed set and nothing compares it with the specification.

### Layer 1 — absent

String literal lexis is stated nowhere. `primitives` gives the numeric literal form and says nothing about string literals. `stdlib-string` states the unknown-escape rule in one clause and cites `<<primitives>>` for it, where it is not.

Writing that rule down turned out to require deciding it first: what the lexer does matches neither language this syntax is grounded in, and E2305 already describes a rule the lexer does not follow. That decision, the compiler change it takes, and the specification of the result are #570. This issue does not write the string literal's lexis; it requires that the grammar inventory reach it.

`if` never states that `else` is required. `CstParser.ifExpr` expects `ELSE_KW`, so an `if` without one does not read and the language has no dangling-else case. The reported question — which `if` an `else` binds to — is not a rule Souther has; what is missing is the sentence saying the question does not arise. `guard` is the one-armed form and the two are not contrasted.

### Layer 2 — written but unreachable

`==== Numeric literals` carries no anchor, so the one place a literal's form is written cannot be asked for. Ten headings in the document are unanchored.

`declaration-syntax` and `expression-syntax` state an obligation — "MUST be written in the form given for it" — and name no destination. E2301 and E2302 land there. The other rejection points do cite sections, and nothing checks that a cited section states a form, which is how `literal-syntax` → `primitives` passes while leaving string literals unanswered.

No name in the document has concrete syntax as its subject. `doc_read {name: "grammar"}` and `{name: "operators"}` do not resolve, and `{name: "syntax"}` suggests the five `*-syntax` rejection points, which are names of refusals rather than of forms.

### Layer 3 — unchecked

[...]

Issue #522, opened 2026-08-09
A field type written as an anonymous union gets a bare syntax error


kawasima:
A field type written as an anonymous union gets a bare syntax error.

## What happens

```
data Holder = { t: A | B }
```

```
-- SYNTAX ERROR  E2301 ---------------------------u.sou:6:22

6| data Holder = { t: A | B }
                        ^

I expected `}` here, but found `|`.
```

The same misconception in a parameter is diagnosed well:

```
-- BOUNDARY TYPE  E1312
parameter `ticket` has an anonymous union type `GeneralAdmission | ReservedSeat`; a parameter type
must be a single named type — declare `data ... = GeneralAdmission | ReservedSeat` and take that name.
```

## Why it matters

A field is where an author most naturally reaches for "either of these kinds", and it is the position that gets no guidance. The repair is the same one E1312 already writes out.

## Expected

Recognize a union in a field type and answer with the E1312 wording, naming the field: declare a `data` over the cases and give the field that name.

Found in round-5 MCP dogfooding (ticketing model).

kawasima:
Fixed in #559. The field's union is one of five cells: which tier refuses a written form was decided by which type production the position calls, and only a form that was read could be named. The grammar is unchanged; each forbidden continuation is recognized where it stands, under E2307 and E2308. Held by a sweep over every type position the specification names crossed with every type form.

PR #550, opened 2026-08-09
The positions a type may be written in are stated once

### LINKED ISSUE #511: The section that lists where an optional may be written names a position the boundary refuses
The section that lists where an optional may be written names a position the boundary refuses.

## What happens

`specification.adoc:663` (`How Option appears`) gives a closed list of five positions:

> `Option<T>` is the name of the type a `?` field has, and it MAY be written wherever a model *reads* an optional: a data field (where `T?` means the same thing), **a behavior's input**, a helper's parameter or declared return, a function type's result, and a local binding's annotation.

`An optional does not stand in a boundary` says the opposite:

> an optional MUST NOT stand anywhere in a behavior's boundary shape — neither as a parameter nor as the output, and at any depth inside a collection either.

The compiler agrees with the second:

```
-- BOUNDARY TYPE  E1313 ---------------------probe2.sou:3:25
3| behavior takesOption : (d: Option<Date>) -> Int
                           ^
Parameter `d` carries an optional (Date?).
```

## Why it matters

This is the page a reader lands on for exactly this question, it presents an enumeration rather than an example, and one of its five entries is false. Trusting it means designing behaviors around inputs the compiler will not accept — a redesign, not a one-line fix, and invisible to anyone who has not yet run the compiler. That is precisely the reader the section is for.

## Expected

Drop `a behavior's input` from the list, or qualify it with the boundary rule that overrides it.

While that paragraph is being edited, two adjacent rules are stated nowhere and were learned only from the compiler:

- `List<T?>` is not a writable type at all — `?` is a field-level suffix and the parser rejects it inside `<>` (E2301). The boundary section's "at any depth inside a collection" implies it is writable and merely refused at a boundary.
- `List<Option<T>>` is writable and has no derivable codec (E1311).

Found in round-5 MCP dogfooding (maintenance contract model).

kawasima:
`How Option appears` gives a closed list of five positions an optional may be written in, and one of them — a behavior's input — is refused by `an-optional-does-not-stand-in-a-boundary`, which the same document states. The compiler agrees with the boundary rule (E1313). This is the page a reader lands on for exactly that question.

Fixing the paragraph alone would have left three known copies of the same defect in place, so this removes the duplication instead.

## What was wrong

Six sections each wrote their own set of type positions, and four of the four that make a claim about where a type form may go disagreed with the compiler:

| section | the list | measured |
|---|---|---|
| `How Option appears` | includes a behavior's input | E1313 refuses it |
| `Qualified reference` | omits a local binding's annotation | `let a: up.Amount = ...` compiles |
| Tuple type | omits a local binding's annotation | `let t: (Int, Int) = ...` compiles |
| Function type | "written in one place: a helper's parameter" | `let p: (Int) -> Int = ...` compiles |

The other two — `a-value-has-the-type-its-position-requires` and its E1301 entry — carry the same list twice and omit a newtype's base and a behavior's output.

The function type's rule was false in its own name. `DiagnosticCode.E1810` cited `a-function-type-is-written-only-on-a-helper-parameter`, `messages.properties` copied the sentence, and `CstParser` carried a third copy as a `spec 13.1` comment. E1810 fires on a different condition entirely: a binding holding a function whose annotation is not a function type (`Elaborator.checkAnnotatedLambdaBindings`). The adjacent rule `a-function-bindings-type-is-known` and E1808's hint both say the opposite of the rule beside them, and both are right.

## What this does

The positions are stated once, in `delimiters`, as `a-type-is-written-in-a-type-position`. They are taken from the parser's `retType()` / `typeRef()` call sites rather than from any of the prose — four of the prose lists were wrong, so a fifth could not be written from the same material.

`delimiters` was already the right home: it separates what a position admits from what a type is (`Writing a type in a position is not being admitted by it`), and it already names the positions requiring an external representation. Every restriction already had an owner — `what-crosses-the-boundary-has-an-external-representation`, `an-optional-does-not-stand-in-a-boundary`, `a-parameter-names-one-type`, `core-privileges-stay-in-the-core`. What was missing was only the set of positions. Each of the six sections now cites the rule and keeps what is specific to its own type.

The restrictions turn out to follow from what the position is, which is what makes the citation sufficient. Measured against the compiler:

| | Option | function | tuple | anonymous union | `?` |
|---|---|---|---|---|---|
| data field | ok | E1311 | E1311 | E2301 | ok (only place) |
| newtype base | E1009 | E1007 | E1007 | — | E1009 |
| behavior parameter | E1313 | E1311 | E1311 | E1312 | E1402 |
| behavior output | E1313 | E1311 | E1311 | ok (only place) | E1402 |
| helper parameter / return | ok | ok | ok | ok | E1402 |
| local annotation | ok | ok | ok | ok | E1402 |
| type argument / tuple member | ok | ok | ok | E2301 | E2301 |
| function type parameter / result | ok | ok | ok | ok | E1402 |

A qualified reference is admitted in every position.

## E1810

The rule, the anchor, the message, the hint, the test names and the parser comment now all say the condition the code checks:

```
The binding `f` holds a function, and its annotation is not a function type.
Hint: Write a function type, such as `(Int) -> Bool`, or leave the annotation off and let the applications settle it.
```

The added test is the one whose absence let the false rule stand: it holds that a function type in a local annotation compiles and runs. The two existing tests pass just as well against a check that refuses every annotation on a function.

## Not in this change

`An anonymous union is written only in an output` (in the codec section) is a fifth instance — helper signatures and local annotations admit one. It reads as true when scoped to boundary positions, so whether to qualify it or drop it is a separate call. `List<T?>`, the codec for `List<Option<T>>`, #520 and the formatter's ownership of the canonical form are on a different axis from the position set.

Nothing here is held by a check yet. The build rule from #543 verifies that a cited anchor resolves, which is reference integrity — it would not have caught any of this.

Closes #511.


kawasima:
Both fixed in 1fb6ef2b.

**P1.** You are right, and the sentence got worse rather than merely staying wrong: it was vague while `type position` was undefined, and became a universal claim the moment the positions were named. The parser carries four shapes, not one, and a `newtype`'s base is a fourth I nearly missed — a tuple and a function are refused where they open (`ANewtypeCannotWrapATuple` / `ANewtypeCannotWrapAFunction`), so it does not read them at all.

- whole type (`retType`) — the `|` chain and a trailing `?`: a behavior's parameter and output, a composition's declared output, a helper's parameter and declared return, a local binding's annotation, a function type's parameter and result
- a data field — `typeRef` and a `?`, no `|` (#522)
- a type argument and a tuple's member — `typeRef` alone
- a newtype's base — narrower again, which the following bullet already owned

Written that way rather than as a taxonomy, as you suggested. The syntax tier and the admissibility tier stay apart, which is what the next bullet states.

**P3.** Two kinds of position, not two positions: the boundary — the parameter and the output — and a newtype's base.

`mvn clean test` green across all modules.

Issue #555, opened 2026-08-09
The layout says where an annotation was written, not where a place is


kawasima:
`Layout` records where a `TokenDoc.At` was written. It is read as where each place of the canonical form is, and those are not the same set: of the 3202 places this repository's corpus makes, 63 have no entry.

`At` is one way an output location is established. It is not what a location is.

## What the 63 are

Sixty-eight construct kinds have an entry for every place they make. Eight do not, and seven of those have an entry for none:

```
construct         with     without
SOURCE_FILE          0          23
LBRACE               0          20
WITH_KW              0           9
ASSIGN               0           4
PRODUCT_BODY         0           3
IDENT               34           2
ELSE_KW              0           1
WITH_CLAUSE          0           1
```

They divide by how a location would be arrived at, not by what kind of thing they are.

**A location the construct wrote.** `PRODUCT_BODY`, `WITH_CLAUSE`, and the two `IDENT`s.

These are missing `At`s and nothing more. A product body writes its own braces, so covering it by the extent of its members leaves them out — measured, the members cover `[ a: Int\n    , b: Int]` where the body wrote `{ a: Int\n    , b: Int\n    }`. A `with` clause is the same shape: its members cover `[dep = two, other = three]` and it writes the `with` as well. The two `IDENT`s are a definition's name — `example helper`, `fake helper` — written as a run, and the other thirty-four `IDENT` places, under `MATCH_CASE`, `SUM_BODY` and `NAME_LIST`, all have one. Two sites out of thirty-six is an omission and not a category.

**A position, and no region.** `LBRACE`, `WITH_KW`, `ASSIGN`, `ELSE_KW` — 34 places.

These name the line a construct opens with, and they carry the comment written at the end of that line. Given `data D =   // about the block` the `ASSIGN` place carries that comment and holds none of the text. There is no region to record and there is a place in the output where it sits.

**The whole output.** `SOURCE_FILE` — one per source.

## So there are three ways a location is arrived at, not four kinds of place

```
written    the extent of what the construct emitted        <- At
point      where a carried comment is inserted             <- the layout's position
global     the canonical output                            <- its bounds
```

The extent of a place's members is not a fourth: the two constructs it looked like a candidate for turned out to write their own brackets, so they belong under `written`.

Wrapping every remaining place in an `At` would therefore be wrong, and wrapping some of them is exactly right. What must not happen is `At` becoming the definition of where a place is — an anchor has no document to wrap and the file's bounds are not something a construct emits.

[...]

PR #556, opened 2026-08-09
One canonicalization run answers for itself

### LINKED ISSUE #551: One canonicalization run answers for itself
Canonicalizing a source makes identities: a `Place` for every place the canonical form writes, a `Doc.GroupRef` for every group, a `Doc.NestRef`, a `Doc.Ref` for every forced break. Everything the run has to say is said over them — `Layout` keys its spans and decisions on them, `Correspondence` keys its relation on the same places.

They belong to that run. Two canonicalizations of one source share none of them:

```
two runs over one source
   places   10 and 10, shared by identity: 0
   groups   4 and 4, shared by identity: 0

a layout from one run, asked about the other run's places
   0 of 10 have a span
   the same question about its own run's places: 10 of 10
```

And the API hands them out a call at a time. `Formatter.document(root)` looks like a function — same input, same shape of answer — and its answer is comparable only with itself. Anyone who needs two of a run's answers calls it twice and gets two runs.

## The failure is a lookup that misses

Nothing rejects the mixture. There is no exception and no type error: a `Place` is a `Place`, a `Layout` takes it, and the answer is that it has no span. The two collections are the same size and disjoint, so a comparison of counts holds and says nothing.

That is not hypothetical. `ALayoutSaysWhereEachPlaceWasWrittenTest.everyPlaceOfTheDocumentHasASpan` builds the document twice — once inside its `layoutOf` and once for the places it walks — and asserts the two are the same size. Asked instead whether the layout has a span for each of those places:

```
expected: <0> but was: <10>
```

None of them. The first test written against this API was misled by it, and the misuse is now pinned as a fixture. `TheLayoutKeepsWhatItsTextCannotSayTest` avoids it only because #545 changed its group assertions from counts to identities after the same weakness was found there.

## What this is

One canonicalization, one result, and every relation it publishes expressed over identities that result owns.

```
canonicalize(source) -> CanonicalForm
        text
        the places, groups and nestings it made
        the correspondence, source elements to places
        the layout — decisions, spans, breaks

format(source) = canonicalize(source).text()
```

`format` stays what it is. What changes is that a caller never has to build twice to hold two of a run's answers, so the identities they hold are the same run's without having to be careful.

This does not make mixing impossible. Java's types will not carry which run a `Place` came from, and someone determined to pass one result's place to another result's layout still can. What goes away is the reason to: today the second run is what a caller does when they need something the first call did not return.

## Acceptance

- One entry point produces the whole result. Nothing produces a part of it on its own that a caller would otherwise have to rebuild.
- Every relation a `CanonicalForm` publishes is over identities that same `CanonicalForm` made. A span answers for its own places; the correspondence answers about them and not about a re-derived set.
- `format` is that result's text, byte-identical to today's over every fixture and every `.sou` in the repository.
- `everyPlaceOfTheDocumentHasASpan` compares the layout against the places of the same run, by identity. Its present form passes over two runs and holds nothing; a count is not the assertion it was written to make. Narrowed to one run it turns out that not every place a run makes is written as a region — an anchor for the line a construct opens with carries a comment and holds no text, a construct writing through its members wraps nothing itself, and the file's own place wraps nothing at all. That is a fixture here and a question for the issue after this one: `spans()` answers where an `At` was written, not where a place is, and what a place's location is when it writes no text is not this issue's to settle.
- A fixture states what two runs share, so that the reason the result is one object is written down rather than assumed.

Refs #444. Comes before the rule identities and the attribution relation of #552.

kawasima:
A canonicalization makes identities — a place for every place it writes, a group for every group, a nesting, a forced break — and everything it has to say is said over them. `Layout` keys its spans and decisions on them; `Correspondence` keys its relation on the same places.

They belong to that run. Two canonicalizations of one source share none:

```
two runs over one source
   places   10 and 10, shared by identity: 0
   groups   4 and 4, shared by identity: 0

a layout from one run, asked about the other run's places
   0 of 10 have a span
   the same question about its own run's places: 10 of 10
```

And the API handed them out a call at a time. `Formatter.document(root)` looks like a function — same input, same shape of answer — and `format` built its own construction separately, so the two were already two runs.

## The failure is a lookup that misses

Nothing rejects the mixture. No exception, no type error: a place is a place, a layout takes it, and the answer is that it wrote it nowhere. Two collections come out the same size and share not one member, so a count of them holds.

`ALayoutSaysWhereEachPlaceWasWrittenTest.everyPlaceOfTheDocumentHasASpan` built the document a second time to walk it and asserted the two were the same size. Asked by identity instead:

```
expected: <0> but was: <10>
```

None of the places it walked had a span in the layout it was checking. The first test written against this API was misled by it, and the misuse was pinned as a fixture. `TheLayoutKeepsWhatItsTextCannotSayTest` escaped only because #545 changed its group assertions from counts to identities after the same weakness turned up there.

## What is here

`canonicalize(file)` is one run and returns what it made — the construction, and the layout keyed on it. `format` is that run's text. The edges from a `SyntaxNode` to a part of a run are gone: `document(SyntaxNode)` and `build(SyntaxNode)` were both a source in and one run's piece out, and a caller could hold either beside another run's layout. A test that wants a document or a construction asks a canonicalization for one, and the longer call is the point — it says at the call site which run the piece belongs to.

The sweep compares the layout against its own run's document, and a fixture states what two runs share, so the reason a canonical form is one object is written down rather than relied on.

## What narrowing it to one run turned up

Of the 3202 places the corpus makes, 63 have no span: the file's own place, an anchor for the line a construct opens with — a `data` declaration's `=`, a block's brace, a match's `with` — and a construct that writes through its members. They are not idle. Given `data D =   // about the block` the `ASSIGN` place carries that comment and holds none of the text.

So `spans()` answers where an annotation was written and not where a place is, and its domain comes from where the construction happened to put one rather than from what a place means. That is held as a fixture here and settled in #555, which also removes the `continue` that has been taking parents without spans out of #545's containment check.

4499 tests pass.

Closes #551

### DIFF
--- a/souther-compiler/src/test/java/souther/compiler/fmt/ADefinitionsBodyOwnsItsCommentTest.java
+++ b/souther-compiler/src/test/java/souther/compiler/fmt/ADefinitionsBodyOwnsItsCommentTest.java
@@ -166,7 +166,7 @@ class ADefinitionsBodyOwnsItsCommentTest {
                     // about the body
                     x
                 """).root();
-        Formatter.Construction built = Formatter.build(root);
+        Formatter.Construction built = Formatter.canonicalize(root).construction();
 
         SyntaxNode lambda = only(root, SyntaxKind.LAMBDA_EXPR);
         SyntaxNode body = childOf(lambda, SyntaxKind.VAR_EXPR);
--- a/souther-compiler/src/test/java/souther/compiler/fmt/ALayoutSaysWhatEachBreakWroteTest.java
+++ b/souther-compiler/src/test/java/souther/compiler/fmt/ALayoutSaysWhatEachBreakWroteTest.java
@@ -24,13 +24,13 @@ class ALayoutSaysWhatEachBreakWroteTest {
     /** A blank line is two breaks and comes back as two. */
     @Test
     void aBlankLineIsTwoBreaks() {
-        Layout layout = Formatter.document(CstParser.parse("""
+        Layout layout = Formatter.canonicalize(CstParser.parse("""
                 module m
 
                 data A
 
                 data B
-                """).root()).resolve().layout(100);
+                """).root()).layout();
 
         List<Integer> offsets = layout.breaks().stream().map(Newline::offset).toList();
         String text = layout.text();
--- a/souther-compiler/src/test/java/souther/compiler/fmt/ALayoutSaysWhereEachPlaceWasWrittenTest.java
+++ b/souther-compiler/src/test/java/souther/compiler/fmt/ALayoutSaysWhereEachPlaceWasWrittenTest.java
@@ -8,6 +8,7 @@ import java.util.ArrayList;
 import java.util.List;
 
 import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertFalse;
 import static org.junit.jupiter.api.Assertions.assertTrue;
 
 /**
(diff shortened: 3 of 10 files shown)


kawasima:
Taken. `document(SyntaxNode)` is gone.

The point that building it through `canonicalize` inside does not help is the one I had wrong. What the boundary is about is which run-local values a caller can obtain on their own, not how any of them is computed — and a shortcut from a `SyntaxNode` to something under a canonical form is the same operation whatever it does underneath.

Six tests asked for a document; they ask a canonicalization for one now. The longer call is the point: it says at the call site that the document is that run's.

The minor is taken the second way. `andTwoRunsShareNothingTheyMade` reads groups as well as places, since #545's group assertions are all within one run and nothing was holding the across-run case. Its name claimed more than its body did.

4498 tests pass — one fewer than before because the removed `document` had a test of its own that is now the same assertion made through `canonicalize`.

kawasima:
All five taken.

The deleted fixture was an accident and you read it right — a replacement I wrote for the cross-run test spanned to the wrong anchor and took `butNotEveryPlaceTheRunMakes` with it. Without it the branch claims only that a place written with an annotation has a span, which is exactly the narrower statement the sixty-three are the counterexample to. It is back.

`build(SyntaxNode)` is private. The argument is the one you made about `document` and I should have applied it to its neighbour in the same commit: what matters is which run-local pieces a caller can obtain from a source, and package-private is not the distinction, since `document` was package-private too. Five call sites in the `#506` tests go through `canonicalize(...).construction()` now.

The cross-run fixture is `andTwoRunsShareNoPlacesOrGroups`. Narrowing rather than widening — nestings and forced breaks are run-local for the same reason, and a fixture claiming them without reading them is the shape this PR is about.

The javadoc on `everyPlaceTheDocumentWritesHasASpan` says the document's places rather than the run's, and points at the fixture and at #555 for the difference.

The PR body's `document goes through it` is gone; it now says both edges are.

4499 tests pass.

PR #557, opened 2026-08-09
An anonymous union is written where a whole type is read

### LINKED ISSUE #553: An anonymous union is admitted in a helper's signature and a local annotation, and two sections say it is written only in an output
An anonymous union is admitted in a helper's signature and a local annotation, and two sections say it is written only in an output.

## What happens

`union-intersection` (`specification.adoc:931`) owns the rule:

> An anonymous union appears *only in a behavior's output*. An anonymous union MUST NOT be written in an input (parameter) type.

The codec section restates it as a premise (`specification.adoc:3603`):

> There is no matching decode. An anonymous union is written only in an output (<<union-intersection>>), so nothing reads one back.

The compiler admits one in a helper's parameter, a helper's declared return, and a local binding's annotation, and the value works — a `match` over it opens both cases:

```
module p exposing (A, B, Out, run)

data A = { a: Int }
data B = { b: Int }
data Out = { v: Int }

let pick (x: A | B) : Int =
    match x with
        | A as u -> u.a
        | B as u -> u.b

let make (n: Int) : A | B = A { a = n }

behavior run : (i: Out) -> Out
    constructs Out, A

let run (i) = {
    let z: A | B = A { a = 1 }
    Out { v = pick(z) + pick(make(2)) }
}
```

This compiles. So the union is not merely parsed and ignored: it is read as a type, a `match` on it is exhaustive over its two members, and the helper is applied at both a local binding and another helper's result.

## What is and is not refused

The refusals the rule is really about are all boundary positions and all have their own diagnostics:

| position | verdict |
|---|---|
| behavior parameter | E1312 |
| behavior output | ok — this is the form's home |
| data field | E2301 (a bare syntax error; #522) |
| type argument, tuple member | E2301 |
| helper parameter / declared return | ok |
| local binding's annotation | ok |

So the sentence is true of the boundary and false of a signature. Read in the codec section it is sound — nothing there decodes a helper's parameter. Read in `union-intersection`, which is where a reader goes to find out where the form may be written, it is a closed claim about the whole language and it is wrong.

## Why it matters

This is the same shape as #511: a section that teaches a type form states where it may be written, in its own words, and the statement is wider than what it can see. #550 took the position set out of four such sections and left this one, because the codec sentence reads as true when scoped to boundary positions and the call was worth making separately.

A reader who takes `only in a behavior's output` at face value writes out a named `data` for a helper that hands two cases to its caller inside one module, where the anonymous form is accepted and says the same thing. That is the opposite error from #511 — the rule is narrower than the language, so it costs vocabulary rather than a redesign — but it is still a rule nobody can check against the compiler by reading.

## Expected

State the restriction where it holds, and say what the position is rather than repeating the list:

- `union-intersection` says an anonymous union is refused at the boundary — a behavior's parameter (E1312) and a data field — and is admitted where a whole type is read (<<a-type-is-written-in-a-type-position>>). A behavior's output is the position it is *for*, which is a different sentence from the only one it may stand in.
- The codec sentence at `:3603` scopes itself: nothing that crosses a boundary is an anonymous union, so nothing reads one back. That is what it needs and it stays true.

Whether a helper signature *should* take one is a separate question from what the document says. If the answer is that it should not, the fix is a diagnostic, not a sentence — and then the sentence becomes true.

Found while closing #511 (PR #550), which left this deliberately.

kawasima:
Closes #553.

## What was wrong

Five sentences said an anonymous union appears only in a behavior's output. The compiler admits one in a helper's parameter, its declared return, a local binding's annotation, and a function type's parameter and result — and it is a type there, not a form that was let through.

The issue names two of the five. The other three are `delimiters` (`An anonymous | MUST NOT be written on input`), the `>->`-composition section (`an anonymous sum MUST NOT be written directly in a parameter type`, emphasised), and `Input and output` (`appears only in a behavior's output`). The shipped E1312 message carries the same vocabulary, and so did the E1313 entry's closing clause.

Two of the five were true only by a convention the document never states: it uses `input` for a behavior's and `parameter` for a helper's. One of them glossed `input` as `(parameter)`, which is the reading that carries a boundary rule into a signature.

## Measured

Every position, by compiling one:

| position | verdict |
|---|---|
| behavior parameter | E1312 |
| behavior output | ok |
| helper parameter | ok |
| helper declared return | ok |
| local binding annotation | ok |
| function type parameter | ok |
| function type result | ok |
| data field | E2301 |
| type argument | E2301 |
| tuple member | E2301 |
| newtype base | not a position — `data N = A \| B` declares a named sum |

This is exactly what `a-type-is-written-in-a-type-position` predicts, item for item. The canonical position set #550 took from the parser agrees with the compiler.

The three E2301s are not a rule about unions: no whole type is read there, which is why `List<T?>` is not written there either. The one deliberate refusal is E1312, produced by `SignatureBoundary` — the class that builds what a behavior's boundary carries — and reached from `PipelineSigs` for behaviors and composed pipelines only. Its spec home is under `External representation`, whose lead sentence scopes its five obligations to what crosses a decoder/encoder boundary. The compiler and the position rule already agreed; only the prose did not.

A union in a signature is fully checked: a `match` missing a member is E1201, an arm for a non-member is E1203, and a non-member argument is E1317. So the other reading of #553 — make the sentence true by adding a diagnostic — is a language change that removes a working type, not a fix. That belongs in its own issue with an ADR if it is wanted.

## What this does

The position list leaves `union-intersection`. Where the syntax is read as a type is `a-type-is-written-in-a-type-position`; what may cross is `a-parameter-names-one-type`; the union section composes the two and keeps the semantics. Being read in a position is not being admitted by it, which is the document's own distinction and now the one the union section is written on. `a-parameter-names-one-type` says *a behavior's* parameter and states in its own words that it is a rule about what crosses rather than about where the form may be written, so nothing rests on the `input` / `parameter` convention. The codec's premise comes from the representation rule rather than from the union section.

A behavior's output stays what the section is about — it is the position the form is *for*, which is a different sentence from the only one it may stand in.

## Held by a check

#550 noted that none of this was. The added test measures the positions as one map and compares the set, so a position dropping out reads as the difference it is. Its fourth case pins that a union in a signature is a checked type, which is the fact the narrower sentence would have to remove.

## Beyond the label

The last commit rewrites the shipped E1312 message in both catalogs. It is produced by `SignatureBoundary` and nowhere else, so `a behavior's parameter must name a single type` is true of every producer, and nothing asserts the text. It is the place a reader meets this rule most often. Revert that commit alone if the message should stay as it was.


kawasima:
Both fixed in d2e829d9.

**The sentence contained its own exception.** You are right, and it is the same mixing in a weaker form: `written wherever a whole type is read and is refused at one position` asserts a set that a behavior's parameter is a member of. The two questions have separate owners and now separate sentences:

> An anonymous union is read as a type wherever a whole type is read (`a-type-is-written-in-a-type-position`). Being read in a position is not being admitted by it, and one of those positions refuses the form: a behavior's parameter, which names what a caller has to write (`a-parameter-names-one-type`).

`Being read in a position is not being admitted by it` is the document's own distinction — `delimiters` already carries `Writing a type in a position is not being admitted by it`, three bullets under the position list. Saying it again where the form is taught costs a clause and puts the two-step in front of the reader who needs it.

`may be written` is gone from `delimiters` for the reason you give: it reads as a verdict rather than as what a reader does. The behavior signature section now says which of a signature's two positions takes one rather than saying where a union is written at all.

**The test named a set it does not measure.** Renamed to `everyNonBoundaryWholeTypePositionTakesOne`, with a note saying what the exception is and where it is held instead. The class note's `four` was a leftover from before the fifth site turned up; it says five.

The class name and the PR title still read `is written where a whole type is read`. As a title that is the ordinary sense and states no closed set, so I left it — say the word if you would rather it matched the sentence.

`mvn clean test` is green.

On the last point: agreed, and it is why the middle commit touches six places rather than five. Replacing the false sentences would have left `union-intersection` owning a position list, which is the thing that produced them.

kawasima:
Fixed in 64d6c2fe, taking the second of your two forms.

`a helper's parameter and every other position reading a whole type take one` was the position list again, written as a complement — and a behavior's parameter is one of the positions reading a whole type, so the clause admitted the form at the position the sentence above it refuses. The same shape as the one fixed in d2e829d9, sitting in the rule that was supposed to have handed the list away.

> This is a rule about what crosses, and not about where the form is read. A helper's parameter is not a boundary and takes one (`a-type-is-written-in-a-type-position`).

An example rather than a complement, for the reason you give: a complement is a second copy of the set, and this PR exists because two copies drift. The boundary rule now names one position, which is the one it is about.

The PR description said `Where the syntax may be written`; it says `is read as a type`, so the description and the document use the same vocabulary for whoever reads this later.

`mvn clean test` is green.

Three rounds, and each finding was the same sentence shape in a different place: a set stated together with its own exception. Worth noting for whoever writes the next rule here — the position set has one owner, and any sentence that reaches for "every position except" is writing a second one.

PR #559, opened 2026-08-09
A narrow type position answers about the form, not the delimiter

### LINKED ISSUE #522: A field type written as an anonymous union gets a bare syntax error

kawasima:
`data Holder = { t: A | B }` answered with ``I expected `}` here, but found `|` ``. The same misconception in a behavior's parameter answers with E1312, which names the union and writes out the repair.

## The cause

Neither wording was chosen. A behavior's parameter reads `retType`, which takes the `|` chain, so the union is built and `SignatureBoundary` refuses it by name. A data field reads `typeRef`, which does not, so the union is never built, nothing downstream has a form to report, and the answer is the token the field list wanted next.

**Which tier refuses a written form is decided by which production the position calls, and only a form that was read can be named.** That is one call apart, and it is not specific to unions — the `?` postfix takes the other route and shows what the route buys: `retType` reads it and `AstBuilder` answers `AnOptionalIsOnlyWrittenOnAFieldOrInTheCore`, so `Int?` on a helper parameter gets a real explanation while `List<Int?>` got ``I expected `>` here, but found `?` ``.

So #522 was one cell of five:

| | `A \| B` | `T?` |
|---|---|---|
| data field | delimiter | read (a field's own optional) |
| type argument | delimiter | delimiter |
| tuple member | delimiter | delimiter |

## What this does

The grammar is unchanged. Each forbidden continuation is recognized where it stands, ahead of the delimiter recovery that erases it — which is the third thing this parser already does, and the one `newtypeBase` uses when it looks ahead and answers `ANewtypeCannotWrapATuple` rather than letting the paren fall through.

The refusals are one helper reading the type a narrow position takes, told only whether a `?` belongs to the type there. A `|` belongs to none of them, so it is not a parameter, and no second copy of the position list lands in the parser.

```
Field `ticket` is written with an anonymous union, which is not a form a field's type is
written in. Declare a `data` over the cases and give the field that name.

An anonymous union is not written inside another type. Declare a `data` over the cases and
name that here.

`?` marks a data field's own optional and is not written inside another type. Write
`Option<T>` for the type, or put the `?` on the field that holds it.
```

`CstError` carries no hint by design — its javadoc says a hint reaching it would be a syntax error under whatever code nothing gave it — so the repair is in the message, as it is for the newtype refusals.

## Two rules, and neither is E1312's or E1402's

- **E2307** `an-anonymous-union-is-not-written-in-a-narrow-type-position`. Stated as a rule about the form: an anonymous union is not written as a data field's type, a type argument, or a tuple's member. The production's narrowness is the mechanism that follows, not the reason — an earlier draft said these positions "take one type" and a union "names more than one", which translates a grammar fact into type cardinality and contradicts the helper's parameter, which takes one type and takes a union. E1312 is a boundary restriction on a union that *was* read, which is a different rule reached from a different tier.
- **E2308** `an-optional-is-not-written-inside-another-type`. Not folded into E1402. E1402 refuses a `?` a *user model* writes outside a field, and the core may write one — `list.sou:193` writes `f: ('a) -> 'b?` on a function type's result. No core source writes a `?` inside `<>` or a tuple, and the production reads none there for anyone, so the reserved namespace is answered the same. `anOptionalInsideATypeIsNotACorePrivilege` is the test that tells the two rules apart.

## Recovery

The refused continuation is consumed so the declaration around it still parses. The tail after a refused `|` is read the way the position reads a type, not with a bare `typeRef` — read that way, a form forbidden in the position went unrecognized inside the recovery meant to prevent exactly that:

```
let f (xs: List<A | B?>) : Int = 1

before:  E2307, then a delimiter complaint and five more as the reading came apart
after:   E2307, E2308
```

`ARefusedContinuationIsConsumedTest` reads `CstParser`'s whole error list rather than the first diagnostic. `Compiler.compile` raises on the first, so a cascade behind it reads as a clean answer from there and as noise to the formatter and the language server, which take the list.

## The sweep

Every type position the specification names, crossed with every type form, asking only whether the answer is about the form. Thirteen positions × six forms.

```
                   before   after
bare delimiter        5        0
```

The five were exactly field × `|`, type argument × `|`, type argument × `?`, tuple member × `|`, tuple member × `?`; nothing else moved.

`NoTypePositionAnswersWithADelimiterTest` holds it, and holds two things about itself, because a sweep this wide can go quiet without saying so:

- every position must take at least one form. A source that never compiles is a template failing for a reason of its own and would answer the test whatever the parser did. Four positions were exactly that when they were first written out — a composition's declared output and its `exposing` form, and a helper's declared return and a local annotation, both given a value of another type. Forms carry a value now and the composition's stages infer an output a form can match.
- the number of refused cells is exact, so the delimiter assertion cannot come to hold by everything compiling. It says nothing about which cells refuse; a change to it is a change to what a position admits.

Taking the field recognition back out makes it fail and name the cell:

```
expected: <[]> but was: <[data field <- A | B  ->  E2301]>
```

## Interaction with #557

#557 landed while this was open and rewrote the union rules over the same three files. The rebase was clean and the conflict was in a test: `thePositionsThatRefuseOneSayWhichRefusalItIs` pinned the three narrow positions to E2301 to hold that they refuse at the grammar rather than at the boundary. That distinction is kept — E2307 is still a syntax error, E1312 is still the boundary — so its expectation is the specific code now, and its javadoc separates the shared mechanism from the two rules.

## Specification

`delimiters` gains the two rules (its stop list goes from five places to seven), and each code gets its section. The sentence #550 added is scoped rather than contradicted:

> What a position *reads* is what it builds as the type standing there; a form it does not read is still recognized where it is written, and named, so that what the author reached for is what they are answered about. Recognizing a form in order to refuse it is neither admitting it nor building it.

## Note on scope

#522's Expected asked for the E1312 wording at a field. That is narrower than the cause: it would have fixed one cell and left four, and it would have meant reading a union into a field to report it. What is here keeps each position's production as it was and closes the coverage hole instead. The issue text still describes the narrow version.

Closes #522.


kawasima:
All five addressed in `4a164fec`.

**1 (blocker).** You are right that this was the same mistake one level down: I replaced a claim about where a union belongs with a claim about how many types it is, and the same PR says a helper's parameter takes one type and takes a union. The rule is about the form now and the anchor says so — `an-anonymous-union-is-not-written-in-a-narrow-type-position`. The production's narrowness follows as the mechanism rather than standing as the reason.

**2 (blocker).** Widened to all thirteen. Adding them surfaced something the count could not: **four positions never compiled at all**, so they were answering the test whatever the parser did — the two just added, and a helper's declared return and a local annotation, which had been given a value of another type since they were first written. The forms carry a value now, the composition's stages infer an output a form can match, and a `neverCompiles` assertion keeps a template from going quiet again.

**3.** Reproduced, and worse than described. `List<A | B?>` raised **seven** diagnostics: the recognition, the delimiter complaint, then five more as the reading came apart and took the rest of the file with it. Fixed at the cause rather than in a test — the refused tail is read the way the position reads a type, so it is E2307 + E2308 and nothing else. `ARefusedContinuationIsConsumedTest` reads the parser's whole error list.

**4.** Exact. 36 of 78.

**5.** PR body rewritten against the current head. The `other fifty cells` line was wrong twice over — 55 then, and the sweep is 78 cells now.

`mvn clean test` green across all modules (4546 + 198 + 109 + 1).

kawasima:
Both done in `aefd9d30`.

**1.** Synced. The sample was the last place the cardinality explanation survived — rule, message, both catalogs and the section body had all been changed, and the sample is the one a reader sees before running anything. Its own section already explained the refusal by the production, so the two disagreed inside one section.

**2.** `narrowType`'s javadoc and the test's now describe what the production reads rather than what the position takes. `newtypeBase`'s "one type rather than a choice between several" is left alone — that is a different rule and not one this branch touches.

One thing worth naming: **nothing holds a specification sample against the catalog it quotes.** Every message key, title and rule anchor is checked, and the sample block under `[#eXXXX]` is not, which is why this drifted while the rest of the rename could not. I tried to measure whether other samples have drifted the same way and could not answer it — a literal sweep cannot separate "the spec quotes a different message under that code", which is by design, from "the spec quotes this one and it has drifted". The check that would answer it is per-code: each sample matched against the messages carrying that code. It does not exist.

`mvn clean test` green across all modules (4546 + 198 + 109 + 1).

PR #560, opened 2026-08-09
Every place is somewhere in the output

### LINKED ISSUE #555: The layout says where an annotation was written, not where a place is

kawasima:
`Layout` recorded where a `TokenDoc.At` was written, and that was read as where each place of the canonical form is. Sixty-three of the corpus's 3202 places had no answer, and a reader of a difference has to be able to point at any of them.

`At` is one way an output location is established. It is not what a location is.

## Three ways a location is arrived at, not four kinds of place

```
written   the extent of what the construct emitted     <- the annotation the layout already had
point     where a carried comment is inserted          <- left by the carrier pass
global    the canonical output                         <- the run's own answer
```

The extent of a place's members looked like a fourth and is not one. A product block writes its own braces and a `with` clause its own keyword, so what their members cover leaves those out — measured, a body's fields cover `[ a: Int\n    , b: Int]` where the body wrote `{ a: Int\n    , b: Int\n    }`. Both were missing an annotation and have one now, which is the whole of that category: two sites.

The anchors — a `data` declaration's `=`, a block's brace, a match's `with`, an `else`, and the name of an `example` or a `fake` — emit nothing and carry the comment written at the end of the line they open. The carrier pass leaves their position where the carrier stood, **before** whatever it carries: a comment written there is at the place, rather than the place being on the far side of the comment. That matters for the reader a witness is for, who wants where the comment attaches.

## Containment compares the ends

```
contains(outer, inner)  =  outer.start <= inner.start && inner.end <= outer.end
```

Not the characters as a set. A position is the empty set of them, and every interval contains the empty set — read that way a point past the end of the file is inside anything.

The corpus cannot show that, since every place it makes is where it belongs, so the rule is held on its own: `[10,20]` contains the positions 12, 10 and 20 and not 30 or 9.

## The check that was passing by leaving

`ALayoutSaysWhereEachPlaceWasWrittenTest.andAPlaceIsWrittenInsideTheOneThatHoldsIt` read

```java
if (above == null) {
    continue;
}
```

so a parent without an entry took its children out of the assertion — the fields of a product body were never compared against the body holding them. It is gone with the reason it skipped for, and every parent and child is compared.

## A correction to the issue's own analysis

The two `IDENT` places under `EXAMPLE_DEF` and `FAKE_DEF` were written up there as ordinary missing annotations, on the grounds that the other thirty-four `IDENT` places have one. They are anchors — `headerLine` makes them for the line an `example` opens with — and a kind agreeing with thirty-four others is not what settles which of the three a place is. They are located as positions like the rest of the anchors.

## What holds it

Every place of every corpus source has an extent; every parent contains its children; an anchor is a position; the file is the output; a comment an anchor carries is found from where the anchor is. Dropping the position the carrier pass leaves fails three of those.

4532 tests pass.

Closes #555


kawasima:
Real, and measured:

```
places() = 4, distinct by identity = 3
   SOURCE_FILE appears 2 times
```

`fileOf` records what it returns, and the construction had already called it, so adding the file's extent made a second entry for the same object.

Reading the file's place is `file()` now and recording it stays `fileOf`'s. Command and query apart, as you say — the name gave no sign that asking wrote anything.

The regression test is `andEachPlaceOnce`, and it is the assertion the rest of the file could not make: both copies are the same place, so both have an extent and both are inside whatever holds them. Every sweep over the places found an answer for each. That is the distinction #551 turned up in another form — every element having an answer, and the elements being the right ones, are two claims and only the first was held.

Putting `fileOf` back fails it by name.

The two comments are updated. `Correspondence.made` said every place is written somewhere in the document and `Place` said a place is somewhere the formatter writes something; both are the misreading this issue is about, since the file's place is written nowhere and neither is an anchor's.

4533 tests pass.

PR #562, opened 2026-08-09
A break opportunity belongs to the group whose decision settles it

### LINKED ISSUE #552: A break opportunity belongs to the group whose decision settles it
The conditional-layout rule answers about a group: it is written on one line, or down the page. A deviation from it is a source that did the other thing. Saying so means finding the group whose decision explains what the source wrote, and nothing relates the two.

## What is missing

`Layout` holds a decision per group (#545) and every break the layout realized. Neither is what a source break has to be matched against.

- `Doc.Line` — the boundary the layout may or may not break — has no identity. There is nothing to name.
- `GroupDecision` says the group's outcome and the column it was measured at, and nothing about which boundaries are inside it.
- Where the canonical form is flat, `Layout.breaks()` holds nothing at all for that boundary. A source that broke there is a deviation and there is no realized break to attribute it to.

That last one is why this is not "a realized break belongs to a group". Both directions have to be explained:

```
source broke            canonical flat      -> the group is flat, and the source broke inside it
source on one line      canonical broke     -> the group broke, and the source kept it whole
```

Only the second has a realized break.

## What it is instead

A **break opportunity** — a place the layout could have broken — belongs to the group whose decision settles it. Then a source gap is matched to the opportunity at that adjacency, and the opportunity names the decision.

```
source gap
   -> the canonical boundary at that adjacency
   -> the break opportunity there
   -> the group whose decision settles it
   -> GroupDecision
```

## Why it is the unit

Over this repository, 1629 differences are a line breaking or not breaking, spread over 565 distinct enclosing constructs — and even that overstates the decisions, because one group written down the page changes every member boundary under it. The rule answers about the group, so those are one witness each and not one per boundary.

Reading it the other way — a witness per differing adjacency — is the 438-differences-are-not-438-decisions finding of #444 in the place it does the most damage.

## Acceptance

- A break opportunity has an identity, and the layout says which group's decision settles each one.
- A group's decision can be reached from any opportunity inside it, and the opportunities inside a group can be reached from its decision.
- Both readings above are answerable, including where the canonical form is flat and the layout realized no break.
- An opportunity inside a nested group belongs to the innermost group that settles it, and a fixture with a group inside a group holds that.
- Nothing about the text changes.
- The relation is built by the run that lays the document out. Working out afterwards which group an adjacency falls inside — by position, or by walking the document again — is the reconstruction this is against.

Refs #444. After #551, before the conditional family of the witness.

kawasima:
The conditional-layout rule answers about a group: it is written on one line, or down the page. A deviation from it is a source that did the other thing, and saying which needs the group that decided. Nothing related the two.

A place the layout may break lowered to one of two shared objects, so there was nothing to name; and a `GroupDecision` said what a group came to and nothing about what was inside it.

## Why a realized break is the wrong thing to match against

Both readings have to be answerable:

```
source broke        canonical flat     -> the group is flat, and the source broke inside it
source on one line  canonical broke    -> the group broke, and the source kept it whole
```

Only the second has a break the layout wrote. Where the canonical form keeps a line whole there is nothing to point at, and the source has still departed from that group's decision.

So what a source gap is matched against is the **opportunity** — the place the layout may break — which is there either way and names the group whose decision settles it.

```
Opportunity(line, settledBy, broke)
```

The group is the innermost one holding it, which is the one the walk was inside when it reached there.

## Why it is the unit

Over this repository 1629 differences are a line breaking or not breaking, spread over 565 distinct enclosing constructs — and even that overstates the decisions, since one group written down the page changes every member boundary under it. A witness per differing adjacency is the 438-differences-are-not-438-decisions finding of #444 in the place it does most damage.

## What holds it

A group that fits still has its opportunities and none of them broke; the same group over the width breaks at the same one; an opportunity names the group that settled it and that group's decision is in the layout; two written the same way are two; and over every corpus source there is one opportunity per boundary the layout may break.

Three mutations fail: recording the enclosing group rather than the innermost one, recording only the opportunities that broke, and giving every place the layout may break one shared identity. The last is the weakness a group's decisions had until #545 read their identities and a nesting's until #548 — the same one, one type further down.

4543 tests pass.

Closes #552

### DIFF
--- a/souther-compiler/src/test/java/souther/compiler/fmt/ABreakOpportunityBelongsToAGroupTest.java
+++ b/souther-compiler/src/test/java/souther/compiler/fmt/ABreakOpportunityBelongsToAGroupTest.java
@@ -5,6 +5,7 @@ import org.junit.jupiter.api.Test;
 import java.util.List;
 
 import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertNotSame;
 import static org.junit.jupiter.api.Assertions.assertTrue;
 
 /**
@@ -92,5 +93,21 @@ class ABreakOpportunityBelongsToAGroupTest {
             case Doc.Concat c -> c.parts().stream().mapToLong(ABreakOpportunityBelongsToAGroupTest::linesIn).sum();
             default -> 0;
         };
+    }
+
+    /**
+     * Two places written the same way are two opportunities. Read by their shape they would be one,
+     * and a witness naming one of them would be naming both.
+     */
+    @Test
+    void andTwoWrittenTheSameWayAreTwo() {
+        Doc doc = Doc.group(Doc.concat(
+                Doc.text("ab"), Doc.line(), Doc.text("cd"), Doc.line(), Doc.text("ef")));
+
+        List<Opportunity> found = doc.layout(100).opportunities();
+
+        assertEquals(2, found.size());
+        assertNotSame(found.get(0).line(), found.get(1).line(),
+                "each is its own, and neither stands for the other");
     }
 }
--- a/souther-compiler/src/test/java/souther/compiler/fmt/ABreakOpportunityBelongsToAGroupTest.java
+++ b/souther-compiler/src/test/java/souther/compiler/fmt/ABreakOpportunityBelongsToAGroupTest.java
@@ -1,0 +1,96 @@
+package souther.compiler.fmt;
+
+import org.junit.jupiter.api.Test;
+
+import java.util.List;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertTrue;
+
+/**
+ * A place the layout may break belongs to the group whose decision settles it.
+ *
+ * <p>The conditional-layout rule answers about a group, and a deviation from it is a source that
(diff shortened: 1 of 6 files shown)


kawasima:
Taken, and measured:

```
a line outside any group:
   settledBy=null broke=true
corpus: 1840 opportunities, 0 with no group
```

Only the ones a group settles are opportunities now. Your reading of why is the one I'd keep: a boundary no group holds is not conditional at all — the outermost context breaks, so it always breaks, which is what a forced break is. An opportunity naming nobody is one a witness could not explain.

I tried refusing it first, on the grounds that this repository tends to fail closed on things nobody decided. That is wrong here and the suite said so: `EveryBoundaryIsAnsweredBeforeADocumentIsRenderedTest` builds a document by hand to check what a boundary beside a comment answers, and `Doc` lays out documents that are not the formatter's. Making the renderer reject one would be answering a question about the formatter inside the renderer.

So the property is stated where it is true. `andTheFormatterWritesNoneOfThem` holds that over every corpus source every boundary the layout may break is one a group settles — which is what says the filter drops nothing real, and is the assertion the count in `andOverASourceEveryOneIsThere` was making by accident.

And `andOneOutsideEveryGroupIsNotAnOpportunity` pins the case itself: the text still breaks there, and there is no opportunity.

4545 tests pass.

Issue #521, opened 2026-08-09
`adequacy: satisfied` is written when the boundary measure derived nothing


kawasima:
`adequacy: satisfied` is written when the boundary measure derived nothing.

## What happens

A model that is four length-and-format bounds:

```
remittance                                     measurement: complete
  validateInstruction    implemented   rows 14   pending 0
    signature   out specified 2/2  observed 2/2  verified 2/2
    partition   axes 0   single-axis 0/0
    boundary    0/0
      · not derivable: raw.bankCode
      · not derivable: raw.accountNumber
      · not derivable: raw.beneficiaryName
      · not derivable: raw.message
    branch      16/16
adequacy: satisfied
```

The behavior takes a `RawInstruction` of plain `String` fields; the bounds live one type-hop away, in the newtypes it builds. Every boundary behavior that validates raw input has this shape — the place bounds matter most. The rows at exactly 48/49, 20/21 and 1 — the rows carrying the model's whole risk — earn no credit, and deleting all of them changes the report not at all.

A second model hit the same thing from the other direction: its four bounds were temporal values compared against another input rather than against literals, so nothing was derived, and `--generate --boundaries` printed no rows.

## Why it is wrong

`example-adequacy` already defines the word for this case — a measure that could have found a gap could not be made is `undetermined`, not `satisfied`. The report is otherwise careful to keep `measurement` and `adequacy` apart, and this is the one place where "I could not measure" is dressed up as "I measured and found nothing".

The cost is that a team reads `satisfied` on a model of four numeric bounds and takes it to mean the bounds were checked.

## Expected

Where the boundary measure derived no axes at all, the verdict is `undetermined`, with the reason on the same line.

Found in round-5 MCP dogfooding (remittance validation and ticketing models).

kawasima:
Merged in #558. The boundary measure answers for itself now: where it derived nothing it says `not measured (no line was derived at any position)` and the verdict reads that answer, so a model whose bounds sit one type away from the position a behavior takes no longer comes out satisfied on a measurement nobody made.

Issue #517, opened 2026-08-09
A behavior whose output is not a sum loses the `signature` line without saying why


kawasima:
A behavior whose output is not a sum loses the `signature` line without saying why.

## What happens

```
behavior look : (t: T) -> Int
```

```
  look                     implemented   rows 1    pending 0
    partition   axes 0   single-axis 0/0
    boundary    1/1
      · not derivable: t.name
    branch      0/0
```

There is no `signature` line. A behavior whose output is a sum gets one:

```
    signature   out specified 3/3  observed 3/3  verified 3/3
```

`example-report-vocabulary` states the rule the report is meant to follow: a measure that has no number must say why. `branch` follows it — a behavior with no body reads `branch not applicable (this behavior has no body)`. `signature` just disappears.

## Why it matters

A reader compares two behaviors in one report and sees one measured on four lines and one on three, with nothing saying the fourth does not apply. The natural reading is that the measurement failed. It also hides a real fact: a behavior answering a bare primitive gets materially less scrutiny than one answering a sum, and the report is the place that should say so.

## Expected

```
    signature   not applicable (this behavior's output is not a sum)
```

Found in round-5 MCP dogfooding (bill-of-materials and maintenance contract models).

kawasima:
Merged in #558, with #521 — one defect in two places. A signature with no sum anywhere in it now reads `not applicable (this behavior's output is not a sum)`, and the reason is the measure's own rather than something the renderer counted back from an empty case set.

Issue #516, opened 2026-08-09
`run` refuses a file that imports another module and calls that module unknown


kawasima:
`run` refuses a file that imports another module and calls that module unknown.

## What happens

`catalog.sou` and `enrollment.sou` sit in one directory. `souther compile catalog.sou enrollment.sou -d out` exits 0, and so does the separate route through `-cp`. Then:

```
$ souther run enrollment.sou --behavior register --input '[]'
-- MODULE ERROR  E1504 ------------------enrollment.sou:11:1

11| import enrollment.catalog as Catalog (
    ^

Unknown module `enrollment.catalog`.
exit=1
```

The module is not unknown. It is in the same directory, it compiles, and its classes are in `out/`. The real rule is stated on `cli/run` — a file that imports another user module cannot be run this way — and the diagnostic does not say it.

## Why it matters

`run` takes one file, no second source, and no `-cp`. So for a model split across modules the behavior a reader most wants to execute has no execution path at all, and the message they get sends them looking for a missing file.

## Expected

Two things, in order of cheapness:

1. Say the actual rule at the point of refusal — that `run` takes a single self-contained file, and name the import that makes this one not self-contained.
2. Let `run` reach a compiled module: accept `-cp` as `compile` and `examples` already do, or accept the same file list `compile` takes.

Found in round-5 MCP dogfooding (course registration model).

kawasima:
Fixed by #567.

`run` now resolves an import of another user module against `-cp` and runs against that path's classes. The three symptoms were one wiring: source cardinality and module-resolution scope were the same decision in `Compilation.ofSource`, and `Runner` composed its own loader instead of the compilation's.

E1504 is unchanged — what it states is true of every compile, however it was started. The command line adds the remediation beside it, once per run and not under `--format json`.

Multiple sources for `run` is a separate step: it needs the entry module's meaning settled first.

PR #564, opened 2026-08-09
A lookup that came back with nothing means one thing

### LINKED ISSUE #513: An unresolved name in application position is reported as an arbitrary Java call
An unresolved name in application position is reported as an arbitrary Java call.

## What happens

```
module t exposing (Amount, f)

data Amount = Int
    invariant value >= 0

behavior f : (n: Int) -> Amount
    constructs Amount

let f (n) = Missing(n)
```

```
-- ARBITRARY JAVA CALL  E1401 --------------------t.sou:9:13

9| let f (n) = Missing(n)
               ^^^^^^^

`Missing` is not a behavior or a builtin.
Hint: Declare a behavior without a `let` and implement it from Java; arbitrary JVM calls are not allowed.
```

The same name in type position is diagnosed correctly:

```
-- NAMING ERROR  E1023
I cannot find a type named `Missing`.
```

## Why it matters

The commonest ways to reach this are a typo and a name left off an import list. Both are naming errors. What the author gets is a title about Java interop and a hint telling them to declare a behavior and implement it from Java — advice that leads away from the fix, and `doc_read {name: "e1401"}` is four lines about `java.time.LocalDate.now`, which does not cover it either.

The tell is that an unresolved name falls into the Java-call bucket in application position and into E1023 in type position, from the same cause.

## Expected

An unresolved name is E1023 wherever it stands, with the same "I cannot find" wording and, where a module in this compilation declares it, a hint to add it to the import list. Reserve E1401 for a name that resolves to something on the JVM.

Found in round-5 MCP dogfooding (course registration model); the name was one the author had left off an import list.

kawasima:
Closes #513.

## What was wrong

`Resolve` had two fallbacks answering the same question, differing in their last line:

```java
private ValueName valueName(...)  { ... : nothing(..., unknownIdentifier(written, bound)); }
private ValueName calledName(...) { ... : nothing(..., notCallable(call.name(), bound)); }
```

`unknownIdentifier` and `notCallable` both asked whether the spelling was a library member, both asked whether a library name had been written bare, both offered a near spelling. Then one said the name is not in scope (E1023) and the other said the name is a call to something on the JVM (E1401), chosen by whether the source had written arguments after it.

The observation is `lookup(name) == nothing`. What follows from it is that the name is not in scope. That the name is a JVM call does not follow from it, and nothing else established it — so a typo and a name left off an import list were reported under a title about Java interop, with a hint to declare a behavior and implement it from Java.

## E1401 had no producer that had looked

Measured rather than assumed. Its one emission site was that fallback, reached two ways, both of them "not found":

| written | before |
|---|---|
| `Missing(n)` — a bare name nothing declares | E1401 |
| `up.noSuch(n)` — a member a real module has not got | E1401 |
| `java.time.LocalDate.now(n)` — **the specification's own E1401 sample** | E1023 |

The documented example of the code does not produce it: `java` is not a namespace, so the chain reads as a field access and its root is a value nothing declares. Nothing in the compiler establishes that a name is on the JVM, so no site could have raised E1401 for the reason it names.

So it is retired rather than kept for a producer that cannot exist. The number was published, so it goes to `RetiredDiagnosticCode` with a `[#e1401-removed]` note; that list has no constructor a site can name, which holds "E1401 means the compiler checked" by construction rather than by a test. That arbitrary Java calls are out of scope stays a design decision (`no-arbitrary-jvm-calls`) — it was never a thing a compile could find.

The build had recorded this once already. `aKeyNamedForACodeIsEmittedUnderIt` notes two codes found reporting a rule other than the one they were filed under; one of the two was this.

## What replaces the second answer

A miss means one thing. How the lookup is made still differs — an application may reach what a bare name may not — and only the meaning of a miss is written once.

Cause and remediation are separate. The near spelling is carried beside the diagnosis; it was the remediation that had chosen the cause, since the advice to implement the name from Java is what made the report say the name was a JVM call.

The second commit adds what the suggestion cannot reach. A spelling suggestion offers names within reach, and a name left off an import list is not within reach — which is the whole problem, and was the real case #513 was found in:

```
I cannot find a value named `Amount` in scope.
Hint: `up` exposes `Amount`. Reach it as `up.Amount`, or name it in an import of `up`.
```

Only where exactly one module exposes the spelling, read off a sorted set. Two makes naming either one a guess, and a guess in a hint is worse than the silence it replaces.

## Held by a check

`AnUnresolvedNameIsTheSameDiagnosisWhereverItStandsTest` holds the property rather than the case: the primary diagnosis of an unresolved name is stable under syntactic embedding. `Missing`, `Missing(n)`, `Missing()`, `twice(Missing)`, `twice(Missing(n))`, `up.noSuch`, `up.noSuch(n)` — one map, one answer. Fixing the reported case alone is satisfied by a second answer written for the others, which is how this arrived.

Two existing tests held the old answer. Both wrote `someJavaMethod(a)`, a name that resolves to nothing; neither established anything about the JVM. What each was for survives — a callee no `depends on` clause names is still an unresolved name.

## Not in this change

Three open issues were checked for the same cause while investigating. #516 is the same defect one level up — `Unknown module` is reported where the producer established only that the module is not on this compilation's path, and the rule that actually refuses it (`run` takes a single self-contained file) is known to a different part of the code. #522 is the sibling but not the same: the parser genuinely observed `expected }, found |`, so it is under-observation rather than overclaim, and the two need different fixes. #521 and #517 are a pair of their own — a measure that found nothing reported as success, and a measure that does not apply omitted in silence.

### DIFF
--- a/docs/adr/0008-asymmetric-java-interop.md
+++ b/docs/adr/0008-asymmetric-java-interop.md
@@ -23,14 +23,17 @@ Souther -> arbitrary Java API              forbidden (only via a non-implemented
 ```
 
 Because Souther cannot reach into Java freely, it also ships no DB/HTTP/file APIs of its
-own — those live behind injected behaviors. Arbitrary JVM calls are a compile error
-(E1401), which points the modeler at declaring a behavior without a `let` and providing
-the implementation from Java.
+own — those live behind injected behaviors. The language offers no path to one: there is
+no syntax and no name-resolution route by which a body reaches a JVM API, so a spelling
+such as `java.time.LocalDate.now` is not a call the compiler recognises and refuses. It
+names no qualifier the language knows, and its root is reported as a name that is not in
+scope like any other. Reaching the outside world is declaring a behavior without a `let`
+and providing the implementation from Java.
 
 Keeping the arrow one-way is what preserves the closed construction paths (ADR-0002) and
 keeps every business branch in the output sum rather than hidden in a foreign call.
 
 ## References
 
-- Specification: `[#asymmetric-interop]`, `[#out-of-scope]`, `[#e1401]`
+- Specification: `[#asymmetric-interop]`, `[#out-of-scope]`
 - ADR-0002 (closed construction paths), ADR-0006 (outside-world via missing implementation), ADR-0007 (unmarked sum output)
--- a/docs/adr/0066-a-helper-is-typed-by-its-body.md
+++ b/docs/adr/0066-a-helper-is-typed-by-its-body.md
@@ -54,7 +54,7 @@ Making the rule uniform surfaced a defect it had been hiding. The shadowing guar
 
 Two more of the rule's own sentences were not what the code did. An arm determines its sibling only where it answers a value, and `unreachable` answers none — its `Never` fits every expectation because it states nothing, so taking it for an answer settled `let choose (b: Bool, v) = if b then v else unreachable "..."` as `Never` and let every call through. Which types say nothing about a value is one question now, asked in one place beside the empty collection's bottom. And a closure was read with no expectation and against no bindings, so `List.find((v) -> v, xs)` left the element open although the step's declared result says `v` answers a `Bool`. A call is solved in the stages `CallElaborator.applySignature` uses — result, then the arguments that are not closures, then the closures — which is also what lets a fold's seed say what its step accumulates.
 
-The report order is settled with it. A helper does not reach a behavior at all (E1401), and asking for the parameter's type first sent the author to write an annotation that changed nothing; the call is reported instead.
+The report order is settled with it. A helper does not reach a behavior at all (E1818), and asking for the parameter's type first sent the author to write an annotation that changed nothing; the call is reported instead.
 
 Measured again after the change: 2194 compiler tests and every module of souther-lang/examples pass, and no existing annotation had to move. What an author gains is that the same shapes their code already takes — a closure over a collection, an argument whose type a sibling argument fixes, an arm beside an arm — no longer ask for a type the body already knows (#302).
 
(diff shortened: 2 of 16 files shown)


kawasima:
All four fixed in f8cf8422, and the sweep turned up three more of the same kind.

**ADR-0008 was the merge blocker and you are right about why.** Replacing E1401 with E1023 there would have kept the false shape — the compiler does not identify an arbitrary JVM call and refuse it. It now says what is true:

> The language offers no path to one: there is no syntax and no name-resolution route by which a body reaches a JVM API, so a spelling such as `java.time.LocalDate.now` is not a call the compiler recognises and refuses. It names no qualifier the language knows, and its root is reported as a name that is not in scope like any other.

The decision is unchanged; what changed is that it no longer rests on a diagnostic that had nothing to look at. `[#e1401]` is out of References rather than repointed — a retirement note is not where the decision is stated.

**The retirement note stands on its own now**, naming the spelling instead of a sample the same commit deleted.

**`moduleOfQualifier`'s javadoc is back on `moduleOfQualifier`.** My insertion went between it and its method.

**`TypeChecker`'s two comments** say E1818 and E1023. The first was the more interesting one: it named E1401 for the rule that a helper does not reach a behavior, which is E1818 — a second code reporting the wrong rule, sitting beside the one being retired for exactly that.

Three more of the same, found by sweeping every mention rather than the ones the review listed:

- `specification.adoc` E1302's retirement note: *a Java method named in an expression is E1401*.
- `docs/adr/0066`: *A helper does not reach a behavior at all (E1401)* — the same wrong code as the `TypeChecker` comment.
- `messages.properties` and its Japanese catalog: a section comment naming E1401 for a section whose keys are E1818's.

What still mentions E1401 says what it was, in the past tense — two ADRs recording what the compiler did before a decision, and three tests recording the report their case used to get. Those stay, which is what the retired list is for: the number was published and a reader who met it still asks what it meant.

On the `TreeSet`: you are right that it cannot change the answer, since only 0/1/many is observed. I left it because the loop reads as "walk the modules in a fixed order" and a later change that returns the first match would otherwise be order-dependent without anything saying so.

`mvn clean test` is green.

Issue #574, opened 2026-08-10
An `example` row breaks like every other chain, and the specification writes it unlike one


kawasima:
## What happens

An `example` row that does not fit is broken with its connectors at the head of each line:

```
example submitTravelRequest
    | "within the ceiling it submits"
        : (Drafting { applicant = EmployeeId("e-001"), estimatedCost = Amount(50000) })
        -> Submitted
```

The specification writes the same row with the connectors ending the line before:

```
example submitTravelRequest
    | "within the ceiling it submits" :
        (Drafting { applicant = EmployeeId("e-001"), estimatedCost = Amount(50000) })
            -> Submitted
```

## Why it is unsettled rather than wrong

The formatter's form is not this construct's own. `Formatter.exampleRow` builds a row as a chain —
description, then the input behind a `:`, then the expected value behind a `->` — and a chain is
written by one rule: broken, each part starts a line one indent in and the connector leads it, so
what joins two parts is visible at the front of the second. A pipeline's `>->`, a value pipe's `|>`,
a union's `|` and a match arm's `->` are all written that way. The `example` row is not an exception
the formatter made; it is the rule applied.

So the specification's rows are the exception, and they are also inconsistent among themselves — the
sample above indents the `->` one level further than the `:`, which nothing else in the document
does.

What is worth arguing is whether a row should be a chain at all. A row has three parts a reader is
meant to see as three, and the trailing-connector form marks the end of each part rather than the
start of the next. Neither form hides anything; they differ in which edge the eye finds.

## What to decide

Either the rows in the specification are rewritten to the chain form and this closes, or an
`example` row gets a rule of its own — and then that rule has to be stated, because a construct that
breaks unlike every other chain is exactly the kind of thing #518 is about being written down rather
than reconstructed by running the tool.

If a row does get its own rule, `Formatter.exampleRow` is where it goes, and the regression test has
to hold the rule against every row shape (with a description and without, with a `with` clause,
with a departure as the expected value), not one example.

## What it holds up

Two of the specification's samples are held back from the `[,souther]` mark that #518 introduced:
`specification.adoc:2458, 2506`.

Named in #518 as out of scope; this is where it is decided.

kawasima:
Fixed by #583 (merged into develop as a741a1e).

Issue #573, opened 2026-08-10
A `guard`'s departure is written wherever its condition happened to end


kawasima:
## What happens

`guard` writes its condition and its departure in a row, so where the departure ends up is wherever
the condition finished:

```
let build (rows) = {
    guard List.length(rows) >= 1 else Small(1)
    Lines(rows)
}

let wide (rows) = {
    guard List.allDistinctBy(x -> x, rows)
        && List.length(rows) >= 1
        && List.length(rows) < 100 else Small(1)
    Lines(rows)
}
```

In the second, `else Small(1)` is at the end of the condition's last continuation line. What the
guard is for — the value it leaves with — is the least visible thing in it, and it reads as the tail
of the third conjunct.

The named-clause form already writes the other shape:

```
    guard Lines(rows) as items else
        | nonEmpty -> Small(1)
        | unique -> Small(2)
```

There the `else` ends the guard's own line and the departures are under it. So one construct is
written two ways, and which one you get depends on whether the departure is one value or a list of
clauses.

Every `guard` in the specification writes `else` on a line of its own:

```
    guard List.length(rows) >= 1
        else NoRows
```

## Why it is unsettled rather than wrong

`Formatter.guardStmt` (`souther-fmt/src/main/java/souther/compiler/fmt/Formatter.java:1979`) writes
`GAP` before the `else`, which is `Break.NEVER`: the layout cannot open a line there whatever the
width. Nothing decided that the departure may not have its own line — it is what writing the parts
in a row gives, the same way the order two lists break in was what nesting them gave (#565).

But the specification's shape is not obviously the answer either. `guard c else X` on one line reads
well and is what most guards are, and a rule that always broke it would spend a line on every short
guard in a body.

## What to decide

Whether `else` may open a line, and if so when:

- always, which is what the specification's samples do;
- only where the guard does not fit on one line, which would leave the short form alone and give the
  long one `else Small(1)` on a line of its own under the condition;
- never, which is what happens today, and then the specification's samples are the ones to change.

The second is the shape the named-clause form already takes, so it is also the one that would leave
`guard` with a single shape rather than two.

Whatever is chosen, `Spacing` and the break rules published in `cli/commands` have to say it, and a
test has to hold the rule rather than one expected string (#565 has the pattern).

## What it holds up

Seven of the specification's samples are held back from the `[,souther]` mark that #518 introduced,
because canonicalizing them today would write a form this may reverse:
[...]

Issue #566, opened 2026-08-09
A block body's paragraph breaks are deleted rather than normalized


kawasima:
## What happens

Every blank line inside a block body is deleted.

```
let f (a: Int) = {              let f (a: Int) = {
    let x = a + 1                   let x = a + 1
                            ->      // --- second section ---
    // --- second section ---       let y = x + 2
    let y = x + 2                   y + 3
                                }
    y + 3
}
```

`Formatter.block` (`souther-fmt/src/main/java/souther/compiler/fmt/Formatter.java:1766`) writes `Break.ALWAYS` between statements and has no path that writes a `HARD_GAP`. At top level the opposite holds unconditionally: `Formatter.blankBetween` puts one blank line between every pair of items. Neither reads what the author wrote. The CST is lossless and holds the blank lines; the formatter reads `WHITESPACE` only to answer whether a line ended.

## What to decide

Preserve or delete is not the only pair. A third rule keeps the canonical form single-valued and keeps the author's paragraphs:

```
between two statements of a block:
  0 blank lines  -> 0
  1 or more      -> 1
```

The presence of a break is the author's; the count is normalized. So

```
validate request

authorize employee

issue ticket
```

survives, and a body written with three blank lines between steps formats to one, and formatting is still idempotent and has one answer per input.

This is what the round-5 report asked for when it said that removing an author's paragraph break takes away the only structural signal a block has — four of six models raised it independently.

Deciding this is separate from #518, which is about the canonical form not being a contract anything is checked against. It is also separate from the comment attachment defect fixed there: that one changes what a comment is about, this one is a layout policy.

Worth settling at the same time: whether the same 0/1 rule replaces the unconditional blank line `blankBetween` writes at top level, or whether top level keeps its forced separator.

kawasima:
The specification's samples now carry a `[,souther]` mark meaning "a complete Souther fragment in
canonical form", and a test holds every marked one against the formatter (#518). Three samples are
held back from that mark by this decision:

- specification.adoc:1968
- specification.adoc:2105
- specification.adoc:3118

Each writes a blank line between the steps of a block body. Canonicalizing them today means deleting
that blank, which is the structure this issue is about keeping — the deleted form stays canonical
under the 0/1 rule too, so nothing would force it back afterwards.

The forced blank line between top-level items does not have that problem: a sample written with the
blank is canonical under the current rule and under the 0/1 rule alike, so the 44 samples
[...]

Issue #565, opened 2026-08-09
Which part of a behavior signature breaks first is decided by group nesting, not by a rule


kawasima:
## What happens

A behavior signature that does not fit breaks its output union first and keeps its parameter list on one line:

```
behavior preApprove : (request: AwaitingPreApproval, approverId: EmployeeId) -> PreApproved
    | NotAuthorized
    constructs PreApproved, NotAuthorized
```

Give it a parameter list that alone will not fit and it writes the other shape, which is the one the specification writes everywhere:

```
behavior preApprove : (
    requestForApproval: AwaitingPreApproval,
    approverEmployeeId: EmployeeId,
    submittedOn: DateTime
) -> PreApproved | NotAuthorized
```

So the formatter can write both. Which one it writes is not a rule anyone stated. The output union is the outer group and the parameter list an inner one, so the union breaks first and the parameters break only when they still do not fit by themselves. The layout falls out of how the document happens to be nested.

## Why that is worth fixing beyond the output

The first shape reads wrongly. `-> PreApproved` puts one member of a symmetric union on the signature line, attached to the arrow, and the rest below it. A reader takes a signature as a name, its inputs, and its output; this splits the output across the boundary between the second and third.

The deeper point is that a nesting order is deciding a readability question. Every construct that holds two breakable lists inherits whatever order it was built in, and the answer changes when the construction is rearranged — a rearrangement that no test would call a change.

## What to decide

The priority itself:

```
break the parameter list before the result union
```

and, having decided it, whether the layout engine carries a break priority (or cost) that a construct declares, or whether this one signature is arranged to nest the other way.

Arranging the nesting gets the output right and leaves the next construct in the same position. Carrying a priority is the larger change and makes the decision a value the formatter holds rather than a shape it happens to produce.

Whichever is chosen, the regression test has to assert the priority rule and not the string that came out, so a later rearrangement fails it rather than silently reversing it.

Split out of #518, which found this but is about the canonical form having no contract, not about what the canonical form should be.

kawasima:
The specification's samples now carry a `[,souther]` mark meaning "a complete Souther fragment in
canonical form", and a test holds every marked one against the formatter (#518). Five samples are
held back from that mark by this decision, because canonicalizing them today would write the shape
this issue may reverse:

- specification.adoc:1673 — the `behavior` topic's own signature
- specification.adoc:1711
- specification.adoc:1794
- specification.adoc:1935
- specification.adoc:2105

[...]

PR #578, opened 2026-08-10
The canonical form is a contract, and three layout questions it exposed

kawasima:
`souther fmt` writes a canonical form and nothing in this repository was required to be in it — not
the standard library, not the specification's samples, and `fmt --check` ran in no CI job and no
build step. Fifteen of the sixteen `.sou` files here were not in it, and 57 of the 83 specification
samples that parse as whole files were not either.

That is what #518 turned out to be about, one level under the round-5 report's "the formatter
contradicts the documentation". Separating what that report had as one finding gave four things, and
this is three of them; the fourth is layout policy and is four decisions of its own, two settled
here and two still open.

## Three layout questions answered by rule

Each was decided by how the document happened to be built rather than by anything written down.

**What a comment is about** was read from whether it started a line and nothing else, so a note
written under a declaration and cut off from the next one by a blank line came back describing that
next declaration — a change of what the comment is about, which the text does not show. `attach`
counts the newlines now, and `Carrier` gains `BELOW`: a comment below what it describes is neither of
the other two, and the set that was short one was the asymmetric one.

**Which part of a behavior signature gives way** (#565) was the output union, because it was the
outer group. It left one member of a symmetric union on the signature line attached to the arrow with
the rest below. The parameters give way first now, and the union breaks only where the line the `)`
left it on is still too narrow.

**A blank line between two members** (#566) was written between every pair of top-level items and
deleted inside every block, both unconditionally. Where a paragraph break is is the author's and how
big it is is not: none stays none, any number comes back as one. A module header and the last import
keep a blank line under them whatever the source had, which is what gofmt does.

## The contract

The ten standard-library modules are canonical and `EverySourceThisShipsIsInCanonicalFormTest` reads
them through `Reserved.MODULES`, so a module added there is covered without being added here.
`souther api --source` prints these to a reader, and shipped source the product's own formatter
rejects is the product contradicting itself.

A specification sample says for itself whether it is Souther: `[,souther]` is a claim that the block
is a whole fragment in canonical form, and a test refutes it if it is not. Membership is what the
document declares and not what happens to parse, so a parser that grows does not pull samples into
the check that nobody marked. Seventy-four are marked; nine are not, each waiting on #573 or #574.

The rules themselves are published in `cli/commands`, which had said a canonical form exists and
nothing about what it is.

The `delimiters` section claimed a leading-comma layout for every sample in the document. The split
is by construct and is the one the formatter already writes; the document's own multi-line record
literal is a counterexample to its own sentence, and that sentence is what six models read before
writing code `fmt --check` then rejected. It is scoped to the construct now.

The bench corpus is left out: it is input to the parser and the compiler rather than model code.

## Not here

`guard … else` (#573) and an `example` row's connectors (#574) are the two layout decisions still
open. The nine unmarked samples are waiting on them.

Refs #518 #565 #566

### DIFF
--- a/souther-compiler/src/main/resources/META-INF/souther-docs/cli/commands.md
+++ b/souther-compiler/src/main/resources/META-INF/souther-docs/cli/commands.md
@@ -49,6 +49,60 @@ canonical form is what the verdict is taken against, so a failing build says wha
 than only which file did, and nothing has to be formatted a second time locally to read it. A file
 already in canonical form is not mentioned.
 
+### What the canonical form is
+
+There is one form per program and the formatter derives it from the tree, so nothing about the
+layout you write is kept. These are the rules it derives it by, so that what the tool will write can
+be read here rather than reconstructed from having run it.
+
+Lines are at most 100 columns and a nesting level is 4 spaces. A line longer than 100 columns is one
+whose content holds nowhere to break — a long string, a single long name, a nesting deep enough that
+the indent takes the width. That is not an exemption for some constructs: everything that has a
+boundary to break at breaks at it, and a 130-column `unreachable` message survives only because a
+string literal has none.
+
+Where a blank line goes is yours; how big it is is not. Between two top-level items, and between two
+steps of a block, one blank line comes back as one and three come back as one, and none stays none —
+so a run of related one-line declarations stays a run and a body written as paragraphs keeps them.
+Two places are the file's rather than yours and get a blank line whatever you wrote: under a module
+header, and under the last import. Nothing else is kept: a blank line inside a construct, between
+the fields of a `data` or the arms of a `match`, is not a paragraph break and does not survive.
+
+A `data` product body writes each field on a line of its own, opening the line with the `{` for the
+first and with a `,` for the rest, and closing with a `}` at the body's own indent. It is written
+that way whenever it has a field, whether or not it would fit on one line; a body with no fields is
+`{}`.
+
+```
+data Employee =
+    { id: EmployeeId
+    , rank: Rank
+    }
+```
+
+Everything else written between brackets — a record literal, an argument list, a parameter list, a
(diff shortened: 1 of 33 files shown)


kawasima:
All three are real, and all three are the same mistake: the rules were written into `cli/commands`
from the formatter as it stood before #565 and #566 landed, and were not read again afterwards.
Fixed in dc3f8c13.

**The signature rule.** It was not published at all, and the bracketed-list rule said what it
contradicts. It is written as a rule now, and the bracketed-list rule names it as the exception:

> A behavior signature lays its inputs and its output out together, so neither is decided on its
> own. Where the whole signature fits, it is one line. Where it does not, the parameter list breaks
> first — one parameter to a line, with the `)` opening the line the output is written on — and the
> output union breaks only where it still does not fit that line. A parameter list short enough for
> a line of its own is therefore written down the page anyway when what follows the arrow leaves it
> no room …

with the three outcomes as an example. Each of the examples in that section is now run through
`fmt --check`.

**One form per program.** You are right that the policy is not what needs changing. The opening now
names the two things a reader writes that are structure rather than layout, and says the form is one
per program taken with those two:

> The formatter derives the layout rather than keeping yours. Two of the things you wrote are
> structure and not layout, and it keeps those: where a paragraph break is, and what a comment was
> written about. Everything else is derived. So there is one form per program taken with those two —
> write the same program with the same paragraphs and the same comments and the same text comes
> back, however differently the rest of it was laid out.

**`examples for`.** Enumeration gap, as you say. It reads "under the header a file opens with — a
`module` or an `examples for` —" now, so the sentence and `blankBetween` name the same set.

On the `O(n²)` in `openedByABlankLine`: agreed on both counts. Both sequences are in source order, so
one cursor makes it linear, and it is not worth holding this. Left as it is rather than changed
without a measurement to point at.

`mvn clean verify` is green (4899) and the three examples added to the document are canonical.

PR #581, opened 2026-08-10
A canonical form is computed, and the deviations from it are named

kawasima:
`souther fmt --check` answered with a diff, which says which lines are not the canonical form's and cannot say why any of them is there. It now answers with the decisions:

```
m.sou:3:12: a construct written down the page writes its members one to a line
    the canonical form: a line ends here; this source: it does not
m.sou:3:25: what goes between two tokens on a line
    the canonical form: nothing; this source: ` `
```

This is #444's steps 2, 4, 5 and 6 — the rules of canonicalization as values, a witness per family, a repair that composes them, and what the command prints. Step 1 (#551) and step 3 (#552) are already in.

## What it comes to

Measured over the sixteen `.sou` files of this repository against the canonical form as it stands on `develop`, source against canonical, one comparison per adjacency of the two:

| | before | after repairing what the rules say |
| --- | --- | --- |
| all differences | 1804 | **0** |
| files not in canonical form | 6 of 16 | 1 of 16 |

The one file that is still not written back exactly differs by the alignment of a comment's continuation line. Every file keeps the comments it started with.

Two rounds, not one: a rule that reads another's result answers about lines the first has not written yet, so the repair is applied until it stops changing the text.

## The rules as values

`Obligation` names the five things a break written whatever the width can be discharging. `TokenDoc.Gap` holds it exactly where the policy is `ALWAYS` — the other two policies say which rule answers them by their policy alone — and there is no constant for a forced boundary any more, since one would be a boundary that always breaks and says nothing about why. `Doc.Ref` carries it to the layout and `Newline.Cause` says of each break either the obligation or the opportunity a group settled, with nothing left saying nothing.

Counted over the corpus, the obligations partition the forced breaks exactly as the construction sites do: 1203 member pairs, 128 brackets, 16 file ends, 533 separations, 1094 comments.

The spacing rule's domain was taken from the boundaries the formatter writes, which is the path being validated: a boundary the canonical form always breaks is one it never asks about, so no row was ever needed and none was missed. The candidates now come from the parser — each corpus source, and its canonical form with one forced break closed up into a space, which is a source that has those two tokens on a line if the language admits it. Seventeen adjacencies come out that the rule has no row for, and what is held of them is that the canonical form breaks every boundary it has for each — the domain total, the expectation partial, and the break decision above it saying where.

## A witness per family, and its unit

Seven families, and the unit is the rule's rather than the difference's:

| rule | unit | over the corpus |
| --- | --- | --- |
| what goes between two tokens | the boundary | 699 |
| one blank line separates two items | the pair of items | 146 |
| one level deeper is one indent further in | the pair of levels | 17 decisions over 150 differing lines |
| a construct too wide breaks | the group | 358 |
| the forced obligations | the boundary | 175 |
| a comment's line, and what is beside it | the comment | — |

A record literal written down the page moves every member boundary it holds and was one decision. A source that indents a construct by two writes every line inside it in the wrong column, and what it got wrong is one step — so the indentation witness carries a step and not a column, and a level moved with the one above it keeps its step and has nothing against it.

The conditional rule is matched against the opportunities rather than against the breaks the layout wrote. Where the canonical form keeps a group whole there is no break of its own to point at, and a source that broke inside it has still departed from the decision; a report built from realized breaks would have nothing to say about the commonest deviation there is.

## Repair composes and writes once

A witness owns no patch. Every expectation becomes what it says about a stretch of the source, the stretches are held against each other, and the text is written once. A stretch inside another's is dropped where the wider one already writes what the narrower asks for — an indent and the break that opens the line it is on — and refused where it does not.

That refusal found a defect. Widening the spacing rule to own the tight boundaries a source broke let it answer about a gap holding a comment, and the repair writes the canonical form's text over the whole of a gap: over this corpus it deleted comments. The comment rule and the spacing rule answered about the same characters and did not agree, which is the conflict this issue's hypothesis says the model does not have and the repair is written to find rather than to settle. `everySourceKeepsItsComments` holds it now.

## What the checks say

**B** is a fixture per family: a source deviating at one unit gives that witness and no other, which also falsifies the applicability edges — a source that broke somewhere the canonical form does not produces no indentation witness, because the line it wrote has no level to be at.

**C** is written out. The first fixture is the one this issue opened with: the padding in front of a `=` and the record body opening down the page are two decisions on one source line.

**D** is written where the rules are values rather than at the end.

**A** holds for the corpus except the three comment placements — and not in the form it was stated in. Asked of a source whose breaks are still wrong, the indentation rule answers about lines that will not exist, so `E` respects the order the rules depend on each other in and the repair is applied until it stops changing the text. Two rounds over this corpus, never more.

## Six places a rule answered about less than its unit

Found by asking what the differences a repair leaves have no witness for. At one point that was 163 differences and ten witnesses, so what was missing was rules and not repairs.

- A group's decision is about **all** of its opportunities; a source that broke some and ran the rest together was being called the same layout.
- A boundary no group settles is the spacing rule's whatever the source wrote there, a line break included.
- A group broken by what it holds is still the conditional rule's. The forced decision settles why it is not flat, not whether it is.
- A source need not have been consistent about a level; the witness carries the steps it wrote.
- Not every line the canonical form opens is a place's — a closing bracket takes a line of its own and is written by the construct.
- A forced rule says how many lines end at a boundary, not whether one does.

## Also

`fmt --check` answered an ordinary source with an internal compiler error: three families hold the two token streams side by side, which the canonicalization that rewrites a definition's lambda makes impossible, and the refusal reached the command. A family that cannot answer is now left out and the report says it is not whole.

The file's document was a bare `concat`, so the adjacency between two top-level items had no construct joining it — the one adjacency a reader closes up by writing two items on a line. It is a `TokenDoc.Node` now; every boundary there always breaks, so nothing the formatter writes changes.

`Repair` switches over `Witness` without a default, so a rule whose expectation is added and not composed does not build. That is how the comment families were caught when they were added.

## What is left

- Rules 1 and 2 — a trailing comma dropped, a first arm's `|` written. No difference over this corpus, and no decision either: the formatter chooses whether to write an optional token and records nothing.
- Rule 3 and the token correspondence under it. Four families hold the two token streams side by side, which the lambda rewrite makes impossible, and that is the only thing `whole` is ever false for.
- One comment continuation line.

Refs #444


kawasima:
All three hold, and all three are fixed. Two of them I reproduced before changing anything.

## 1 — the file's own break reached neither the report nor the repair

Reproduced. A file that is canonical but for its final newline was answered with a diff and "no rule accounts for yet", when the rule was a value and the witness had been made.

The cause is the one you name: what turns a witness into a stretch of source pairs a boundary with the two tokens it stands between, and the file's break has no token on its far side. It is written now, from the last code token to the end of the text, and a fixture holds both directions — none, and one too many — and that repairing it writes the canonical form.

```
noeol.sou:3:17: a file ends with one newline
    the canonical form: one line ends here; this source: no line ends here
```

## 3 — the forced witness kept the answer it had before the rule changed

Reproduced, and it printed a false statement:

```
blank.sou:5:18: a construct written down the page writes its members one to a line
    the canonical form: a line ends here; this source: it does not
```

A line does end there. `Forced` carries both answers now and the record says what the rule says:

```
    the canonical form: one line ends here; this source: two do
```

You are right that this is the one that matters most. The point of the feature is the two answers, so a repair that is correct while the answer is untrue is not the feature.

## 2 — a family that cannot answer said so with the exception a broken invariant throws

`Witnesses.NoCorrespondence` is a kind of its own, thrown only by the four families that hold the two token streams side by side. `Deviations` catches that and nothing else. Everything else a family throws reaches the command, which says so for that file and shows the difference rather than judging it as though the rules had been asked. A test pins that the two kinds are unrelated.

That separation earned itself immediately. Merging `develop` in, the indentation rule threw its own invariant — a level written at two columns — because a blank line's break writes no indent and its zero was read as one. Caught by the wider net that would have been a report saying the source was merely unusual. `Newline` carries whether the break indents now.

## The conflicts

Resolved by merging `develop` in once rather than replaying fifteen commits over it, since the residual measurement and the suite are a stronger oracle for the integration than a conflict-free replay would be.

`develop` and this branch both gave `TokenDoc.Gap` a second component, and they say different things — why a break is written, and whether it indents the line it opens — so it holds both. `BLANK_LINE` becomes `TokenDoc.blank(Obligation)`: a break that leaves a line empty is written for a reason like any other, the separation rule's under a header and the members rule's between two steps of a block.

What the separation rule says has changed under this branch — a paragraph break is the author's and the canonical form keeps it — and the witnesses needed nothing, since they read the decision rather than a rule written down beside it. The fixtures asserted the old rule and say the new one now, and so does the sentence `fmt --check` prints.

## Where it stands

Over the sixteen `.sou` files, against the merged canonical form: **1804 differences, and 0 after repairing what the rules say.** Every file repairs to its canonical form exactly except one comment continuation line, and every file keeps the comments it started with. Full build green.

What is still open is unchanged: rules 1 and 2 have no decision behind them yet, and rule 3's token correspondence is the only thing `whole` is ever false for.

kawasima:
Both cases held, and both are fixed. I reproduced them before changing anything.

```
eofA.sou: a file ends with one newline
    the canonical form: one line ends here; this source: no line ends here
    ... and more that no rule accounts for yet:      <- A
eofB.sou: internal compiler error: StringIndexOutOfBoundsException: Index -1   <- B
```

## Where each stretch comes from

Your call on the responsibility boundary is the one I took: the EOF newline is the file's. A comment that ends the file is written above nothing, so the distance `CommentAbove` answers about is not there, and it now says nothing about such a comment.

**A.** The file's rule was projected onto the text from the last *code* token, which holds the comment where one ends the file. It is taken from the last thing written now — comment or token — so the comment stays where it is and only what follows it is written.

**B.** With that, `CommentAbove` no longer computes a stretch from a newline that is not there, and the minus-one offset that reached the projection is gone.

## What the two boundaries past the last token are

There are two, and answering for both against one stretch is what said the same newline twice when I first split them. They are held against two now: the lines above the comment by the rule that put it on a line of its own, and the lines after it by the file's.

```
eofA.sou:3:31: a file ends with one newline
    the canonical form: one line ends here; this source: no line ends here

eofB.sou:3:17: nothing shares a comment's line
    the canonical form: one line ends here; this source: two do
eofB.sou:5:21: a file ends with one newline
    the canonical form: one line ends here; this source: no line ends here
```

Both `whole`, no diff appended, no exception.

## The fixtures

`AFileThatEndsWithACommentStillEndsWithOneNewlineTest`, over both of your sources, holds that `Deviations.of` does not throw, that everything the source deviates by is named, that exactly one rule answers for how the file ends, that the repair keeps every comment, and that what it writes is the canonical form.

## The description

Updated to the merged tree: 1804 differences against the sources and **0 after repairing what the rules say**, 1 of 16 files not written back exactly rather than 6. You are right that a stale measurement is worse here than stale prose — the measurement is the argument.

PR #594, opened 2026-08-10
Whether a size can be built is answered by what builds it

### LINKED ISSUE #593: A boundary on a measure of a value gets no row, and the rows offered already meet it
A boundary drawn on a measure of a value — `String.length(n) = 1` from a `min` invariant — is
reported as an edge nothing composes a value at, without a candidate being tried. The rows the same
generation offers for other obligations are already standing on that edge.

## What happens

```souther
module bm

data Name = String

data Size = Int
    invariant value >= 1

data Tag = Big | Small

behavior label : (n: Name, s: Size) -> Tag

let label (n, s) = if s.value >= 5 then Big else Small

example label
    | (Name("abc"), Size(9)) -> Big
```

`souther examples bm.sou --generate --boundaries`:

```
    boundary    0/4
      · no row is at label/String.length(n) = 1 (invariant Name (min))
      · no row is at label/s = 1 (invariant Size (min))
      · no row is at label/s = 5 (guard@13:20)
      · no row is at label/s = 4 (guard@13:20)

// generated by `souther examples --generate`: 4 rows to fill what nothing covers.
// example label
//     | "s=1 <= x < 5" : (Name("x"), Size(1)) -> <?>
//     | "s = 1" : (Name("x"), Size(1)) -> <?>
//     | "s = 5" : (Name("x"), Size(5)) -> <?>
//     | "s = 4" : (Name("x"), Size(4)) -> <?>
// no row for `String.length(n) = 1` in `label`: nothing here composes a value at that edge, which does not make the edge unwritable
```

Three numeric edges get a row. The fourth does not, and `Name("x")` — written into all three of
those rows — is a string of length one, which is the value that edge asks for.

## What it costs

Measured on the two bench corpora, with the CLI, since a compile does not run the measure:

| | uncovered boundaries | of which a measure of a value | boundary rows offered |
|---|---|---|---|
| `corpus/crm` (43 behaviors) | 25 | 18 | 3 |
| `corpus/issuetracker` | 8 | 8 | 0 |

Every one of the 26 is `String.length(x) = 1`, from `invariant String.length(value) >= 1` on
`CompanyName`, `PersonName`, `UserId`, `CampaignName`, `IssueId`, `Label`. None of those types
carries another clause, so `"x"` meets each of them, and the generator writes exactly that value
into the partition rows of the same run.

The issuetracker corpus reports `boundary 0/8` and `--boundaries` offers nothing at all.

## Why

`Generator.probe` (`souther-compiler/src/main/java/souther/compiler/partition/Generator.java:298`)
returns before it composes anything, for every axis whose term is not the value at the position:

```java
// A line on something taken of the value is not met by writing that number at the position:
// a length of five is a string of five characters, and nothing here composes one (#528).
if (!(axis.term() instanceof NumericTerm.ValueOf)) {
    return new BoundaryAttempt.Unresolved(new UnresolvedCombination(List.of(label),
            UnresolvedCombination.Reason.NOTHING_COMPOSES_ONE));
}
```

`NumericTerm` has two cases and this admits one of them. Every `SizeOf` axis — the measures #530
made boundary axes out of — leaves here. The reason is accurate about what this method does and not
about what can be written: the report says the edge is not unwritable, and it is right, and no row
is offered for it anyway.

## What a fix has to hold

- A boundary whose axis term is `NumericTerm.SizeOf` composes a candidate of that size at the
  position and puts it through the same `CandidateCheck` every other candidate goes through.
- A candidate refused at construction — a type carrying both a `String.matches` clause and a length
  clause, where the composed witness meets the second and not the first — is reported as
  `ALL_CANDIDATES_REJECTED`. `NOTHING_COMPOSES_ONE` stays for a measure nothing can compose against,
  and stops standing for one nothing tried to.
- `List.length` reaches the same path as `String.length`. Neither corpus draws a line there today,
  so this is not measured by the numbers above.
- On `corpus/issuetracker`, `--boundaries` offers rows for all eight edges; on `corpus/crm`, for the
  eighteen.
- The numeric path is untouched: the rows offered at `s = 1`, `s = 4`, `s = 5` above are the same
  rows.

kawasima:
Closes #593.

A line drawn on a count taken of a value got no row, and the rows the same generation offered were
already standing on it. The capability was not missing — it was decided at the caller.

## What was deciding it

`Generator.probe` read the kind of the term and returned before composing anything:

```java
// a length of five is a string of five characters, and nothing here composes one
if (!(axis.term() instanceof NumericTerm.ValueOf)) {
    return new BoundaryAttempt.Unresolved(new UnresolvedCombination(List.of(label),
            UnresolvedCombination.Reason.NOTHING_COMPOSES_ONE));
}
```

`Witnesses` composes exactly that — `"x".repeat(n)`, a collection of `n` — and the same generation
writes such values into the rows it offers. The sentence stopped being true when a candidate began
being proposed from the rule rather than the carrier alone, and the guard was not revisited.

`Intervals.classesOf` held the same assumption on the partition side, marking every class of a range
of lengths ungeneratable. A behavior whose obligations are all length-shaped was offered no row of
any kind:

```
// generated by `souther examples --generate`: 0 rows to fill what nothing covers.
// no row for `c=1 <= x < 3` in `pick`: no value can be written there
// no row for `String.length(c) = 1` in `pick`: nothing here composes a value at that edge
```

after:

```
//     | "c=1 <= x < 3" : (Code("x")) -> <?>
//     | "String.length(c) = 1" : (Code("x")) -> <?>
//     | "String.length(c) = 5" : (Code("xxxxx")) -> <?>
```

## What this changes

`Witnesses` owns the question and answers both halves of it, as it already did for a floor:

| | what it hands back |
|---|---|
| `holding(carrier, least)` | proposals for a floor of `least`: the count where the type has it, and what the type does have where it has less, for the decoder to refuse in the rule's own terms |
| `ofSize(carrier, size)` | values whose measure is `size`, and nothing where none was built |
| `reasonForSize(carrier, size)` | why no value of that count was built |

`holding` makes no promise about the measure of what it returns — a set of two under a floor of three
is exactly what it is for.

Neither is written in terms of the other: the builder carries the count of each value it made, and
each promise reads that build for itself — `ofSize` keeps what holds the count, `holding` keeps all
of it. That they agree for a string or a list is a property of the one builder rather than a call
from one contract to the other, so a cheaper witness for a floor cannot move a boundary. They differ
already for a set the type has too few values for, and at zero — a rule asking for at least none of anything has asked for no value in
particular, while a line at zero is an edge the empty value stands on, which is one more row on the
corpora.

`Generator.probe` composes through it, puts what comes back through the same `CandidateCheck` as any
other candidate, and settles the position's own number only where the line is on the position's own
content. `Intervals.classesOf` asks it for a class's representative.

## Measured

The bench corpora, with the CLI, since a compile does not run the measure:

| | rows offered | `nothing composes` notes |
|---|---|---|
| `corpus/crm` | 41 → 84 | 78 → 3 |
| `corpus/issuetracker` | 7 → 18 | 11 → 0 |

Diagnostics are unchanged on both. The three remaining notes mean it: a list of one element whose
element is a record nothing proposes a value for.

Five behaviors in `crm` now stop for a different and accurate reason — `lead.reach: ContactPoint`
has no value that can be written — where the guard used to stop them before the row was walked.

## Tests

`WhetherASizeCanBeBuiltIsAnsweredByWhatBuildsItTest` holds the boundary an author sees: a string of
the length the line names, a line further up met at its own length, an element count met by a
collection, a sized value the format refuses reported as refused rather than as unbuildable, and the
numeric path unchanged.

`WhatBuildsASizeSaysWhatItCouldNotBuildTest` holds the owner's own answers, including the two
promises coming apart at zero. A count larger than a row is worth carrying cannot be pinned through
the report: the measure calls such a row undecided before the generator is asked.

`mvn clean test` passes.

### DIFF
--- a/souther-compiler/src/main/java/souther/compiler/partition/Generator.java
+++ b/souther-compiler/src/main/java/souther/compiler/partition/Generator.java
@@ -1085,14 +1085,13 @@ public final class Generator {
                     : new Edge(List.of(at), null, numberOf(value));
         }
         int size = countOf(value);
+        if (size < 0) {
+            return Edge.none(UnresolvedCombination.Reason.NOTHING_COMPOSES_ONE);
+        }
         Type carrier = TypeOps.base(axis.type(), symbols);
-        List<FixtureTemplate> built = size < 0 ? List.of()
-                : Witnesses.ofSize(carrier, size, symbols, Set.of());
+        List<FixtureTemplate> built = Witnesses.ofSize(carrier, size, symbols, Set.of());
         if (built.isEmpty()) {
-            UnresolvedCombination.Reason why = size < 0 ? null
-                    : Witnesses.reasonForSize(carrier, size, symbols);
-            return Edge.none(why == null
-                    ? UnresolvedCombination.Reason.NOTHING_COMPOSES_ONE : why);
+            return Edge.none(Witnesses.reasonForSize(carrier, size, symbols));
         }
         List<FixtureTemplate> out = new ArrayList<>();
         for (FixtureTemplate each : built) {
--- a/souther-compiler/src/main/java/souther/compiler/partition/Generator.java
+++ b/souther-compiler/src/main/java/souther/compiler/partition/Generator.java
@@ -97,10 +97,14 @@ public final class Generator {
              * Nothing here knows how to compose a value of the shape asked for.
              *
              * <p>Apart from {@code NO_REPRESENTATIVE}, which says the position has no values. This
-             * says the search has no way to write one — a string of a given length is the case
-             * (#528) — and that is a fact about this compiler rather than about the model. Told
-             * apart because the first licenses "no value can be written there" and this one does
-             * not: the edge may be the easiest row in the file to write by hand.
+             * says the search has no way to write one — a collection of more elements than a row is
+             * worth carrying is the case — and that is a fact about this compiler rather than about
+             * the model. Told apart because the first licenses "no value can be written there" and
+             * this one does not: the edge may be the easiest row in the file to write by hand.
+             *
+             * <p>Which is why nothing decides this from the shape of the question. A reason read off
+             * the kind of term outlives whatever made it true, and says a value cannot be composed
+             * while the same generation composes one in the row above.
              */
             NOTHING_COMPOSES_ONE,
             /** Every value tried was refused at construction. */
@@ -290,32 +294,22 @@ public final class Generator {
                     UnresolvedCombination.Reason.NO_REPRESENTATIVE));
         }
         String label = axis.term() + " = " + written(obligation.value());
-        // A line on something taken of the value is not met by writing that number at the position:
-        // a length of five is a string of five characters, and nothing here composes one (#528).
-        // Said as having no value to offer rather than by handing the decoder an integer where a
-        // string goes — which it refuses, and "every value tried was refused" would then report an
-        // opinion the decoder never had about a boundary a row could easily meet.
-        if (!(axis.term() instanceof NumericTerm.ValueOf)) {
-            return new BoundaryAttempt.Unresolved(new UnresolvedCombination(List.of(label),
-                    UnresolvedCombination.Reason.NOTHING_COMPOSES_ONE));
-        }
-        FixtureTemplate at = valueOf(obligation.value(), axis.type(), subject.symbols());
-        if (at == null) {
-            return new BoundaryAttempt.Unresolved(new UnresolvedCombination(List.of(label),
-                    UnresolvedCombination.Reason.NO_REPRESENTATIVE));
+        Edge edge = edgeOf(axis, obligation.value(), subject.symbols());
+        if (edge.values().isEmpty()) {
+            return new BoundaryAttempt.Unresolved(
+                    new UnresolvedCombination(List.of(label), edge.reason()));
         }
         List<FixtureTemplate> inputs = new ArrayList<>();
         // What the rest of the row has to sit beside. A field of a record is not chosen from its
         // own type once another field of that record is fixed: the rule relating them says what
         // is left, and taking the bottom of the type's range instead is how a boundary that can
         // be written came back as one every value tried was refused at.
-        BigDecimal settledAt = numberOf(obligation.value());
-        Map<String, BigDecimal> settled = settledAt == null ? Map.of()
-                : Map.of(axis.path().toString(), settledAt);
+        Map<String, BigDecimal> settled = edge.settledAt() == null ? Map.of()
+                : Map.of(axis.path().toString(), edge.settledAt());
         for (int p = 0; p < subject.parameters().size() && p < subject.types().size(); p++) {
             Map<String, List<FixtureTemplate>> here =
                     TermPath.of(subject.parameters().get(p)).head().equals(axis.path().head())
-                            ? Map.of(axis.path().toString(), List.of(at)) : Map.of();
+                            ? Map.of(axis.path().toString(), edge.values()) : Map.of();
             Outcome tried = valueAt(subject, p, here, settled, check);
             if (tried.value() == null) {
                 return new BoundaryAttempt.Unresolved(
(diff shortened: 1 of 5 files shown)


kawasima:
Confirmed and fixed in e40ee544. The blocker was real, and worse than the contract table being wrong: the undersized value was **offered as a row**, not just returned internally.

Minimal model, four lines:

```souther
data Flags = Set<Bool>
behavior pick : (f: Flags) -> Tag
let pick (f) = if Set.size(f.value) >= 3 then Big else Small
example pick | (Flags([true])) -> Small
```

Before the fix, on this PR's head:

```
//     | "f=3 <= x" : (Flags([true, false])) -> <?>
//     | "Set.size(f) = 3" : (Flags([true, false])) -> <?>
//     | "Set.size(f) = 2" : (Flags([true, false])) -> <?>
```

Two rows of size two, one under a line drawn at three and one under the class `3 <= x`. Nothing
refuses them — the count comes from a guard, not an invariant — so `CandidateCheck` was never going
to catch it, and `Coverages` reads a built attempt as a witness that the edge can be written. The
claim reached further than the row did. Both of those rows are a regression this PR introduced: on
`develop` the same model offers nothing.

After:

```
//     | "Set.size(f) = 2" : (Flags([true, false])) -> <?>
// no row for `f=3 <= x` in `pick`: no value can be written there
// no row for `Set.size(f) = 3` in `pick`: nothing here composes a value at that edge
```

## What changed

The builder now carries the count of each value it made, and the two promises read the same build
differently rather than one calling the other:

```java
List<FixtureTemplate> exactly(int size)   // ofSize: those holding the count
List<FixtureTemplate> all()               // holding: all of them
```

`holding` keeps being offered the two booleans for a floor of three, which is the value the rule
refuses, put to the decoder so the refusal is said in the rule's own terms. `ofSize` gets nothing.
That they agree for a string or a list is now a property of the one builder rather than a call from
one contract to the other, which is the point you made about a cheaper floor witness moving every
boundary.

On the failure reason: `reasonForSize` now names one wherever `ofSize` offers nothing, so a set the
type has too few values for reports `NOTHING_COMPOSES_ONE` rather than being built, refused, and
reported as `ALL_CANDIDATES_REJECTED`. It stays deliberately weak about which of the three cases it
is — the value enumerator is bounded, so "the type universe is exhausted" is not something this can
tell from "the stepper ran out", and `NOTHING_COMPOSES_ONE` is documented as not licensing "no value
can be written there". Splitting it further needs the enumerator to say whether it finished, which is
a separate change.

`Generator.probe` also stopped supplying a default for the silence — that was the last place a caller
still decided what `Witnesses` is here to answer. `heldBackFor` keeps its own reading rather than
delegating: a floor nothing was built for is a position offering what it ordinarily offers, and
naming a reason there would put `NOTHING_COMPOSES_ONE` under every position with no floor at all.

## Tests

`WhatBuildsASizeSaysWhatItCouldNotBuildTest` now pins the cardinality contract:

```
Set<Bool>       size 2 -> [true, false]      size 3 -> no candidate
Map<Bool, Int>  size 2 -> a candidate        size 3 -> no candidate
holding(Set<Bool>, 3)  -> [true, false]      (the floor still gets it)
```

Neither bench corpus draws a line on the size of a set or a map, so their numbers are unchanged —
this is a defect the corpora could not have found.

`mvn clean test` passes.

kawasima:
Both fixed in 566e9e15.

## 1. `SEARCH_LIMIT` lost at the boundary

The code fact is exactly as you describe: `edgeOf` took `ofSize`'s values and dropped `Built.heldBack()`,
so a refusal of the sixty-four pairings that were built was reported as `ALL_CANDIDATES_REJECTED` —
which the generator's own reading says it is not, since values never built are not values refused.

`ofSize` now answers with both halves and `Edge` carries the second to the one place a refusal becomes
a reason:

```java
record Sized(List<FixtureTemplate> values, Generator.UnresolvedCombination.Reason heldBack) {}

UnresolvedCombination.Reason refused() {
    return heldBack == null ? Reason.ALL_CANDIDATES_REJECTED : heldBack;
}
```

`reasonForSize` reads the same pair instead of asking `sized` a second time.

**One correction to the report, and it goes against me:** I could not build a model that reaches it,
and I do not think one exists today. The pairing cap needs `keys × values > 64`, where both counts are
`proposalsFor(...)` — `Partitions.representativesOf`. A sum type proposes **one** value however many
cases it has: `Map<Case, Case>` with nine cases each gives one pairing, not eighty-one. I checked that
directly rather than reasoning about it. Nothing else I could find proposes more than a handful, so
`heldBack != null` with values in hand looks unreachable, and I could not write a test that fails
before the change and passes after.

So this is not a live defect being fixed — it is the same shape as the defect this PR is about, at a
place I put it: a caller reading half of what the owner answered and deciding the other half itself.
I have written it that way in the commit message rather than claiming a repro I do not have. If you
would rather not carry a branch nothing reaches, say so and I will drop it; my reading is that it costs
one field and removes the need to think about it again when the builder does learn to hold something
back.

## 2. Stale Javadoc

All three updated, and you are right that this one matters more than usual here — a stale sentence
about what could be composed is what the whole issue came from.

- `ofSize`: "the primitive the other two are asked through" → the narrower of the two promises, with
  `holding` reading the same build for what it asks and neither written in terms of the other.
- `holding`: "this asks for exactly the minimum" now sits with, rather than against, the paragraph
  below it — a value is built at the minimum, and one short of it is offered too where the type has no
  more to give, both being this method's reading of the build rather than `ofSize`'s.
- The test class header no longer says the second is answered by asking for the first.

## Verification

`mvn clean test`: 3837 + 1699 + 202 + 77 + 1, no failures. Corpus numbers unchanged from e40ee544
(crm 84 rows / 3 notes, issuetracker 18 / 0), and the `Set<Bool>` model still offers only the row at
`Set.size = 2`.

PR #600, opened 2026-08-10
A citation names the section rather than the position it sits at

### LINKED ISSUE #546: Source comments cite section numbers no lookup accepts
408 references of the form `spec 13.1`, `spec section 20`, `spec 8` sit in source comments across 155 files. A section number is not a name any lookup takes — `souther doc 13.1` answers "no section" and offers numeric lookalikes — and the numbers move: the chapter that `spec 19.5` named is the twenty-first now.

#543 fixed this for what a diagnostic shows a reader, and holds it with a build rule: every `(spec <name>)` in the catalog is a name `SpecDocument.section` resolves, which is the same call `souther doc` makes. Comments were left out of that change deliberately — they are maintainer-facing, and repointing one file would leave the tree half-converted.

The same rot applies to them. A maintainer following `spec 12.5` from a comment lands nowhere, and has no way to tell whether the number was wrong when written or drifted since.

## What this asks for

Repoint the comment references at anchors, and extend the build rule to cover them so a comment cannot cite a section that is not there. The counts by target:

```
44 spec 13.1    24 spec 12.5    17 spec 7.3     17 spec 4
16 spec 14.1    13 spec 8.3     13 spec 18.1    12 spec 19.8
12 spec 18.2    11 spec 8.7     11 spec 16.3    11 spec 13.2
```

The mapping is not mechanical. `spec 19.8` in the catalog turned out to point at declared sum data where the diagnostic was about a behavior's anonymous union output — a different rule — so each number needs reading against what the comment claims, not a table lookup from an old table of contents.

Worth checking while doing it whether a comment needs the citation at all. Several of these sit next to code that already says what the rule is.

kawasima:
Closes #546.

## What was wrong

432 comments pointed at the specification by section number. A number is not a name the document gives a section: it is produced from the heading's position by `:sectnums:`, so inserting a heading re-issues every number after it and nothing in the text changes. Over the nine days before this was written, 171 of 172 anchors survived unchanged while only 13 of them kept their number. 148 of the 432 named a number no section has.

The issue reads this as rot the catalog fix (#543) left behind in comments. It is narrower than that. The numbers were names once: until 5fc980aa the specification was Markdown with the number written into the heading, so `spec 7.3` was something a reader could search for. The conversion to AsciiDoc turned every one of them into a computed value and the citations stayed where they were. Nothing has been able to check one since — the number appears nowhere in the source, so neither an author nor a reviewer can look it up, and `souther doc` answers a name.

That is why numbers kept being written after the conversion. `UnreachableReached.java` was given `spec 16.3` on 2026-08-01, two weeks after 16.3 stopped meaning `match`, and a comment carrying a stale number is indistinguishable from one carrying a live one.

## Repointing them

The conversion preserved the numbering exactly — 125 of its 127 headings hold their number across it, the other two differing only in `+...+` escaping — so the number a comment was written against resolves to the heading it named, and from there to the anchor that heading carries today.

Read per occurrence rather than applied as a table. `spec 18.1` named the String section and is written on arithmetic; `spec 22` named the compile errors and is written on a fake's rows; `spec 8` named `data` and is written on absence at the boundary. Around thirty say something the number could not, and those are decided from the comment.

The spelling matches the 167 citations the tree already wrote as names. Three of those resolved to nothing and are repointed here.

## Keeping them

Two rules over every text file in the repository, in `ASpecificationCitationNamesASectionRatherThanItsNumberTest`.

- Every `§name` resolves through `SpecDocument.section`, which is the call `souther doc` makes.
- No section number is written where a citation goes.

Neither catches the other's failure. A number is not a name, so it is not a name that fails to resolve; and forbidding numbers says nothing about an anchor that was deleted. A third test holds the walk to the citations there are to find, so neither rule can pass over an empty scan.

Two occurrences write a number because they say why not to. They are allowed by path and by count — a file that may quote the mistake is not a file that may make it. The rule's own text writes no example of what it refuses, so it holds over itself.

## Left out

Six of the rewritten parentheticals also carry an ADR or issue number, which the repository's own convention keeps out of comments. That is a different rule and is not mixed into this diff.

`mvn -o clean test`: BUILD SUCCESS.


kawasima:
Both are real. Fixed.

**1. Elided citations.** `Ast.java` had three, and a sweep of every parenthesised number on a comment line found 19 across nine files — `Requirements.java`, `PipelineSigs.java` ×3, `SpecChecker.java` ×2, `TypeChecker.java` ×3, `Backend.java` ×3, `DataChecker.java`, `Ast.java` ×4, and two tests. All repointed.

On the convention: the same sweep says the shape is readable on its own. Where a parenthesised number on a comment line carries a dot it is a section, every time; the rest are a complexity or an argument (`O(1)`, `Amount(500)`, `Probability(10)`). Nothing dotted turned up in Markdown or XML prose. So the rule reads comment lines in Java and Souther sources only, and leaves prose elsewhere alone, where a version number is written the same way.

**2. `spec §13.1`.** The pattern now allows for the sign and refuses it:

```
\b[Ss]pec(?:ification)?s?\s+(?:sections?\s+)?§?[0-9]+(?:\.[0-9]+)*
```

Pinned in `aNumberIsRefusedWhetherOrNotItIsWrittenAsIfItWereAName`, against the patterns rather than against the tree — the tree only says what is there today. Both shapes were also injected into `Lists.java` and confirmed to fail at tree level.

The test writes neither shape: the word and the number are joined (`word + " §13.1"`) so the file holds no occurrence and the rules keep holding over their own text. Same reason the elided pattern's Javadoc describes the shape instead of showing it.

`mvn -o clean test`: BUILD SUCCESS, 3846 + 202 + 1. Rebased onto develop (#599 had landed meanwhile); the three new commits brought no numeric citations with them.

PR #603, opened 2026-08-10
Admit a fixture against the position it supplies

### LINKED ISSUE #597: A fixture standing in for a dependency is not typed, and a Java cast failure becomes the example's result
## What happens

A fixture standing in for a dependency's result is not checked against that dependency's declared
output type. Whatever is written is handed to the behavior, and the mismatch surfaces as a JVM cast
failure printed where the row's evaluated value belongs.

```souther
module repro exposing ( Stamp, Mark, now, today, stampAt, stampOn )

data Stamp = { at: DateTime }
data Mark = { on: Date }

behavior now : () -> DateTime
behavior today : () -> Date

behavior stampAt : () -> Stamp
    depends on now
    constructs Stamp
let stampAt (now) = Stamp { at = now() }

example stampAt
    | "a bare string" : () with now = "2026-07-20T09:00" -> Stamp
```

```
-- EXAMPLE  E1905 ---------------------------repro.sou:15:57

15|     | "a bare string" : () with now = "2026-07-20T09:00" -> Stamp
                                                                ^^^^^

This example does not hold.

It is:
    aborted: class java.lang.String cannot be cast to class java.time.LocalDateTime (java.lang.String and java.time.LocalDateTime are in module java.base of loader 'bootstrap')
But it needs to be:
    Stamp
```

It is not particular to strings, and not particular to `with`. Every one of these reaches the same
place:

```
with now = 1                        java.lang.Long cannot be cast to java.time.LocalDateTime
with now = Date("2026-07-20")       java.time.LocalDate cannot be cast to java.time.LocalDateTime
with today = "2026-07-20"           java.lang.String cannot be cast to java.time.LocalDate
```

and a `fake` table's output is unchecked in the same way:

```souther
behavior at : (key: String) -> DateTime

fake at
    | ("k") -> "2026-07-20T09:00"
```

```
It is:
    aborted: class java.lang.String cannot be cast to class java.time.LocalDateTime
```

The correct form, `with now = DateTime("2026-07-20T09:00")`, compiles and passes.

## Why it matters beyond the message

Three things are wrong at once, and the third is the reason to fix it rather than reword it.

A fixture's type is known before anything runs. `now` is declared `() -> DateTime`; the row writes a
`String`. That is a mismatch the checker holds every other value to, and it is being deferred to a
`ClassCastException` at evaluation.

The failure is reported against the wrong thing. E1905 says "this example does not hold", which
means the model computed one value and the row expected another. Here the model computed nothing —
it was handed a value it could not have been given. The row is blamed for a disagreement about the
result when the disagreement is about the fixture. E1903 is the code for a fixture that cannot be
built, and this is that.

A Java class name is not a diagnostic. `class java.lang.String cannot be cast to class
java.time.LocalDateTime (java.lang.String and java.time.LocalDateTime are in module java.base of
loader 'bootstrap')` names the runtime representation of two Souther types and the module system of
the language the compiler happens to be written in. Nothing in it says `now`, `DateTime`, or which
of the row's fixtures was wrong. In a row with several `with` clauses there is no way to tell from
the message which one to fix.

## Where the wrong form comes from

`example-fakes` writes it:

```
example preApprove
    | (AwaitingPreApproval { ... }, EmployeeId("boss-1")) with now = "2026-07-20T09:00" -> PreApproved
```

That block is marked `[,text]` rather than `[,souther]` and elides a value as `{ ... }`, so it is not
claiming to be a compilable row. It is still the only place the section shows a `with`, and a bare
string beside a `DateTime` dependency is the form a reader takes from it. The rule that a fixture is
written the way a fixture is written — `DateTime("…")`, `Date("…")` — is stated under `stdlib-date`,
which a reader looking up how to fake a dependency has no reason to open.

This is how the defect was reached: the section was read, the form in it was copied, and the
compiler answered with a `ClassCastException`. Whether or not the checking is added, the sample
should use the form that compiles, and `example-fakes` should say that a `with` value is a fixture
and is built like one.

## What to decide

Check a fixture against the type it stands in for, at the same point every other fixture is checked,
and report E1903 naming the dependency, the type it declares and the type the row wrote. Then no row
that reaches evaluation can produce a cast failure from its fixtures, and the `ClassCastException`
path becomes unreachable rather than reworded.

The regression test needs a row per fixture surface — a `with` value and a `fake` output — and per
kind of mismatch, at minimum a wrong primitive and a wrong newtype-like type that share a JVM
representation family (`Date` for a `DateTime` dependency), since those are the ones a reader gets
wrong by writing something plausible rather than something absurd.

## How it was found

Writing a medical-insurance claim adjudication model in Souther, using the MCP server as the only
source of language knowledge. In the model as first written, only the two rows reaching the
current-date dependency failed and the other five passed, so it read as a problem with those rows
rather than with a fixture — which is what the diagnostic said it was.

kawasima:
Closes #597.

A `with` value and a `fake` output were not held to the type the dependency they
stand in for answers with. Whatever the row wrote was handed to the behavior, and
the model met it as a JVM cast failure printed where the row's evaluated value
belongs.

```
15|     | "a bare string" : () with now = "2026-07-20T09:00" -> Stamp

It is:
    aborted: class java.lang.String cannot be cast to class java.time.LocalDateTime …
But it needs to be:
    Stamp
```

Two changes, and the second is why the first alone would not have been enough.

## Admit a stand-in against the output it stands in for

`buildFixture` used the reader that builds an *expected* value. That reader is
deliberately permissive: a row may expect a case the behavior does not answer
with, and reporting that disagreement is its job, so it builds what the row wrote
and leaves the rest to the comparison. A stand-in has no such licence — it is the
dependency's answer while the row runs.

The mismatch was already computed. `caseOfValue` answers whether the built value
is one of the output's cases, and a null answer was read as a cue to fall back to
the case the row wrote rather than as the refusal it is.

Building and admitting are now separate:

- a construction builds through the case the row named;
- a helper's answer is taken as returned — reading it back through the output's
  decoder states nothing for a stand-in written as an application;
- a literal is read through the output's own shape, so `with quote = 2` builds the
  `Amount` a dependency answers with, the way `(1)` already does at an input.

What each produces is then admitted: by representation where the whole answer is
one shape, and by case membership where it is several types and no one decoder
reads it. A sum has no class of its own, so a nominal shape admits any leaf it
holds.

Refused before the behavior is applied, reported as **E1908** with the dependency
named, at the `with` or the `fake` that wrote it. E1903 stays what it is: a
fixture of the row itself. The issue asked for E1903; E1908 is the boundary the
spec already draws, and it keeps the dependency name and position the callers
hold.

## Read a fixture from what Souther writes, not from what a boundary carries

With admission alone, `with now = "2026-07-20T09:00"` would decode cleanly and the
row would pass. A derived Decoder reads more than the form Souther writes a value
in — a temporal also arrives as ISO text, because that is how one crosses a
boundary over JSON and out of a JDBC row — and a fixture built through it was
admitted by everything it reads. `FixtureReader` states the opposite directly
above `leafDecoder` ("a temporal arrives parsed … not the ones that parse text"),
and the spec gives one form for writing one, so this was neither intended nor
written down. raoh is unchanged; what a fixture may hand it is what narrowed.

The reader also had a second reading of its own: a newtype over a temporal took
the string its base is written from, `貸出日("2026-07-01")`, parsed here rather
than by a decoder. That is not a coercion leaking out of raoh — it fails the build
on a malformed value like any written temporal — but it is a spelling the fixture
language had and the language it writes fixtures for does not, since a model body
writing it is E1317. The same string had one meaning inside a temporal newtype and
another everywhere else. It takes what it wraps now, which is what a model body
writes and what `FixtureTemplate` already generates for it.

So one rule holds everywhere:

```souther
"2026-07-20T09:00"                    // String
DateTime("2026-07-20T09:00")          // DateTime
貸出日(Date("2026-07-01"))             // a newtype over one
```

## What the issue's rows do now

```
with now = "2026-07-20T09:00"  E1908  a `DateTime` is written `DateTime("…")`;
                                      a string is how one arrives from outside,
                                      not how a fixture writes it
with now = 1                   E1908  expected date-time
with now = Date("2026-07-20")  E1908  expected date-time
fake at | ("k") -> "…"         E1908  a `DateTime` is written `DateTime("…")` …
with now = DateTime("…")       compiles and passes
```

No fixture reaches the model unadmitted, so the `ClassCastException` path is
unreachable rather than reworded.

## Spec

`example-fixtures-are-buildable` named only inputs and the expected result;
stand-ins are now named too. `fixture-is-written-not-carried` is new — without it
"through that type's derived Decoder" reads as sanctioning the transport form.
`temporal-literal` states the newtype form, `example-fakes` no longer shows a bare
string as a `with` value (the issue names that sample as where the wrong form came
from), and E1903/E1908 say which fixtures each is for.

## Tests

`AStandInIsAdmittedByWhatItStandsForTest` — `with` and `fake` against Scalar,
Nominal, named sum and unnamed union outputs, plus guards that a base literal, a
helper-produced value and the output's own form still stand in.
`ATemporalFixtureIsNotWrittenAsTextTest` — every position a fixture writes a
temporal at, and the forms that still build. `CompileDateLiteralTest` gains the
refusal beside the form it keeps.

`mvn -o clean test` is green across every module.


kawasima:
Both blockers reproduced, and a third case with them. Fixed in 2c9b8fec.

```
helper returning String at a DateTime input      compiled clean
helper returning String in a DateTime field      compiled clean   <- also open
bare string in a DateTime? field                 compiled clean
bare string in a List<DateTime> field            E1903 (control)
```

You were right that the cause is where the check sits, not what it covers. `Ast.StringLit` is a node of one of the two walks that produce the neutral form; the other one — `NeutralForm.of`, re-materialising what a helper returned — has its own scalar path that returns a live value before looking at the position. And the walk that finds the temporal a position is opened newtypes but not optionals, so the written side leaked through `DateTime?` as well.

`temporalUnder` now opens both and lives on `NeutralForm`, whose class comment already says why: the rules that decide the form itself belong there so the two directions cannot answer differently about one position. Both readings ask it, and `notWrittenAsATemporal` words the refusal once.

Regressions added: a bare string at a `DateTime?` field, a helper answering with a `String` at an input, and the same helper inside a construction — the last so the rule is not fixed only for a whole fixture. Guards added too: an optional temporal still takes one and still takes `None`, and a helper answering with a temporal still writes one.

On the spec — agreed, and it was wrong in a second way. "Every fixture ... through that type's derived Decoder" would make a helper's answer owe a second decode, which is exactly what states nothing for a stand-in written as an application. It also flattened the expected result, which is allowed to name a case the behavior does not answer with. It now says what each of the three positions is held to, separately. `fixture-is-written-not-carried` says it reaches a helper's answer and a `?` field.

`mvn -o clean test` green across every module (compiler 3881).

kawasima:
Blocker reproduced and fixed in 7467ef43, along with both non-blocking items.

```
whole expected = helper answering String,       at DateTime        E1905 -> E1903
whole expected = helper answering List<String>, at List<DateTime>  E1905 -> E1903
stand-in       = helper answering String,       at DateTime        already E1908
stand-in       = helper answering List<String>, at List<Int>       already E1908
```

The two stand-in rows were already closed by admission — worth saying, because it means the hole was specific to the branch that returns the answer with nothing else asked of it, not to helpers in general.

Putting that answer through `NeutralForm.of` would have closed it and asked for more than the rule: `of` also refuses a value it cannot re-materialise, and being unreadable is not the same as being in the wrong form — that distinction is why the whole-answer branch exists. `requireWrittenForm` holds only the `temporalUnder` rule. It walks containers, since a position's element type is part of what the position says and a written `[ "…" ]` at `List<DateTime>` is already refused, and stops at anything nominal, whose fields are the types they were declared as. Both places a helper can answer for the whole fixture call it, so the expected side and the stand-in side word the refusal identically.

On the collection regressions — you were right that nothing tested `each()`. My first attempt at a probe did not reach it either: a `let` bound to a list literal is not a helper application, so it went through the literal branch and failed in the decoder instead. With an application it reports `a \`List\` does not stand in for \`List<Int>\``, which is the contract you wanted pinned — right container, wrong element. Added positive and negative.

On `builds the same way` — agreed, and it was pointing at the one difference that matters. It now names it: a `with` value or `fake` output is built from what it writes, a literal at the type the dependency answers with, so `with quote = 2` builds the `Amount` while an expected `2` stays a `2`.

`mvn -o clean test` green across every module (compiler 3886).

PR #608, opened 2026-08-10
Point an argument's type mismatch at the argument

### LINKED ISSUE #605: An argument's type mismatch is drawn under the callee, so the argument it names can be lines away
An argument whose type does not fit is reported with the caret under the callee. The message names
the argument by position; the region names the call. When the call is written down the page, the two
are lines apart, and the reader is told to look at something the compiler declined to point at.

## What happens

```souther
module repro.multiline4 exposing (notice)

behavior notice : (a: String, b: String) -> String

let notice (a, b) =
    String.append(
        a,
        String.length(b)
    )
```

```
$ souther compile multiline4.sou -d out --color never
-- TYPE MISMATCH  E1317 ------------------multiline4.sou:6:5

6|     String.append(
       ^

argument 2 of String.append: expected String, but got Int.
```

`argument 2` is `String.length(b)` on line 8. The region is line 6, column 5 — the `S` of the callee.
Nothing on the line the compiler prints is wrong.

The same anchoring shows up on a single line, where it costs less but still points away from the
operand:

```
5| let label (n) = String.append("book: ", n)
                   ^

argument 2 of String.append: expected String, but got Int.
```

E1322 does it with a literal. The message quotes the text that is wrong and the region covers the
constructor:

```
5| let closedOn (n) = Date("2026-02-30")
                      ^^^^

`2026-02-30` is not a Date. Write `Date("2026-07-01")` / `DateTime("2026-07-01T09:00")`.
```

E1318 does it with a list, and here the offending element is two lines below the caret:

```
14|         [ loan.copyId
            ^

The elements of this list do not all have the same type.
Hint: One element is String and another is Int; make every element the same type.
```

The list is `[loan.copyId, loan.memberId, loan.renewalCount]` written one element to a line;
`loan.renewalCount` is the `Int` on line 16.

## That the region can be the operand

A record field of the wrong type anchors at the field, which is on the same line as the value and
next to it:

```
16|         renewalCount = loan.copyId
            ^^^^^^^^^^^^

Field `renewalCount` expects Int, but got String.
```

This is the same diagnostic code as the call case, E1317, so the code is not what decides the
region. Whatever picks the region for a field could pick the argument for a call.

## What it costs

A reader who follows the caret reads the callee and finds it well typed. To find the argument the
message names, they count arguments by hand — across lines, and past any nested call that supplies
one. The longer the argument list, the further the caret is from what it is about; there is no bound
on the distance.

The convention is consistent, so this is a choice of region rather than an off-by-one: the region is
the name the message has already spelled out in words, and the operand — the part the reader cannot
recover from the message alone — is what goes unmarked.

## What to expect instead

The region covers the expression whose type did not fit: `String.length(b)` for `argument 2`, the
`"2026-02-30"` literal for E1322, and the element that differs for E1318.

kawasima:
Closes #605.

An argument whose type did not fit was reported with the caret under the callee, so on a call written down the page the message named `argument 2` and the region named line 6.

## Where the region came from

Not from the diagnostic code, and not from the kind of failure. Three paths type a call, and each answered differently:

| written | path | region |
|---|---|---|
| `String.append(a, String.length(b))` | `settledByValues` → `TypeOps.unify`, at `call.pos()` | callee |
| `shout(String.length(a))`, a helper | `Substitution.hold` | argument |
| `shout(...)`, a behavior called by name | `CallArgs.require` → `Elaborator.requireType` | argument |

Two of the three already point at the argument. The third does not because `settledByValues` — the walk that settles the signature's type variables — refuses as well, and it is handed two types and the position of the call. The argument survives only as the words in the `what` string.

The check it refuses with is one `applySignature` runs again three lines later: `requireTyped` requires each non-function argument against the parameter type the walk settled, at the argument's own position. That one never ran.

## What changed

`TypeOps.bindVars` settles variables and refuses nothing; `settledByValues` calls it. The refusal stays where the argument is in hand.

`parseTemporal` takes a `Region` rather than a `SourcePos`, and its two callers pass the literal's. The message quotes the text, so the caret covers the text.

```
8|         String.length(b)
           ^^^^^^^^^^^^^

argument 2 of String.append does not have the type it needs here.
```

```
8|     on = Date("2026-02-30")
                 ^^^^^^^^^^^^

`2026-02-30` is not a Date.
```

## What it does not change

The set of programs refused. `requireTyped` runs the same `assignable` on the settled parameter type for every non-function parameter. `unify`'s variable branch is the one asymmetry — it checks both directions where `requireTyped` checks one — and the case it admits is one `requireTyped` already rejects today.

E1318 is left alone. Which of two elements is the offending one is a question the list rule has to answer first, and the message currently declines to ("One element is String and another is Int").

The caret covers `String.length`, not `String.length(b)`: `Ast.Apply` carries a start position and no end, so the whole call is not a region anything can name yet. The frontend has it — a CST node answers `start()` and `end()`, and `AstBuilder` reads only the start — so carrying it is a change of its own.

## Tests

`ADiagnosticPointsAtTheOperandThatSuppliedTheValueTest` holds all three call paths and the temporal to one rule. The kernel and temporal rows were red before this change; the helper and behavior rows were green before it and are the positive control that the rows are not empty probes.

`CompileImportedNameTest` asserted `of trim:`, the colon belonging to the sentence this no longer reaches. It now asserts `argument 1 of trim `, and its point — that the report quotes `trim` and not `String.trim` — is unchanged.

`mvn clean install` is green across the reactor: souther-compiler 3893, souther-fmt 2190, souther-lsp 202, souther-syntax 87, souther-runtime 109.

### DIFF
--- a/souther-compiler/src/main/java/souther/compiler/FixtureReader.java
+++ b/souther-compiler/src/main/java/souther/compiler/FixtureReader.java
@@ -2,6 +2,7 @@ package souther.compiler;
 
 import souther.compiler.ast.Ast;
 import souther.compiler.check.CallElaborator;
+import souther.compiler.check.Elaborator;
 import souther.compiler.check.Symbols;
 import souther.compiler.check.TypeOps;
 import souther.compiler.observe.Limits;
@@ -974,7 +975,7 @@ public final class FixtureReader {
             if (c.args().size() != 1 || !(c.args().get(0) instanceof Ast.StringLit lit)) {
                 throw new FixtureException("`" + c.written() + "` takes one written string");
             }
-            return CallElaborator.parseTemporal(c.written(), lit.value(), c.pos());
+            return CallElaborator.parseTemporal(c.written(), lit.value(), Elaborator.region(lit));
         }
         if (!neutral.isNewtype(c.written())) {
             String reached = helperKey(c);
--- a/souther-compiler/src/main/java/souther/compiler/check/CallElaborator.java
+++ b/souther-compiler/src/main/java/souther/compiler/check/CallElaborator.java
@@ -162,6 +162,7 @@ public final class CallElaborator {
             arity(call, params.size());
         }
         Map<String, Type> bind = settledByValues(call, params, kept.result(), expected, ca::type, ctx);
+        requireValueArgs(call, params, ca, bind);
         for (int i = 0; i < params.size(); i++) {
             if (params.get(i) instanceof Type.FnOf declared) {
                 Type.FnOf at = (Type.FnOf) TypeOps.substitute(declared, bind);
@@ -172,12 +173,6 @@ public final class CallElaborator {
                 // kept because its meaning belongs to whoever reads it, not to this.
                 TypeOps.unify(declared.result(), answered, bind, ctx.symbols(),
                         call.pos(), "argument " + (i + 1) + " of " + call.written());
-            }
-        }
-        for (int i = 0; i < params.size(); i++) {
-            if (!(params.get(i) instanceof Type.FnOf)) {
-                ca.requireTyped(i, TypeOps.substitute(params.get(i), bind),
-                        "argument " + (i + 1) + " of " + call.written());
             }
         }
         return new Core.PreservedCall(call.denotes(), ca.cores(),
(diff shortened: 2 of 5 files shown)


kawasima:
## [P1] Confirmed and fixed in d17f6b61

The reproducer as written does not fail — `List.find(x -> x > 0, n)` still reports `argument 2 of List.find` at `n`. A comparison tolerates a `Type.Var` operand, so with `'a` unbound the predicate types cleanly and the value argument is reached.

The class of failure is real. A block body that needs a concrete type does refuse first. Four probes on `List.sortBy(<block>, n)` with `n : Int`, before and after the first commit:

| block | before | after |
|---|---|---|
| `x -> x > 0` | argument 2 of List.sortBy | argument 2 of List.sortBy |
| `x -> String.length(x) > 0` | argument 2 of List.sortBy | argument 1 of String.length |
| `x -> x.flag` | argument 2 of List.sortBy | E1321 `I cannot read a field flag` |
| `x -> String.append(x, "a")` | argument 2 of List.sortBy | argument 1 of String.append |

Three of four regressed, and the point holds: the refusal moved to where the argument is in hand, but it moved past the block as well.

`applySignature` now requires the value arguments as soon as `settledByValues` returns, before any function argument is resolved. All four probes answer at `n` again. `preservedCall` is held in the same order, being the same application.

On the fold concern: `List.fold((acc, x) -> acc ++ [x * 2], [], xs)` still compiles, and `List.fold((acc, x) -> acc ++ [x], [], n)` on a non-list reports `argument 3` at `n`. The empty seed binds its variable to the bottom in `settledByValues`, and `assignable` absorbs the bottom, so requiring it early asks nothing the seed cannot answer.

`aValueArgumentIsRefusedBeforeTheBlockThatWasTypedFromIt` pins the `String.length` row.

## [P3] Reworded

`preservedCall` is a fourth path, and it does still carry `TypeOps.unify(declared.result(), answered, ..., call.pos(), ...)` for a function argument's result. That is the same shape as `HelperTyping`'s `"block result"`, and both belong to the `pos + what:String` cleanup rather than here.

The comment now reads:

> The rows below cover the three paths a call written in a body is typed by — a kernel applied against its declared signature, a helper expanded, a behavior called by name — which answered three different ways for one code and one kind of failure. They are the paths that exist now; a fourth is not held to anything here.

## Tests

`mvn clean install` green across the reactor: souther-compiler 3894, souther-fmt 2190, souther-lsp 202, souther-syntax 87, souther-runtime 109.

Issue #602, opened 2026-08-10
A result with nothing in it is reported as a model with nothing in it


kawasima:
Four ordinary things a model can say are not read. The four are worth fixing, but they are symptoms:
what they share is that an empty result is turned into a statement about the model. This issue is
about that, and the four readings are what comes after it.

## What happens

The model below is five behaviors. `approvalSimple` and `approvalCompound` differ only in the
`r.cost >= 0 &&` in front of the same comparison.

```souther
data Kind = Domestic | Overseas
data Request = { kind: Kind, cost: Int }

let approvalSimple (r)   = if r.cost <= 100000 then Auto else Manual
let approvalCompound (r) = if r.cost >= 0 && r.cost <= 100000 then Auto else Manual

data CardNumber = String
    invariant String.matches("[0-9]{16}", value)
data Amount = Decimal
    invariant value >= 0.0m
data Card = { number: CardNumber }
data Cash = { amount: Amount }
data Payment = Card | Cash

let feeFor (p) = match p with
    | Card -> Amount(1.0m)
    | Cash -> Amount(0.0m)

let freshness (on)         = if on < Date("2026-01-01") then Stale else Fresh
let retryVerdict (retries) = if retries == 3 then GiveUp else Again
```

Each behavior has one example row. `souther examples repro.sou --generate --boundaries`:

```
  approvalSimple           implemented   rows 1    pending 0
    partition   axes 2   single-axis 2/4   pairs 1 reached / 1 known reachable, 3 untried
    boundary    0/2
      · no row is at approvalSimple/r.cost = 100000 (guard@27:5)
      · no row is at approvalSimple/r.cost = 100001 (guard@27:5)
  approvalCompound         implemented   rows 1    pending 0
    partition   axes 1   single-axis 1/2
    boundary    not measured (no line was derived at any position)
      · not derivable: r.cost
  feeFor                   implemented   rows 1    pending 0
      · no row uses `Card`
    partition   axes 1   single-axis 1/2
    branch      1/2
      · no row goes through `case Card` (56:5)
  freshness                implemented   rows 1    pending 0
    partition   not measured (no partition axis was derived at any position)
      · not derivable: on
  retryVerdict             implemented   rows 1    pending 0
      · not derivable: retries
```

and the generated block ends with

```
// no row for `p=Card` in `feeFor`: no value can be written there
```

A value can be written there. One row added by hand to the same file closes every finding about
`Card` at once:

```souther
[...]

PR #614, opened 2026-08-10
Offer one row per row, and answer for every gap a build refuses

### LINKED ISSUE #606: `--generate` exposes one row per boundary obligation, and handles only one of four adequacy-gap kinds
`souther examples --generate --boundaries` writes one row per uncovered edge. Several edges are met
by one input, so the rows it offers repeat the same input tuple and differ only in the description.
The count in the header counts those repeats. And of the four findings that decide `adequacy`, the
block offers rows for one and says nothing at all about the others, so adopting every row it offers
does not reach `adequacy: satisfied`.

What is underneath the three is one thing: `--generate` does not produce an input set that would
satisfy adequacy. It exposes the sequence of probes the boundary measurement already made, one per
boundary obligation, directly as user-facing rows.

## What happens

```souther
module repro.d1 exposing (Rate, Cap, Policy, fee)

data Rate = Int
    invariant nonNegative = value >= 0

data Cap = Int
    invariant nonNegative = value >= 0

data Policy =
    { rate: Rate
    , cap: Cap
    }

behavior fee : (days: Int, policy: Policy) -> Int

let fee (days, policy) = {
    let accrued = days * policy.rate.value

    if accrued > policy.cap.value then policy.cap.value else accrued
}

example fee
    | "five days at ten a day, under the cap" : (5, Policy { rate = Rate(10), cap = Cap(500) }) -> 50
```

```
$ souther examples d1.sou --generate --boundaries --color never
repro.d1                                                 measurement: complete
  fee                      implemented   rows 1    pending 0
    signature   not applicable (this behavior's output is not a sum)
    boundary    0/2
      · no row is at fee/policy.rate = 0 (invariant Rate (min))
      · no row is at fee/policy.cap = 0 (invariant Cap (min))
      · not derivable: days
    branch      1/2
      · no row goes through `then` (19:5)

1 behavior: 1 implemented, 0 injected; 0 rows waiting for a `let`.
adequacy: not satisfied
// generated by `souther examples --generate`: 2 rows to fill what nothing covers.
// example fee
//     | "policy.rate = 0" : (0, Policy { rate = Rate(0), cap = Cap(0) }) -> <?>
//     | "policy.cap = 0" : (0, Policy { rate = Rate(0), cap = Cap(0) }) -> <?>
```

The two rows carry the same input, `(0, Policy { rate = Rate(0), cap = Cap(0) })`.
`Generator.probe` (`partition/Generator.java:288`) is called once per `BoundaryObligation` and builds
a whole input tuple each time, and the positions the edge does not name are filled by a search that
knows nothing about the other obligations. Where these minimum obligations are satisfiable together,
the independent probes can arrive at the same tuple, because each free position is filled from the
bottom of its own domain. Whether they always do is not the point: independent probe generation
provides no normalisation for candidates that coincide, however often that happens.

Not because both invariants happen to have the same minimum. Give `Cap` the invariant
`value >= 100` instead, and the rows are still identical, because the free position takes the bottom
of its own domain and the minimum edge names exactly that:

```
      · no row is at fee/policy.rate = 0 (invariant Rate (min))
      · no row is at fee/policy.cap = 100 (invariant Cap (min))
//     | "policy.rate = 0"  : (0, Policy { rate = Rate(0), cap = Cap(100) }) -> <?>
//     | "policy.cap = 100" : (0, Policy { rate = Rate(0), cap = Cap(100) }) -> <?>
```

`Generator.java:280` chose this unit deliberately — "One row per boundary rather than one row
covering several, because a row is a question put to a person and a row sitting on three edges at
once is three answers they have to separate". That reasoning is about not composing several edges
into one row on purpose. It does not cover rows that coincide, and what comes out is the thing it
was written to avoid: rows a reader cannot tell apart at all.

## Adopting every row offered does not satisfy

Uncommenting both rows and answering them:

```
$ souther examples d1b.sou --generate --boundaries --color never
  fee                      implemented   rows 3    pending 0
    boundary    2/2
      · not derivable: days
    branch      1/2
      · no row goes through `then` (19:5)

adequacy: not satisfied
$ souther examples d1b.sou --strict >/dev/null; echo $?
1
```

No block is printed at all this time. The report names a gap, the build refuses it, and the
generator has nothing to say about it in either direction.

That is not a missing branch generator. `Adequacy.Kind` (`query/Adequacy.java:1106`) holds four
findings whose `isAdequacyGap()` is what `--strict` refuses on:

```
OUTPUT_CASE_UNSPECIFIED
INPUT_CASE_UNSPECIFIED
BOUNDARY_UNMET
ARM_UNREACHED
```

`Adequacy.Filling` (`query/Adequacy.java:928`) holds two slots, `pairs` and `boundaries`. They are
the two generators that exist, not a division of the findings: `pairs` fills axis classes and the
pair space, which are not themselves adequacy gaps — a row it offers may close one along the way —
and `boundaries` is the only one aimed at a finding that fails a build. Nothing in the generator
reads a `Finding`. So there is no mapping from the gaps
to what the generator can offer — not a partial one, none — and a gap of a kind no generator was
written for produces no row and no note, because `GeneratedRows.notes()` (`report/GeneratedRows.java:179`)
builds its lines from what a generator returned.

## The header is counted from a different collection than the block

```java
// report/GeneratedRows.java:88
int rows = asked.stream().mapToInt(e -> e.getValue().rows().size()).sum();
```

counted from the per-obligation lists, before `blocks()` renders anything. The block and its
description are two readings of two collections, and the two lines above the block come from the
same place. A result with explanations and no rows prints:

```
// generated by `souther examples --generate`: 0 rows to fill what nothing covers.
// no row for `policy.rate = 0` in `fee`: every value tried was refused at construction, which does not make the combination impossible
// no row for `policy.cap = 0` in `fee`: every value tried was refused at construction, which does not make the combination impossible
```

"Replace each `<?>`" over a block holding no `<?>`. `toWrite` (`report/GeneratedRows.java:94`) is
true when there are rows *or* explanations, and the preamble it guards is written for the first case
only. One boolean where the block has three shapes: rows, explanations, and both. (Why that
particular search failed is #602's subject and not this one's; what it shows here is the shape of
the block.)

## What it costs

The line above the block reads `Replace each <?> with what the system actually answers, then
uncomment`, which invites pasting the block verbatim. A reader who does that gets duplicate rows —
in a larger model, four identical rows on one behavior — and a file that still fails `--strict`,
with no indication of which part of the shortfall the generator had decided not to help with.

The header number is the one thing a reader can use to judge how much work is left, and it is not
the number of rows they need to write in either direction: it double-counts inputs that serve
several edges, and it excludes gaps that get no row.

## The four stages, of which two exist

```
Adequacy findings
       │
       ▼
[1] disposition          AdequacyGap -> GenerationOutcome
       │
       ▼
[2] candidate generation obligation -> candidate input
       │
       ▼
[3] normalisation        rendered candidates -> the obligations each row stands on
       │
       ▼
[4] presentation         rendered rows, and what could not be offered
```

Today [2] exists through two generator paths, `pairs` and `boundaries`, of which only `boundaries`
targets one of the four adequacy-gap kinds directly, and it is read by a presentation layer that
also does its own counting. Each symptom is one missing stage:

| symptom | missing |
| --- | --- |
| duplicate rows | [3] |
| header count, and `Replace each <?>` over no `<?>` | [4] reads [2]'s internals |
| a gap the block says nothing about | [1] |

[1] is the one that has to be total. Every finding that `isAdequacyGap()` gets one of

```
Generated(candidates)   an applicable strategy produced candidates
CannotGenerate(reason)  an applicable strategy could not produce a candidate
NotSupported(reason)    no strategy is applicable to this gap or to the shape it would need
```

and none of them may be absent. What separates the second from the third is applicability, not what
the search did: whether candidates were attempted belongs in the reason, and does not decide the
kind. A strategy may recognise a shape as its own and stop before enumerating anything, which is
`CannotGenerate`; another may read half a condition before finding a form it does not handle, which
is `NotSupported`. Keying the three on what a search happened to touch would derive a semantic state
from a search history again, which is what #602 is about, one layer over. `NotSupported` is where the branch gap in the repro lands: a row that
reaches `then` needs a partition axis on that guard, and `GuardThresholds.read()`
(`partition/GuardThresholds.java:87`) reads a threshold only where one side of the comparison is a
constant. `accrued > policy.cap.value` has a constant on neither side. Saying that is enough to close
this issue; deriving a witness for it is not part of it.

Although the patches go 1, 2, 3, the types should be written in the other direction from the start.
`GenerationOutcome` is the outer answer and the per-obligation candidates live inside
`Generated(candidates)`. Building a new abstraction around `Filling.boundaries()` first would mean
replacing it in the third step.

## The identity to group by is the rendered row, and it should say so

There is no canonical value equality to group inputs by today. `FixtureTemplate` is
`record (String text, Ast.Expr value)`; `Ast.IntLit` carries a `SourcePos`; `Ast.Apply` carries a
`ConstructionOrigin` and an `appliedAs`. Record equality over those is equality of provenance as
much as of value. Writing a canonical form for a value graph is the shape this codebase has already
had to correct three times (#507, #523, #563), and it is not what this issue is for.

So the criterion is about what is rendered, not about identity of inputs. Grouping by rendered text
cannot collapse two textually distinct rows that a reader would paste, and the worst it can do is
leave two rows that mean the same thing, each naming its own edge. Claiming "one row per
distinct input" would claim a guarantee that is not held. This issue is about having confused
obligation identity with input identity; stating the fix in terms of a value identity that does not
exist would repeat it one layer over.

## What to expect

1. Obligations whose candidates render the same row are offered as one row: no two rendered rows in a
   block are identical.
2. That row carries every obligation it stands on, and its description names them.
3. The header's row count is the number of rows actually rendered.
4. Where the block holds no `<?>`, the line telling the reader to replace each `<?>` is not printed.
5. Every finding that `isAdequacyGap()` has a disposition: `Generated`, `CannotGenerate`, or
   `NotSupported`.
6. A gap no generator was written for is named, with its reason, rather than left out.
7. Extending what a witness search can reach is out of scope. `position op position` guards get
   `NotSupported`, and deriving an axis from one is a separate issue.

## Order against #602

#602 does not fix any of the three, and it is not their cause. What it does is increase the number of
discovered obligations, and with generation keyed on obligations that means more duplicate rows —
pressure on a defect that is already here, not a new one.

The abstraction above is also the place its results will land, so it is worth having first. 1 and 2
in particular are independent of everything #602 changes.

One thing found while investigating this belongs to #602 and not here: a representative search that
composes nothing removes the boundary obligation from the denominator, and `--strict` passes on a
model whose own report names the unmet edges. That is about whether a gap may disappear, which is one
step before what this issue answers.

Related: #593, which is where the boundary rows above began to be offered at all.

kawasima:
Closes #606.

`--generate` exposed the sequence of probes the boundary measurement makes — one per
obligation — directly as user-facing rows, and answered only for the gaps it happened to
have a strategy for.

## What changes

**One row per row.** A probe fills the positions its own edge does not name from the bottom
of each position's domain, which is the value a minimum edge names, so obligations that are
satisfiable together arrive at the same tuple. Candidates that render the same row are now
one row, and its description names every class and edge it stands on.

```
- // generated by `souther examples --generate`: 2 rows to fill what nothing covers.
- //     | "policy.rate = 0" : (0, Policy { rate = Rate(0), cap = Cap(0) }) -> <?>
- //     | "policy.cap = 0" : (0, Policy { rate = Rate(0), cap = Cap(0) }) -> <?>
+ // generated by `souther examples --generate`: 1 row to fill what nothing covers.
+ //     | "policy.rate = 0 x policy.cap = 0"
+ //         : (0, Policy { rate = Rate(0), cap = Cap(0) })
+ //         -> <?>
+ // nothing offers a row for `then` in `fee`: nothing here composes an input for the arm it would reach
```

Grouped on what is written rather than on the values. There is no identity for a value here —
a fixture carries the position it was parsed at and the path it was constructed through — and
two rows meaning the same thing and spelt differently staying two rows is the direction to be
wrong in.

**The header describes the block.** The count was taken from the per-obligation lists before
anything was rendered, and the line telling a reader to replace each `<?>` was printed over
blocks that hold none. Both now read the rows that were printed.

**Every gap a build refuses gets an answer.** `Adequacy.Kind` holds four findings whose
`isAdequacyGap()` is what `--strict` refuses on, and the generator read none of them. Each now
carries a `GenerationOutcome`:

```
Generated(candidates)   an applicable strategy produced candidates
CannotGenerate(why)     an applicable strategy could not produce one
NotSupported(reason)    no strategy is applicable to this gap or to the shape it would need
```

The reason a strategy could not produce one is carried as the attempt recorded it. Classes that
were not there and classes that would not link are two things the assessment tells apart, and
printing either as the other is this compiler choosing which of them it saw.

Which of the three is a question about strategies and not about searches, so a gap moves
between them as strategies are written and never with what one run happened to touch. The
switch over the kinds is exhaustive: a kind added later does not compile until somebody has
said which of the three it is.

## Two things the implementation settled

**Candidate generation is not keyed on gaps.** Offering boundary rows only through the
dispositions broke three existing tests, all on models with no `example` rows at all: a line is
measured against the rows, so such a behavior has no gap at any of its lines while every one of
them is still a row worth offering. The strategies' own results stay the source of what is
offered; the dispositions are what makes the answer total.

**`CannotGenerate` at a boundary has two ways in, and only one of them is open.** Writability
is what makes an unmet edge a gap, and a search that composes a row is one way to establish it —
so where the search is what failed, there is no gap, and #602 is what opens that route.
`Coverages.writabilityOf` also proves writability from the projection whatever the attempt did,
so an edge the projection settles and the attempt could not build reaches `atEdge` today. That
is where `LINKAGE_FAILED` arrives.

## Out of scope

Deriving a witness for a guard comparing two positions (`accrued > policy.cap.value`).
`GuardThresholds.read()` reads a threshold only where one side is a constant, so such an arm is
`NotSupported` and says so.

## Verification

`mvn clean test` — 3902 tests, 0 failures. Each new case was checked against a mutation that
defeats it: the old count expression, an `atCase` that answers every case the same way, the two
build failures printed as one sentence, and the stop reasons appended without the set that keeps
a fact from being said twice.


kawasima:
Pushed `10887f6a`.

## 1 [P1] — fixed

The catch was around the whole answer, so a `LinkageError` in the search skipped the walk over the findings entirely and `Filling.stopped()` returned no gaps. It is now around the search alone:

```java
Generator.GenerationResult pairs;
try {
    pairs = pairsFor(...);
} catch (LinkageError _) {
    pairs = new Generator.GenerationResult(List.of(), List.of(),
            List.of(new GenerationReason.NothingToBuildAgainst(spec.name())));
}
out.put(spec.name(), new Filling(pairs, offered(spec.name(), edges), dispositions(..., pairs, ...)));
```

`Filling.stopped()` is gone. `atCase` asks whether the search had nothing to build against **before** it reads the row list, so the reason is the one that happened rather than the one the empty result suggests. Applicability is unchanged by linking: an arm stays `NotSupported` either way.

## 5 — fixed

An edge with no assessment behind it now throws instead of answering `CannotGenerate(NOTHING_TO_BUILD_AGAINST)`. The gap is established from an assessment, so the two readings disagreeing is the pipeline contradicting itself, and rounding it to a generation failure is the substitution this type exists to prevent.

## 4 — added, and one more thing it turned up

All three answers for a case of an input are reachable, and are now tested:

| model | gap | answer |
| --- | --- | --- |
| sum of unit cases | `Overseas` at 1 | `Generated` |
| case carrying a record field | `Card` at 1 | `CannotGenerate(NO_REPRESENTATIVE)` |
| 13 divided positions, `MAX_AXES` is 12 | `Overseas` at 13 | `NotSupported(NO_AXIS_AT_THIS_POSITION)` |

and `OUTPUT_CASE_UNSPECIFIED` → `NotSupported(NO_STRATEGY_FOR_AN_OUTPUT_CASE)`. Checked against a mutation that answers every case the same way: three of the four fail.

The second row is #602's `feeFor`/`Card` symptom — a case that is writable by hand, reported as one no value can be written for. It arrives here as `CannotGenerate`, which is the right kind; only its reason is the false one, and that is #602's to fix.

**The fallback you flagged was worse than defensive.** `POSITION_NOT_READ` named a cause — a row's value at the position could not be read — that nothing on that path had established. It is now `NO_REASON_RECORDED`: *the search that takes this class left no reason, which is this compiler failing to say rather than anything established about the class.*

It is also not reachable today, for the same reason `CannotGenerate` at a boundary is not: a position is withheld only when some row went unread, which makes the measurement `PARTIAL`, and `Finding.isAdequacyGap()` requires `COMPLETE`. So both dead branches trace to one rule — a gap needs a completed measurement — and both go live from the same direction. I have left them as values rather than throws: unlike the edge case in 5, nothing here proves an inconsistency, and a report is the wrong place to find out.

`mvn clean test` — 3901 tests, 0 failures.

kawasima:
Pushed `de0f4d65`. Confirmed the path first: `Coverages.writabilityOf` returns `ProvenByProjection` from `knownWritable` whatever the attempt did, and `isUnmetGap()` is `Missed && writability.known()`, so an unmet gap whose attempt is `NotAttempted(LINKAGE_FAILED)` reaches `atEdge`.

`GenerationReason.LinkageFailed` and `UnresolvedCombination.Reason.LINKAGE_FAILED` now carry it the whole way:

```java
case BoundaryAssessment.Attempt.NotAttempted absent -> switch (absent.reason()) {
    case NO_CLASSES    -> cannotGenerate(NOTHING_TO_BUILD_AGAINST);
    case LINKAGE_FAILED -> cannotGenerate(LINKAGE_FAILED);
    case A_ROW_IS_ALREADY_THERE, NOT_MEASURED -> throw new IllegalStateException(...);
};
```

The `LinkageError` around the pair search records `LinkageFailed` rather than the other one, and `atCase` maps whichever of the two the search recorded. The two sentences are now:

```
// generation stopped for `pick`: the generated classes would not link, so the decoders a candidate is built through were out of reach
// generation stopped for `pick`: there was nothing to build a candidate against
```

**One more place had the same collapse**, which you did not name and which had the same cause: `offered()` — the edge strategy's own result, the one that survives when a behavior has no rows — mapped both reasons to `NothingToBuildAgainst` too. Fixed with the same switch.

`GenerationReason.NothingToBuildAgainst`'s own doc had argued the two were one fact, on the grounds that *the difference changes nothing an author would do, since neither leaves a row to offer*. That is true about what to do and beside the point about what happened; the doc is rewritten to say what each one is and that having no row to offer is all they share.

Test: the report sentence for each reason, checked against a mutation that prints the collapsed one for both. I could not drive a real `LinkageError` through a test, so what is pinned is the mapping at the report boundary; upstream, both switches are exhaustive over the assessment's four reasons, so a reason added there does not compile until it is placed.

`mvn clean test` — 3902 tests, 0 failures.

kawasima:
Pushed `ed22da87`, and corrected the PR body.

## P2 — fixed

The lines about what could not be written already went through the set that keeps a fact from being said once. The stop reasons did not, and the pair search and the edges are two searches of one behavior — nothing to put a candidate through stops both, each result carries the reason, and the reader was told twice that generation stopped.

Routed through the same `say()` rather than `distinct()`, per your second suggestion: line by line, so the reason's own type does not have to differ for the sentence to be recognised as one already said. `RowsNotRead` yields several lines and each of them goes through it separately.

Pinned by `oneReasonBothSearchesStoppedForIsSaidOnce` — the block for a reason carried by one search equals the block for the same reason carried by both. Checked against a mutation that appends instead of saying.

## The PR body was wrong, and you caught my own counterexample

> `CannotGenerate` at a boundary is not reachable today.

I wrote that, and then two rounds later established the opposite myself while confirming your `LINKAGE_FAILED` finding. The claim was true only of one of the two routes in. The body now says:

> Writability is what makes an unmet edge a gap, and a search that composes a row is one way to establish it — so where the search is what failed, there is no gap, and #602 is what opens that route. `Coverages.writabilityOf` also proves writability from the projection whatever the attempt did, so an edge the projection settles and the attempt could not build reaches `atEdge` today. That is where `LINKAGE_FAILED` arrives.

`NO_REASON_RECORDED` in `atCase` is still not reachable, and for the reason I gave: a withheld position comes from an unread row, which makes the measurement `PARTIAL`, and a gap needs `COMPLETE`. That one I have not contradicted.

Also removed a test case that was saying something its neighbour already said — the row count, where the case beside it holds the whole written block.

## On CI

You are right that I should not have called it green. What I verified is local: `mvn clean test`, 3902 tests, 0 failures, on `ed22da87`'s tree. The Actions run is the repository's to report.

## The four layers

Your summary is the one I would keep:

```
Adequacy      what is missing
Disposition   whether the generator has a capability for it
Generation    what it could and could not compose this time
Presentation  how many rows a person actually writes
```

The three symptoms in #606 read as separate bugs because those four were one path.

Issue #615, opened 2026-08-10
A block-result mismatch is reported at the block's parameters, not at the expression it answered with


kawasima:
A block handed to a helper's function parameter is checked against what that parameter declares. When
the block answers something else, the report is drawn under the block's parameter list rather than
under the expression the block answered with.

> A diagnostic is anchored by the operand that contributed the rejected value, not by the construct
> that happened to initiate the check.

## What happens

```souther
module R5

let applyTwice (f: (Int) -> Int, x: Int) =
    f(f(x))

let used =
    applyTwice(
        (n) ->
            "not an int"
    , 1
    )
```

```
-- FUNCTION  E1805 ----------------------------block.sou:8:9

8|         (n) ->
           ^

The block passed to `f` of `let applyTwice` must return Int, but returns String.
```

`"not an int"` is the `String`, on line 9. The caret is on line 8, under `(n) ->`. The message names
what the block returns and the caret points at what takes its parameters.

## Where it is decided

`HelperTyping.checkFunctionArg` reaches the same constructor from both of its result checks:

```java
// HelperTyping.java:593 — the declared result still carries a type variable
try {
    TypeOps.unify(want.result(), got, bind, symbols, lambda.pos(), "block result");
} catch (CompileException _) {
    throw blockReturnMismatch(h, paramName, want.result(), got, lambda.pos());
}
...
// HelperTyping.java:600 — the declared result is concrete
if (!TypeOps.assignable(got, want.result(), symbols)) {
    throw blockReturnMismatch(h, paramName, want.result(), got, lambda.pos());
}
```

```java
// HelperTyping.java:547
private static CompileException blockReturnMismatch(Ast.FnDef h, String paramName, Type want,
                                                    Type got, SourcePos pos) {
    return CompileException.of(Diagnostic.at(pos)
            .say(new HelperMessage.TheBlockAnswersAnotherType(paramName, h.name(),
                    Type.show(want), Type.show(got)))
            .build());
}
```

`got` is `Elaborator.typeOf(inliner.inline(lambda.body(), ...), lenv, ...)` — the type of
`lambda.body()`. The expression whose type is being refused is in hand at both call sites, and
`Elaborator.region(Ast.Expr)` already builds a region from it. `lambda.pos()` is passed instead.

## Not the same defect as #613

#613 is about `TypeOps.unify` rendering a diagnostic from inside the solving walk. That is visible
here — the `:593` branch assembles `"block result"` and then throws the resulting exception away —
[...]

Issue #611, opened 2026-08-10
A list's element mismatch is drawn under the opening bracket, and the type it conflicts with has no operand to name


kawasima:
A list whose elements disagree is reported with the caret under the opening bracket, and the message
does not say which element disagreed. The reader is told two types are present and left to find both.

## What happens

```souther
module Repro

data Loan =
    { copyId: String
    , memberId: String
    , renewalCount: Int
    }

let ids (loan: Loan) =
    [ loan.copyId
    , loan.memberId
    , loan.renewalCount
    ]
```

```
-- TYPE MISMATCH  E1318 -------------------listcase.sou:10:5

10|     [ loan.copyId
        ^

The elements of this list do not all have the same type.
Hint: One element is String and another is Int; make every element the same type.
```

`loan.renewalCount` is the `Int`, on line 12. The caret is on line 10, under the `[`. Neither the
region nor the message names an element, so a list written one element to a line is searched by
hand, and a long one is searched from the top.

## Where it comes from

`Elaborator` folds the element types left, and hands the fold two types and the list's own position:

```java
// Elaborator.java:349
elem = elem == null ? c.type() : BottomInfer.unifyElem(elem, c.type(), lit.pos());
```

```java
// BottomInfer.java:174
static Type unifyElem(Type a, Type b, SourcePos pos) {
```

The element expression `el` and its Core `c` are in hand at the call. What crosses into `unifyElem`
is `Type`, `Type`, `SourcePos`.

`unifyElem` has a second caller, and it is the one that settles the design question below:

```java
// BinaryElaborator.java:128 — `++`
Type.list(BottomInfer.unifyElem(lo.element(), ro.element(), bin.pos()))
```

## The accumulated type is not an operand

The obvious fix — give `unifyElem` a `Region` for each of the two types — does not have a value to
pass for the left one.

**A left fold's accumulator has no element.** By the time the fold fails, the elements it already
joined have been replaced by their join:

```souther
data Role = Admin | Member | Guest

let rs =
    [ Admin
    , Member
    , 1
    ]
```

```
Hint: One element is Member | Admin and another is Int
```

`Member | Admin` is no element's type. The left side of the mismatch is evidence gathered from a
prefix of the list, not a thing written in the source.

**`++` compares types that no operand has either.** The two types handed to `unifyElem` are
`lo.element()` and `ro.element()` — element types — while the only regions available are
`bin.left()` and `bin.right()`, whole lists. The printed type and the region that would carry it
[...]

PR #616, opened 2026-08-10
Answer a solve with the types, not with a report about them

### LINKED ISSUE #613: A type operation renders a diagnostic instead of returning the failure, so the report is decided where the operand is gone
A type operation knows why solving failed. It does not know which construct in the source should own
the report. Today it decides both: it renders a `Diagnostic` and throws it, at a position and with a
sentence its caller had to supply before the answer was known. Every caller that only wanted to know
whether the types fit pays for a report it discards, and one caller reads a rendered diagnostic back
with a regular expression to recover a fact the operation had in hand.

> Type operations report types. The elaborator that still owns the operand decides where and how a
> mismatch is reported.

#605 established the anchoring half of this — a diagnostic is anchored by the operand that
contributed the rejected value, not by the construct that initiated the check — and moved two
refusals to callers that still had the operand. The API that made those moves necessary is unchanged.

## The two APIs

```java
// TypeOps.java:610
public static void unify(Type param, Type arg, Map<String, Type> bindings,
                         Symbols symbols, SourcePos pos, String what)

// Substitution.java:166
void decide(Type declared, Type actual, Symbols symbols, SourcePos pos, String what)
```

`Substitution.constrain(...)` and `Substitution.hold(...)` (`:80`, `:94`) have the same signature.
All of them take a `SourcePos` and English prose and throw
`TypeMessage.ItExpectedOneTypeAndGotAnother(what, expected, got)` from inside the solving walk.

### 1. One method reports one rule two ways, and the only variable is which API had the operand

`CallElaborator.preservedCall` (`CallElaborator.java:157`) types a kept call's arguments. Value
arguments and function arguments are the same rule about the same call, nine lines apart:

```java
// :165 — value arguments
requireValueArgs(call, params, ca, bind);
        // → Elaborator.requireType, anchored at Elaborator.region(args.get(i))
        // → ItDoesNotHaveTheTypeItNeedsHere("argument N of X")

// :174 — function arguments
TypeOps.unify(declared.result(), answered, bind, ctx.symbols(),
        call.pos(), "argument " + (i + 1) + " of " + call.written());
        // → ItExpectedOneTypeAndGotAnother("argument N of X", ...) at the callee
```

| | value argument | function argument |
|---|---|---|
| semantic rule | same | same |
| argument, callee, program, caller | same | same |
| region | the argument | `call.pos()` |
| sentence | `ItDoesNotHaveTheTypeItNeedsHere` | `ItExpectedOneTypeAndGotAnother` |

`requireValueArgs` states the rule in its own javadoc — "Each value argument held to the parameter it
was given to, at its own position — the refusal `settledByValues` leaves to whoever has the argument
in hand." Nothing about the position is unavailable at `:174`; the API does not take it.

### 2. The solving layer already has a non-reporting mode

```java
// TypeOps.java:624
public static void bindVars(Type param, Type arg, Map<String, Type> bindings, Symbols symbols) {
    unify(param, arg, bindings, symbols, null, null, false);
}
```

The private walk is already parameterised by `refusing`. The split exists; it is exposed as an
all-or-nothing choice — refuse and report, or bind and say nothing — and `bindVars` has one caller
(`CallElaborator.java:237`).

### 3. Five callers use the exception as a boolean

They want the third thing neither mode offers: bind, and tell me whether it fit.

| site | `what` assembled | what happens to the exception |
|---|---|---|
| `CallElaborator.java:174` | `argument N of X` | escapes — the only one |
| `HelperTyping.java:509` | `argument N` | `try 506-510` → `return` |
| `HelperTyping.java:594` | `block result` | `try 593-595` → replaced by `blockReturnMismatch` |
| `HelperParams.java:921` | `closure` | inside `solved()`'s `try 955-986` → `return sig.params()` |
| `HelperParams.java:977` | `argument` | same |
| `BottomInfer.java:98` | `result of X` | `try 97-100` inside `pinResultTypeVars` → swallowed |

`Substitution.decide` is the same picture: `HelperParams.java:782` (`closure`) and
`HelperParams.java:791` (`closure result`) are caught at `793`; `Elaborator.java:655`
(``the result of `...` ``) is not, and reports at `ex.pos()`.

### 4. A structural fact is rendered to prose and recovered by regex

`CallElaborator.java:447` catches a step-typing failure and has to know whether it failed because an
accumulator's element type was never fixed:

```java
} catch (CompileException stepError) {
    int seed = BottomInfer.untypedEmptySeed(args, signature, bind, call.pos());
    if (seed < 0 || !BottomInfer.reportsUnresolvedBottom(stepError)) {
        throw stepError;
    }
```

`BottomInfer.reportsUnresolvedBottom` (`BottomInfer.java:108`) answers it by matching
`(^|[^\p{Alnum}])_([^\p{Alnum}]|$)` against `diagnostic().diff().actualType()`, `expectedType()`, and
every value in `diagnostic().values()`. The round trip is complete:

```
structural type fact → diagnostic → CompileException → rendered type string → regex → structural type fact
```

The check that spells the bottom as `_` and then looks for `_` is the abstraction boundary pointing
the wrong way.

Removing it is not part of this issue. `stepError` is whatever `Elaborator.resolveStepBinding`'s
`elaborateBlockArg` threw — an arbitrary diagnostic from elaborating the whole step body, at
arbitrary depth — not a `TypeOps.unify` failure. Giving the solving walk a failure value does not
give this caller the fact it is looking for; that needs an unresolved bottom to be reported
structurally wherever it arises. The function stands as evidence of the boundary, and its removal
is separate work.

### 5. The prose is discarded at five of six sites

`what` folds four things into one string — which operand, where it is written, what kind of position
it is, and the wording — and none of them can be read apart again. At five sites it is assembled and
then dropped by a `catch` in the same method. It is a symptom of the boundary, not the defect.

## What to expect instead

The solving walk returns success or a structural failure, and the elaborator that still owns the
operand turns a failure into a report. The failure carries what the walk knows — the two types, and
which rule was broken where two rules share a walk — and nothing about source or wording.

The exact type is not fixed here. What this issue requires is that a type operation returns a
structural failure value (or an equivalent internal representation carrying the same information)
rather than a rendered diagnostic, and that `SourcePos` and prose leave the solving layer.

`Substitution.constrain` and `Substitution.hold` come with `decide`: `constrain` is `decide` then
`hold`, so `decide` cannot change shape alone, and `hold` throws
`DeclarationMessage.ItExpectsAnotherType(what, ...)` from the same walk under the same defect. All
three are in scope.

## A failed solve must not commit partial bindings

`unify` writes into `bindings` as it walks and throws part-way, so on failure the map is partly
updated. Four callers get away with it by discarding the map; `pinResultTypeVars` does not, and
already implements the contract by hand:

```java
// BottomInfer.java:96
Map<String, Type> probe = new HashMap<>(bind);
try {
    TypeOps.unify(result, expected, probe, symbols, pos, what);
    bind.putAll(probe);
} catch (CompileException _) {
    // the expected type does not fit this result; leave bind untouched
}
```

This is the semantics the speculative callers want: try solving, commit what was learned on success,
leave nothing behind on failure. Returning a failure value without settling this only replaces
exception control flow with value control flow, and leaves each caller to decide whether its map is
trustworthy. A failed solve must not leave observable partial bindings.

## Not in scope

- The block-result report at `HelperTyping.java:594` is anchored at `lambda.pos()` by
  `blockReturnMismatch`, which the caller constructs after catching. Fixing the solving API does not
  move it. Tracked in #615.
- `BottomInfer.unifyElem(Type, Type, SourcePos)` has the same shape, but the question it raises —
  what source evidence corresponds to an accumulated element type, given that `++` compares
  `lo.element()` and `ro.element()` while only `bin.left()` and `bin.right()` exist in the source —
  is a diagnostic evidence question rather than this boundary. Tracked in #611.

kawasima:
Closes #613.

A type operation knew which two types did not go together, and was made to decide
the report as well: it was handed a `SourcePos` and an English sentence before the
answer was known, and threw a rendered `Diagnostic` from inside the solving walk.
The position it could offer was whatever the caller had spare, and which operand
the report was about survived only as prose.

## What changed

`Fit` — the answer a solving walk gives. `Fits`, or `Disagrees(expected, actual)`
carrying the two types at the position that disagreed. No source, no wording. It
takes the shape `ArithmeticCheck` already takes in this package, for the same
reason: the answer carries what was broken, and the diagnostic is chosen from it
rather than re-derived beside it.

`TypeOps.unify(param, arg, bindings, symbols)` returns a `Fit`. `SourcePos` and
`String what` are gone, and so is the `SourcePos` import.

`Substitution.decide` and `hold` return a `Fit`. `constrain` moved out of
`Substitution` into `Elaborator`, where it still bundles the two rules but each one
states itself once and draws its report at the operand. `Substitution` now imports
nothing from `diag`.

`Elaborator` gained the three reporting functions — `doesNotFit`, `decide`, `hold`
— so one rule has one sentence and every region comes from an operand.

**A failed solve commits nothing.** The walk stopped at the first position that
disagreed and left behind whatever it had bound on the way there, which a caller
could not tell from what was in the map before. `pinResultTypeVars` was already
working around it with a scratch copy; the copy is the contract, so it is written
where the walk is and the workaround is gone.

## What a reader sees

An argument held to what an expansion's binding declares is now underlined whole
rather than at its first character:

```
-- TYPE MISMATCH  E1317 -------------------------exp.sou:9:9

9|         String.append(a, b)
           ^^^^^^^^^^^^^
```

A kept call's function argument is refused where the block is written, by the
sentence a value argument is refused by, instead of at the callee by a second
sentence.

## Tests

- `AFailedSolveCommitsNothingTest` — a variable settled before the disagreement is
  not kept, a walk that fits keeps what it settled, and the answer names the
  position that disagreed rather than the pair the walk started from.
- `AKeptCallsFunctionArgumentIsRefusedWhereItIsWrittenTest` — the block is
  underlined where it is written, and refused by the same sentence as a value
  argument.
- `AnExpansionRefusesAnArgumentWhereItIsWrittenTest` — the argument's whole region,
  not its first character.

`mvn clean install`: 6515 tests, all green.

## Left out on purpose

`BottomInfer.reportsUnresolvedBottom` stays. It recovers a structural fact by
matching a regular expression against a rendered type string, which is this same
boundary pointing the wrong way — but the exception it inspects comes from
elaborating a whole step body at arbitrary depth, not from a solve, so a failure
value at this boundary does not supply what that caller asks for. Removing it needs
an unresolved bottom to be reported structurally wherever it arises.

The block-result report at `HelperTyping:594` is anchored at the lambda by a
diagnostic the caller builds after catching, so this change does not move it —
#615. The list-element mismatch is #611, which is blocked on a different question:
what source evidence corresponds to an accumulated element type.

### DIFF
--- a/souther-compiler/src/main/java/souther/compiler/check/BottomInfer.java
+++ b/souther-compiler/src/main/java/souther/compiler/check/BottomInfer.java
@@ -84,22 +84,17 @@ public final class BottomInfer {
     }
 
     /** Best-effort: bind the type variables of {@code result} from an {@code expected} type the context
-     * pushed down, into {@code bind}. Unifies into a scratch copy and merges only on success, so an
-     * expected type that does not fit the result leaves {@code bind} untouched and the ordinary checks
-     * report the mismatch rather than this throwing early. Shared by the checker's call typing and the
-     * backend's fold materialisation so both pin the same accumulator type (issue #70). */
+     * pushed down, into {@code bind}. An expected type that does not fit the result leaves
+     * {@code bind} untouched and the ordinary checks report the mismatch rather than this refusing
+     * early — which is what a solve that does not fit leaves behind anyway, so nothing is asked of
+     * the walk here beyond its own contract. Shared by the checker's call typing and the backend's
+     * fold materialisation so both pin the same accumulator type. */
     public static void pinResultTypeVars(Type result, Type expected, Map<String, Type> bind,
-                                         Symbols symbols, SourcePos pos, String what) {
+                                         Symbols symbols) {
         if (expected == null) {
             return;
         }
-        Map<String, Type> probe = new HashMap<>(bind);
-        try {
-            TypeOps.unify(result, expected, probe, symbols, pos, what);
-            bind.putAll(probe);
-        } catch (CompileException _) {
-            // the expected type does not fit this result; leave bind untouched
-        }
+        TypeOps.unify(result, expected, bind, symbols);
     }
 
     /** Whether a step-typing error is the unresolved-bottom error (an operand/branch reported as the
--- a/souther-compiler/src/main/java/souther/compiler/check/CallElaborator.java
+++ b/souther-compiler/src/main/java/souther/compiler/check/CallElaborator.java
@@ -116,8 +116,7 @@ public final class CallElaborator {
         }
         Type declared = entry.signature().result();
         Map<String, Type> bindings = new HashMap<>();
-        BottomInfer.pinResultTypeVars(declared, expected, bindings, ctx.symbols(),
-                v.pos(), "the type of " + lib.qualified());
+        BottomInfer.pinResultTypeVars(declared, expected, bindings, ctx.symbols());
         return new Core.Call(new ReachName.OfLibrary(lib), List.of(),
                 TypeOps.toBottom(TypeOps.substitute(declared, bindings)), v.pos());
     }
@@ -171,8 +170,15 @@ public final class CallElaborator {
                 // signature and nothing more. The fold rule that reads a step's result as an
                 // accumulator to grow is one operation's meaning, and an operation kept standing is
                 // kept because its meaning belongs to whoever reads it, not to this.
-                TypeOps.unify(declared.result(), answered, bind, ctx.symbols(),
-                        call.pos(), "argument " + (i + 1) + " of " + call.written());
+                //
+                // Refused here rather than inside the walk, and by the sentence a value argument is
+                // refused by: both kinds of argument are one rule, and this is the reader that
+                // still has the argument to point at.
+                if (TypeOps.unify(declared.result(), answered, bind, ctx.symbols())
+                        instanceof Fit.Disagrees d) {
+                    throw Elaborator.doesNotFit(call.args().get(i), d.actual(), d.expected(),
+                            "argument " + (i + 1) + " of " + call.written());
+                }
             }
         }
         return new Core.PreservedCall(call.denotes(), ca.cores(),
@@ -210,8 +216,7 @@ public final class CallElaborator {
                                              CheckContext ctx) {
         Map<String, Type> bind = new HashMap<>();
         if (params.stream().anyMatch(Type.FnOf.class::isInstance)) {
-            BottomInfer.pinResultTypeVars(result, expected, bind, ctx.symbols(),
-                    call.pos(), "result of " + call.written());
+            BottomInfer.pinResultTypeVars(result, expected, bind, ctx.symbols());
         }
         // Each value argument is asked once, here, in the order it is written. What the ordering
         // below decides is which of them settles a variable first, and nothing about how many times
--- a/souther-compiler/src/main/java/souther/compiler/check/Elaborator.java
+++ b/souther-compiler/src/main/java/souther/compiler/check/Elaborator.java
@@ -652,14 +652,14 @@ public final class Elaborator {
         if (declaredResult != null) {
             // What the body answers decides a variable the arguments left open — a result the
             // signature relates to a function parameter rather than to a value one.
-            decided.decide(declaredResult, body.type(), ctx.symbols(), ex.pos(),
+            decide(decided, declaredResult, ex.body(), body.type(), ctx.symbols(),
                     "the result of `" + shown(ex) + "`");
             if (ex.callee() instanceof ValueName.Local) {
                 // A function the caller supplied, expanded where the callee applies it. What it was
                 // declared to answer is what the caller was held to when it handed the function
                 // over, so a body answering something else is the caller's error and is reported
                 // here — a helper's own declared return is its promise, and is reported against it.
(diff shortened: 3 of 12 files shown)


kawasima:
Added the contract test you asked for, and the javadoc note.

`whatWasSettledBeforeTheWalkIsStillThereAfterwards` starts from a map that already holds a reading and asserts it is still there after a walk that disagrees. You were right that the empty-map case does not pin the contract: with the rollback written as `bindings.clear()` the three tests that were there all pass and only the new one fails —

```
Tests run: 4, Failures: 1
AssertionFailedError: a walk that did not fit left the map as {} ==> expected: <{'settled=STRING}> but was: <{}>
```

On `Substitution.decide`: agreed it is not a blocker and agreed it should not be made symmetric with `TypeOps.unify`. The javadoc now says why — what a walk decides is the application's decision from the moment it is made, whereas the map `unify` settles belongs to its caller — and states what a caller may do after `Disagrees`, which was the part nothing had written down. Every caller today either refuses at once or drops the whole `Substitution`.

`mvn clean install`: 6516 tests, green.

PR #618, opened 2026-08-10
Let an expression say where it ends, not how wide it looks

### LINKED ISSUE #612: An expression starts somewhere and ends nowhere, so a report guesses how wide it is
Every `Ast.Expr` carries where it starts and nothing about where it ends.

```java
// Ast.java:1481
record Apply(Expr function, List<Expr> args, ConstructionOrigin origin, String appliedAs,
             SourcePos pos) implements Expr {}
```

So a report that wants to underline an expression has to guess how wide it is, which is what
`Elaborator` does — a table of node kinds, each measured from what it happens to know:

```java
// Elaborator.java:1437
public static int width(Ast.Expr e) {
    return switch (e) {
        case Ast.Var v -> v.name().length();
        case Ast.StringLit s -> s.value().length() + 2;
        ...
        case Ast.Apply c -> c.written().length();
        default -> 1;
    };
}
```

## What it costs

An application is underlined as far as its callee name and no further. #605 moved an argument's
mismatch onto the argument, and where the argument is itself a call the best available answer is:

```
8|         String.length(b)
           ^^^^^^^^^^^^^

argument 2 of String.append does not have the type it needs here.
```

The line is right and the operand is named. `(b)` is part of the expression whose type did not fit
and is not covered.

A literal is measured from its decoded value, so an escape makes the underline short:

```
8|     on = Date("2026-02-30\tx")
                 ^^^^^^^^^^^^^^

`2026-02-30	x` is not a Date.
```

The literal is 15 characters of source and 14 are underlined — `\t` is two characters written and
one in the value `width` measures.

Anything the table has no case for is one column, which is most of the language: a binary operation,
a match, a construction, a comprehension.

## What is already there

The CST carries both ends of every node:

```java
// SyntaxElement.java:16,19
int start();
int end();
```

`AstBuilder` reads only the first:

```java
// AstBuilder.java:1582
private SourcePos pos(SyntaxNode n) {
    return lines.posOf(firstMeaningfulToken(n).start());
}
```

and elsewhere in the same file it already turns an end offset into a `SourcePos` to build a region:

```java
// AstBuilder.java:1607
.map(close -> new Region(open, lines.posOf(close.end())))
```

So this is carrying a value the frontend computes and drops, not deriving one that is missing.

## What to expect instead

An expression answers where it ends. `Elaborator.width` goes away, and a report underlines the
expression it is about rather than a per-kind guess at it.

The cost is that every `Ast.Expr` record gains the field and every construction site passes it,
including the lowering passes that synthesize nodes — which have no source text and have to say so
rather than inventing an end.

kawasima:
Closes #612 once merged.

## What was wrong

Nothing in the tree said how far an expression reached, so a report worked the width out where it was
written — a table of node kinds in `Elaborator`, each measured from whatever that kind happened to
know. Of the 21 expression kinds it named seven and a companion `region` named three, leaving
fourteen at one column. The two disagreed about the three they both answered: `region(Var)` gave the
occurrence the parser read and `width(Var)` gave the length of the canonical name, so the defect
`AReportUnderlinesWhatTheAuthorTypedTest` exists to stop was fixed on one path and live on the other.

Carrying an end would not have been enough. A binary operation is positioned at its operator, a field
read at its field, and an application built from a pipe at the callee on the right of it — each of
those is where a report about that node belongs, so a region built from the position and any width at
all starts inside what the author wrote.

The other half was the measuring. A width read off a decoded value is a claim about the value, and
the two part company wherever the source spells something the value does not keep.

## What it does now

Where a report is anchored and what it covers are two questions, and `Ast.Expr` answers both. The
extent is read once in `AstBuilder` off the first and last meaningful token; the position stays where
it is, operator and field included.

Parentheses are part of the extent. `PAREN_EXPR` is reduced away and its characters are still in the
file, so the reduced node is stamped with the bracketed node's region — a report can narrow an extent
it was given but not widen one it was not.

A node no one wrote has no extent, rather than a zero-width region at its anchor. `WrittenName`
carries the same distinction: a synthetic name keeps an anchor and holds no segments, so it no longer
takes the canonical name's length at a place holding something else. Reports ask `reportedAt()`,
which is the characters where there are some and the anchor where there are none.

Rewrites carry the extent or replace it by the rule that already decided the position:
`HelperInliner.keepsItsPositions` stamped a copied prelude body with the call site so an error inside
it points at the user's call, and it now stamps both ends of it.

`Elaborator.region` and `Elaborator.width` are deleted; the 19 sites that read one of them read the
expression's own region.

## Before / after

```
6|     String.append(a, String.length(b))          6|     String.append(a, String.length(b))
                        ^^^^^^^^^^^^^                                     ^^^^^^^^^^^^^^^^

6|     Date("2026-02-30\tx")                       6|     Date("2026-02-30\tx")
            ^^^^^^^^^^^^^^                                 ^^^^^^^^^^^^^^^

6|     if a > 0 then a + 100 else b                6|     if a > 0 then a + 100 else b
                       ^                                                ^^^^^^^

8|     String.append(a, c.count)                   8|     String.append(a, c.count)
                          ^^^^^                                          ^^^^^^^

6|     String.append(b, [a, a, a])                 6|     String.append(b, [a, a, a])
                        ^                                                 ^^^^^^^^^

6|     String.append(a, 007)                       6|     String.append(a, 007)
                        ^                                                 ^^^
```

## Two defects the fabricated width had been covering

Three passes rebuilt a `FieldInit` or a `FieldAccess` from the canonical name and a position when
only the value or the target had changed, replacing the occurrence the author made with a synthetic
name. This was invisible while `synthetic` fabricated a region of the same width. They rewrite
through `withValue` and `withTarget` now.

An application whose callee a lowering replaced with a binding had nowhere to point, the binding
being written nowhere. The binding carries the callee's extent, and `Apply.appliedAt()` reads it from
the callee rather than from the name.

## Not in this change

`HumanRenderer` still draws one caret under a region that ends on a later line. The extent a
multi-line expression carries is right and what a terminal draws under it is not yet; the acceptance
test pins the region so a published one and an editor reading it get the whole of it today.

## Tests

`AnExpressionIsUnderlinedOverWhatWasWrittenTest` — nine rows, each cutting the region out of the
source and comparing the characters, since an assertion on the number of columns passes on a region
of the right size in the wrong place. It covers an application over its arguments, a binary operation
from its left operand, a field read over its target, a list, an escape, a decomposed spelling, a
leading zero, a parenthesized expression, and a multi-line one.

`ADiagnosticPointsAtTheOperandThatSuppliedTheValueTest` expected an argument to be underlined as far
as its callee's name in two rows. Its own failure messages named the whole argument; they now expect
it.

Full clean build: 6,505 tests, no failures.

## ADR

ADR-0102 — an anchor is not an extent.


kawasima:
Both blocking points confirmed in the code and fixed in 48b8d8ec. The first turned out to have a
third instance the review did not name, which the regression test found.

## 1. `renameVar`

Confirmed. It rebuilt every name from `v.name()` and a position, which is `WrittenName.of` measuring
the canonical spelling at the anchor — the rule this branch removes, on the other side of a rewrite.

Split three ways, by whether the characters the name stands at still spell it:

- a parameter read becoming the binding the expansion made — a different name, written nowhere;
  `Copy.of(Ast.Binder)` already says exactly this about the binder;
- a name in a copy stamped with a call site — the occurrence is in the callee's file;
- anything else — the occurrence comes through with the positions it keeps.

## The third instance

`Apply(String fn, ...)` built its callee through `WrittenName.of` too. `List.map` expanded into a
caller left `List.foldFrom` — a name the prelude rewrite chose — claiming thirteen characters of the
call site, `authored()` and all. Both spelling-taking `Apply` constructors build a name written
nowhere now, and `e |> f` hands the callee over as the `Var` it is rather than as its canonical
spelling.

`WrittenName.of` reads a spelling as if it were the source at that place. That is true of a token the
parser has just read and false of every other caller, which is what `Var.respelled` and
`FieldAccess.restamped` are for.

## 2. `FieldAccess`

Confirmed and fixed, symmetric with the `FieldInit` beside it, through `FieldAccess.restamped`.

Being straight about the evidence: **I could not construct an input that observes it.** A field read
inside a foreign body needs a foreign helper that has one, and the prelude has none; a helper
parameter read only through a field must be annotated (E1811), so the read is settled in the callee's
own module and a report about it does not arise at expansion. So this is written for the rule and for
the occurrence travelling into another file's tree, not for a diagnostic I can show. If you know a
route to one, I will add the row.

## 3. Literals and `unreachable`

Measured rather than argued. Stamping them costs nothing — the full suite is green with it — so the
exemption was not paying for anything, and it made the stated rule false: a `Date("…")` inside a
prelude helper reported at a line of `souther.list` while everything around it reported at the call.
The copy is stamped whole now, and the ADR says so rather than being weakened.

## The test

`AnExpansionKeepsTheSpellingItCopiedTest` — a property rather than the two cases, since the two cases
would not have found the third instance: every name an expansion says the author wrote has to be
spelled by the characters it points at, part by part against the source. Part by part because
`WrittenName.region()` is one continuous stretch and a qualified name written over a line break has
the break inside it; only the segments are the name.

Three rows: a decomposed parameter copied into a caller, a qualified name written over a line break,
and a body copied out of another module. Each was checked against a mutation — reverting `renameVar`
fails the first, reverting the `Apply` constructor fails the third — so neither is an empty probe.

Full clean build: 6,508 tests, no failures.

kawasima:
All five confirmed in the code and fixed in 5f88e714.

One correction on the premise, because it changes how to read the previous round: `e640ac1` **was** the
head and it **did** contain the earlier fixes — `git merge-base --is-ancestor 48b8d8ec e640ac11` is
true, and `48b8d8ec` is the second of the three commits. Nothing was unpushed. So Blocking 3 was a
live finding about pushed code, and a correct one: routing the fix through the `Apply` constructor is
what handed the call's region to the callee. That was mine, introduced in the previous round.

## 1. `ExampleStatements`

Confirmed and fixed. `Statement` carries a `Region` now; `SourceRef` + `int width` is gone, and
`Output` passes the region straight to `Diagnostic.at` and `secondaryIn`.

One correction to the failure mode, because it changes what the test can hold. The Binary case you
describe is not reachable: a fixture must be a literal or a construction (E1908), and every one of
those is anchored where it starts, so `pos + width` and the region agree. What the old shape did lose
is the second line: `Region.ofWidth` ends on the line it began on, so a row stating a construction
written over three lines was marked at its first character. That is what the two new rows hold, and
reverting `Output` to `Region.ofWidth(region.start(), region.sourceSpan())` fails both.

## 2. `Var.denoting`

Confirmed and fixed, in `denoting` and in the callee `Resolve.applied` settles — the latter rebuilt
the callee rather than answering it, which loses `(f)(x)` the same way.

`aParenthesizedNameIsUnderlinedOverItsParentheses` is the row: `String.append(b, (a))`. Reverting
`denoting` to the three-argument constructor gives `a`.

## 3. Callee extent vs application extent

Confirmed and fixed. The spelling-taking `Apply` constructors give the callee no extent — they have
only the application's, and it is not the callee's — and the two callers that do have it,
`HelperNames.qualifyHelpers` and the prelude rewrite in `HelperInliner`, build the `Var` from
`call.function()`'s own place.

## 4. `WrittenName.then`

Confirmed and fixed. `then` makes the whole name unwritten when either part is, keeping the anchor,
and the constructor refuses a spelling-less name that holds segments.

## 5. The `Var` invariant

Confirmed and fixed: an authored name in an expression written nowhere is refused. `Region.encloses`
is vacuously true when either side is absent, which is right for what it answers and is not the
question here, so the check is its own.

## Full build

6,571 tests, no failures.

## One thing worth deciding before the next round

Three rounds, and the same cause each time: a place that works the extent out instead of carrying it.
The locations differ every time and the cause has not. What has actually been finding them is one
property — every name an expansion says the author wrote is spelled by the characters it points at —
which caught the prelude rewrite that neither of us had named.

That property covers the `WrittenName.of` family. It does not cover the two other shapes this round:
a region narrowed by a rewrite that keeps the same name (`denoting`), and a region rebuilt from a
start and a count (`Statement`). The `Var` constructor now closes the first for names; nothing closes
the second in general.

If you would rather not run a fourth round of the same, the alternative is a check over the corpus
that holds every expression's region against the source it claims, rather than more cases. I have not
written it, because it is a different size of thing from this branch and I would rather you decide
whether it belongs here or after.

Issue #624, opened 2026-08-10
Coverage counts the same helper obligation once per inlined occurrence


kawasima:
`souther examples` counts a non-recursive helper's arms once **per call site**, so the number of arms a behavior owes is a product of its call graph rather than a count of what its author wrote. §example-branch-coverage says the measure is not that wide.

## What it does today

A three-arm table, written once, called from one behavior and twice from another:

```
module example.arms

data Grade = Int
    invariant value >= 0

data A
data B
data C
data Band = A | B | C

// 3-arm table, written once.
let rate (band: Band): Grade =
    match band with
        | A -> Grade(1)
        | B -> Grade(2)
        | C -> Grade(3)

// Called once.
behavior once : (band: Band) -> Grade
    constructs Grade
let once (band) = rate(band)

// Called twice from one body.
data Left
data Right
data Side = Left | Right

behavior twice : (side: Side, band: Band) -> Grade
    constructs Grade
let twice (side, band) =
    match side with
        | Left -> rate(band)
        | Right -> rate(band)

example once
    | "A" : (A) -> Grade(1)
    | "B" : (B) -> Grade(2)
    | "C" : (C) -> Grade(3)

example twice
    | "left A" : (Left, A) -> Grade(1)
    | "left B" : (Left, B) -> Grade(2)
    | "left C" : (Left, C) -> Grade(3)
```

```
  once                     implemented   rows 3    pending 0
    branch      3/3
  twice                    implemented   rows 3    pending 0
    branch      4/8
      · no row goes through `case Right` (31:5)
```

`twice` writes two forks. It is measured against eight arms: its own two, plus `rate`'s three counted twice. Only one of the four missing arms is named, because the other three are `rate`'s second copy and nothing has reached the fork they sit under.

## Why this is not what the measure says it is

> This is *branch-arm* coverage and nothing wider. Going through both arms of two nested conditions is four arms and says nothing about whether their combinations were tried, so covering every arm is a lower bound on covering what a body can do rather than the same thing.

Eight arms in `twice` is exactly the combination — `side` × `band` — that sentence disclaims. Whether that is the better measure is a separate question; today the specification describes one measure and the report computes another.

The inlining itself is stated and is not what this is about:

> What a call answers is not read. A non-recursive helper is already inlined into the body that uses it, so its arms are here.

Reading a helper's arms as the caller's is defensible. Counting them once per syntactic call site is what turns a fixed obligation into a multiplied one.

## The cost is not readable from the source

[...]

Issue #622, opened 2026-08-10
The adequacy report draws boundary lines on Int, Decimal and Date, but not on String or DateTime


kawasima:
A comparison in a body is what `souther examples` reads to derive a partition axis and the boundary rows on it. It reads that comparison for `Int`, `Decimal` and `Date`, and does not read it for `String` or `DateTime`. All five are ordered types — `<`, `<=`, `>`, `>=` are defined on all of them — so an author who writes the same rule over a `String` or a `DateTime` silently gets no partition, no boundary, and no generated row.

## What it does today

One file, one shape of body throughout — `guard <position> < <literal> else ...`:

```
module example.ordering

data Amount = Int
    invariant value >= 0
data YearMonth = String
data Rate = Decimal

data Bigger
data Smaller
data Size = Bigger | Smaller

data Newer
data Older
data Era = Newer | Older

behavior classifyAmount : (a: Amount) -> Size
    constructs Bigger, Smaller
let classifyAmount (a) = {
    guard a.value < 5000 else Bigger
    Smaller
}

behavior classifyRate : (r: Rate) -> Size
    constructs Bigger, Smaller
let classifyRate (r) = {
    guard r.value < 0.5m else Bigger
    Smaller
}

behavior classifyDate : (d: Date) -> Era
    constructs Newer, Older
let classifyDate (d) = {
    guard d < Date("2026-08-01") else Newer
    Older
}

behavior classifyMonth : (m: YearMonth) -> Era
    constructs Newer, Older
let classifyMonth (m) = {
    guard m.value < "2026-08" else Newer
    Older
}

behavior classifyMoment : (t: DateTime) -> Era
    constructs Newer, Older
let classifyMoment (t) = {
    guard t < DateTime("2026-08-01T00:00:00") else Newer
    Older
}
```

`souther examples ordering.sou --generate --boundaries`:

| position | partition | rows generated |
|---|---|---|
| `Amount` (`Int`) | `axes 1` | `Amount(0)`, `Amount(5000)`, `Amount(4999)` |
| `Rate` (`Decimal`) | `axes 1` | `Rate(-0.5m)`, `Rate(0.5m)` |
| `Date` | `axes 1` | `Date("2026-07-31")`, `Date("2026-08-01")` |
| `YearMonth` (`String`) | none | none |
| `DateTime` | none | none |

The two that fail report the same line:

```
  classifyMonth            implemented   rows 0    pending 0
    partition   not measured (no partition axis was derived at any position)
[...]

PR #630, opened 2026-08-10
Count an arm once for the fork it was written as, not once per call site

kawasima:
Fixes the measure in #624: `souther examples` counted a helper's arms once per call
of that helper, so what a behavior owed was a product of its call graph rather than a
count of the forks its author wrote.

## What was wrong

The measure is taken over the tree that runs, and by then every non-recursive helper
call has been replaced by a copy of the callee's body. `Ast.Expansion` marks that
boundary but is lowered to a chain of `Core.LetIn`, so nothing downstream could say
two forks came from one. `CoverageSites` numbered a site per fork instance and
`Adequacy` counted those directly.

It was not only the arms. A guard inside a helper drew its line once per call too, so
the boundary measure listed the same value twice, spelled identically, with nothing in
either line to tell them apart. Both are the same defect.

And it could make a false statement at a real line: an arm was named as unreached at a
position a row demonstrably goes through, because the copy meant was the one under an
arm nothing had reached.

## What changed

Three commits.

1. **Carry where a fork was written.** `CoverageOrigin(module, ordinal, lowered)`,
   minted in `AstBuilder` where a construct is recognised and carried through every
   rewrite and every copy. Nothing reads it and no number moves.
2. **Count an arm once for the fork it was written as.**
   `CoverageSites.Obligation(behavior, origin, part)`, with the quotient taken in
   `BranchEvidence` and `Coverages.assess` and nowhere else. Occurrence probes are
   unchanged, so `GuardEdge`, `ProducedCases` and `GuardReachability` keep reasoning
   one occurrence at a time — a proof about a copy under one call site is not a proof
   about the copy under the next.
3. **Take the structural hash out of identity.** `Site.fingerprint` was dead and could
   not have served: it says whether two arms are alike, not which one the author wrote.

Both readings are existential across a fork's occurrences. An arm is owed where any
occurrence can be reached and covered where a row went through any of them, so a row
that keeps its element at one `List.filter` and drops it at the next has been through
both arms of the one fork they copy.

## Numbers

| model | before | after |
| --- | --- | --- |
| the issue's `twice` | `branch 4/8` | `branch 4/5`, one arm named |
| helper calling a helper | `branch 3/18` | `branch 3/7` |
| two sequential `List.filter` | `branch 2/4` | `branch 2/2` |
| guard inside a helper | `boundary 0/5` | `boundary 0/3`, no duplicate line |
| two alike forks written separately | `branch 2/6` | `branch 2/6` (must not move) |
| `--generate --boundaries` rows | 5 | 5 (must not move) |

## On the issue's framing

"the combination coverage the measure says it is not" describes the first reproduction
but not all of it: two sequential calls of one stdlib helper are neither exclusive nor
a combination, and they doubled as well. The cause is that obligations were
call-site-sensitive; forcing a Cartesian product is one consequence of that.

## Schema

`schemaVersion` stays at 1. `arms` and `covered` change value and `unreached` changes
length; no field is added, removed or renamed.

## Checks

`mvn clean test` is green on every module at each of the three commits, and each was
built and run on its own before the next was written.


kawasima:
Thanks — the second finding is right and is fixed. On the first, the identity was
wrong in the way you describe, but the merge it would cause cannot occur, and finding
out why changed what the fix should be.

## 1. Boundary identity: the premise was wrong, the merge is not reachable

I built both shapes: an own higher-order helper, and two `List.filter` calls with
different numeric thresholds. Neither derives a guard boundary at all.

```
viaParam    partition   not measured (no partition axis was derived at any position)
            boundary    0/1        ← the invariant's minimum, nothing else
viaHelper   boundary    0/3        ← guard@14:5, twice, as expected
```

The reason is one step earlier than the partition. Applying a function parameter
reduces to the predicate's body under a binding for its argument, so the fork's
condition is a `Core.LetIn` — I probed it to be sure. `CoverageSites.comparisons`
numbers a comparison only where the condition is one or is built from `&&` and `||`,
and it does not descend a binding. An injected predicate gets **no comparison site**,
so there is no boundary obligation for two of them to be merged into one.

So the old key was sound. What it rested on was a line in a walk that says nothing
about why, which is the part worth removing rather than the count.

`Ast.Binary` and `Core.Binary` now carry a `CoverageOrigin` of their own, minted where
the operator is read and carried through every rewrite, and a comparison's obligation
is keyed on it. The key no longer assumes the comparison was written where the fork
was. A walk that later learns to descend a binding needs nothing here changed.

A comparison the discharge reader rebuilds out of a preserved call carries
`CoverageOrigin.unwritten()`, and numbering refuses one rather than letting it pass
for an obligation.

### On the regression rows you asked for

The one you named — two separately written predicates, same value, two obligations —
cannot be written today: there are zero comparison sites either way, so it passes on
any identity. I did not add it.

The two that do discriminate are in
`AComparisonBelongsToWhoWroteItNotToTheForkTestingItTest`: a helper's own comparison
copied at two call sites is one line, and two comparisons of one condition stay two
through the same copies (that second one fails if a key collapses a fork's
comparisons together). The class states why the third case has no row and what would
have to change for it to get one.

## 2. `CoverageOrigin.ordinal` wording

Correct — the builder does not number in source order. A statement guard folds the
rest of the block before taking its own number, and a comprehension takes its number
after building its parts. The Javadoc no longer claims source order or that adding a
fork above moves the one below; it says a deterministic builder-local ordinal, which
is what identity needs. `fork()` is also renamed to `construct()`, since it numbers
comparisons now too.

## 3. On `whicheverSawMore`

Agreed, and left as is. If `Attempt` and `Writability` gain provenance independent of
`Coverage`, the merge has to become a join per axis rather than winner-takes-all on
the assessment.

## Rebase

`develop` moved (#622 / #629) and is merged in. The overlap was `GuardThresholds` and
`Adequacy`, in hunks far from mine, and both merged clean. Worth noting for this
thread: #622 widened what a line can be drawn on to `Carrier.ofValue`, which is
exactly the kind of upstream widening that would have turned the latent identity into
a live one. I re-measured the injected case after merging; it still derives nothing.

`mvn clean test` green on every module.

PR #633, opened 2026-08-11
Give a carrier's count a type, and draw a line on an enumeration

### LINKED ISSUE #628: A carrier's count is not a value, and an enumeration is a carrier
The interval algebra carries every ordered value as a `BigDecimal`, and a model's own numbers are `BigDecimal` too. Nothing separates them, so the count a carrier is made of can be written wherever a value of the position belongs, and the compiler has no reason to object. Both carriers added so far went wrong here, in different ways: the day count by two readers of it disagreeing, and the second count by reaching a row.

An enumeration is the next carrier, and its count is an ordinal — so a leak there surfaces as a small integer, which is the most plausible-looking of the three wrong values and the one a report is least likely to give away.

These are one piece of work because the second is the reason to want the first, and doing them in the other order adds the leak before the type that stops it.

## The count leaks, and it has

`Date` joined the algebra by counting days from an epoch (6dcc9484). The commit message records four copies of one projection found while doing it — reading a number out of an observed value was spelled in `NumericTerm`, `BoundaryAssessment`, `Coverages` and `Generator` — and unified three of them onto the term's reader.

`DateTime` joined it on #622 and found a fifth, which was the other direction: a count reaching a row rather than two readings of one disagreeing. `Generator.valueOf` wrote a temporal as a date where it could read one and fell through to `numberOf` where it could not:

```java
if (value instanceof ObservedValue.Temporal date && Dates.dayOf(date.iso()) != null) {
    return Witnesses.wrapped(type, FixtureTemplate.date(date.iso()), symbols);
}
BigDecimal number = numberOf(value);
```

So a boundary row at a date-time carried `1785542400` — the epoch second, as an `Int` fixture. The decoder refused it and the report said this:

```
// no row for `t = 2026-08-01T00:00:00` in `classifyMoment`: every value tried was refused
// at construction, which does not make the combination impossible
```

which is true of the search and says nothing about what went wrong. The fix was one arm; the reason it was reachable at all is that `BigDecimal` is both what the ranges hold and what a model writes.

There are three places a count leaves the algebra — `Generator.valueOf`, `Intervals.standingIn`, `Partitions.numeric`/`standing` — and each of them is a switch a carrier has to be added to. Today a missing arm is a wrong value; with the count as its own type it is a compile error.

## An enumeration is ordered and gets nothing

`souther examples` on a body comparing an enumeration:

```souther
data Prospecting
data Qualified
data Won
data Stage = Prospecting | Qualified | Won

behavior classifyStage : (s: Stage) -> Size
    constructs Bigger, Smaller, Qualified
let classifyStage (s) = {
    guard s < Qualified else Bigger
    Smaller
}
```

```
  classifyStage            implemented   rows 0    pending 0
    partition   axes 1   single-axis 0/0
      · not read: s (it is compared against values no line can be drawn on here)
```

An enumeration has everything the algebra asks of a carrier. It is ordered by the order its cases are declared (§primitives). Its count is the ordinal, which is order-preserving. Both neighbours exist — `Prospecting` is below `Qualified` and `Won` is above it — so unlike a `Decimal` or a `DateTime` it can even answer the neighbour question.

What excludes it is the shape of the table rather than anything about the values: `Carrier.ofValue` is keyed on `Type.Prim`, and an enumeration is a `Type.Ref` to a sum. A `String` is excluded for a reason about the values — there is no count to embed into — and an enumeration is excluded for no reason at all.

The line is the only thing missing. The position already has an axis, because the sum's cases are classes.

## The cut adds no class

`s < Qualified` cuts the values into `{Prospecting}` and `{Qualified, Won}`. The case partition is `{Prospecting} | {Qualified} | {Won}`, which is finer, so the meet of the two is the case partition and the cut adds nothing to it.

So a threshold on an enumeration gives boundary obligations and no classes. That is a rule ADR-0090 does not currently state, because until now every threshold divided a position its type had left whole.

## Acceptance

- A carrier's count has a type of its own, and a value of the position does not. The three write-back sites are exhaustive over the carriers, so a carrier added without one is a compile error rather than a refused row.
- `Carrier` is keyed on the ordered types rather than on `Type.Prim`, and an enumeration is one of them. `TypeOps.base` still reduces a single-value newtype first, so `data StageN = Stage` answers as `Stage` does.
- `guard s < Qualified` draws a line. The obligations are the value and its neighbour, both written as case names and never as ordinals.
- The classes at that position do not change: the cut is the coarser partition and the meet is what the cases already stated. Recorded in ADR-0090.
- `OneCarrierTableAnswersForEveryOrderedTypeTest` gains the enumeration rows and every other cell is identical. The invariant row and the guard row agree about the enumeration, as they now do about the four carriers already in the table.

## Out of scope

**`String`.** The type this introduces is the one a `Text` case would go in, and adding it stays cheap. Whether to add it is a separate decision and not a technical one: measuring a `String` position turns its classes and its line into rows `--strict` refuses a build over, and today the report instead says the position was not read, which is what led the one model that hit this to `{ 年: Int, 月: Int }` — the better model. That trade is worth deciding on its own.

**A newtype over a sum loses the sum's cases.** `data StageN = Stage` comes back as `not derivable` where the bare `Stage` beside it has three classes and three rows. That is the classes reader stopping at the name, not the carrier table, and it reaches every single-value newtype over something with classes — a `Bool`-backed one included. Recorded in the matrix test as it stands; it wants its own issue.

kawasima:
Closes #628.

The interval algebra carried every ordered value as a `BigDecimal`, and a model's own numbers were `BigDecimal` too. Nothing separated them, so the count a carrier is made of could be written wherever a value of the position belonged. Both carriers added so far went wrong there, in different ways: a day count read by two readers that disagreed about it, and a second count reaching a row as an `Int`.

Four commits: the separation, the enumeration it was for, and two rounds of review fixes.

## `Count`, and one reader for an ordered rule

`Count` is the algebra's coordinate. `Endpoint`, `NumericDomain.Bounds`, `Threshold`, `Intervals.Interval`, `Cut` and `BoundaryObligation` hold counts; the linear form inside `NumericDomain` stays on `BigDecimal` and is wrapped at the boundary. One type for every carrier, because what has been confused is the algebra's coordinates against the model's numbers, not one carrier's counts against another's.

`Carrier` is the only place either direction is crossed, and it moves to `check`, where the rest of the type classification is (`check` had no import of `partition`, and this does not add one). `OrderedLiteral` folds into it; `Dates` and `DateTimes` move to `numeric`, which is what they count in.

What that closes: `NumericTerm.numberOf`, which sniffed a written temporal for a `T` to decide whether it was counting days or seconds; `Generator`'s four readers of a boundary value; `BoundaryAssessment.value`; `Coverages.sameNumber`; `Partitions.numeric`, `standing`, `written`, `holds` and `same`. `BoundaryDomain` had four implementations each checking the shape of a value before it could step; it is now one, taking the neighbours from the carrier's spacing and the ends from what the carrier holds — so a step off the end of a day count is refused where only an `Int`'s was before.

Whether an ordered rule became a bound is one question with one reader. `InvariantBound.of` takes the carrier and nothing else; the overload keyed on `Int` and `Decimal` is gone.

## The defect that found

`everyRuleBecameABound` asked the reader that maps rules to the checks a decoder emits, which has no word for a comparison of dates, while the algebra read the same clause as a bound without difficulty. So a `Date` field with an invariant made its record's reading incomplete, and every boundary under that parameter — its `Int` siblings included — was downgraded from a gap to an edge nothing promises is writable.

```
A: data Span = { a: Slot }            boundary 0/2
B: data Span = { a: Slot, d: Day }    boundary 0/0, all four "not known to be writable"
```

`Slot` and the rules about `a` are identical in both. An axis nobody had touched stopped being measured because of what its neighbour was declared as. Controls: a `String` newtype beside the same pair does not do this, and neither does one bounded by `length(value) >= 1`. It is `Date` alone.

Fixed, and pinned by `ACarrierNothingReadsUnmeasuresItsSiblingsTest`.

**Still open:** a rule relating two `Date` fields (`invariant low.value < high.value`) is unread, because the atoms of the projection are not counts yet. `seeded` is what fails there rather than `everyRuleRead`. Out of scope here and stated in the test.

## The enumeration

An enumeration has everything the algebra asks of a carrier: ordered by declaration, the order is a count, and both neighbours exist. What excluded it was the shape of the table — the carrier was keyed on `Type.Prim` and an enumeration is a `Type.Ref` to a sum.

`Carrier` becomes sealed rather than an enum, because the fifth one carries something: a count means nothing without the cases it counts in. Every switch that reads or writes a count is over the sealed cases, so the ordinal had to be answered for at each of them before this compiled. That is the property the first commit bought — an ordinal is 0, 1, 2, the most plausible-looking wrong value of the five.

A threshold on an enumeration gives boundaries and no classes: `s < Qualified` divides the cases into `{Prospecting}` and `{Qualified, Won}`, which is coarser than the three the type already states, so the meet is the cases. Recorded in ADR-0090, and the specification no longer says an enumeration is not read.

A newtype over a sum still loses the sum's cases, so `data StageN = Stage` gets its line and no classes. That is the classes reader stopping at the name; it reaches every single-value newtype over something with classes and wants its own issue. The matrix test writes down what it does today.

## What moves for a reader

One line of the examples corpus, out of 3341 over 13 modules:

```
- · no row is in `terms.weeklyHours/0 <= x < 20.0`
+ · no row is in `terms.weeklyHours/0 <= x < 20`
```

A class label kept the places a decimal was written to while every other spelling of the same line dropped them, so one line printed two ways depending on which rule reached it first. The number is the identity and the spelling is not.

`--strict` tightens on enumerations, which now carry boundary obligations. No example moves — none compares an enumeration with an ordering operator — so the evidence for that half is the carrier table's rows and the obligations test rather than the corpus.

## What review found

Five defects, each reproduced from a written model before being fixed, and each pinned by a test that fails when its fix is reverted.

- A strict bound was sharpened onto the count beside the one it names using the range of a `long` — one carrier's idea of where counts stop. `invariant value > Won` landed one past the last case and crashed the reader that writes an obligation as a value. Both ends, and `Date` past the calendar too, which predates this branch.
- The carrier of a position was taken from the order the position can be *compared* on, which holds of a single case as well. A position declared as one case took the whole enumeration's counts and was asked for a row at a value it cannot hold.
- A boundary row wore one name where the position declares two. The temporal path is a regression from this branch; the numeric path was one layer deep before it and is not any more. `FixtureTemplate.on` now writes the bare value and every caller wraps through `Witnesses.wrapped` — one rule for how many names a value wears instead of two.
- `onTheGrid` corrected a count instead of answering about it: half a nanosecond came back as a nanosecond.
- `onTheGrid` for a date-time asked the writer before asking where the calendar stops, so a count past either end threw out of a question whose job is to answer no. Reached by a different route from the first item — a dense carrier has nothing to sharpen onto, so the bad count appears only when something looks inside the range the rule leaves.

`BoundaryDomain.midpoint` went with them: nothing outside its own test had called it, here or on develop.

## Checks

4036 compiler tests green. Every fix and every new test was mutation-checked — reverting one fix at a time fails exactly the intended test and nothing else. One test written during this work was deleted for failing to catch its own mutation: with no runtime to probe against, the boundary rows drop out of the generated output, so a sweep over them was blind to the path it claimed to cover.

None of the review defects moves the examples corpus. No example declares a newtype over a newtype, a position of one case, or a bound past a carrier's end — so nothing measured would have said any of it had stopped working.


kawasima:
All four reproduced before fixing, and each fix is pinned by a test that fails on a mutation of it. `da506267`.

## P1-1 — `stepped()` kept its own idea of where counts stop

Confirmed, and worse at the low end than the report says. Both directions crash:

```
invariant value > Won         Cut at=3   ArrayIndexOutOfBoundsException: Index 3 out of bounds for length 3
invariant value < Prospecting Cut at=-1  ArrayIndexOutOfBoundsException: Index -1 out of bounds for length 3
```

The `Date` variant is the same shape and predates all of this, as you say:

```
invariant value > Date("+999999999-12-31")
  Cut at=365241780472   DateTimeException: Invalid value for EpochDay (valid -365243219162 - 365241780471)
```

`ordered()` now takes the carrier and steps through `onTheGrid`; `LONG_MIN`/`LONG_MAX` are gone, and `ofSize` passes `Carrier.WHOLE` rather than a boolean.

On the semantics you flagged: I did not make "past the end" mean bottom. `Optional.empty()` today makes the position report a comparison it could not read, which is inaccurate for an unsatisfiable rule — but it is exactly what an `Int` past `Long.MAX` has always reported, and inventing a third answer for one carrier would put the two out of step again. Written down in ADR-0090 as a thing not decided rather than left implicit. If you want bottom, it is a decision about all five carriers and I would rather take it separately.

## P1-2 — carrier resolution used the comparability question

Confirmed:

```
axis s : Qualified
  carrier   Ordinal[Stage, [Prospecting, Qualified, Won]]
  obligation AT = Won        <- a value the position cannot hold
  obligation BELOW = Qualified
```

Fixed the way you suggest: `ofValue` no longer calls `orderingEnumeration`. It requires the base to be the sum itself (`isUnitOnlySum`), so a case and a union are not positions on the order, and `TypeOps.base` still lets a newtype over the enumeration through. `comparisonEnumeration` is untouched.

I could not get a sub-union parameter to compile, so that half of your example is unconfirmed — the case-typed position is enough to establish the defect and the fix covers both.

A case-typed position now reports "compared against values no line can be drawn on here", which is not accurate — a line *can* be drawn on `Stage`. The report vocabulary has no reason meaning "this position holds one value". Left as is and recorded in ADR-0090 rather than adding a fifth reason inside this PR.

## P1-3 — one name where the position declares two

Confirmed, and the `Int` path was wrong too (one layer on develop, as you said, so pre-existing rather than a regression):

```
before   ShippingDay(Date("2026-08-01"))     ShippingAmount(500)
after    ShippingDay(Day(Date("2026-08-01")))  ShippingAmount(Amount(500))
```

I took the responsibility split rather than only the boundary path: `FixtureTemplate.on(carrier, count)` writes the bare value and no longer takes a wrapper, and every caller — `Generator.edgeOf`, `Intervals.standingIn`, the singled-out classes, the reserve tier, the displaced representative — wraps through `Witnesses.wrapped`. That removes the second rule for "what names does this value wear": `isSingleValueNewtype` at some call sites, `data.newtype()` at others.

## P2 — `onTheGrid` was a correction

Confirmed. `Carrier.MOMENT.onTheGrid(0.5ns)` returned 1ns; it returns null now, and a whole nanosecond still returns itself. Round-tripped and held to itself, as you wrote it.

While checking the blast radius: **`BoundaryDomain.midpoint` has no caller in main, here or on develop.** Only its own test exercised it. Deleted, with that test case — which also removes the one place where a snapped count was returned without being re-checked.

## Measurements

4033 tests green. The examples corpus is unchanged by all four fixes — no example declares a newtype over a newtype, a position of one case, or a bound past a carrier's end, so nothing measured would have told us any of this had broken. The only corpus movement in the PR remains the decimal label spelling.

Each fix was mutation-checked. Reverting one fix at a time fails exactly the intended test and nothing else:

```
stepped() does not ask the carrier   -> the three "leaves no end" cases
carrier from the comparable order    -> aPositionDeclaredAsOneCaseIsNotOnTheEnumerationsOrder
onTheGrid snaps                      -> aCountBetweenTwoMomentsIsNotOneOfThem
one layer of names                   -> both nested-newtype cases
```

## On your closing point

Agreed that the first two are one root. The carrier now carries all three of order, representable counts and encoding, and the two bypasses are closed: `InvariantBound` steps through `onTheGrid`, and `ofValue` no longer picks an order wider than the position's values. The part still missing is that a position narrower than its carrier — one case, a sub-union — cannot be measured at all rather than measured against its own values. That needs `Ordinal` to hold the admitted subset beside the full order, and it is not this PR.

kawasima:
Fixed in `060399a0`, and the two non-blocking items with it.

## The blocker

Reproduced, with one correction to the path.

Direct call, before the fix:

```
Carrier.MOMENT.onTheGrid(MAX + 1s)   DateTimeException: Invalid value for EpochDay (valid -365243219162 - 365241780471)
Carrier.MOMENT.onTheGrid(MAX + 1ns)  DateTimeException
Carrier.MOMENT.onTheGrid(MIN - 1s)   DateTimeException
Carrier.MOMENT.onTheGrid(MIN - 1ns)  DateTimeException
Carrier.MOMENT.onTheGrid(1e21)       ArithmeticException: Overflow
```

Both edge literals parse, so the model that reaches them is writable:

```
+999999999-12-31T23:59:59.999999999 -> 31556889832780799.999999999
-999999999-01-01T00:00:00           -> -31557014135596800
```

The correction: `obligationsOf` does not reach it. A strict bound over a dense carrier stays on the count it names, so the cut is at `MAX` itself and nothing steps past it — your description of why the discrete carriers are caught earlier and this one is not is exactly right, but the consequence is that no boundary asks. What does ask is the value that stands for the position:

```
Partitions.representativesOf(PastMoment)
  -> insideTheNewtype -> inside(bounds, MOMENT)
  -> Endpoint.valueBetween(exclusive(MAX), null, DENSE) = MAX + 1
  -> onTheGrid -> DateTimeException
```

Reproduced from source, and worth recording how it looked from outside: **the exception was swallowed and the report came back missing its `partition` and `boundary` lines altogether**, with nothing said about why. Through the CLI it reads as a model that simply has nothing to measure.

`DateTimes` now names `MIN` and `MAX` and answers `holds`; the carrier asks that before it asks the writer anything. Your four cases are pinned, plus a fifth so the first four are not passing on an empty answer:

```
DateTime.MIN / MAX          -> itself
MIN - 1s / -1ns             -> null
MAX + 1s / +1ns             -> null
```

and at the model level, `value > MAX` and `value < MIN` offer nothing from inside the range rather than throwing, while a rule inside the calendar still offers a value from inside itself. Reverting the range check fails both the unit case and the model case and nothing else.

One thing I did not assert the way you wrote it: `representativesOf` on those two models is not empty. What the newtype wraps still stands for it — a candidate is a proposal the decoder answers, and this one is refused there — so the assertion is that the value *from inside the range* is gone, which is where the bad count came from. Asserting emptiness would have been asserting something the design does not say.

## Non-blocking, both done

The `Carrier` javadoc no longer says an enumeration is asked of `orderingEnumeration`. As you say, that distinction was the defect itself, so it now reads: the enumeration is the sum, checked as one; comparability is a wider question with its own answer.

PR body updated — four commits, 4036 tests.

## Where this leaves the carrier

Order, representable counts and encoding are now all the carrier's, and every producer of an end goes through the third. What is still true is that a position narrower than its carrier — one case, a sub-union — cannot be measured at all rather than measured against its own values, and that a bound past a carrier's end reports as a comparison that could not be read rather than as a rule nothing satisfies. Both are written down in ADR-0090 as decisions not taken rather than left implicit.

Issue #634, opened 2026-08-11
A JSON report names a source by an id it does not explain


kawasima:
The adequacy JSON writes a reason about a source as

```json
{ "code": "observation_absent", "scope": "source", "subject": "0" }
```

`subject` is the compilation's identity for that source: the position in the list a build passed, or the document URI an editor is keyed on. #627 left it that way deliberately. It is what makes two reasons the same reason, and a display name is not a substitute for it — what to call `a/model.sou` depends on whether some other file in the same run is also called `model.sou`, so a name is chosen from the set of files in front of a reader and is neither stable across runs nor a key. Replacing the identity with the name would move the defect rather than remove it: the document would say what a person should be shown and stop saying which source it is about.

What is left open is that the document cannot be read on its own. A consumer that also holds the argument list can recover which file `"0"` was; the JSON itself says nothing that lets anyone else. Every other subject in this field is a name the author wrote — a behavior, a module, a position — and this one is not.

Two shapes that keep both halves:

```json
{
  "sources": { "0": "other.sou", "1": "zeroname.sou" },
  "modules": [ { "incompleteness": [ { "scope": "source", "subject": "1" } ] } ]
}
```

or

```json
{ "scope": "source", "source": { "id": "1", "name": "zeroname.sou" } }
```

The first adds a field and leaves `subject` alone; the second replaces a string with an object and is a rename, which the schema's own rule says raises `schemaVersion` (`example-report-vocabulary`: a version is raised when a field is removed or renamed, and not when one is added). That difference is most of the decision.

Whichever is chosen, the name has to come from the same place the human report and the diagnostics get theirs, so a document cannot disagree with the terminal that produced it.

kawasima:
Done in #637.

PR #635, opened 2026-08-11
Name a source by what the caller calls it, not by the id it is held under

### LINKED ISSUE #627: A report that could not read a source names it `0`
When a compile error stops the rows being read, the report names the source as `` `0` ``.

```
    · no rows were read from `0`, so what they cover is unknown
// generation stopped for `countKept`: no rows were read from `0`, so what they cover is unknown
```

`0` is where a file name belongs. Anything that raises a diagnostic before the rows are evaluated reproduces it — the shortest is an over-declared `constructs`:

```
module example.zeroname

data Amount = Int
    invariant value >= 0

behavior passThrough : (a: Amount) -> Amount
    constructs Amount
let passThrough (a) = a

example passThrough
    | "through" : (Amount(1)) -> Amount(1)
```

```
$ souther examples zeroname.sou --generate
-- OVER-DECLARED CONSTRUCTION  E1006 ------zeroname.sou:6:1
...
    · no rows were read from `0`, so what they cover is unknown
```

The sentence is otherwise the right one to print — <<example-report-emission>> is explicit that a compilation which raised a diagnostic still reports whatever it observed, and "unknown" is the honest word for rows that never ran. It is only the name that is missing, and a report that cannot say which file it failed to read is least useful in exactly the case it is printed for: a run over several sources where one of them broke.

kawasima:
Closes #627.

A compile that raised a diagnostic before its rows ran reported the source as `` `0` ``:

```
$ souther examples zeroname.sou --generate
-- OVER-DECLARED CONSTRUCTION  E1006 -------zeroname.sou:6:1
    · no rows were read from `0`, so what they cover is unknown
// generation stopped for `passThrough`: no rows were read from `0`, ...
```

## What it was

`0` is the compilation's source id, not a name. `Compilation.ofSources` calls the i-th source `String.valueOf(i)`; a build hands its sources over as a list and gets their positions back, an editor hands over documents and gets its own URIs back. A reason should carry that — two reasons are the same reason when they are about the same source, and only an identity settles that. `Reasons.said` read it as a name.

The id-to-name map already existed and only the caller had it. `Main.sourcesOf` builds one with `SourceNames.of` and gives it to the diagnostic renderer, which is why the same run headed the error `zeroname.sou:6:1` and named the file `0` four lines below. The report and the `--generate` block are rendered from the compilation alone and were given nothing.

A single-source compile made it worse: `Compilation.ofSource` sets `namesSources = false` so that diagnostics name no source, and the report does not go through that — it reads the ids off `exampleSourcesOf`. The run that had decided not to name a source named one anyway.

## What this does

Keeps the identity in the result and gives the name at the point of rendering.

- `SourceNameResolver` (souther-syntax/diag): what to call a source id, and `identity()` for a caller with none to give.
- `Target.shown(names)`: the sum answers what it is shown as. Which subjects are identities and which are already names is what the kinds differ in — a behavior, a module and a position are written by the author and shown as they are — and a renderer that asks for the subject and prints it cannot tell them apart.
- `human`, `GeneratedRows.of` and `Reasons.said` take the resolver rather than defaulting to one. A no-arg convenience would leave the trap that was sprung here: the caller that forgets is the one holding the only copy of the names. The language server passes `identity()`, its ids being the URIs it calls its documents.

Holding the name with the source would have been the wrong half to move. A name is the shortest tail of a path no other file in the same compile shares — `a/model.sou` is `model.sou` on its own and `a/model.sou` beside a `b/model.sou` — so it is a fact about the set of files in front of a reader rather than a property of one of them.

## After

```
    · no rows were read from `zeroname.sou`, so what they cover is unknown
// generation stopped for `passThrough`: no rows were read from `zeroname.sou`, ...

$ souther examples a/model.sou b/model.sou
    · no rows were read from `a/model.sou`, ...
    · no rows were read from `b/model.sou`, ...
```

## The JSON is unchanged

Its subject is what identifies the source. Turning it into a name would move the defect rather than remove it: the document would say what a person should be shown and stop saying which source it is about. The schema now says that, where it said only `"type": "string"`. That the document cannot be read without the argument list beside it is a separate question, raised as #634.

## Tests

On the boundary rather than on the sentence — `AReportNamesASourceTheWayItsCallerDoesTest`: a single source is named and never numbered, several are named apart, two files called `model.sou` are told apart by as much of their paths as it takes, the generated note reads the way the report does, and the JSON still carries the id.

`AReasonSaysOnlyWhatWasEstablishedTest` expected the sentence to read `` `1` `` for a reason about source `1` — a test of what a person is shown, holding the identity it was handed. That is why nothing caught this, and it is the assertion this replaces.

`mvn clean install` green.


kawasima:
All three addressed in 89028796.

**1. `namesOf` no longer goes through `indexOf`.** You are right that it was the one place the boundary was broken, and the reason is the one you gave: `indexOf` answers for a diagnostic, which may name no source at all, so the single file is the answer however the diagnostic is tagged. A reason in a report always names one. The report's lookup now matches on the id alone and falls back to the id, which is what `SourceNameResolver` says it does. `indexOf` is untouched.

The test names the difference rather than the fix:

```java
SourceNameResolver names = Main.namesOf(List.of(Path.of("a", "zeroname.sou")));

assertEquals("zeroname.sou", names.nameOf(Compilation.idOfSourceIndex(0)));
assertEquals("elsewhere.sou", names.nameOf("elsewhere.sou"));
assertEquals("1", names.nameOf("1"));
```

Measured against the old lookup rather than argued: put `indexOf` back and this fails on the second assertion (`expected: <elsewhere.sou> but was: <zeroname.sou>`) and nothing else in the class moves.

**2. `Compilation` said name where it meant id.** Fixed — "named by their position" → "identified by their position", `idOfSourceIndex` no longer "calls" the i-th source, a build "identifies its sources" by where they are in the list. The field `namesSources` is `saysWhichSource`: whether a diagnostic states which source it is in is a question about identity, and that name was doing both jobs at once. I also wrote on `idOfSourceIndex` that nothing downstream may print one at a person, which is where the next reader will be.

**3. The specification is two rules now**, one per form, with what separates them written under them rather than left as an exception in the following sentence.

I did not touch `shown()`'s name. `mvn clean install` green.

PR #637, opened 2026-08-11
Explain the source identities a report writes, where they are written

### LINKED ISSUE #634: A JSON report names a source by an id it does not explain

kawasima:
Closes #634.

The adequacy JSON says which source a reason is about with the identity the caller handed the source over as, and #627 left it that way on purpose: an identity is what makes two statements about one file the same statement, and a name is chosen from the files in front of one reader, so it is neither stable across runs nor a key. What was left open is that the document could not be read on its own. A consumer handed `{"scope": "source", "subject": "1"}` and nothing else has no way to reach the file.

Of the two shapes the issue named, this is the first: a `sources` table at the root. Nothing moves, so a reader written against this shape reads what it read before and `schemaVersion` stays at 1 — the schema's own rule is that a version is raised when a field is removed or renamed, and not when one is added.

```json
{
  "schemaVersion": 1,
  "modules": [ { "incompleteness": [ { "scope": "source", "subject": "1" } ] } ],
  "sources": { "1": "b/model.sou" }
}
```

## Where the identities are collected

A reason's `subject` was one place an identity is written and the `sourceId` of an `at` was another. Collecting those two would be a fix aimed at the places somebody had noticed, and the next field to carry an identity would be outside it.

So everything that writes an identity asks `DocumentSources` for the string to write, and is recorded by asking. A field that is emitted is a field that is explained.

- `Target.carried(DocumentSources)` answers which subjects are a source's identity, for the reason `shown(SourceNameResolver)` does: which subjects are identities and which are already names is what the kinds differ in, and a renderer reading the scope would be that list of kinds a second time.
- The shape of an `at` is written once, rather than at each of the two places that point into a source.
- `json(SourceNameResolver)` replaces `json()`. The resolver is not an optional dependency of this rendering, so there is no overload to pick between.

## What the table is of

The identities this document carries, and not the sources the compile was handed. A source the report says nothing about has no entry: a list of every input would be a second thing for the document to be about, and a consumer reading it as the report's subject would read in files nothing here says anything about. `the-json-explains-the-identities-it-carries` states this.

## Tests

`ADocumentExplainsTheIdentitiesItCarriesTest` holds three properties of the document rather than expected values per field:

- every source identity anywhere in the document has an entry — found by reading the document for identities, and asserting that both kinds were in fact present, so the walk cannot pass by finding nothing;
- the names are the ones the same run's human report and diagnostics use, measured over two files whose basenames collide;
- three sources handed over, one written about, one entry.

`mvn clean install` is green.


kawasima:
Review is addressed in 8f6c252c.

**The sum no longer takes the renderer's table.** `carried(DocumentSources)` is gone; `Target` answers `Optional<String> sourceIdentity()`, and writing that down plus recording that this document owes an account of it is the renderer's own line. `DocumentSources` moves from `souther-syntax/.../diag` to `souther-compiler/.../report`, since nothing under `observe` refers to it any more.

The renderer still tests no scope:

```java
g.put("subject", gap.sourceIdentity().map(sources::written).orElseGet(gap::subject));
```

And the reasoning in the description above was wrong where it said a renderer reading `scope()` would be a second list of kinds. `scope()` is the classification, already made. What `sourceIdentity()` buys is that the renderer does not have to know `SOURCE` means "the subject is a source id" — it consumes the classification as a value instead of restating what it implies.

**"Stable" is out.** A position in a list means another file when the list changes. An identity is what a compilation refers to a source by, and so what makes two reasons about one file the same reason — which is the whole of what is needed here, and is why a name cannot stand in for it. `DocumentSources`, `json(...)` and the test's own account all said more than that and now do not.

**The walk claims what it does.** It covers the two spellings the schema gives an identity — an `at`'s `sourceId`, and a `subject` under `scope: source` — wherever either of them appears, so a new place either is written is covered. A third spelling is not, and the comment now says where it would have to be said. What holds for every field is the one call every writer of an identity goes through, which is the writing side of this.

`mvn clean install` is green.

Issue #625, opened 2026-08-10
A guard that compares two positions asks for no boundary row, so the line it turns on is never evaluated


kawasima:
Where a `guard` compares one position against another, the report says the position was not read and no boundary is asked for. Every other measure can be full, `--strict` can exit `0`, and nothing in the model has ever been evaluated at the line the rule turns on. Flipping `>` to `>=` there survives.

## What it does today

```
module example.relational2

data Charge = Int
    invariant value >= 0
data Ceiling = Int
    invariant value >= 1000

data NoBenefit
data Benefit = { amount: Charge }
data Result = NoBenefit | Benefit

behavior benefitOf : (charge: Charge, ceiling: Ceiling) -> Result
    constructs NoBenefit, Benefit, Charge
let benefitOf (charge, ceiling) = {
    guard charge.value > ceiling.value else NoBenefit
    Benefit { amount = Charge(charge.value - ceiling.value) }
}

example benefitOf
    | "over the ceiling" : (Charge(100000), Ceiling(80000))
        -> Benefit { amount = Charge(20000) }
    | "under the ceiling" : (Charge(50000), Ceiling(80000)) -> NoBenefit
    | "charge = 0 x ceiling = 1000" : (Charge(0), Ceiling(1000)) -> NoBenefit
```

```
  benefitOf                implemented   rows 3    pending 0
    signature   out specified 2/2  observed 2/2  verified 2/2
    partition   not measured (no partition axis was derived at any position)
    boundary    2/2
      · not read: charge (the comparison relates it to another position rather than dividing it)
      · not read: ceiling (the comparison relates it to another position rather than dividing it)
    branch      2/2
```

`souther examples --strict` exits `0`. Both output cases are verified, both arms are taken, and the two boundaries that were asked for are the ones the *invariants* drew — `Charge`'s minimum and `Ceiling`'s. The line the behavior is about, `charge == ceiling`, is named by nothing.

Change one character:

```
-    guard charge.value > ceiling.value else NoBenefit
+    guard charge.value >= ceiling.value else NoBenefit
```

No row fails. The two models disagree:

```
$ souther run relational2.sou --behavior benefitOf --input '[1000,1000]'
{"type":"NoBenefit"}
$ souther run rel2mut.sou --behavior benefitOf --input '[1000,1000]'
{"amount":0,"type":"Benefit"}
```

[...]

PR #638, opened 2026-08-11
Draw a line on a string, and leave out only the row below it

### LINKED ISSUE #622: The adequacy report draws boundary lines on Int, Decimal and Date, but not on String or DateTime

kawasima:
Closes #622 — the `String` half. The `DateTime` half went in with #629.

A `String` is ordered — the language defines `<` on it lexicographically (§primitives) — and a rule written over one was not measured at all: no axis, no boundary, no generated row, and a report saying the comparison was against *values no line can be drawn on*, two lines below a body that compares them.

## What was actually out of reach

Four things a measure produces at a line, and only the last needs something a string has not:

| | |
| --- | --- |
| the line | the literal the body already wrote |
| the classes either side of it | as derivable as they are for an `Int` |
| a row at the line | the literal again |
| a row just below the line | a string has no value before another one |

The first three need the order. The fourth needs a neighbour, and a carrier with no smallest step already gets that answer — a `Decimal` and a `DateTime` have had it all along. So the issue's own table is what this implements, and the fourth row is the only one still empty.

## Two commits

**`Place` separates where a value sits on its carrier's order from the number it counts to.** The algebra was written for one number per position, with arithmetic on it wherever a range had to give up a value. Everything it actually needs in order to *divide* values — compare two, tell two apart, write one down — is `Place`; `Count` is the case that has a number, and the arithmetic stays there. What holds a place is what the divisions are made of (an end, a threshold, an interval, a cut, an obligation, what a term reads off a row); what holds a count is the domain that proves what a position can hold, which reasons over differences.

Two questions move onto the carrier with it, because both are answers only a carrier has: *a place inside a pair of ends*, and *a place that is none of the ones a body singled out*. They were computed from the numbers directly, which is where they could only ever be a number's answer.

Nothing about what is measured changes in that commit, and the corpus is identical line for line.

**`Text` is the fifth carrier.** A line is drawn at the string, a row carries the string, and the class below a line is offered the empty string — the least there is, not one this compiler chose. Above a strict bound it offers nothing: every string with that one as a prefix is greater, and picking which would put a character nobody wrote into a row somebody has to read.

## The one thing that had to be got right

`String` is the only type two measures answer for — its own order, and the length of it. `measure()` looked at the carrier before the size, so making `String` a carrier would have turned every rule anyone ever wrote about the length of a string into a rule about the string: **134 lines of the examples corpus**, silently. Which measure a position gets is now decided by what its rules are about.

The same mistake surfaced once more on the way: `everyRuleBecameABound` asked the ordered reader about a length rule, and every `String.length` boundary became *not known to be writable*. Same shape as the defect #628 was filed for. Both readers are asked for a `String` now, and one for every other carrier.

## What this closes

The same rule and the same two rows:

```
before   YearMonth = String   partition not measured    adequacy: undetermined    exit 0
         Month     = Int      boundary 1/2              adequacy: not satisfied   exit 1

after    YearMonth = String   boundary 0/1              adequacy: not satisfied   exit 1
         Month     = Int      boundary 1/2              adequacy: not satisfied   exit 1
```

They differ only in the denominator: the string is owed the row at its line and not the one below it. The type an author reaches for no longer decides whether the boundary rule applies to them.

## Measurements

4049 tests green. `String.length` in the corpus: 134 before, 134 after. The corpus moves two lines, both message text:

```
- · not read: owner (it is compared against values no line can be drawn on here)
+ · not read: owner (a comparison here is written in a form this does not read)
```

Both are `==` between two positions on a `String` newtype. The old sentence was about the carrier for something that is not about the carrier; the new one is closer to what happened.

## Also recorded

ADR-0090 and §a-line-is-drawn-where-the-values-can-carry-one now say a line is drawn on every ordered value, and that what differs between them is the value *beside* a line. Both also state the decision from #628 that a position which is one case of an enumeration, or a union of some of them, carries no line — it is comparable on its sum's order and ranges over less than it, so a line from the sum's places would ask for a row at a value the position cannot hold.

## Not in this

The general question #622 raises about `--strict` — that a model with an unread rule passes every gate — is not fixed here, and is not fixable on the reasons as they stand. Measured across the corpus: 119 unread rules, of which 65 are a comparison form this does not read, 43 are a comparison relating two positions (#625), 9 are the depth limit, and 2 are the ones above. Refusing over all of them would refuse most of the corpus over the compiler's own limitations. That wants the reasons to name what happened before a gate can read them, and it is its own issue.


kawasima:
Both blocking items reproduced and fixed in `558353b9`, with items 3 and 4. One correction to the report, and one thing item 2 turned up that neither of us had.

## 1 — a range closed at the top

Confirmed, and it is worse than the example: the upper end was never looked at at all, so an exclusive lower end answered "nothing" whatever was above it.

```
before   a < x <= b   ->  class `a < x <= b`  ungeneratable
after    a < x <= b   ->  class `a < x <= b`  ["b"]
```

Your framing is the one I took: `"b"` is a value the model wrote, and taking it is not the same act as inventing a character. The restraint stays exactly where it belongs — above a strict bound, where every string with it as a prefix is inside and none of them is one this names.

## 2 — the least string for "everything else"

Confirmed.

```
before   x == "foo"   ->  class `/= foo`  ungeneratable
after    x == "foo"   ->  class `/= foo`  [""]
```

Your three-way rule is what it does now, including the third case: `x == ""` leaves `/= ""` with nothing, because the empty string is the one singled out and there is no next one to invent. Held in the table so it stays that way.

**One correction.** `x != "foo"` is not Souther — `!` is a syntax error, the operator is `/=`. My first probe of it crashed, and I reported that to myself as a compiler bug before finding it was my own harness calling `behaviors().get(0)` on a model that had not compiled. The compiler does not crash there. `x /= "foo"` measures identically to `x == "foo"`, which the table now fixes.

## 3 — `onTheGrid` and carrier identity

Agreed, and taken. `Dense` and `Text` were answering yes to anything, including a place of the other carrier. You are right that this is the wrong operation to skip the check in: it is where a carrier says which places are its.

```java
case Dense _ -> count instanceof Count ? count : null;
case Text _  -> count instanceof Text ? count : null;
```

`Seconds` had the same hole one level down — it went straight to a range comparison that would have thrown rather than answered — so it checks too.

## 4 — the code-point comment

You are right, and I checked which way it had to be fixed rather than assuming. The runtime orders strings with `String.compareTo`, both in `Representations` and in the generated bytecode, so the comparison is UTF-16 code units and matches what a row's branch does. The comment was wrong, not the code.

Rewritten, and it now says why the measure beside it counts code points instead: the length matches what the decoder admits, the order matches the branch a row takes, and each matches what it is a measure of. Left implicit, two units in one measure is exactly what gets "corrected" later by someone who finds only one of them.

The first version of that comment cited a spec anchor I had invented. `ASpecificationCitationNamesASectionRatherThanItsNumberTest` caught it — worth saying that the check exists and works.

## 5 — the `Text` special case in `everyRuleBecameABound`

Agreed and not done. It is the same shape as the `measure()` precedence you singled out, and both are instances of one thing: a type with more than one measure candidate, where which measure applies is what the rules are about rather than what the type could carry. `String` is the only such type today, so generalising now would be writing the abstraction from one example. Left as it is, and the reasoning is in the ADR rather than only in the diff.

## The table

Added, over every shape you listed. Classes, what stands for each, and what a build is owed — three numbers per row, because a change that gets the classes right and the obligations wrong is a partition nothing has to cover.

```
x <  "b"        x < b        [""]     b <= x       ["b"]      AT b
x <= "b"        x <= b       ["b"]    b < x        none       AT b
x >  "a"        x <= a       ["a"]    a < x        none       AT a
x >= "a"        x < a        [""]     a <= x       ["a"]      AT a
"a" <  x <= "b" x <= a ["a"] a < x <= b ["b"]      b < x none AT a, AT b
"a" <= x <  "b" x < a  [""]  a <= x < b ["a"]      b <= x ["b"] AT a, AT b
x == "foo"      = foo ["foo"]          /= foo ["" ]           AT foo
x /= "foo"      = foo ["foo"]          /= foo [""]            AT foo
x == ""         =     [""]             /=     none            AT
```

The asymmetry between the two ends is deliberate and the test says so, so that nobody makes the rows match later.

Both fixes are mutation-checked: reverting the upper-end witness fails two rows, reverting the least-string witness fails one, and nothing else moves.

## Measurements

4054 tests green. `String.length` in the corpus: 134 on develop, 134 here. The corpus diff against develop is still the same two message lines and nothing else.

PR #641, opened 2026-08-11
Ask for the row that tells a rule written > from one written >=

### LINKED ISSUE #625: A guard that compares two positions asks for no boundary row, so the line it turns on is never evaluated

kawasima:
A `guard` comparing one position against another draws a line — the place where the two are equal — and nothing asked for a row on it. Every other measure could be full, `--strict` exited `0`, and flipping `>` to `>=` survived. Found modelling Japan's high-cost medical expense benefit, where ten positions of a finished model were reported this way and they are the ten the domain is about.

Fixes #625.

## What was losing it

`GuardThresholds.readOne` reads a term on one side and a constant on the other. Returning nothing where there is no constant dropped the threshold, both arms' edges and the obligation together, and what was left was a note about the partition — which carries no diagnostic code, is no adequacy gap, and is counted nowhere.

Underneath that, every boundary was "one position at one value". `BoundaryObligation` held an axis and a place on it, and the readers that check a row, build a candidate and print a line all assumed one. A line between two positions is on neither of them, so there was nowhere to put it.

## What is here

**Where a line is is its own shape.** `BoundaryObligation` holds a `BoundaryTarget`: `AtPlace`, a place of one position, or `EqualTerms`, two terms of one carrier holding the same place. The second has no place of its own — the value a row happens to satisfy it with belongs to that row, and written into the obligation one row on the line would name every other row on it as a different boundary. The three readers that assumed a place are switches over the shape, so a shape added is one each of them has to answer for.

**Boundary assessments come from two sources.** A line at a place comes off the position's axis; a line between two positions comes off the comparison. A behavior can draw one while having no axis at all — two plain `Int` parameters have no invariant, no threshold and no axis, and the body still draws a line between them. That model is the one the old reading cannot be patched into, and it is in the tests.

**Recognised by the carrier, not by the type.** Two operands compare only when they are of one type, and a type is not what makes a line measurable. `Charge` against `Ceiling` is two types on one carrier and has a line; two cases of one sum order against each other and carry no places, so they have none. `String` arrived on its own when #622 gave it a carrier.

**Whether a row can be written on it is proven, not searched for.** A place both positions admit is the two ranges read together, which the rules answer without anything being built. Where they leave none, the line is reported and not counted — a search coming back empty never says a row cannot be written (#602).

**The classes measure is unchanged and says so where it is.** A rule relating two positions is still a partition of neither. The note saying so used to print two rows beneath a boundary count that was counting the line that same comparison had drawn; it now sits under the partition.

## Report

```
      · not read: charge (the comparison relates it to another position rather than dividing it)
    boundary    2/3
      · no row is at benefitOf/charge = ceiling (guard@15:5)
```

`--strict` exits `1`. `--generate --boundaries` offers `| "charge = ceiling" : (Charge(1000), Ceiling(1000)) -> <?>`; with that row written the measure is `3/3`, and the `>=` mutant fails it with E1905.

## Specification

The specification described a `guard`'s line only as a comparison against a constant and said nothing about a comparison of two positions, so there was no rule to be breaking. `a-comparison-of-two-positions-draws-a-line-between-them` now says what such a comparison draws, that it is a partition of neither position, that the carrier is what makes it measurable, and that the row *beside* the line is not asked for — that needs a value one step from where the difference falls, and neither position has a neighbour there.

## Document

Every boundary gains a `kind` of `at_value` or `between_positions`. A line between two positions writes the other position where a line at a place writes the value, and the two read alike — a consumer reading `value` as a value would read a position's name as one. `value` stays required and is documented as the other side of the line, so a document of this version keeps its shape.

## Where the three answers come apart

A term that is a measure of a value rather than the value — `String.length(a) > String.length(b)` — draws the line, is measured against the rows, and **is counted**: the rules prove a length both positions admit, and that is what says a row can be written on it. What is missing is only the offer, because nothing here writes a value from a length — four is not what goes at the position, it is four characters somebody has to choose. The block says that about itself rather than about the model: `nothing here composes a value at that edge, which does not make the edge unwritable`. Same for a position whose declared type nothing here can reach.

## Checks

`mvn clean install` then the full suite: 4121 tests, 0 failures, 0 errors. `ALineBetweenTwoPositionsIsStillALineTest` holds twelve rows — two newtypes with different minima, the row on the line, two plain `Int`s with no axis, an enumeration, a `String`, a comparable pair with no carrier, an operand that is not a position, two ranges with nothing in common, where the note is printed, and what the offered row contains.


kawasima:
Three of the four are in `a754be41`. On the second I took a different fix than proposed, and the reason is below.

**Line quotient.** Correct, and the asymmetry was the whole of the difference — `assessBetween` collected into a list while the unary path merges under `OriginRef.Line`. It now goes through the same key.

I could not build a model that produces two readings of one relational comparison. A guard inside a non-recursive helper compares that helper's own parameters, and `pathOf` only accepts a `Core.Read` naming a parameter of the behavior — so the caller's positions are not named and no line is drawn at all. That holds for a unary threshold in a helper too: `guard a > 100` inside `pick(p)` called twice reports `p` as not derivable and draws nothing. So there is no behavioural test to add here; a helper model would assert "one line" on a model that produces one line for a different reason, which is an empty probe. The key is here for the reason the unary one is — 13f0406e says a walk that learns to descend a binding needs nothing changed — and I have said that in the commit rather than dressed it as a fix for something reproducible.

**`SizeOf`.** Reproduced, and it was worse than described. The report counted the line and the block said:

```
// no row for `String.length(a) = String.length(b)` in `cmp`: no value can be written there
```

Two contradictory statements about one line, and the second is false — two strings of one length are the easiest row in the file to write by hand.

I did not narrow the projection to `ValueOf`. The proof `commonPlace` gives is that a length both positions admit exists, and that is sound for `SizeOf`: `String.length` is a whole number, the ends come from the same rules any other term's do, and the row is writable. Restricting `known` would drop an obligation the model genuinely owes on the grounds that this generator has no route to it — which is a search failure removing an obligation, the shape #602 fixed. The report would then understate what the rows owe, and `--strict` would pass on a model missing a row an author can write by hand.

What was wrong is the sentence, so that is what changed: `NOTHING_COMPOSES_ONE` instead of `NO_REPRESENTATIVE`. The two exist for exactly this distinction — the first says this compiler has no way to write one, the second licenses "no value can be written there". Now:

```
// no row for `String.length(a) = String.length(b)` in `cmp`: nothing here composes a value at that edge, which does not make the edge unwritable
```

The three answers stay apart: the line is drawn, the rows are measured against it, and nothing here writes a value from a length. `aLineOnAMeasureIsCountedAndOfferedNoRow` holds all three.

You are right that the PR description was inconsistent with this — it said such lines are "not counted while nothing promises them". They are counted, because the rules promise them. I have corrected the description.

**`kind` required.** Correct and unambiguous — the schema's own description says a key added since is optional so that an earlier document is still one of this version, and requiring it refuses every boundary object written before this PR. It is optional now, its absence documented as the line being at a value. `aBoundaryWrittenBeforeAKeyExistedIsStillOfThisVersion` holds the boundary object's required keys to what one carried before the key existed; as you say, the word test cannot see this, since the words were right and the document carrying none of them was the one refused.

**Specification.** The paragraph was written before the rebase onto #622 and I did not re-read it afterwards, so it stated the opposite of what the code does. It no longer restates a list: which ordered values a line can be drawn on is `a-line-is-drawn-where-the-values-can-carry-one`'s, and this paragraph says that is what the two sides must share rather than a type — wider than a type in one direction (`Charge` against `Ceiling`) and narrower in another (two positions each one case of an enumeration). The claim is now an *ordering* comparison, with `==` named as not one of these and why. The writability paragraph says failing to compose a row settles nothing either way, which is the `SizeOf` case stated where it belongs. `count` is `place` throughout the javadoc and the schema description.

4098 tests, 0 failures.

kawasima:
The blocker is real and reproduced exactly as described. Fixed in `c246c65a`.

Before:

```
    boundary    0/1
      · no row is at cmp/p.a = p.b (guard@13:5)
```

under `invariant a < b`, where the diagonal holds nothing. A row demanded that cannot exist, and `--strict` refusing the build for not writing it.

You named the reason precisely: the intersection of two unary projections proves each position admits the place on its own, and that is not the pair. `FieldDomains` already keeps the two questions apart, and I was using the first as an answer to the second.

**What changed.** The projection settles it only where nothing can relate the two positions, which is two positions under different parameters. A behavior's parameters are constructed one at a time and a model has no way to write a rule across them — the generator says so about its own search — so a value each admits is a pair both admit. Under one parameter that does not hold, and the line is settled by a witness instead: a row already on it, or a value the module's decoder took. The decoder is the authority the `Probe` interface already names for this exact question — "an invariant relating two fields refuses a pair each field would have accepted alone".

I did not add a joint projection. The decoder answers the same question and is already wired into the candidate attempt, so the completion re-ask would be a second reader of one rule set — and where the two disagreed, the one that has actually built a value is the one to believe. The cost is that a `SizeOf` line between two fields of one record now stays unknown rather than counted, since nothing composes a string from a length to offer the decoder. That is the conservative direction and it is reported as such, not dropped.

**The pair you asked for, both held:**

```
invariant a < b   → · not known to be writable: cmp/p.a = p.b — nothing composed one:
                      every value tried at p.a = p.b was refused
                    boundary 0/0, no row demanded

invariant a <= b  → · no row is at cmp/p.a = p.b
                    boundary 0/1, row demanded
```

`aRuleRelatingTheTwoPositionsIsNotAnsweredByTheirRangesOverlapping` and `aRuleThatAdmitsTheDiagonalStillOwesTheRow`. The second is the one that matters for your worry about escaping into "any relational invariant ⇒ unknown": one character apart, and they answer differently.

The independent cases are unchanged — `Charge`/`Ceiling` 2/3, two plain `Int`s 0/1, the enumeration 0/1, `String` 0/1, `String.length` between two parameters 0/1.

**Specification.** The paragraph stated the rule your counterexample breaks, so it is rewritten: whether a row can be written on such a line is whether the two can hold one value *at once*, which is not what their ranges overlapping says, with the `Pair` model as the case in point and the two settlements named.

**The details.** The historical "String has no count / until it has a carrier" in the test's class comment is gone; it now names the two directions the carrier differs from the type in. PR body Checks says 4121.

Whole repo: 4121 in souther-compiler, 0 failures, 0 errors across all modules.

kawasima:
Reproduced, and following it through found the same hole in a third place. Fixed in `ab1e2de2`, by a wider change than proposed — the reason is below.

Your model, before:

```
    boundary    0/1
      · no row is at cmp/a = b (guard@16:5)
```

`Zero` admits only `0` and `NonZero` refuses it, so the diagonal holds nothing and the build was refused for not writing it. Your diagnosis of `uncertain` is exact: it is recorded only where a cut was made, which is where an edge used to come from and no longer is.

**Why I did not fix it there.** I made your change first — recorded the reading per parameter, dropped the `cuts.isEmpty()` guard, asked it by path so a `SizeOf` of an unread value was covered too. Your model went to `0/0`. Then I tried the regex pair you raised, in the form where no common length exists:

```souther
data A = String invariant String.matches("a", value)
data B = String invariant String.matches("bb", value)
guard String.length(x.value) > String.length(y.value)
```

Still counted. `allRulesRead()` is `everyClauseRead`, and that is `read &= !owed.unreadable()` over the clauses — it says the checker *understood* each clause, not that each clause reached the ranges. `String.matches` is read perfectly well and bounds nothing, so a position admitting one string looks unbounded from there. The same is true of the disequality in your own model: it was caught by the per-parameter change for a different reason than the one that matters.

The reading that would have to be consulted — whether the bound at a position was derived losslessly — is the one `FieldDomains` names in its javadoc as having no caller. Adding it means reaching into `InvariantChecker` to tell "read as an obligation" from "taken into the numbers", which is a change to the reading every measure depends on, made to serve one new consumer.

**What I did instead.** A line between two positions is settled by a witness and by nothing weaker: a row already on it, or a value the module's decoder took, which is the one thing here that reads every rule of a value at once. No projection proof for this shape at all.

That is the rule the specification already states for an edge nothing can prove — counted when a witness is found, never dropped for want of one — and it closes all three cases at once rather than the two that were reported.

**What it costs, stated plainly.** A line between two lengths — `String.length(a) > String.length(b)` on two bare parameters — was counted and now is not, because nothing composes a string from a length to offer the decoder. A person can obviously write that row. It is reported, not dropped:

```
· not known to be writable: cmp/String.length(a) = String.length(b) — nothing composed one:
  nothing here composes a value at String.length(a) = String.length(b)
```

I took the under-report over the over-report because that is the direction the existing rule picks for writability, and because the over-report refuses a build for a row nobody can write.

**Where it lands:**

| | before | now |
|---|---|---|
| `/= 0` against `== 0`, different parameters | 0/1 | **0/0** |
| two patterns with no common length | 0/1 | **0/0** |
| two patterns, values compared | 0/1 | **0/0** |
| `Pair invariant a < b` | 0/0 | 0/0 |
| `Pair invariant a <= b` | 0/1 | 0/1 |
| `Charge`/`Ceiling` | 2/3 | 2/3 |
| two plain `Int`s | 0/1 | 0/1 |
| two `A`/`B` both `>= 0` | 0/3 | 0/3 |
| enumeration | 0/1 | 0/1 |
| `String` | 0/1 | 0/1 |
| two lengths | 0/1 | **0/0** |

Every line that a decoder-accepted value witnesses is still counted, so this has not degenerated into "any invariant ⇒ unknown".

Tests: `aRuleTheRangesCouldNotTakeInIsNotAProofEither` holds your model, and `aLineOnAMeasureIsReadAndPromisedByNothing` holds the cost. The `a < b` / `a <= b` pair and the `A`/`B` both-`>= 0` positive control stay as they were.

Specification updated: the ranges do not settle it, and what refuses the pair need not be in either range — a rule relating the two, or a rule the ranges could not take in.

Whole repo green: 4122 in souther-compiler, 0 failures, 0 errors.

Issue #643, opened 2026-08-11
A gap no strategy is written for is reported as an input that cannot be composed


kawasima:
`--generate` says of an uncovered arm that nothing composes an input for it. What is true is that nothing here has a strategy that composes an input for the sake of an arm — and the row is one a person writes in a line.

The distinction is the one #467 drew for boundaries: what the rows cover, what can be written, and what was tried are three answers. `GenerationOutcome.NotSupported.Reason` already knows this — its javadoc says "a fact about which strategies are written, not about the model. Each of these is a strategy that could exist and does not. None of them says the gap cannot be met" — and the sentence it renders says the opposite of its own contract.

## Reproduction

Split Okinawa out of `shippingfee`'s region enumeration and give it a fee of its own:

```souther
data 本州
data 北海道
data 沖縄
data 離島
data 配送地域 = 本州 | 北海道 | 沖縄 | 離島

let 離島以外は無料 (地域: 配送地域): 送料無料 | 送料あり =
    match 地域 with
        | 離島 -> 送料あり { 金額 = 1500 }
        | 本州 -> 送料無料
        | 北海道 -> 送料無料
        | 沖縄 -> 送料無料

let 地域別送料 (地域: 配送地域): Int =
    match 地域 with
        | 本州 -> 500
        | 北海道 -> 800
        | 沖縄 -> 1200
        | 離島 -> 1500
```

The existing rows then leave the new case uncovered, which is reported exactly as it should be:

```
    partition   axes 3   single-axis 7/8   pairs 16 reached / 16 known reachable, 4 untried
      · no row is in `沖縄`
    branch      10/12
      · no row goes through `case 沖縄` (61:5)
      · no row goes through `case 沖縄` (54:5)
```

`--generate` offers two rows for the class, and then says this of the arms:

```
// nothing offers a row for `case 沖縄` in `送料を求める`: nothing here composes an input for the arm it would reach
```

Both arms are reachable, and writing the rows by hand closes the gap:

```souther
    | "沖縄・一般・5000以上" :
        (注文 { 番号 = 注文番号("0000-000009"), 合計 = 商品合計(5000), 地域 = 沖縄, 会員 = 一般 })
            -> 送料無料
    | "沖縄・一般・5000未満" :
        (注文 { 番号 = 注文番号("0000-000010"), 合計 = 商品合計(4999), 地域 = 沖縄, 会員 = 一般 })
            -> 送料あり { 金額 = 1200 }
```

```
[...]

Issue #642, opened 2026-08-11
A newtype over a sum loses signature case coverage even though the wrapped sum is coverable


kawasima:
A behavior parameter whose type is a single-value newtype over a sum has the same possible cases as the sum it wraps, but signature case coverage does not see them.

This is separate from #631. That issue concerns partition axes and the `classesOf` / representative / classifier path, and is fixed. This one concerns signature case coverage and has its own pair of readers: `Adequacy.coverableCases` derives the cases to cover, while `FixtureReader.constructedCase` reads which case a row actually supplied.

## What it does today

```souther
module demo

data Ok

data Approved = { id: Int }
data Rejected = { why: String }
data Decision = Approved | Rejected
data DecisionN = Decision

behavior bare : (decision: Decision) -> Ok
    constructs Ok
let bare (decision) = Ok

behavior wrapped : (decision: DecisionN) -> Ok
    constructs Ok
let wrapped (decision) = Ok

example bare
    | (Approved { id = 1 }) -> Ok

example wrapped
    | (DecisionN(Approved { id = 1 })) -> Ok
```

```
  bare                     implemented   rows 1    pending 0
                in #1 specified 1/2
      · no row uses `Rejected`
    partition   axes 1   single-axis 1/2
      · no row is in `Rejected`
  wrapped                  implemented   rows 1    pending 0
    signature   not applicable (this behavior's output is not a sum)
    partition   axes 1   single-axis 1/2
      · no row is in `Rejected`
```

The two parameters differ by a wrapper and nothing else. The bare one has two cases to cover and one covered. The wrapped one has no input line at all: the only thing signature coverage says about it is about its output, so the position reads as one with nothing to cover, beside a partition measure that says on the same line that it divides two ways.

Writing more rows does not bring the line back. The position is not counted rather than counted as uncovered.

## Where it comes from

`Adequacy.coverableCases` asks whether the written type itself is a sum. A `Type.Ref` naming `DecisionN` is a newtype, so this reader stops at the name instead of reading through it to `Decision`.

Widening that alone is not the fix, and does not read as a smaller version of it. With `coverableCases` reading through the names — and nothing else changed — the same model reports:

```
  wrapped                  implemented   rows 2    pending 0
                in #1 specified 1/2
      · no row uses `Approved`
      · no row uses `Rejected`
```

[...]

Issue #631, opened 2026-08-10
A single-value newtype loses the classes of what it wraps, and the position is reported as one the model does not divide


kawasima:
A position typed as a single-value newtype has the classes of what it wraps — its values are those values under a name — and it comes back with none. The report says *not derivable*, which is defined as a fact about the model (§example-report-vocabulary, ADR-0090), and the model has divided the values three ways two lines above.

## What it does today

Two behaviors differing by the wrapper and by nothing else:

```souther
module example.enumwrap

data Ok
data Prospecting
data Qualified
data Won
data Stage = Prospecting | Qualified | Won
data StageN = Stage

behavior bare : (x: Stage) -> Ok
    constructs Ok
let bare (x) = Ok

behavior wrapped : (x: StageN) -> Ok
    constructs Ok
let wrapped (x) = Ok
```

`souther examples enumwrap.sou --generate`:

```
  bare                     implemented   rows 0    pending 0
    partition   axes 1   single-axis 0/0   (1 not measured: no row names this behavior)
  wrapped                  implemented   rows 0    pending 0
    partition   not measured (no partition axis was derived at any position)
      · not derivable: x

// example bare
//     | "x=Prospecting" : (Prospecting) -> <?>
//     | "x=Qualified" : (Qualified) -> <?>
//     | "x=Won" : (Won) -> <?>
```

Three rows for the bare position and none for the wrapped one. No `guard` is involved: the classes a sum states are the type's, and wrapping the type takes them away.

## Not only sums

The same, with `Bool`:

```souther
data Flag = Bool

behavior bare    : (x: Bool) -> Ok      // partition axes 1, rows for true and false
behavior wrapped : (x: Flag) -> Ok      // not derivable: x
```

So the shape is *a single-value newtype over anything that states classes*, and today that is a sum and a `Bool`. A newtype over an optional is refused at the declaration, and a newtype over a number has no classes to lose — everything outside an invariant's bound is refused at construction, so that case draws a boundary and no partition, which is why this went unnoticed.

## Where it comes from

`Partitions.classesOf` branches on `TypeOps.isSumType(type, symbols)` and on `type == Type.BOOL`. A `Type.Ref` naming a newtype is neither, so it falls through to the numeric reading and comes back with nothing. There is no arm that reduces a single-value newtype to what it wraps.

[...]

Issue #626, opened 2026-08-10
A threshold compared against a field of a list element draws no boundary, though the same threshold on a bare position does


kawasima:
The same literal threshold, written the same way, is a boundary when the value is a parameter and is nothing at all when the value is a field of a list element. A position typed `List<T>` reports `not derivable`, and the report stops there rather than looking for lines inside the element.

## What it does today

Both behaviors below compare against `21000`:

```
module example.incollection

data Charge = Int
    invariant value >= 0

data Item = { charge: Charge }

data Total = Int
    invariant value >= 0

// The same literal threshold, twice: once on a bare position, once on a field
// of a list element.
behavior overThreshold : (charge: Charge) -> Total
    constructs Total
let overThreshold (charge) = {
    guard charge.value >= 21000 else Total(0)
    Total(1)
}

behavior countOverThreshold : (items: List<Item>) -> Total
    constructs Total
let countOverThreshold (items) =
    Total(List.length(List.filter(i -> i.charge.value >= 21000, items)))

example overThreshold
    | "under" : (Charge(1000)) -> Total(0)
    | "over" : (Charge(50000)) -> Total(1)

example countOverThreshold
    | "one under, one over"
        : ([ Item { charge = Charge(1000) }, Item { charge = Charge(50000) } ]) -> Total(1)
```

```
  overThreshold            implemented   rows 2    pending 0
    partition   axes 1   single-axis 2/2
    boundary    0/3
      · no row is at overThreshold/charge = 0 (invariant Charge (min))
      · no row is at overThreshold/charge = 21000 (guard@16:5)
      · no row is at overThreshold/charge = 20999 (guard@16:5)
    branch      2/2
  countOverThreshold       implemented   rows 1    pending 0
    partition   not measured (no partition axis was derived at any position)
    boundary    not measured (no line was derived at any position)
      · not derivable: items
    branch      2/2
```

`overThreshold` gets an axis and three rows offered. `countOverThreshold` gets `not derivable: items` and nothing else. On its own it passes `--strict` with `0`, and this survives:

```
-    Total(List.length(List.filter(i -> i.charge.value >= 21000, items)))
+    Total(List.length(List.filter(i -> i.charge.value > 21000, items)))
```

No row fails, though the two disagree on an item charged exactly `21000`.

## Why `not derivable: items` is the wrong place to stop

[...]

PR #644, opened 2026-08-11
Make partition derivation exhaustive

kawasima:
A position the derivation could not answer for was reported as a position the model does not divide. Five issues in a row were that one sentence — #602, #610, #629, #631, #626 — each fixed where it was found, each coming back one reader over. What they have in common is that a report about a model was produced by a reader that had run out, and nothing in the code told the two apart.

So the walk is a protocol now. Each phase answers as a value, an answer says which of the two it is, and the sentence "the model divides this position no way" is what completing the last of them produces.

```
Type
 ↓ TypeView            what the position is, and what it is written as, read once
 ↓ PartitionInput      proof that a value can stand at it at all
 ↓ LocalInspection     Evidence(classes, cuts) | Exhausted
 ↓ StructuralInspection Children | Leaf | Blocked(reason)
 ↓ BodyCutInspection   Evidence | Exhausted | Blocked(reason)
 ↓ PendingPosition.complete
                       Derived | CannotDerive(reason) | Absent
```

## The four cases it is for

| case | before | after | what it holds |
| --- | --- | --- | --- |
| a declaration reachable from itself | Absent | TYPE_UNRESOLVED | not knowing is not a fact about the model |
| a threshold on a list's elements (#626) | UNSUPPORTED_SYNTAX | UNSUPPORTED_TRAVERSAL | the cause follows where the reading stopped |
| `guard List.length(items) < 3` | Derived | Derived | a pending block is still rescued by evidence at the same position |
| a sum under a name (#631) | Absent | Derived | reading through a name reaches the readers that need it |

#626 was fixed by the walk running on the structural answer rather than by anything aimed at it. What did it was a precedence that had never been exercised: a reason from the reading that stopped outranks one offered by a reader of some rule that names the position.

## What each commit closes

- **Read a position's type once** — `TypeView` says what a position is and what it is written as. `Type.Ref` was one constructor for four things, so a walk that answered every constructor had still not answered every kind of position. Fifteen shapes, five of which are ways of not having determined one.
- **Prove a position is one a partition can be derived from** — the premise `Absent` needs, in the input's type rather than in a check the walk repeats.
- **Say a block is the reason a position would be left with** — a leaf and a block are both pending; they differ in what is reported if nothing else answers.
- **Preserve nominal projection in partition classes** — `data DecisionN = Decision` divides into `Decision`'s cases, a row writes `DecisionN(Approved { id = 1 })`, and a row already written is read back through the same names. `RepresentativeSource` became an algebra so that a name can be carried over a value nothing has composed yet.
- **Make local partition inspection explicit** — the classes a type states and the lines its rules draw come back as one answer, so "local evidence ran out" is a value and not two empty lists to be noticed.
- **Settle why a rule could not be a line where the rule is read** — the reasons a comparison stopped are `BlockReason`s, recorded by the producer and projected to the published word in one place.
- **Make an absence what completing a position produces** — `Why.Absent` has no constructor to call; it is reached through a `PendingPosition`, which is package-private cases and all.
- **Say of the derivation only what its types forbid otherwise** — an audit of the universal claims in the protocol's own words. It found the hole above: closing the terminal value was not enough while the introduction rules of the proof were public.

## What it does not change

No report changes. The examples corpus was rendered with the jar from `develop` and with the jar from this branch, and every module's output is identical apart from the sentence issue #626 is about. The reason the corpus has no wrapped sum at an input position is why the two readings stayed apart for so long — the tests are what hold it.

`schemaVersion` stays at 1. The schema states its own rule: a change that removes or renames anything raises it, and a word added to an enumerated field does not, since a document written before that word existed is still a document of this version. Two words were added — `type_unresolved` and `unsupported_traversal` — and both are now asserted to reach a document rather than only to exist in the vocabulary.

## Follow-ups

- #642 — a newtype over a sum loses signature case coverage. The same nominal projection at a different measure, with its own pair of readers; widening one alone reports one of two covered and neither of the two covered.
- #643 — `--generate` says nothing composes an input for an arm, where what holds is that no strategy composes one for the sake of an arm. `GenerationOutcome.NotSupported.Reason` already documents itself as the first and its words say the second.
- The class producer and the value producer still call each other. A class needs a representative value and a representative value needs classes, so the recursion is real and moving the code does not remove it.

Nothing here closes an issue automatically; #631 and #626 are closed by hand after this merges.

### DIFF
--- a/souther-compiler/src/main/java/souther/compiler/check/Shape.java
+++ b/souther-compiler/src/main/java/souther/compiler/check/Shape.java
@@ -24,8 +24,25 @@ import java.util.Set;
  * through it, which is what left a {@code data StageN = Stage} with no classes while the line drawn
  * on the same position read straight through the name.
  *
+ * <p><b>This is a total classification, and some of its cases are not value shapes.</b> A reading
+ * that answers "a record", "a sum", "a list" for the types it knows and nothing for the rest is the
+ * defect, not the base case: what a reader does with a type it cannot interpret is decided by the
+ * reader, and every reader has decided it silently. So the failures to determine a shape are cases
+ * here, of the same standing as the shapes, and a reader answering all of them has answered every
+ * type it can be handed.
+ *
+ * <p>Which is why {@link Uninhabited}, {@link Bottom}, {@link Erroneous}, {@link Undecided} and
+ * {@link Unresolved} are five cases and not one {@code Unknown}. Two of them say something about the
+ * values — no value arrives, or the element type of an empty collection is not fixed yet — and three
+ * say something about this compiler. Collapsing them would put a reader in the position of having to
+ * tell a semantic fact from an analysis failure after the distinction was thrown away, which is the
+ * shape of the defect this exists to remove: a position nothing could read was reported as one the
+ * model divides no way.
+ *
  * <p>Nothing here says what a position divides into or what may be written at it. Those are
- * questions asked of a shape, by readers that must answer every case of it.
+ * questions asked of a shape, by readers that must answer every case of it — and a reader that finds
+ * itself writing this switch out again wants a question of its own asked once, the way
+ * {@link Carrier#ofValue} asks whether a line can be drawn, rather than a copy of this one.
  */
 public sealed interface Shape {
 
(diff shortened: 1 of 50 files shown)


kawasima:
Both blockers confirmed and fixed in `42ea6005`.

**`pending` dropped by the second phase** — reproduced before touching anything. A blocked position under a rule that is read and divides nothing:

```souther
behavior outside : (items: List<Item>) -> Ok constructs Ok
let outside (items) = { guard List.length(items) < -1 else Ok
    Ok }
```

```
· not derivable: items                      ← before: an absence
· not read: items (its values are held inside something this does not reach into)   ← after
```

The threshold is understood — it is a threshold, on a term this measures — and no value of the position is on the far side of it, so `reachable` is empty and the axis is rebuilt with nothing to say it had stopped. `Axis` has `measuredAt` and `carrying` now, so moving a position to another term and giving it what a body's rules found are things that happen to an axis rather than a constructor call at the site. The row is in `WhatStoppedADerivationIsWhatIsReportedTest`, beside the three it belongs with.

**`PartitionInput` not at the entry** — the reachability was measured rather than assumed. An anonymous union is refused at a parameter (E1312) and at a field (E2307), so no compiling model reaches the walk with a `Shape.Cases`; moving the proof to the front refuses nothing that is written today and makes the disagreement loud if a boundary ever regresses. `LocalInspection.inspect` takes the `PartitionInput` now, so the phase cannot be handed a position that was not admitted — which is the part a test could not hold — and the record is package-private, since a public record is a canonical constructor anything can call.

Corpus output is unchanged from the previous head, which is what both fixes have in common: they are about models nothing in it has.

kawasima:
Both fixed in `1ebc6c42` (blocker 1) and the commit before it (blocker 2).

**wrapped product** — reproduced first, against `develop` to see which way it moved:

```
develop : wrapped … · not derivable: x                     ← an absence
42ea6005: wrapped … axes 1  single-axis 0/2                ← the axis, and no row readable at it
          · a row's value at `wrapped/x.flag` could not be read
          measurement: partial
1ebc6c42: wrapped … axes 1  single-axis 1/2  · no row is in `false`   ← what `bare` says
```

`BehaviorInputs` carries what a behavior takes — names, declared types, symbols — and owns the walk into a row's values, since none of the three is enough alone. It is what `Generator.Subject` is made of now, rather than a second triple beside it.

The rule you named is the one implemented: the names come off **on the way** to the next field and never at the end. What comes back is the value as the position wears it, because which names the position itself is written under is what `Classifier.under` reads there — peeled to the base, a `data Flag = Bool` field would arrive at its classifier as a bare `Bool`.

The other two directions: the generator reads the position through the same `TypeView` and writes the names back on what it composed (`(SlotN(Slot { flag = false }))`), and a record's clauses about its fields are read off the record under the names — with the declaration as written where there is no record, which three existing tests caught: a position that is not a record has no fields for a clause to relate, and its own rules still say what a reading could not turn into a range.

Held end to end in `APositionUnderANameIsReachedThroughItTest`, against the bare form in every row, plus a `PairN` owing the same edges a `Pair` does.

**`PendingPosition.of`** — `StructuralInspection.Pending` is a sealed sub-interface of `Leaf | Blocked`, `Axis.pending` is typed as that, and the walk switches on the two kinds. So `Children` cannot be there to be read as a leaf, and `null` is a fail-fast rather than an absence.

Corpus output is unchanged from the previous head.

kawasima:
Fixed in `0caeb0d6`.

```java
public record Subject(BehaviorInputs inputs, List<Axis> axes)
```

with `parameters()` / `types()` / `symbols()` delegating, so what reads a subject does not have to know where the three came from — but there is one value holding them. `Adequacy` builds the inputs once per behavior and hands the same value to both halves: the rows already written are read by it, and the rows offered are generated from it.

You are right that the javadoc was the only thing saying the subject was made of the inputs. That is the same failure the whole branch is about, one level up: a claim in words that the types did not carry.

Corpus output unchanged.

PR #645, opened 2026-08-11
Say what a generator has no strategy for as that

### LINKED ISSUE #643: A gap no strategy is written for is reported as an input that cannot be composed

kawasima:
`--generate` said of an uncovered arm that nothing composes an input for it. Splitting a case out of an enumeration and running it reports:

```
// nothing offers a row for `case 沖縄` in `送料を求める`: nothing here composes an input for the arm it would reach
```

while the row is two lines, and writing it closes the gap:

```souther
    | "沖縄・一般・5000未満" :
        (注文 { 番号 = 注文番号("0000-000010"), 合計 = 商品合計(4999), 地域 = 沖縄, 会員 = 一般 })
            -> 送料あり { 金額 = 1200 }
```

`GenerationOutcome.NotSupported.Reason` already documents itself as "a strategy that could exist and does not", and that "none of them says the gap cannot be met". Its words said the opposite. Each of the three now says what is missing here:

```
// nothing offers a row for `case 沖縄` in `送料を求める`: rows here are composed for classes and for boundaries, and nothing composes one for the sake of an arm
```

The words are the contract at this point, so they are held one by one in `EveryGapTheBuildRefusesGetsAnAnswerTest` — there is no property of a sentence that says it is about a compiler, and what there is, is a person reading each of them next to what it is written for.

Closes #643

### DIFF
--- a/souther-compiler/src/main/java/souther/compiler/partition/GenerationOutcome.java
+++ b/souther-compiler/src/main/java/souther/compiler/partition/GenerationOutcome.java
@@ -48,10 +48,13 @@ public sealed interface GenerationOutcome {
          *
          * <p><b>Which the words have to carry too.</b> A sentence saying nothing composes an input
          * for an arm is read as there being no input that reaches it, and an author who writes the
-         * row in one line has been told something false — the row for a case somebody added to an
-         * enumeration is two lines, and this said no input for it could be composed (issue #643).
-         * So each of these says what is missing here, and what a reader does about it is write the
-         * row that this cannot.
+         * row by hand has been told something false (issue #643). So each of these says what is
+         * missing here, and what a reader does about it is write the row that this cannot.
+         *
+         * <p>Including where what is missing is upstream of the strategies. No axis at a position
+         * is no classes <em>derived</em> there, and the position may well have cases — the one this
+         * is written for is a sum past the axis limit, whose cases are in the declaration and in no
+         * partition. Said as there being no classes, it is the model that reads as having none.
          */
         public enum Reason {
 
@@ -67,7 +70,7 @@ public sealed interface GenerationOutcome {
 
             /** The position the case belongs to is not one any axis was derived at. */
             NO_AXIS_AT_THIS_POSITION("no axis was derived at the position this case belongs to, so"
-                    + " there are no classes here to compose a row from");
+                    + " no classes were derived there to compose a row from");
 
             private final String said;
 
--- a/souther-compiler/src/main/java/souther/compiler/partition/GenerationOutcome.java
+++ b/souther-compiler/src/main/java/souther/compiler/partition/GenerationOutcome.java
@@ -45,18 +45,29 @@ public sealed interface GenerationOutcome {
          *
          * <p>Each of these is a strategy that could exist and does not. None of them says the gap
          * cannot be met, and none is read as evidence about the model.
+         *
+         * <p><b>Which the words have to carry too.</b> A sentence saying nothing composes an input
+         * for an arm is read as there being no input that reaches it, and an author who writes the
+         * row in one line has been told something false — the row for a case somebody added to an
+         * enumeration is two lines, and this said no input for it could be composed (issue #643).
+         * So each of these says what is missing here, and what a reader does about it is write the
+         * row that this cannot.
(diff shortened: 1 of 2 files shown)


kawasima:
Fixed in `819bb2e5`, with the reading you gave: `no classes were derived there to compose a row from`.

The `WIDE` model is the case that shows it. Position 13 is a `Kind = Domestic | Overseas`, so `Overseas` is a case of the position's type and in no partition — what is missing is the derivation, and `there are no classes here` said it was the classes. The enum's javadoc now says that where what is missing is upstream of the strategies, it is still what is missing here.

The `one line` / `two lines` clash is gone too — `an author who writes the row by hand`. The reproduction's row is two lines and the sentence did not need a count.

Issue #646, opened 2026-08-11
A newtype parameter accepts a value written without its nominal constructor


kawasima:
A parameter whose type is a single-value newtype accepts a row that does not write the nominal constructor, and accepts one that writes a different name over the same base. The value the behavior runs on is right either way — the decoder supplies the position's own name — so what is not held is the row's own spelling.

## What it does today

```souther
module demo

data Ok

data Approved = { id: Int }
data Rejected = { why: String }
data Decision = Approved | Rejected
data DecisionN = Decision
data OtherN = Decision

behavior wrapped : (decision: DecisionN) -> Ok
    constructs Ok
let wrapped (decision) = Ok

example wrapped
    | (Approved { id = 1 }) -> Ok
    | (OtherN(Approved { id = 2 })) -> Ok
```

Both rows compile with no diagnostic. Neither writes a `DecisionN`.

An unrelated value at the same position is refused, so the position is being held to something: `(Ok)` is *E1903* `an input could not be built`, and so is `(AmountN(3))` for a `data AmountN = Int` — `expected object`. What is admitted is anything the base's decoder reads, whatever name the row put over it or left off.

That the value comes out right is checkable on the row itself. With

```souther
behavior echo : (decision: DecisionN) -> DecisionN
let echo (decision) = decision

example echo
    | (Approved { id = 1 }) -> DecisionN(Approved { id = 1 })
    | (OtherN(Approved { id = 2 })) -> DecisionN(Approved { id = 2 })
```

both rows hold: what the behavior was applied to was a `DecisionN` in each case. So the name is not being taken off the value. It is being supplied by the decoder to a row that did not write it.

## Where it comes from

An input fixture is built at the shape the boundary established — `FixtureShape.Nominal(DecisionN)` — and decoded through the derived decoder for that name, which reads the neutral form of its base. A neutral form spelling `Approved { id = 1 }` is one the base reads, so it decodes and is wrapped; a neutral form spelling `OtherN(Approved { id = 2 })` is another, since `OtherN` has the same base. Nothing between the written text and the decoder holds the row to the name the position is written under.

## Why it is worth its own issue

It is the opposite direction from #642, at the same boundary. Signature coverage has to read *through* the name to say what case a row supplied; admission must not, or a row states a value the position could not have been given. Reading them as one rule is what #642 was about, so they are worth separating here too:

| | reads through the name |
|---|---|
| signature input coverage | yes |
| value construction | no |
| parameter admission | **no — and does today** |
[...]

PR #647, opened 2026-08-11
Read an input's case at the position it is written at

### LINKED ISSUE #642: A newtype over a sum loses signature case coverage even though the wrapped sum is coverable

kawasima:
Closes #642 (a PR onto `develop` does not close it automatically — close by hand after merge).

Signature input coverage was reading which nominal constructor built a fixture and counting that as which case the position sees. For a `data DecisionN = Decision` those are two different names, so a position the declaration divides two ways was reported as one the model divides no way.

Both ends of the measure read the nominal one, which is why widening either alone moves the defect rather than removing it — with the denominator read through the name and nothing else changed, every row lands outside the set it is counted in. So:

* `Adequacy.coverableCases` becomes `inputCoverableCases` / `outputCoverableCases`, sharing `casesOfSum` and differing by which type each hands it. Input reads through the names the position writes its values under; output keeps the old rule.
* `FixtureReader.caseUnder` is the position-relative reading. `constructedCase` is unchanged, because its nominal answer is what picks the shape a fixture is built against and what an expected arm is held to.

The names come off on the same terms `Classifier.under` takes them off an observation, and the denominator's terminal and the numerator's layers are the two halves of one `TypeOps.newtypeSpine`.

## Not in this change

An output written under a name is still answered with that name. Whether signature coverage should read a newtype-over-sum output through its name is a separate question and is stated as unanswered in the javadoc rather than settled by this change.

That a row may write `Approved { id = 1 }` bare, or `OtherN(Approved { .. })`, at a `DecisionN` position is #646. Held at admission, those rows never reach this measure — no filter was added here, which would have made that fix untestable at this measure.

## Verification

* `SignatureCoverageReadsAPositionThroughItsNamesTest`, 8 rules: the coverable set, either case read at the position, a count agreeing with the cases beside it, a doubly-worn name, a name the position does not write not being read through, an output under a name still answered with it, and a bare sum unchanged. 6 of the 8 were red before the change; `aBareSumIsUnchanged` and the output guard were the positive controls.
* `mvn -o clean test` green across all modules.
* The corpus reports are unchanged: every single-value newtype in `souther-examples` wraps a primitive (`Int`, `Decimal`, `String`, `Date`), so no position there is a name over a sum.

### DIFF
--- a/souther-compiler/src/main/java/souther/compiler/ExampleVerifier.java
+++ b/souther-compiler/src/main/java/souther/compiler/ExampleVerifier.java
@@ -9,6 +9,7 @@ import souther.compiler.check.BoundaryOutput;
 import souther.compiler.types.Type;
 import souther.compiler.types.TypeName;
 import souther.compiler.check.TypeOps;
+import souther.compiler.check.TypeView;
 import souther.compiler.coverage.Probe;
 import souther.compiler.diag.Diagnostic;
 import souther.compiler.diag.msg.ExampleMessage;
@@ -634,7 +635,7 @@ public final class ExampleVerifier {
                 state.failed(FailurePhase.INPUT_FIXTURE);
                 return;
             }
-            state.inputCases.add(caseWritten(fixtures, row.inputs().get(i)));
+            state.inputCases.add(caseWritten(fixtures, row.inputs().get(i), ins.get(i).type()));
             state.inputs.add(fixtures.observed(args[i]));
         }
         // validate the expected arm/value against the output cases before running. Which case the row
@@ -726,11 +727,19 @@ public final class ExampleVerifier {
         state.failurePhase = FailurePhase.NONE;
     }
 
-    /** The case an input fixture constructs, or null where its text does not say. A form this cannot
-     * read is not a reason to lose the row: the measure that reads it says it could not classify. */
-    private static TypeName caseWritten(FixtureReader fixtures, Ast.Expr fixture) {
+    /**
+     * The case an input fixture supplies at {@code position}, or null where its text does not say. A
+     * form this cannot read is not a reason to lose the row: the measure that reads it says it could
+     * not classify.
+     *
+     * <p>Read at the position and not off the fixture alone. A position writing its values under a
+     * name divides into what its base divides into, so what a row there supplies is the case under
+     * that name — the same projection, in the direction that reads rather than the one that derives.
+     */
+    private TypeName caseWritten(FixtureReader fixtures, Ast.Expr fixture, Type position) {
         try {
-            return fixtures.constructedCase(fixture);
+            return fixtures.caseUnder(TypeView.of(position, symbols).wrappers().stream()
+                    .map(TypeOps.Layer::named).toList(), fixture);
         } catch (RuntimeException e) {
             if (overspending(e) != null) {
                 throw e;   // the row's budget is gone; it is not a form that could not be read
--- a/souther-compiler/src/main/java/souther/compiler/FixtureReader.java
+++ b/souther-compiler/src/main/java/souther/compiler/FixtureReader.java
@@ -426,7 +426,28 @@ public final class FixtureReader {
      * spelling or nothing (issue #206).
      */
     TypeName constructedCase(Ast.Expr e) {
-        return constructedCase(e, new LinkedHashSet<>());
+        return constructedCase(e, new LinkedHashSet<>(), List.of());
+    }
+
+    /**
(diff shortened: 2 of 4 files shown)


kawasima:
Correcting my note above: this PR was already merged when I wrote it, so there is no branch to rebase and no cleanup to add here. #652 has since landed on top (`6f290901`), and both are on develop.

What stands is the observation, as a follow-up on develop. An input fixture is now held to the position it is written at, before anything reaches a decoder, so a row writing `OtherN(Approved { … })` at a `DecisionN` position is *E1903* and never reaches the measure. `caseUnder`'s permissive branch — `wears()` returning false and the written case being returned as it stands — no longer describes a fixture that can arrive, and its comment states a case that cannot happen. If one did arrive, answering "the row supplied `OtherN` at this position" would be a false statement about a position whose cases do not include it, so `null` is the answer that keeps measurement from stating what admission has already refused. The diagnostic stays where it is.

develop with both merged: `mvn -o clean test`, 6858 tests, no failures.

PR #652, opened 2026-08-11
Hold a fixture to the name its position is written under

### LINKED ISSUE #646: A newtype parameter accepts a value written without its nominal constructor

kawasima:
Closes #646 once merged.

A derived decoder reads the form its base reads and answers with a value of its own type, so the name a position is written under is one the decoder supplies rather than one it holds a row to. At a `data DecisionN = Decision`, the neutral form of `Approved { id = 1 }`, of `OtherN(Approved { id = 1 })` and of `DecisionN(Approved { id = 1 })` is the one value — measured, it is `{id=1, type=Approved}` in all three — and each decoded, so the behavior ran on a value the row never stated.

## What was admitted

The issue names a newtype over a sum. The reading had no nominal rule at all, so the same hole was open wherever a fixture writes a value:

| position | written | before |
| --- | --- | --- |
| `Int` | `AmountN(1)` | admitted |
| `List<Int>` | `IntList([1])` | admitted |
| `data AmountN = Int` | `1` | admitted |
| `data AmountN = Int` | `OtherAmountN(1)` | admitted |
| `data AmountNN = AmountN` | `AmountN(1)`, `1` | admitted |
| `data A = { id: Int }` | `B { id = 1 }` (same fields) | admitted |
| `A`, a case of `data S = A \| B` | `B { id = 1 }` | admitted |
| `Order { amount: AmountN }` | `Order { amount = 1 }` | admitted |
| `List<AmountN>` | `[1, 2]` | admitted |
| helper parameter `(a: AmountN)` | `1` | admitted |
| `AmountN` | a helper answering `Int` | admitted |
| `List<AmountN>` | a helper answering `List<Int>` | admitted |
| `Ok`, a unit case | `None` | admitted |

Ordinary code refuses every one of these (E1317), so the fixture reader was the one place a name could be invented.

## The rule

A fixture the model is given states a value of the position it is written at, and is asked before anything reaches a decoder.

A position answers one of four things. `OneOf(names)` for a nominal position — `TypeOps.leafCases` of the type it is written under, which expands a sum to its cases and stops at everything else, so a record takes itself, a sum takes the cases it lists and a newtype takes its own name, with no arm for any of the three. `NoName` for a primitive or a collection, which is as much a rule as the first: an `AmountN` is one wherever it is written, and its representation reading as an `Int` does not make it one. `Unsaid` where the position is not declared, or is one a fixture is refused at before this. And `OrAbsent(held)` for an optional, which takes absence or what it holds — read as one of the answers rather than opened before the answer is read, since an opened optional no longer says it made room for anything, and `None` at a unit case decoded as that case.

What a frame writes answers the same way — a `Name`, a value under `NoName`, `Absence`, or said `Elsewhere` — and the two are held against each other, so both directions of the coercion are the one rule.

An application says nothing in the text, so what it answered with is read instead, before the value becomes a neutral form: a helper's declared result is never consulted, and one that declares nothing is held to the same rule as one that declares everything. The reading of that value walks the position's type, so a helper answering a `List<Int>` has not answered a `List<AmountN>`, and an optional is absence or what it holds.

It is held at every position a value stands at rather than at the outermost one — the reading already recurses with the type of each field, element and argument, so a field, a list element, a newtype's inner value and a helper's argument each get it.

A spread is the one frame that states no value of its own. `Filed { ...d, filedOn = on }` copies a `Document`'s fields, and holding `d` to being a `Filed` would refuse how the language writes one record from another. That frame alone is exempt; every position under it is written the way any other is.

## What a row is told

```
-- EXAMPLE  E1903 ------------------------------nom.sou:12:7

12|     | (1) -> Ok
          ^

Input #1 of `takesAmount` could not be built: nothing here names a value of `AmountN`; one is written `AmountN(...)`

-- EXAMPLE  E1903 ------------------------------nom.sou:13:7

13|     | (OtherN(2)) -> Ok
          ^

Input #1 of `takesAmount` could not be built: `OtherN` is written where a value of `AmountN` stands; one is written `AmountN(...)`
```

## Fallout

Four kinds of red were possible; three appeared.

**Fixtures the hole admitted.** The issue-tracker corpus writes `assignee = "kawasima"` at an `Assignee?` field (six places) and `AStandInIsAdmittedByWhatItStandsForTest` asserted that `with quote = 2` stands in for a dependency answering `Amount`. Both are written out. The second stated the hole as a rule, in its name and in the specification, and is inverted with a positive control beside it.

**A refusal overtaken.** Admission runs before the reading, so a row naming a behavior (`(base)`) was told about a type instead of being told that a behavior is not a value a fixture can name. The arm for a bare name now branches on what the name denotes: a binding and a value open a frame at the same position, and everything else reaches the refusal that is about it.

**Generated candidates.** None. `FixtureTemplate` already writes the constructor, and `souther examples --generate` over the corpus emits `assignee = Assignee("x")`.

**The expected side.** Not held, at any depth, and filed as #653. A row may state what the behavior does not answer with, and reporting that disagreement is what the row is for, so what a row expects is built and compared. An anonymous union output settles it rather than scope: `Int | Rejected` takes a value under no name and `AmountN | Rejected` does not, and whether a nameless value belongs at such a position is a question about the value's type — the comparison's question, not admission's. So `-> 1` at an `AmountN` output stays E1905, and an expected `Receipt { total = 1 }` is still coerced by the decoder. That is the same hole on the side whose fix is not admission but not coercing — a different question, and #653 states it. `CompileAttemptedConstructionTest` keeps its `Accepted { code = "AB" }` row for that reason: it characterizes the boundary this PR draws.

## Specification

`BOUNDARY_VERSION` is not moved. It versions what a jar declares to another compiler, and an `example` does not leave its module.

No ADR: `docs/adr/README.md` excludes an implementation catching up with a rule already stated, which is what ADR-0014 and ADR-0032 stated. The specification gains `[[fixture-names-what-it-supplies]]`, loses the `with quote = 2` sentence that stated the hole, and names this among E1903's causes.

## Verification

`mvn -o clean test` across every module: 6850 tests, no failures. The characterization test is 25 rows of the matrix above; before the change its positive rows passed and every negative row admitted what it was written to refuse.

The later commits close what review found: a name written at a position that wears none, a helper's answer held only at its root, an optional a helper answered with, a spread reached through a value alias or a helper application, and `None` written where a position holds a value — which a unit case's decoder read as that case. Both directions of the coercion are one rule — a position answers `OneOf(names)`, `NoName`, `OrAbsent(held)` or `Unsaid`, and what a frame writes answers `Name`, `NoName`, `Absence` or `Elsewhere`.

One property is not covered by a test: a helper applied in a fixture runs exactly once. There is no way to observe an execution count from Souther, and approximating it with the evaluation budget would be unstable. It holds structurally — `applied` calls `answered` once, and the written-side admission only reads the helper table.

## Not in this PR

#647's `caseUnder` consumes the names a position wears permissively, and says in its comment that a fixture not wearing them is read as it stands. After this, such a fixture does not reach the measure. Making it strict is a cleanup that follows from this rule rather than from merge order, and it belongs with whichever of the two lands second — with the measurement answering `null` rather than throwing, since admission owns the diagnostic.


kawasima:
All five reproduce. Two of them are regressions the first commit introduced, and no test covered them — the measurement before the fix:

```
[1a Int <- AmountN(1)]                      ADMITTED
[1b AmountN <- AmountN(OtherAmountN(1))]    ADMITTED
[1c List<Int> <- IntList([1])]              ADMITTED
[2a List<AmountN> <- helper List<Int>]      ADMITTED
[2b Set<AmountN> <- helper Set<Int>]        ADMITTED
[2c Map<String,AmountN> <- helper Map]      ADMITTED
[3a AmountN? <- helper Some]  E1903 `pickAmount` answered with a `Some`    <- regression
[3b AmountN? <- helper None]  E1903 `pickAmount` answered with a `None`    <- regression
[4b spread through alias]     E1903 `Document` is written where a `Filed`  <- regression
[4d spread through helper]    E1903 `viaHelper` answered with a `Document` <- regression
[5a expected <- helper Int]   E1905 (never reaches admission)
```

One correction to the report, which does not change the finding: a helper cannot declare `AmountN?`. `let f (…): AmountN? = …` is E1402, and `None` in a helper body is E1303 — the optional has to be inferred. The case is still reachable, by returning a `?` field: `let pickAmount (o: Held) = o.amount`. That is what rows 3a/3b use, and both were refused.

`91a6e819` fixes all five. The diagnosis was right about the abstraction: `Set.of()` meaning "ask nothing" was two facts under one name.

## What changed

`admits(position)` answers one of three things now — `OneOf(names)`, `NoName`, `Unsaid` — and what a frame writes answers the same way: `Name`, `NoName`, `Absence`, `Elsewhere`. Holding the two against each other closes both directions with one rule, so `AmountN(1)` at an `Int` is refused for the same reason `1` at an `AmountN` is:

```
Input #1 of `takesInt` could not be built: `AmountN` is written where a value of `Int` stands; a value of `Int` wears no name
```

A union stays `Unsaid`, and deliberately: a fixture never reaches one, because a position of several types is refused where its shape is read, saying to write the case it is. Answering for it here would be a second rule about a position this cannot be asked about.

The built side is a walk over the position's type now, and it is the walk a stand-in already went through — `representedBy(Object, FixtureShape)` became `holds(Object, Type)`, so the stand-in path and the helper path cannot come to different answers, and the optional case is written once. That closes 2 and 3 together: an optional is absence, or what it holds, read through `Some`; a collection is its container holding what its element type holds.

`Admission` is three-valued, and its propagation is the rule you named: `HELD` at a frame that states a value, `HELD_BELOW` for a spread's own frame, `UNHELD` for a reading that compares. `below(admission)` maps the first two to `HELD` and leaves `UNHELD` alone, and it is applied where a child position is opened — a field, an element, a tuple part, a newtype's inner value. The forms that go on at the same position — a name standing for a body, a `let`, an application — carry the frame's own value, which is what 4 was about.

## The expected side is dropped

Taking your suggestion, and the anonymous union is what decides it rather than scope. `Int | Rejected <- 1` has to pass and `AmountN | Rejected <- 1` has to fail, so a rule stated as "which names does this position take" cannot answer for it: whether a nameless value belongs there is a question about the value's type, which is the comparison's question and not admission's.

So an expected value is read `UNHELD` at every depth, and `-> 1` at an `AmountN` output stays E1905 as it was before this branch. `CompileAttemptedConstructionTest` is reverted to `Accepted { code = "AB" }`. The characterization keeps a row for the boundary, so where the rule stops is pinned rather than left unsaid.

`Receipt { total = 1 }` as an *expected* value is therefore still coerced by the decoder. That is the same hole on the side where the fix is not admission but not coercing, and comparing faithfully is its own change. I have not filed it — say the word and I will.

## Characterization

20 rows to 24, covering the eight families you listed: nominal-at-a-nameless-position (three rows), helper-returned `List`/`Set`/`Map`, an optional field taking a `Some`, a `None` and a helper answering either, a spread reached through a value alias and through a helper, and the expected-side boundary.

`mvn -o clean test` across every module: 6849 tests, no failures.

## The stale comment

Fixed in the same commit. It says what the line does now: a stand-in is read through the decode an input goes through and is held to it as an input is, because a stand-in is what the dependency answers while the row runs.

kawasima:
Both confirmed. `1fce3b27`.

## `None` at a position that holds a value

Measured before the fix:

```
[Ok(unit) <- None]           ADMITTED
[AmountN  <- None]           E1903 ... is required
[Int      <- None]           E1903 ... is required
[AmountN? <- None (control)] ADMITTED
```

The unit case is the one that mattered, exactly as you said: `AmountN` and `Int` were stopped by their decoders finding nothing to read, not by the rule, and a unit decoder reads nothing at all — so `(None)` at an `Ok` was handed an `Ok` the row never wrote.

The cause was `admits()` opening the optional before answering, which dropped the room the optional had made. An optional is one of the answers now rather than something opened before the answer is read:

```java
case Type.OptionOf o -> new Admits.OrAbsent(admits(o.element()));
```

and `Stated.Absence` is admitted against `Admits.OrAbsent` and nothing else. After:

```
Input #1 of `takesOk` could not be built: `None` is the absent value of a `?` position, and this position holds a value of `Ok`
```

All four rows are in the characterization, the unit one included.

## The revert

You are right, and my report was wrong. I ran `git checkout <path>`, which restored from the index — where the change was already staged from the first commit — so the file never went back. It is `git checkout develop -- <path>` now, and the file is identical to develop:

```
$ git diff develop --stat -- .../CompileAttemptedConstructionTest.java
(no output)
```

Your reason for keeping it is better than my reason for changing it: `Accepted { code = "AB" }` is now a characterization of the boundary this PR draws, so the known hole stays written down where it can be read.

## The expected side is filed

#653. Framed as you put it — not "admit before decode" but how an expected value is built as a value of its own type and compared against what came out without either side being coerced into the other. It records the anonymous union as part of why it is not the input rule, and the three things that have to be settled: which decode an expected value goes through, what a mismatch shows when the two sides are of different types, and what a row writing a base literal at a newtype field is told, under which code.

`mvn -o clean test`: 6850 tests, no failures. The characterization is 25 rows.

PR #656, opened 2026-08-11
Read a floor on how much a field holds wherever the rule stating it is written

### LINKED ISSUE #650: A floor on how much a field holds is read from the record and lost before the value is built
`--generate` builds a value for a field out of what the field's own type stands for. A rule on the record that says how much the field holds — `List.length(xs) >= 1`, `String.length(name) >= 3` — is read into the numeric domain and then has no way to reach the field, so the value built is the one the carrier stands for, and the construction refuses it. A row fills every position at once, so that refusal takes every other position's boundary row down with it.

## Reproduction

```souther
module demo

data Yen = Int invariant value >= 0
data Bag = { xs: List<Int> } invariant List.length(xs) >= 1
data Ok = { n: Int }

behavior bareList : (bill: Yen, xs: List<Int>) -> Ok
    constructs Ok
let bareList (bill, xs) = Ok { n = bill.value + List.length(xs) }

behavior wrappedList : (bill: Yen, bag: Bag) -> Ok
    constructs Ok
let wrappedList (bill, bag) = Ok { n = bill.value + List.length(bag.xs) }

example bareList    | (Yen(5), [1]) -> Ok { n = 6 }
example wrappedList | (Yen(5), Bag { xs = [1] }) -> Ok { n = 6 }
```

```
$ souther examples r3.sou --generate --boundaries
  bareList     boundary 0/1   · no row is at bareList/bill = 0 (invariant Yen (min))
  wrappedList  boundary 0/1   · no row is at wrappedList/bill = 0 (invariant Yen (min))

// example bareList
//     | "bill = 0" : (Yen(0), []) -> <?>
// no row for `bill = 0` in `wrappedList`: every value tried was refused at construction, which does not make the combination impossible
```

The gap being filled is `bill = 0`, and `bill` has nothing to do with the list. The sentence is accurate about what happened — `Bag { xs = [] }` was tried and refused — and `Bag { xs = [0] }` is a value that would have been accepted.

It is not about collections. A string field carrying a length rule fails the same way, and so does a field of a field:

```souther
data Named = { name: String } invariant String.length(name) >= 3
data Tagged = { tags: Set<Int> } invariant Set.size(tags) >= 2
data Inner = { name: String } invariant String.length(name) >= 3
data Outer = { inner: Inner }
```

```
// no row for `bill = 0` in `byName`: every value tried was refused at construction, ...
// no row for `bill = 0` in `byTags`: every value tried was refused at construction, ...
// no row for `bill = 0` in `nestedMeasured`: every value tried was refused at construction, ...
```

## The same rule one layer in does build

Move the rule onto the collection's own type and the row appears:

```souther
data NonEmpty = List<Int> invariant List.length(value) >= 1
data Wrapped = { xs: NonEmpty }
```

```
//     | "bill = 0" : (Yen(0), Wrapped { xs = NonEmpty([0]) }) -> <?>
```

So the line is not whether a collection is empty. It is whether the rule sits on the position's own type or on the record that has the position.

## A number written the same way does build

Nor is it that a record's rules fail to reach its fields. A scalar field bounded by the record's own invariant is answered at the bound, at any depth:

```souther
data Scalar = { n: Int } invariant n >= 5
data Inner  = { n: Int } invariant n >= 5
data Outer  = { inner: Inner }
```

```
//     | "bill = 0" : (Yen(0), Scalar { n = 5 }) -> <?>
//     | "bill = 0" : (Yen(0), Outer { inner = Inner { n = 5 } }) -> <?>
```

What reaches the field is the atom of the field's value. `InvariantChecker.name` records one only where `Terms.atomOf` names one, which is null for a type that is not an affine scalar, so a `List` or `String` field gets no entry and `FieldDomains.at` answers null for it. The bound itself is not lost — seeding `Bag` leaves `lo = {@0: 1}` under the size atom with `atoms` empty. It is held under a name nothing can ask for by field.

Both ends of the missing connection already exist. `NumericMeasures.takenOf` names what counts a value of a type — `String.length`, `List.length`, `Set.size`, `Map.size`, and nothing for anything else. `Partitions.leastHeld` reads a floor off that measure, and `Witnesses.holding` builds a value of a given size from it. `leastHeld` takes a type, so that path is reachable only where the rule is written on the position's type, which is the newtype above.

## Expected

Where a record's rules put a lower bound on the measure `NumericMeasures.takenOf` names for a field's type, candidate synthesis for that field honours it — the same answer for `List.length`, `String.length`, `Set.size` and `Map.size`, since which of them applies is already read off the type.

A bound above one asks for that size: `List.length(xs) >= 2` composes `[0, 0]` and not `[0]`. A field of a field is answered like a field, as `FieldDomains.at` already answers a nested path.

The value built for the floor is offered before the one the carrier stands for, which is what `insideTheNewtype` already does — the search over a row's positions is budgeted, so that order is what keeps rows at positions the rule has nothing to do with.

## Not this

The row losing `bill = 0` is a consequence and not the defect. A row fills every position at once, so a position with no value it can build takes the rest with it; that follows from what a row is, and it goes away when the floor arrives.

Upper bounds on a measure, domains carried into a collection's elements, and the shape of row composition are all outside this.

So is a floor no value can meet. `data Impossible = { xs: Set<Bool> } invariant Set.size(xs) >= 3` is the opposite case — the floor reaches synthesis and no witness exists — and it should stay reported as it is rather than be mixed into this.

kawasima:
Closes #650.

A rule counting a field — `List.length(xs) >= 1` on the record that has `xs` — was read into the numeric domain and then had no way back to the field, so `--generate` offered the value that holds nothing, the construction refused it, and because a row fills every position at once the refusal took every other position's boundary row with it.

```
- // no row for `bill = 0` in `wrappedList`: every value tried was refused at construction, ...
+ //     | "bill = 0" : (Yen(0), Bag { xs = [0] }) -> <?>
```

## Where it was lost

The same rule written on the field's own type already built. So did a number bounded by the record's own invariant, at any depth — `data Scalar = { n: Int } invariant n >= 5` answers at `Scalar { n = 5 }`. What was missing was neither the record reading its fields nor collections in particular: the reading that names each field names the ones that are numbers, and a rule counting a list bounds the count rather than the list. The bound was in the domain the whole time, under an atom no reader could ask for.

`String.length(name) >= 3` fails identically, and so does a field of a field, so the line is not collections. It is whether the measure `NumericMeasures.takenOf` names for the position's type has a floor the position can be asked for.

## What this does

`Terms.takenAtomOf` names that atom, from the measure the type gives and the key the container has. It recognises rather than interns: naming a shape creates it, and a reader asking what was said about a field would otherwise be handed an atom nothing bounds. The key is a location — `demo.Bag.fields.0`, not `xs` — so this is the one place that puts the two halves together.

`seedFields` records it per field path, and `FieldDomains` resolves it to bounds while the reading that named it is still alive, since an atom means nothing afterwards. It comes out as `Held`, its own type: the numbers are the same numbers, and a bare `Bounds` gives a caller nothing to stop it reading how much a value holds as what the value is. A field is in `at` or in `heldAt`, never in both.

`Partitions.leastHeld` takes the higher of the type's floor and the record's — both are rules the construction has to satisfy, so a value clearing one and not the other is refused either way. A newtype and a record are no longer separate readings; they are one floor with two places it can come from. What an end comes to as a count is `Counts.leastFrom`, which both go through, so `> 3` cannot be four on one side and three on the other.

Both halves of the answer read that one floor:

- what a position is offered puts the value built for the floor ahead of the one the carrier stands for, which is the assembly `insideTheNewtype` already made. The order is not taste — the search over a row's positions is bounded.
- why nothing was built reads it too. `notBuilt` knowing only the type said "every value tried was refused" of a position whose values were never built, which is a sentence about the model where the fact is about this compiler. Its narrow form is replaced rather than left beside the new one.

The measure decides what a floor comes to, so a set of two is two elements no two of which are equal and a map of two is two entries no two of which share a key.

## What is not here

Upper bounds on a measure, element domains carried into a collection, and the shape of row composition. A row losing its other boundary rows follows from a row filling every position at once, and goes away when the floor arrives. A floor no value can meet — `Set<Bool>` with `Set.size >= 3` — is the other side and is left as it is.

A floor this cannot resolve to a number, such as one stated against another field, propagates nothing and behaves as before.

## Verification

`mvn clean test` green on every module, and each of the two commits builds and tests green on its own.

The new tests were held to failing first. One of them failed for the wrong reason to begin with — the refusing `CandidateCheck` refused every parameter, so the row reported the leading `Bool`'s reason rather than the position under test — and was aimed at the position before being trusted. The six added last were written after their code, so they were checked against the floor disabled: seven of nine went red, including the merge case dropping from three elements to two, and the two that stayed green are the controls that should not move.

Every layer has a control that does not move: no rule counting the position still offers the empty one, no floor still says "every value tried was refused", a rule bounding only `List.length(a) + List.length(b)` is a floor under neither, and a floor already on the type is not lowered by a smaller one on the record.

souther-examples is byte-identical before and after against this branch's merge base, so nothing was pushed out of the bounded search. It is also not evidence that the new path ran: the corpus has nine record-field measure rules and all of them are on result types, which `--generate` never composes. The reproduction and the unit tests carry that half.

### DIFF
--- a/souther-compiler/src/main/java/souther/compiler/check/FieldDomains.java
+++ b/souther-compiler/src/main/java/souther/compiler/check/FieldDomains.java
@@ -27,9 +27,12 @@ public final class FieldDomains {
 
     /** Nothing known of any field. */
     public static final FieldDomains NONE =
-            new FieldDomains(Map.of(), false, true, NumericDomain.top(), () -> true);
+            new FieldDomains(Map.of(), Map.of(), false, true, NumericDomain.top(), () -> true);
 
     private final Map<String, NumericDomain.Bounds> byField;
+    /** What each field has to hold, kept apart from what each field is. Same numbers, different
+     * question — see {@link Held}. */
+    private final Map<String, NumericDomain.Bounds> heldByField;
     private final boolean infeasible;
     /** Whether the reading that produced these bounds ran to the end. Kept because it is what that
      * reading knows about itself: a walk that fell over produces the same empty domain as a value
@@ -39,10 +42,12 @@ public final class FieldDomains {
     private final java.util.function.BooleanSupplier reading;
     private volatile Boolean everyRuleRead;
 
-    private FieldDomains(Map<String, NumericDomain.Bounds> byField, boolean infeasible,
+    private FieldDomains(Map<String, NumericDomain.Bounds> byField,
+                         Map<String, NumericDomain.Bounds> heldByField, boolean infeasible,
                          boolean seeded, NumericDomain numbers,
                          java.util.function.BooleanSupplier reading) {
         this.byField = byField;
+        this.heldByField = heldByField;
         this.infeasible = infeasible;
         this.seeded = seeded;
         this.numbers = numbers;
@@ -91,14 +96,59 @@ public final class FieldDomains {
                 out.put(field, bounds);
             }
         });
+        // Resolved here rather than handed over as atoms. An atom is a name the seeding gave a shape
+        // and means nothing once the reading that named it is gone, so a caller holding one could
+        // only ask the domain it came from — which is this one, while it is still here.
+        Map<String, NumericDomain.Bounds> holds = new LinkedHashMap<>();
+        seeded.held().forEach((field, atom) -> {
+            if (data.newtype()) {
+                return;
+            }
+            NumericDomain.Bounds bounds = seeded.numbers().boundsOf(atom);
+            if (!bounds.isEmpty()) {
+                holds.put(field, bounds);
+            }
+        });
         // Classifying the rules is a second reading of every one of them, and the bounds are the
         // whole of what a caller filling a row needs. Asked when the answer is, and not before.
-        return new FieldDomains(Map.copyOf(out), seeded.numbers().isBottom(),
+        return new FieldDomains(Map.copyOf(out), Map.copyOf(holds), seeded.numbers().isBottom(),
                 seeded.everyClauseRead(), seeded.numbers(),
                 // Classifying the rules is a second reading of every one of them. Asked when the
                 // answer is, and not before: a caller filling a row wants the bounds and nothing
                 // else.
                 () -> InvariantChecker.everyRuleRead(named, data, symbols));
+    }
+
+    /**
+     * How much a value at a position has to hold, which is not what the value there is.
+     *
+     * <p>Its own type because the numbers are the same numbers. {@code >= 2} at a field is a range of
+     * that field's values where the field is a number, and a count of what it holds where a rule
+     * counts it — and a caller handed a bare {@link NumericDomain.Bounds} has nothing to stop it
+     * reading one as the other. There is one such mistake in this already: the atom a size rule
+     * bounds is the size's, and writing it under the field is how a list would come to be told it
+     * must be at least 2.
+     */
+    public record Held(NumericDomain.Bounds bounds) {
+
+        public Held {
+            if (bounds == null) {
+                throw new IllegalArgumentException("a floor with no bounds is no floor");
+            }
+        }
+    }
+
+    /**
+     * What the rules say the value at {@code path} holds, or {@code null} where they count it in no
+     * way this read.
+     *
+     * <p>Read off the measure the position's own type names ({@link NumericMeasures#takenOf}), so a
+     * field this can answer for is one whose values are counted by something. A field of a number has
+     * no such measure and is answered by {@link #at} instead; the two never speak about one field.
+     */
+    public Held heldAt(String path) {
+        NumericDomain.Bounds bounds = heldByField.get(path);
+        return bounds == null ? null : new Held(bounds);
     }
 
     /**
--- a/souther-compiler/src/main/java/souther/compiler/check/InvariantChecker.java
+++ b/souther-compiler/src/main/java/souther/compiler/check/InvariantChecker.java
@@ -240,12 +240,17 @@ public final class InvariantChecker {
      * bounds that field on its own; and both land in one domain over the same atoms, which is what
      * lets a bound reach one field through another.
      *
+     * @param atoms the atom each field's own value is, for the fields that are numbers
+     * @param held the atom the count of each field is, for the fields whose values are counted by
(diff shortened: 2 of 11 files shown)


kawasima:
Fixed in 2b302a1b. You were right, and counting the readers turned up a third one.

**`heldBack`** now takes the same `settled` the candidates were chosen under. Confirmed the premise first: `FieldDomains.of(..., {n: 7})` on `List.length(xs) >= n` gives `heldAt("xs")` a floor of 7, so the two halves really were reading different rules.

**`candidatesAt`** was the third. Values are chosen twice over — the product pass with only what the caller settled, then a position at a time with each earlier choice asserted — and the second pass called `displacedRepresentativesOf`, which read `at` and not `heldAt`. A row that has to choose `n` before `xs` was offered the empty list however large the `n` it had just chosen. Reachable: with `Count` carrying a hole the domain cannot keep, the product pass offers a zero the model refuses, the row falls to the second pass, and `bill = 0` got no row at all. It builds now.

All three take their floor from one reading of the record's rules under this row's settled positions.

Two things worth flagging:

**My earlier "green" on the `heldBack` fix was an empty run.** I used `-Dtest='A+B'`, which is not a surefire separator, and with `failIfNoSpecifiedTests=false` it ran zero tests and reported success. The regression I had written for it was also mis-designed — no axis settled the sibling, so it never exercised the path. It is replaced by a boundary-based one where the edge is what makes the floor concrete, and that one goes red when `settled` is dropped from `heldBack`.

**One existing test changed fixture.** `whatTheSearchDidNotReachIsNotReportedAsRefused` used `String.length(a.value) > 1000` on a record, which with a floor being read now composes a thousand x's — a value that satisfies both patterns the field carries. The same rule on a newtype already composed that string before this PR, so that model was unbuildable only because of the asymmetry this fixes. Its width is unchanged; what refuses every value is now a pattern nothing derives a value from.

Verification: `mvn clean test` green on every module (4265 in the compiler). souther-examples byte-identical against the merge base, 19,543 lines. The `Set<Bool>, size >= 3` negative control is still not there — happy to add it if you want it in this PR rather than left as the documented other side.

PR #658, opened 2026-08-11
Build an expected value as what the row wrote, at every position

### LINKED ISSUE #653: An expected value is decoded into its position, so a row can assert a value it did not write
An expected value is decoded into the position, so a row can assert a value it did not write. A field written as the base of the newtype the position declares comes back as that newtype, and the comparison it is then held against is against a value the row never stated.

## What it does today

```souther
module demo

data AmountN = Int
data Receipt = { total: AmountN }

behavior receiptOf : (n: Int) -> Receipt
    constructs Receipt, AmountN
let receiptOf (n) = Receipt { total = AmountN(n) }

example receiptOf
    | (1) -> Receipt { total = 1 }
```

The row holds. `1` is decoded into an `AmountN` on the way in, so what is compared is `Receipt { total = AmountN(1) }` — which is right, and is not what the row wrote.

The same row on the input side is refused: an input states a value the model is given, so it is admitted before anything is decoded, and `Order { amount = 1 }` names no `AmountN` at the field that takes one. The expected side has no such reading, deliberately (see below), so nothing there notices.

At the top of an expected value it does not arise: `-> 1` at an `AmountN` output is not decoded and is reported as a mismatch (*E1905*), which says what the row wrote against what came out. It is only under a name — a field, an element, a map value — that the decode happens and the coercion with it.

## Why it is not the input rule

An input is a value the model is given, so there is nothing to disagree about: a fixture that states no value of its position is one the position could not have been given, and refusing it before the decode is the whole rule (#646).

An expected value is not that. A row may state what the behavior does not answer with, and reporting that disagreement is what the row is for. So an expected value is built and compared, and admitting it would refuse rows that are doing their job.

Holding it to its position also cannot be stated as the input rule is. A behavior's output may be an anonymous union, where `+Int | Rejected+` takes a value under no name and `+AmountN | Rejected+` does not — whether a nameless value belongs at such a position is a question about the value's type, which is the comparison's question rather than admission's.

## What the question is

Not "admit before decode" but: how is an expected value built as a value of its own type, and compared against what came out without either side being coerced into the other?

Some of what has to be settled:

* Which decode, if any, an expected value goes through. Today a row naming a case decodes through that case, which is what makes the two sides comparable — and what coerces.
* What a mismatch shows when the two sides are of different types. *E1905* renders both through the encoder, which assumes they are of one type.
* What a row that writes a base literal at a newtype field is told, and under which code. It is not *E1903* — the value builds — and calling it a mismatch has to say what differs, which is a name and not a value.

## Acceptance

* `+Receipt { total = 1 }+` expected of a behavior answering `+Receipt { total = AmountN(1) }+` does not hold, and what the row is told names the field and the two types.
* `+Receipt { total = AmountN(1) }+` holds, and what is compared is what the row wrote.
* A row that expects a case the behavior does not answer with still reports that disagreement rather than a fixture error.
* An anonymous union output takes an expected value of any of its cases, whether or not the case wears a name.

kawasima:
Closes #653.

An expected value was read through the decoder of the position it stood at. A derived decoder reads a base representation and answers with a value of its own type, so under a name — a field, an element, a map's value — the name came from the position rather than from the row:

```souther
data AmountN = Int
data Receipt = { total: AmountN }

example receiptOf
    | (1) -> Receipt { total = 1 }      -- held
```

`1` was decoded into an `AmountN` on the way in, so what was compared was `Receipt { total = AmountN(1) }`. The row asserted a value it had not stated. Two names over one base reached the decoder alike, so `Receipt { total = OtherAmountN(1) }` held as well, and a top-level collection output took `-> [1]` for the same reason.

## Where it came from

The name a row wrote was read in exactly two places, and both read only the outermost frame of the expectation: the case check that reports *E1904*, and the choice of which decoder to run. Everything under that went through the neutral form a decoder reads, in which a standalone newtype is its inner value — so `1`, `AmountN(1)` and `OtherAmountN(1)` were one form.

What had read a written name at every depth was admission, and an expected value is deliberately not held to it: a row may state what the behavior does not answer with, and reporting that disagreement is what a row is for. Carrying the name lived entirely inside admission, so switching admission off left nothing to distinguish the three.

## What changed

An expected value is built as the value the row wrote and compared against what came out, with neither side read as the other. The rules the build holds to are stated above `assertedExpected`:

1. A part is built at the type its construction requires of it, never at the position the construction stands in.
2. A finished node may be decoded as its own written type, never as the position it stands in.
3. What renders a value reads the names it wears; an encoder writes the representation a value crosses a boundary in, which is not that.

The second rule is what keeps `AmountN("x")` a fixture that cannot be built (*E1903*) and what runs a newtype's invariant. The position still resolves what the text does not say on its own — whether `[ ]` is a list, a set or a map — and supplies no name.

Rendering moves off the encoder. A newtype's representation is the base it wraps, so both sides of a mismatch were written alike and the difference the row was reported for could not be read in what it was shown:

```
This example does not hold.
It is:
    Accepted { code = Code("AB") }
But it needs to be:
    Accepted { code = "AB" }
At $.code the row states String, but the behavior answered with Code.
```

## Behaviour changes

A temporal expected as text is now the disagreement it is (*E1905*) rather than a fixture that could not be built (*E1903*). Holding an expected value to the form its position declares was the last place a position decided what a row may state, and no general rule was left for it to stand on. Three rows in `ATemporalFixtureIsNotWrittenAsTextTest` move; the input side is unchanged.

*E1904* stays the root preflight it is. A name the output does not list at all is reported as the case it is not, before anything is compared; a nominal disagreement under that root is *E1905*. The boundary is asserted rather than left to be discovered.

## How it was checked

The comparison was replaced before the build was, so the two halves could fail separately.

**Phase A** ran the new comparison beside `Values.equal` over the whole corpus with both sides still built as before, and recorded where the two disagreed: zero divergences, zero throws, build green. `ValueMatchTest` and `ValueRenderingTest` then fix what the corpus does not exercise — a set against a list at one pair of sequences, empty collections per position, decimal scale, map matching by key, and that two observations a limit stopped are not one value.

**Phase B** replaced the build. Before running it the expectation was one row: an `admitWritten`-based probe over the whole corpus found exactly one holding row relying on the coercion, `CompileAttemptedConstructionTest`'s `Accepted { code = "AB" }`, which is itself an instance of this defect. What changed was that row, the three temporal rows, and three assertions that were pinning the encoder's name-erasing rendering.

The old expected-value path is removed rather than left unused — `builtExpected`, `builtExpectedOrNull`, `describeExpected`, `ExampleVerifier.matches` — so there is no position-driven decode left to reach for.


kawasima:
Pushed `fb67de2e` for the four findings. All four reproduced first; the fifth did not, and why is below.

## 1 — `Set` / `Map` absorbed by the position

Confirmed, and it was not a patch. `Set.fromList([1])` and `Set.empty` both held at a `List<Int>` output.

The root was reusing `ObservedValue` for what a row wrote. That type leaves out which collection a sequence is *because the type behind an observation says it* — true of a value a behavior produced, and not true of a row's expectation, which stands at a position it is free to disagree with. So one `position` answered for both sides.

What a row wrote is now its own type, `Asserted`. It carries which collection the row said it was — `Set.fromList`, `Set.empty`, `Map.fromList`, `Map.empty` and a helper's answer say it; `[ … ]` does not and takes the answer's reading, which is the one thing the position still decides.

A helper's answer reads its container kind from the value itself rather than from its declared result type, which may be inferred.

## 2 — a newtype's own decoder

Confirmed, and the boundary was wrong rather than missing. It was keyed on whether the argument was a literal, so `Positive(-1)` was E1903 and `Positive(neg(1))` was E1905 — one written meaning, two outcomes.

It is now keyed on what the newtype wraps. A scalar is settled by the value standing there however it got there, so literal and helper answer are one case. Where it wraps something with parts, its decoder would read those parts at what the newtype declares, which is the reading this change exists to remove, so the parts stand as written.

## 3 — duplicate map keys

Confirmed. The check went with the decode and nothing took it over.

It is back in the assertion builder, and — as you note — it asks the keys **as the row wrote them**: two keys are one key under the same comparison everything else uses, so `1` and `AmountN(1)` are two keys. Reading them into the position's `K` first would rebuild the original defect.

## 4 — required fields

Confirmed. Every missing field was read as absence; that is the rule for an optional and only for one. A construction with no value for a required field now states no value of its type (E1903), and an omitted optional still reads as `None`.

## 5 — record invariants: not a regression

`CodecGen.invariantsOf` returns `Invariants.NONE` for anything that is not a newtype, so a record's invariant was never a decoder constraint. `Floored { n = 1 }` was E1905 before this branch too. Worth its own issue; not this one.

## What the four had in common

Your diagnosis holds: the nominal name was preserved and the rest of what a construction states of itself was not. That is now the third invariant above `assertedExpected`, and stated in the specification:

> What a construction states of itself is the form's and not the position's: which collection a row wrote, that a map's keys are distinct, that a record has a value for every field it has. The position answers only what the form leaves open.

`ValueMatchTest.aRowThatSaidNothingAboutWhichCollectionIsReadAtTheAnswers` and `aRowThatSaidWhichCollectionItWroteIsHeldToIt` now fix both directions, so the earlier test no longer justifies one position for both sides.

Full `mvn clean test` green: 4314 in the compiler module.

kawasima:
Correction to what I wrote about record invariants above.

I said they were never checked on either side, reading `CodecGen.invariantsOf` returning `Invariants.NONE` for anything that is not a newtype. That is true of the decoder's *constraints* and not of the rule: a record's invariant runs when the value is **constructed**, which the decode did on the way past. The input side still refuses `Span { from = 5, to = 1 }` with `invariant violated on demo.Span`, which is the evidence I should have taken.

So it is a regression from this PR, not a pre-existing gap, and the two sides of a row had come to answer differently about the same value. #660 restores it, through the same two stages a newtype goes through.

Issue #651, opened 2026-08-11
A boundary row is given up on where a companion position is a sum with no unit case, though the pair rows build one


kawasima:
Filling a boundary row means putting some value in every other position. Where one of those is a sum,
the value chosen is a unit case, and where the sum has none the row is given up on — in the same
report where the pair rows built a record case of that very sum.

## What happens

```souther
module demo

data Yen = Int invariant value >= 0
data Ok = { n: Int }

data Boxed = { x: Int }
data Labelled = { s: String }
data EveryCaseARecord = Boxed | Labelled

data Plain = { x: Int }
data OneCaseUnit = Plain | Bare

behavior recordCases : (bill: Yen, v: EveryCaseARecord) -> Ok
    constructs Ok
let recordCases (bill, v) = Ok { n = bill.value }

behavior unitCase : (bill: Yen, v: OneCaseUnit) -> Ok
    constructs Ok
let unitCase (bill, v) = Ok { n = bill.value }

example recordCases | (Yen(5), Boxed { x = 1 }) -> Ok { n = 5 }
example unitCase    | (Yen(5), Plain { x = 1 }) -> Ok { n = 5 }
```

```
$ souther examples r4.sou --generate --boundaries
// example recordCases
//     | "v=Labelled" : (Yen(0), Labelled { s = "x" }) -> <?>
//
// example unitCase
//     | "v=Bare x bill = 0" : (Yen(0), Bare) -> <?>
// no row for `v: EveryCaseARecord` in `recordCases`: no value can be written there
```

`Labelled { s = "x" }` is a value of `EveryCaseARecord` written two lines above the sentence saying
none can be written. The two lines come from two strategies: the class-coverage one builds a case,
and the boundary one wants a representative and takes only a unit case. `unitCase` shows the second
strategy succeeding, and succeeding by way of `Bare`.

The message is also about the wrong thing. Nothing is uncomposable about `v`; what could not be
composed is the row, and it could not because the representative for `v` was looked for in a smaller
set than the one the other strategy draws from.

## Where it was found

Dogfooding a mobile-order model through the MCP. `Fulfilment = Delivery | Takeout | TableOrder` has
three record cases and no unit one, which is ordinary — each way of fulfilling an order carries
something. Every boundary on every other parameter of a behavior taking a `Fulfilment` went
ungenerated:

```
  kitchenInstructionAt     boundary 0/2
      · no row is at kitchenInstructionAt/cook = 1 (invariant CookMinutes (min))
[...]

PR #659, opened 2026-08-11
Decide a line by what a clause reaches, not by where it is written

### LINKED ISSUE #649: A bound a record states about one of its own fields draws no line, while the same bound on a newtype does
A bound written on a newtype draws a line the boundary measure asks for a row at. The same bound
written by a record about one of its own fields draws nothing, so the rows carrying it earn no
credit and `--generate` never offers one.

## What happens

```souther
module demo

data Name  = String            invariant String.length(value) >= 1
data Bag   = { xs: List<Int> } invariant List.length(xs) >= 1
data Tally = { n: Int }        invariant n >= 1

data Ok = { size: Int }

behavior onName : (v: Name) -> Ok
    constructs Ok
let onName (v) = Ok { size = String.length(v.value) }

behavior onBag : (v: Bag) -> Ok
    constructs Ok
let onBag (v) = Ok { size = List.length(v.xs) }

behavior onTally : (v: Tally) -> Ok
    constructs Ok
let onTally (v) = Ok { size = v.n }

example onName  | (Name("a"))        -> Ok { size = 1 }
example onBag   | (Bag { xs = [1] }) -> Ok { size = 1 }
example onTally | (Tally { n = 1 })  -> Ok { size = 1 }
```

```
  onName                   implemented   rows 1    pending 0
    boundary    1/1
  onBag                    implemented   rows 1    pending 0
  onTally                  implemented   rows 1    pending 0
      · not derivable: v.n
```

The three rows are the same row: each is at the smallest value its type admits. One of them is
counted.

## The same rule, spelled one type-hop away, is measured

```souther
data Cart = List<Int> invariant List.length(value) >= 1

behavior onCart : (c: Cart) -> Ok
    constructs Ok
let onCart (c) = Ok { size = List.length(c.value) }

example onCart | (Cart([1])) -> Ok { size = 1 }
```

```
  onCart                   implemented   rows 1    pending 0
    boundary    1/1
adequacy: satisfied
```

So "this list is non-empty" is a measured rule or an unmeasured one depending on whether the
aggregate holding it has a second field. An aggregate that does — an order with lines and a
customer, a store payment with splits and a till — has to state the rule about its own field, and
that is the spelling that disappears.

## What the specification says about it

`example-partition` covers the case and gives a reason that does not reach it:

> A record's rule takes an edge in and does not put one there. A position its own type leaves open
> is one the model draws no line through, whatever a rule relates it to, because a rule between two
> fields is not a partition of one of them.

The reason is about a rule *relating two fields*, which is the `startsAt < endsAt` shape #427 is
about, and it is a good reason there: neither position is divided. `List.length(xs) >= 1` and
`n >= 1` relate one field to a constant. They are the same shape as an invariant on a newtype, which
the paragraph above them says draws a line at 1.

## What would settle it

Either read a clause a record states about a single one of its fields the way the same clause on a
newtype is read, or say in `example-partition` why a one-sided bound is different depending on where
it is written. As it stands the measure rewards a modelling choice — wrap every constrained
collection in a newtype of its own — that has nothing to do with the rule being stated.

Found dogfooding a mobile-order model through the MCP: `data Cart = { lines: List<OrderLine> }
invariant List.length(lines) >= 1` and `data PayAtStore = { splits: List<StoreSplit> } invariant
List.length(splits) >= 1` are both invisible to the boundary measure.

kawasima:
Closes #649.

A bound was read off the newtype chain of the position's own type and off nothing else, so the same
rule was measured or unmeasured according to where the author had room to write it.

```
data Cart = List<Int>              invariant List.length(value) >= 1   boundary 1/1
data Bag  = { xs: List<Int> }      invariant List.length(xs) >= 1      boundary not measured
data Tally = { n: Int }            invariant n >= 1                    boundary not measured
```

An aggregate with a second field — an order with lines and a customer, a store payment with splits
and a till — has no choice but the second spelling. The measure rewarded wrapping every constrained
field in a newtype of its own, which is a modelling choice and says nothing about the rule stated.

## The rule this replaces it with

ADR-0090 said *a record's rule takes an edge in and does not put one there*. The reason given was
`startsAt < endsAt`, which is a good reason there and does not reach `n >= 1`. It also did not match
the implementation: `{ n: Count } invariant n.value >= 5` already moved the line to 5, so the record
was participating in where the edge sits while being said not to place one.

So the subject moves from the declaration to the clause:

> A clause that bounds one term against a constant without relying on another term may place an edge
> there. A clause relating terms may take an existing edge in and may not place one.

Both halves matter, and the second is what the older wording was protecting. `{ a: Int, b: Int }
invariant a < b invariant b <= 10` leaves `a` running to 9 and draws no line through it: that 9 is
where `b` stops and not something any clause said about `a`. Placing an edge and taking one in stay
separate acts, and only the first may be a line.

## How a clause is read

`own` — which is what says a line exists and who it is named by — is now every coordinate-local
bound reaching the position, from the newtype chain of its type, from the record it is a field of,
and from the declarations under that record it sits inside. `projected` still only narrows.

Which coordinate a clause is about is the naming the discharge check already made, so `n`,
`n.value` over a newtype, and `List.length(xs)` are each the one coordinate they name, and an import
writing a size call without its qualifier is read like any other. Recognising them by their spelling
would put a second expression grammar beside `Terms`, and the two would answer differently the first
time they disagreed.

The alternative considered was seeding one clause at a time and reading what it bounded. It does not
separate locality from propagation: with `data B = Int invariant value <= 10` and `{ a: Int, b: B }
invariant a < b.value`, seeding that clause alone still has `B`'s maximum under it, so `a <= 9` falls
out of the clause's own derivation and reads as an edge the clause placed. That case is held green
in the tests before anything else was written.

Two consequences of the same rule. A line is named by the declaration the clause is written on, so
the `Moved` case above reports `invariant Moved (min)` where it used to say `invariant Count (min)
within Moved` — a narrowed origin is now only for a clause relating terms, which is what it was for.
And a cut no longer requires the position to be a `Type.Ref`: that condition happened to hold
whenever an end existed, and a bare `Int` field and the length of a bare `List<Int>` have ends now.

## What a clause reaches, and from where

A clause reaches a position from the position's own type and the names round it, the record it is a
field of, the declarations under that record, and the names wrapped round any of them. That last one
is not a list entry so much as a consequence: a name worn is not a step of the coordinate's path.
`Location.isStep` has always said so and `Terms` reads by it, so the atom of `w.value.n` *is* the
atom of `w.n` — and the one reader that counted `value` as a step filed positions under a wrapper
where nothing asks for them, or stopped at the name and never reached them.

Naming it the way the rest of the check does means a wrapper at a parameter, a wrapper on a field,
and a stack of names are not separate cases. It also means a wrapper is read as one governing
declaration rather than as an extra source of ends, so all three of the things a declaration does
arrive together: it places ends, it projects ranges onto the positions under it, and it can hold a
clause this could not read. `data Wrapped = Base invariant value.a < value.b` places no edge and
still leaves `a` one below where `b` stops.

A narrowed edge is named by the declaration whose clause moved it, which need not be the value the
position sits in. Read off the value, a `Held` holding such a `Wrapped` reported the edge as taken in
by `Held` — a declaration with no clause about the pair — and the name is part of what tells one line
from another, not only what is printed.

## The writability proof

The `Set<Bool>` control came back `ProvenByProjection`: the numeric domain knows 3 is a number the
rules leave and has no term for there being two booleans. Reachable only through the route this adds
— the newtype spelling and a record holding such a newtype both answer `Unknown` — so it is a row an
author would be told to write and could not, introduced here.

Declining the projection at every count fixes it and the whole suite stays green. It is also wrong:
over `souther-examples` it takes twenty `String.length` edges out of the denominator, owed falling
from 75 to 55. So does declining it wherever the count is of distinct things, which is the other
place the line suggests itself — "a list repeats an element" is an answer only once there is an
element, and a `List<Loop>` of one needs a `Loop` that something inhabits.

What a range cannot answer is whether a value of that count exists. A string's length is the one
count every value has, since a string of any length is a character repeated, so
`NumericMeasures.everyCountHasAValue` is true there and nowhere else. The reason the line is drawn
there rather than at every count is measured rather than argued, and recorded in the ADR, since the
suite cannot hold it.

## Verification

`mvn clean test` green on every module, 4340 tests. The eleven cases of the matrix were written and
held failing first, seven red and four green, and the four are the controls that must not move:
`startsAt < endsAt` places nothing, `n >= min` places nothing, `a < b.value` under a bounded `B`
gives `b` its line and `a` none, and the same rules on newtypes are measured as before. Review added
the rest: a wrapper at a parameter, on a field and stacked; a wrapper's relation narrowing without
placing, beside the bare record where the same edge does not move; the coordinate a rule from outside
may not choose; and a length over an element type nothing inhabits.

Over `souther-examples`, four positions gain a line and boundaries owed go from 75 to 78. All four
are one declaration — a CRM lead stating `touches >= 1` beside a rule relating two of its dates,
which is the shape the issue was found on. No adequacy verdict moved and nothing that was measured
stopped being. The corpus has few record-stated bounds on input types, so it is evidence that nothing
was lost rather than evidence of how common the shape is.

One position left the report: a behavior that gained an axis pushed another past the axis limit,
reported as omitted rather than dropped in silence.

An existing test's model caps a field on its record, so its report gains the row that cap is owed.
The row is written into that test rather than the shared model, since the arm cases below it move the
guard to where a `b` of 10 would reach the arm they are about.

## Not here

What `--generate` says about a floor no value can meet stays as #651 left it: the generator makes no
judgement that a value cannot exist, and reports `NOTHING_COMPOSES_ONE`.

A position still has one coordinate, and which one is settled by which the model wrote about.
Reading the rules from more declarations does not change that selection. Where a position turns out
to have rules about both its value and its length, nothing here chooses between them.


kawasima:
All three reproduce, and all three are fixed in f411a867. Taking them in the order they bite.

## [P1] A name wrapped round a record

Confirmed, and the diagnosis is right on both counts: `fieldDomainsOf` reads `recordIn(type)`, and
`seedFields` was returning no directs for a newtype. `NonEmptyBag` and `data Wrapped = Base invariant
value.n >= 1` both came back `boundary not measured`.

Fixed as you suggested, by putting it through the model `seedAt` already uses rather than by lifting
the `data.newtype()` guard. The clauses of each layer of the parameter's newtype chain are read, and
the path they are filed under has the `value` they are written through taken off — a newtype's value
is the same location as the newtype, so `value.xs` is the `xs` the walk names. An end that strips to
nothing is the newtype's own bound, already read where its value is a position, and is dropped rather
than owed twice.

They are read beside the record's ranges rather than through them. A wrapper places ends and leaves
no ranges — `at()` of a newtype is empty by design — so merging the two into one `FieldDomains` would
mean choosing a name to file the answer under. `Placed` now carries both and answers with both.

```
onWrapped/v.n = 1                 invariant Wrapped (min)
onNonEmpty/List.length(v.xs) = 1  invariant NonEmptyBag (min)
```

## [P1] Coordinate selection

Confirmed, and this one is a regression rather than a gap. Your `Person` case exactly:

```
before this PR:  no row is at onPerson/v.name = m (invariant Name (min))
with this PR:    boundary 1/1          <- the length edge; the `m` line is gone
```

An axis the report had been printing, taken away by a change meant to add lines. Your reading of the
cause is right — the record's size direct was feeding `bySize`.

Fixed the way you set out. The position's own type answers first and its answer stands; where it says
nothing, one such rule may settle it. Where they arrive about both coordinates nothing chooses, and
the position is left as one nothing divides:

```
onPerson   no row is at onPerson/v.name = m (invariant Name (min))   <- unchanged from before
onR        · not derivable: v.s
```

Worth saying plainly: the ADR text in this PR already claimed this ("Reading those rules from more
declarations does not change that selection"). The code did not do it. The declared semantics was
right and untested, which is the worse of the two ways to be wrong.

## [P2] The writability proof

Confirmed, including the example. `data Loop = { next: Loop }` under `List.length(xs) >= 1`:

```
onBag/List.length(v.xs) = 1  writability=ProvenByProjection  attempt=NotAttempted[NO_CLASSES]
```

Proven writable with nothing even attempted. And your objection to the axis is right: distinctness is
only one way for the count not to be reachable. "A list repeats an element" is an answer only once
there is an element.

So the predicate is now `NumericMeasures.everyCountHasAValue`, true for `String.length` and nothing
else, which is your recommendation. A string of any length is a character repeated and a character is
always to be had; every other measure counts things that may not be there.

The reason the line is drawn there rather than at every count is measured rather than argued: over
`souther-examples`, declining the proof at every count takes twenty `String.length` edges out of the
denominator and the whole suite stays green while it happens. That is recorded in the ADR, since the
suite cannot hold it.

## What the corpus says now

Unchanged by all three fixes — boundaries owed 75 → 78 over `souther-examples`, the same four
positions on the same CRM declaration, no adequacy verdict moved. So nothing in the corpus was
relying on the proof at a `List.length` or `Set.size` edge.

## On the tests

The two new models are held in modules of their own rather than added to the matrix. One unreadable
row makes its whole module partial, and a declaration nothing can build leaves the module with
nothing to build against — either would have read as the matrix failing. Both are asserted on the
assessments rather than on report text, which is also what lets them do without rows.

`mvn clean test` green on every module, 4293 tests.

kawasima:
Both are real, both reproduce, and they are one cause. Fixed at that cause in 659fcf16 rather than by
widening the lifting — the lifting is gone.

## The cause

`InvariantChecker.name` counted a newtype's `value` as a step of the path and stopped at a newtype.
`Location.isStep` — which the rest of this reads by, and which `Terms.pathKey` already honours — says
a name worn is not a step:

```java
return !(field.equals("value") && TypeOps.isSingleValueNewtype(readFrom, symbols));
```

So the atom of `w.value.n` *is* the atom of `w.n`, and one reader was naming positions by a different
rule from the one that names atoms. Everything you found follows: positions under a wrapper were
filed one step deeper than anything asks for, or were never reached because the walk stopped at the
name.

Naming it the way the rest of the check does dissolves both findings at once. A wrapper at a
parameter, a wrapper on a field, and a stack of names stop being separate cases — there is nothing to
special-case once the naming agrees:

```
onHeld/v.w.n = 1                    invariant Wrapped (min)      # Held { w: Wrapped }
onHeldBag/List.length(v.b.xs) = 1   invariant NonEmptyBag (min)  # HeldBag { b: NonEmptyBag }
onStacked/v.w.n = 2                 invariant W2 (min)           # W2 = W1 = Base
```

`placedByWrappers` and the `wrapped` list on `Placed` are removed. `fieldDomainsOf` reads the written
name, and everything a wrapper contributes arrives through the one `FieldDomains` the record's own
rules arrive through.

## Which is what fixes the second one

You were right that lifting ends could never have covered it, and the reason is that a wrapper does
all three things a governing declaration does. Only one was arriving. Reading it as one declaration
brings the other two with it — the projection and the reading completeness — for free rather than as
two more lifts.

The control is the interval shape written one name up:

```souther
data A = Int invariant value <= 10
data B = Int invariant value <= 10
data Base = { a: A, b: B }
data Wrapped = Base invariant value.a.value < value.b.value
```

```
onBare/v.a = 10     invariant A (max)
onWrapped/v.a = 9   invariant A (max) within Wrapped
```

The bare record beside it is there on purpose: an edge that does not move proves nothing unless the
same edge moves elsewhere.

Two notes from writing it. My first attempt used a `guard` past the narrowed range, and it was an
empty probe — a behavior with no rows has no body read, so no guard line is drawn either way and the
assertion passed on nothing. And `data W2 = W1 invariant value.n >= 2` does not reach `n`: `value` is
a `W1`, and this language does not deref through a name, so it is `value.value.n`. The clause was
silently unread, which is its own small lesson about how these are written.

## `within` now names the value that was read

Falling out of the same change: which declaration's rules reach the positions, and which declaration
is said to have taken an edge in, are answers about the same value. Read apart, an edge a wrapper
narrowed was reported as narrowed by the record under it — a declaration that may hold no clause
about the pair at all. One name for both questions now.

## Verification

`mvn clean test` green on every module, 4339 tests. Over `souther-examples` the corpus is unchanged
by this commit — owed still 75 → 78, the same four positions, no adequacy verdict moved. Nothing in
the corpus wraps a record it states rules about, so what holds this is the tests rather than the
corpus, and I would rather say that than quote a number that moved for other reasons.

## What this does not do

A position still carries one coordinate. `#650`'s multi-coordinate shape is not here, and where rules
arrive about both of a position's coordinates from outside, nothing chooses — the position is left as
one nothing divides.

kawasima:
Reproduced, and fixed at the cause rather than at the name. 02765e76.

## [P1] `within` named the value, not the clause

```
onHeld/v.w.a = 9   invariant A (max) within Held
```

`Held` holds no clause about the pair. Your reading of why is right: `Placed.value` is one name for a
whole parameter and was standing in for provenance, and it cannot be that.

I did not move `Placed.value` down the walk, for the reason you gave. What changed is that the
reading which already walks every clause with the declaration it is written on now records, per
coordinate, which declarations relate it to something else. Whether a comparison places an end or
relates a coordinate is one question, so it is asked once in the same pass — a separate walk would
have to agree with this one about which comparisons it had already accounted for.

```
onWrapped/v.a = 9   invariant A (max) within Wrapped
onHeld/v.w.a = 9    invariant A (max) within Wrapped
```

Writing that turned up a second thing. Deciding "places an end" by whether the other side names a
coordinate is wrong: what a clause compares a coordinate to may be a position deeper than the naming
reaches — `invariant ordered = n < l1.l2.leaf.x` is one, and it is already in the suite — or
arithmetic over several. Neither is a number an edge can be put at. So the end is attempted and what
fails to be one may have moved one. The deep case went red on the first version and is what found it.

On which of several clauses moved it: not asked. A bound arrives along a path through the differences
and several can be on it. What is known is that some declaration wrote a relation about this
coordinate, and naming one that did is the whole of it — that is in the ADR rather than left to be
inferred from the code.

## [P2] The second spine walk

Gone. `TypeOps.newtypeChain` drives it, cycle detection and all, and the local `HashSet + while +
fieldTypes(...).get("value")` is removed. You are right that it was the same shape as the defect the
commit before it fixed — two readings of "how far a name reaches", one of which could change alone.

## PR body

Rewritten. It carried `countsDistinct`, "a list repeats an element", and 4288 tests, all of which
this branch has since rejected or moved past. It now says what the code says, including why the line
is drawn at `String.length` and not at distinctness.

## Verification

`mvn clean test` green on every module, 4340 tests. Over `souther-examples` nothing moved with either
of these: owed still 75 → 78, the same four positions on the same CRM declaration, no adequacy
verdict changed. Neither shape occurs there, so what holds them is the tests.

The new control is the one you asked for, and it is the nested spelling rather than the root one:

```souther
data Wrapped = Base invariant value.a.value < value.b.value
data Held    = { w: Wrapped }
```

with the bare `Base` beside it, since an edge that does not move proves nothing unless the same edge
moves elsewhere.

kawasima:
Both reproduce, and both are fixed in 583ff467.

```
before   onOuter/v.a = 7   invariant A (max) within Outer
         onBoth/v.a  = 3   invariant N (min) within Both
         onBoth/v.a  = 7   invariant N (max) within Both

after    onOuter/v.a = 7   invariant A (max) within Inner
         onBoth/v.n  = 3   invariant N (min) within Both
         onBoth/v.n  = 7   invariant N (max) within Upper
```

## On your question about `NumericDomain`

Instrumenting the propagation is the root of *exactness*, and it is not the root of *this defect*.
The defect is that `get(0)` claimed to know which declaration decided the edge. Not claiming it is a
complete fix for the wrongness; tracking derivations in the domain is a capability that would let us
say more than "one of these". Only the first is a defect, and it is the one #649's own principle is
about — read what the model says, do not invent.

But there is a third option between the two, and it is what I took: the domain does not have to
*record* the derivation for us to *ask* it. Take a candidate declaration's clauses away, seed again,
and see whether the end moves. That is the same reader answering a second question, not a second
reader re-deriving the first answer — which is the distinction the last two rounds of this review
were about.

It resolves both of your cases exactly. `Outer`'s clause reaches nothing `a` had not already passed,
so removing it moves nothing and it names nothing. `Upper` holds the maximum and `Both` the minimum,
and asking per end says so.

What it does not resolve is over-determination: two clauses each saying what the edge says, where
removing either alone changes nothing. There the answer is the set, handed over together, and
`NarrowedOrigin.within` is now a list for exactly that reason — including in `Line`, so identity does
not turn on a choice nobody made. That is the honest limit of asking from outside, and the ADR says
so rather than leaving it to be inferred.

So I would not put provenance into `NumericDomain` for this. It is the hottest shared structure in
the compiler, it is read by the discharge check and the generator as well as by this, and what it
would buy over the above is resolving the over-determined case — which changes no count, no verdict
and no edge value, only whether a name is one or two. If something later needs the exact derivation —
explaining *why* a value was refused, say — that is the issue to open, and this case is its
acceptance test.

## Cost

The re-reading runs only where an end actually moved and there is a candidate, so `souther-examples`
is byte-identical to the previous commit: owed still 75 → 78, the same four positions.

## Controls

Your "tight Inner + redundant Outer" is in, as you wrote it, plus the two-ends model. Both assert the
origin string rather than the count, since the count is what does not move when this is wrong.

`mvn clean test` green on every module, 4342 tests.

kawasima:
Reproduced with your model, fixed in 08179c8b.

```
before   onW3/v.x = 7   invariant X (max) within W3 or W2 or W1
after    onW3/v.x = 7   invariant X (max) within W1 or W2
```

You are right that "nobody is essential" does not make every candidate equally explanatory, and that
this was the same defect coming back through the branch where drop-one finds nothing. Adding `W3`
still changed a line's identity while the domain, the edge and the clauses holding it stayed where
they were.

So where nobody is essential, each candidate is now asked what it does as the only one left. One that
moves the end nowhere is out whatever the rest are doing. Same seeding, same `without` predicate,
one more question — no new reader.

I stopped at that boundary rather than going further, and said so in the ADR instead of leaving it to
be read off the code. What is left is clauses that reach an end only together: taking those apart is
a search for the smallest sufficient subsets, it is combinatorial, and which constraints a bound came
from is the closure's to record and it does not. The set is the answer there. Your distinction is the
one I tried to write down — "we cannot tell, so here is the set" is honest, "everything we collected
is a member" is not, and those are different claims.

One thing your case made me fix that you did not raise: the names are now sorted. Several of them are
one answer and the answer is part of what tells one line from another, so an order read off the walk
that collected them would have made two readings of one edge into two lines. Your `W3 or W2 or W1`
output is what showed me the order was walk-derived.

## Verification

`mvn clean test` green on every module, 4343 tests. `souther-examples` is byte-identical to the
previous commit — the extra question only runs where no candidate is essential, which does not happen
there.

kawasima:
Reproduced with your model, fixed in 08179c8b.

```
before   onW3/v.x = 7   invariant X (max) within W3 or W2 or W1
after    onW3/v.x = 7   invariant X (max) within W1 or W2
```

You are right that "nobody is essential" does not make every candidate equally explanatory, and that
this was the same defect coming back through the branch where drop-one finds nothing. Adding `W3`
still changed a line's identity while the domain, the edge and the clauses holding it stayed where
they were.

So where nobody is essential, each candidate is now asked what it does as the only one left. One that
moves the end nowhere is out whatever the rest are doing. Same seeding, same `without` predicate,
one more question — no new reader.

I stopped at that boundary rather than going further, and said so in the ADR instead of leaving it to
be read off the code. What is left is clauses that reach an end only together: taking those apart is
a search for the smallest sufficient subsets, it is combinatorial, and which constraints a bound came
from is the closure's to record and it does not. The set is the answer there. Your distinction is the
one I tried to write down — "we cannot tell, so here is the set" is honest, "everything we collected
is a member" is not, and those are different claims.

One thing your case made me fix that you did not raise: the names are now sorted. Several of them are
one answer and the answer is part of what tells one line from another, so an order read off the walk
that collected them would have made two readings of one edge into two lines. Your `W3 or W2 or W1`
output is what showed me the order was walk-derived.

## Verification

`mvn clean test` green on every module, 4343 tests. `souther-examples` is byte-identical to the
previous commit — the extra question only runs where no candidate is essential, which does not happen
there.

PR #660, opened 2026-08-11
Read a construction's own invariant of the value the row wrote

kawasima:
A regression from #658, found while scoping what I had reported there as a pre-existing gap. It is not one — the two sides of a row had come to answer differently about the same value.

## What it does

```souther
data Span = { from: Int, to: Int }
    invariant from <= to
```

| where the value stands | before #658 | after #658 |
|---|---|---|
| an input | E1903 `invariant violated on demo.Span` | E1903 (unchanged) |
| an expected value | E1903 | **E1905, the row is reported as disagreeing** |

`Span { from = 5, to = 1 }` is a value that cannot exist. Expected of a behavior, it was reported as a value the behavior did not answer with.

## Why it moved

A record's invariant is not among the constraints its derived decoder carries — `CodecGen.invariantsOf` returns `Invariants.NONE` for anything that is not a newtype. It is run when the value is *constructed*, which the decode used to do on the way past. #658 stopped putting a record through its decoder, so nothing constructed one and nothing read the rule.

Reading `invariantsOf` and not the construction is how I came to report this as never having been checked. The input side, which still refuses it, is the evidence that it was.

## What changed

The same two stages a newtype goes through, for the same reason.

1. Whether every field states what this type declares it to be is asked of what the row wrote.
2. Only where it does is there a value of this type for the invariant to be about — and only there can constructing one read nothing as something else.

A row whose field states another type has not failed an invariant. `Receipt { total = 1 }` at a field declaring an `AmountN` is the disagreement it reports (E1905), and asking a rule about the value it would have been coerced into would answer for a value nobody wrote.

## Checked

`AConstructionMeansWhatItsFormMeansTest.aRecordReadsItsOwnInvariantToo` fixes both directions. Full `mvn clean test` green: 4320 in the compiler module.

### DIFF
--- a/souther-compiler/src/main/java/souther/compiler/FixtureReader.java
+++ b/souther-compiler/src/main/java/souther/compiler/FixtureReader.java
@@ -728,7 +728,7 @@ public final class FixtureReader {
      *  is why it is not asked of a decoder: one reads a whole number where a `Decimal` stands, and a
      *  row writing `1` there wrote an `Int`. */
     private static boolean spells(ObservedValue v, TypeName name) {
-        return name.name().equals(ValueRendering.primitiveNamed(v));
+        return name.primitiveKind() != null && name.primitiveKind() == ValueRendering.primitiveOf(v);
     }
 
     private boolean each(List<Asserted> elements, Type element) {
--- a/souther-compiler/src/main/java/souther/compiler/FixtureReader.java
+++ b/souther-compiler/src/main/java/souther/compiler/FixtureReader.java
@@ -522,7 +522,28 @@ public final class FixtureReader {
             }
             fields.put(f.getKey(), new Asserted.Value(new ObservedValue.Absent()));
         }
-        return new Asserted.Built(built, fields);
+        Asserted.Built whole = new Asserted.Built(built, fields);
+        admitsItself(whole, nd, built);
+        return whole;
+    }
+
+    /**
+     * A construction's own invariant, over the value it states.
+     *
+     * <p>The same two stages a newtype goes through, for the same reason. Whether every field states
+     * what this type declares it to be is asked of what the row wrote; only where it does is there a
+     * value of this type to ask the invariant about, and only there can building one read nothing as
+     * something else. A row whose field states another type is not a value whose invariant failed —
+     * it is the disagreement it reports, and asking a rule about a value nobody wrote would answer
+     * for a value nobody wrote.
+     */
+    private void admitsItself(Asserted.Built whole, Ast.NewData nd, TypeName built) {
+        if (TypeOps.effectiveInvariants(declared(built), symbols).isEmpty()
+                || !states(whole, Type.ref(built))) {
+            return;
+        }
+        decode(FixtureShape.of(Type.ref(built), symbols),
+                neutral.shaped(raw(nd, Type.ref(built), Admission.HELD), Type.ref(built)));
     }
 
     private Asserted assertedApply(Ast.Apply c, Type position) {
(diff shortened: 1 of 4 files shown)


kawasima:
Pushed \`4610374b\`. Reproduced first, and the root is one notch behind where you found it.

## What it did

\`\`\`souther
data Priced = { at: Decimal }
    invariant at > 0m
\`\`\`

\`Priced { at = -1 }\` expected of a behavior answering \`Priced { at = 1m }\` was **E1903**, reported for the rule \`at > 0m\`. The row wrote an \`Int\` at a \`Decimal\` field; it should be the disagreement it reports.

Exactly what the PR body says it does not do — reached through the check that was meant to prevent it.

## The root

\`spells()\` asked *what a row wrote* and answered it with *what a decoder takes*. Raoh's \`decimal()\` reads any \`Number\`, because that is how a boundary carries an amount. A row writing \`1\` there wrote an \`Int\` — the language makes the difference written, \`1m\` being the other spelling.

The same leniency stood over the temporals: one written form answered for all four, so a row could write \`Date("2026-07-20")\` where a \`DateTime\` stands and pass.

This is the same shape as everything else on this branch. A rule about what a row wrote was read off a mechanism that exists to read something else.

## What changed

Which primitive a written value with no parts is of now has one answer, and the three readers that wanted it share it:

- whether a row wrote a value of a position's type (`states`)
- whether two values are of one type (`ValueMatch.leaf`)
- what to call them where they are not (`typeShown`)

Each was working it out from what it had, and one of them had a decoder. `ValueMatch.leaf` now compares that answer first and reports `TYPE` when it differs, so a `Date` against a `DateTime` is a difference of type rather than of contents — it was `VALUE` before, while the message said `Date` and `DateTime`.

## Tests

`anInvariantIsNotReadOfAValueTheRowDidNotWrite` fixes both halves of the rule: `{ at = -1 }` and `{ at = 1 }` are E1905, `{ at = 1m }` holds. `oneTemporalIsNotAnother` fixes the temporal case.

Full `mvn clean test` green: 4322 in the compiler module.

kawasima:
Pushed \`6f6845ef\`. CI green.

## The test now proves what the commit claims

You were right that it did not. `Stamped` carried no invariant, so `admitsItself` short-circuited before `states()` and the gate under test was never reached — and a `Date` at a `DateTime` field was a mismatch under the old reading too, by its text rather than by its type.

`Stamped` now carries `invariant from <= to`. Against the reading this change removed, both rows are red:

```
anInvariantIsNotReadOfAValueTheRowDidNotWrite → E1903  invariant violated on demo.Priced
oneTemporalIsNotAnother                       → E1903  from: expected date-time
```

Ran as a control by putting the lenient `spells()` back, so the two are detectors and not restatements.

The temporal row also fixes the `TheTwoAreOfDifferentTypes` note carrying `Date` / `DateTime` at `$.from`, so the difference is reported as one of type rather than of contents.

## `String` → `Type.Prim`

Taken. `Type.Prim.shown()` is the one table a primitive is spelled from and `TypeName.primitiveKind()` reads it back; answering with a string beside it was a second table free to drift from the first. The three readers now share the primitive itself:

```java
name.primitiveKind() == primitiveOf(v)   // states
primitiveOf(v) == primitiveOf(o)         // ValueMatch
primitiveOf(v).shown()                   // rendering
```

## `ObservedValue.Temporal` keeping its kind

Agreed that this is where the model closes, and I have left it out of this PR. It is not a correctness bug today — the four canonical forms are mutually distinguishable, which is what the recovery reads — and the record is public IR with construction sites in `ObservedValues`, `Carrier` and `GuardThresholds`, none of which this branch is about. It wants its own change.

Issue #661, opened 2026-08-11
E1013 decides a cycle, not whether a type has a value


kawasima:
`checkNoUninhabitableCycle` walks for a recursion with nowhere to stop. That is
the part of "has this type a value" that a walk from one declaration can answer.
Three shapes have no value and compile, and each of them needs an answer about a
whole type where the walk has only a position.

## What is admitted

Held today as passing tests in
`souther-compiler/src/test/java/souther/compiler/check/WhatTheConstructionCheckDoesNotDecideTest.java`.

A sum whose every case holds the sum:

```souther
data Leaf = { s: Shape }
data Branch = { s: Shape }
data Shape = Leaf | Branch
```

The walk stops at a sum because a cycle through one may bottom out in a case the
cycle did not come through. Where no case bottoms out there is none. The same
holds with a record between the sum and the cycle:

```souther
data Leaf = { t: Trunk }
data Branch = { t: Trunk }
data Shape = Leaf | Branch
data Trunk = { s: Shape }
```

A set asking for more values of its element than the element has:

```souther
data One = Int
    invariant only = value >= 1 && value <= 1
data Pair = Set<One>
    invariant two = Set.size(value) >= 2
```

## Why one issue and not two

The local causes differ — a sum needs a disjunction over its cases, a set needs
the number of distinct values its element has against the number it must hold —
but neither is another rule in the walk. Both ask whether a type has a value,
which the walk cannot ask because it visits positions. Closed separately, the
first one closed reads as the whole of it.

## The judgement

E1013 is no longer a reachability check for a mandatory cycle. It consumes a
conservative upper bound on the number of distinct values a type can have. A type
is refused only when that upper bound is zero.

The concrete value sets of recursive data are least fixed points, so the analysis
starts every recursive component at zero and rises monotonically as constructors
show values to exist. The abstract values are upper bounds at every iteration;
starting at zero is therefore about the least-fixed-point semantics of the
datatype, not a greatest-fixed-point analysis of bounds.

The transfer is by type shape:

```text
scalar / newtype   the carrier's upper bound, narrowed by the invariant domain;
                   zero where the seeded domain is infeasible

optional T         1 + |T|

record             the product of its field-occurrence upper bounds

sum                the sum of its cases' upper bounds

List<T>            Σ |T|^k over admitted lengths;
                   UNKNOWN when an admitted length beyond the enumeration bound
                   remains possible and |T| >= 1

[...]

PR #662, opened 2026-08-11
Read whether a collection may be empty before calling it a base case

### LINKED ISSUE #654: E1013 accepts recursive collection types whose invariants remove the empty value needed to bottom out
E1013 rejects a data whose construction needs a value of itself with nothing to bottom out, and
names two ways out: make the self-referring field optional (`?`) or hold it in a collection.
Collection containment is treated as a recursion break structurally, even where the invariants on
the collection remove the empty value that is the actual bottom.

## What happens

```souther
module demo

data Kids = List<Tree>
    invariant nonEmpty = List.length(value) >= 1

data Tree = { kids: Kids, v: Int }
```

Accepted. No value of `Tree` exists: every `Tree` holds a `Kids`, every `Kids` holds at least one
`Tree`, so nothing finite bottoms out.

`Set` and `Map` are accepted the same way, so this is not about `List`:

```souther
data Kids = Set<Tree>          invariant nonEmpty = Set.size(value) >= 1
data Kids = Map<String, Tree>  invariant nonEmpty = Map.size(value) >= 1
```

## Where it is decided

`DataChecker.mandatoryReaches`
(`souther-compiler/src/main/java/souther/compiler/check/DataChecker.java:455`) forms an edge only
where a field's type is a `Type.Ref`. A collection-typed field is not one, so what the collection
holds is never reached and the walk stops there. The javadoc above it states the rule being
applied:

> An optional (`?`) field or a `List`/`Map` field is a base case (`None` or the empty collection
> breaks the cycle), so it does not count as a mandatory edge.

The cause is not which collection it is. It is that reachability into a collection's contents is
left out of the judgement, on the strength of a bottom the invariants may have taken away.

## Why the optional is not the same case

`None` is a value of every optional, and the language keeps it that way: E1009 refuses a newtype
wrapping an optional (`data Next = Tree?`), so there is nowhere to write a rule that removes
`None`. An optional's bottom is held by a language rule. A collection's empty bottom is held only
by whatever the invariants on it admit, and a size invariant removes it. So `?` is an
unconditional way out and a collection is not, and the difference is stated in the language rather
than assumed by this check.

## Where it was found

While investigating #651. The generator cannot compose a value for this type, and the question
there was whether a search that failed may be reported as "no value can be written there". It may
not — but this is the type where that sentence would be true, and nothing in the compiler
establishes it. Filed apart from #651: that one is about a synthesis path too narrow for types that
do have values, and this is about what the front end claims when it calls a data constructible.

## Expected

A recursive path through a collection is a base case only where the collection can be empty under
the invariants that apply to it.

How to get there is not settled here. The local reading is to ask for the minimum size where a
collection is being taken as the base case; the general one is an inhabitance judgement E1013
consults, which is a larger change and closing this does not require it. Either way the compiler
already reads a cardinality lower bound off these same invariants — `Partitions.leastHeld` derives
one for the generator — so the fact the check needs is one it can already obtain.

kawasima:
Closes #654.

E1013 refuses a data whose construction needs a value of itself with nothing to
bottom out. Which occurrences it could stop at was decided from the shape of the
type written there: a reference was a mandatory edge, an optional or a collection
was a base case.

That is right about the optional and an assumption about the collection. `None`
is a value of every optional and the language keeps it that way -- a newtype may
not wrap an optional, so there is nowhere to write a rule that removes it. A
collection's empty value is held by nothing but the rules on the collection, and
a rule about how much it holds removes it. After that the recursion runs through
the collection like any other, and there is no first value:

```souther
data Kids = List<Tree>
    invariant nonEmpty = List.length(value) >= 1

data Tree = { kids: Kids, v: Int }
```

## What changes

The walk reaches one occurrence at a time rather than one type at a time, because
whether a position is mandatory is not a property of the type standing there. The
same `Kids` is an edge at `kids: Kids` and none at `kids: Kids?`, and a list is a
base case or not according to what the rules reaching it admit.

Four rules, each with cases of its own:

- an optional stops the walk before what it holds is looked at;
- whether the value at an occurrence can be empty carries into what a name wraps,
  which is the same value;
- it is read afresh at what a collection holds, which is a value of its own;
- a collection's element is entered as a value with fields, not stopped at.

**Whether a collection can be empty is asked of the rules, not matched against
their shapes.** Seeding a declaration relates every rule that reaches a position
-- the record's clauses and each field's own type's, at whatever depth they name.
The question goes there: settle the count at none and see whether anything is
left. So all of these are read, and none of them is a case in the checker:

```souther
List.length(value) >= 1        List.length(kids.value) >= 1
List.length(value) > 0         List.length(kids) >= least, least >= 1
List.length(value) == 1        List.length(kids) /= zero, zero == 0
List.length(value) /= 0        Set.size(value) >= 1
List.length(value) < 0         Map.size(value) >= 1
```

No spelling is privileged, `/= 0` included. It used to be dropped on the way into
the domain, on the grounds that a disequality is a disjunction and the domain
holds conjunctions of bounds -- true in general, and not true where one of the
two sides is already out. A count is never below none, so `count != 0` is
`count > 0`. Asked of what the domain knows rather than of the shape of the form,
so it holds against a literal, against a sibling settled at none, and through a
name the field wears.

The walk also does not take an ill-typed rule as evidence. `List.length(value)`
on a newtype over another newtype is E1317, and a reader could still get a count
out of it -- so a type would be refused on the strength of a clause the author had
already been told to fix. The cycle is decided only where every declaration stated
rules that checked. A real cycle beside a wrong declaration is reported one build
later, and there is a test naming that cost.

## Commits

The first two change no behaviour. They exist because the reading E1013 needs was
inside the partitioning, and a checker that wrote its own would be the second
reader of one invariant -- which is what the classes it reads were gathered to
prevent.

| | |
|---|---|
| `5eed769a` | `DeclaredBounds` holds the reading of the rules; `TypeBounds` keeps composing it with a projection. `CountDomain` holds the step from an endpoint to a quantity. The duplicate `&&`-flattening is gone: five callers now share one. |
| `ea5e7987` | The count floor moves beside the bounds it derives from, and the generator asks there. |
| `104cbcd3` | The walk becomes occurrence-sensitive. |
| `8b7373cd` | The types with no value that the check still admits, held as passing tests. |
| `6fba9ba6` | The spec, at the three places that stated the old rule. |
| `cc8b0369` | The shortest cycle, and what the generator's descent test rests on. |
| `1fd58558` | Whether a collection can be empty is not where its counts stop; and an ill-typed rule is not evidence. |
| `1463f444` | That question goes to the domain the rules seed, not to a table of comparisons. `DeclaredBounds` goes back to being a range. |
| `4cc83317` | A disequality with one side already out is a bound, not a dropped disjunction. The last clause-reading goes. |

## Not in scope

E1013 still decides a cycle rather than inhabitance. A sum whose every case is on
the cycle has no value and is admitted, as does a set asking for more distinct
values than its element has. Both are held as passing tests naming the gap, and
filed as #661 -- one issue, because neither is another rule in the walk and
closing one alone would read as the whole of it.

## Two rows that turned out not to be programs

The reasoning called for a case of an optional inside a collection and a case of a
count rule on a name over another name. Neither is writable: E2308 and E1317. Both
are held as the diagnostics they are, so the absent cases are absent for a reason.

## One expectation moves

`aTypeWrittenInTermsOfItselfIsGivenUpOn` asked what the generator's descent does
with a type written in terms of itself, spelled as a newtype holding a non-empty
collection of itself. The front end now refuses that model, so the descent never
sees it. The question is asked through a sum instead, which is the spelling that
still reaches the generator, and the test says it rests on that gap.

## Verified

Full suite green: compiler 4355, lsp 202, bench 1.

souther-examples compiled clean against this branch's jars (`clean compile`,
worktree-local repository): zero E1013 across all 11 modules. `member` fails on a
pre-existing `会員 as _` parse error (E2303) at the same site under a develop-era
jar. A positive control -- the model above, dropped into an example -- is refused
here and accepted by the develop jar, so the comparison discriminates.

https://claude.ai/code/session_019GSZdJqs1yDbifBZ3P69a6


kawasima:
Both confirmed and fixed in 1fd58558. The first was wider than the report and the second narrower in cause, so here is what measurement said.

## The floor reading

Your trace of `ofSize` -> `ordered` -> `EQ` -> empty is right. Measured:

```
leastCountOf(Ref(Kids))    >= 1  -> 1
                           >  0  -> 1
                           == 1  -> 0
                           /= 0  -> 0
```

`== 1` at a plain field was nevertheless refused, because `FieldDomains.heldAt`
runs a solver that does read equality and returned 1. So the type-level reading
had already lost it and the record-level reading was standing in. Four
occurrences with no record to stand in were admitted, each with no value to
build:

| occurrence | rule | was |
|---|---|---|
| newtype holding a collection of itself | `== 1` | admitted |
| collection's element | `== 1` | admitted |
| field | `/= 0` | admitted |
| record's own rule about a field | `/= 0` | admitted |

The fix takes your suggestion: `ofSize` is untouched and keeps answering only
for comparisons a range can hold. The question the walk actually needs is not
"where do the counts stop" but "is the empty collection admitted", which is
`DeclaredBounds.mayHoldNothing`, read off the operator as written -- a floor
above none, an equality away from none, a disequality at it. A rule stating only
how many at most leaves the empty one, since none is under every such rule.

The shape of a size comparison is now recognised once
(`InvariantBound.sizeComparedIn`) and read in two places, so `ofSize` and this
cannot come to disagree about which clauses are about a count.

Dropping the floor also dropped the `FieldDomains` call from the walk: reading
the record's conjuncts directly answers this question, and is conservative where
the solver was informative (`List.length(a) + List.length(b) >= 1` matches no
subject, so it removes nothing -- which is the right answer).

## The cascade

You were right that I was hiding it. That assertion is `assertEquals` now.

The cause is narrower than a validity gate on the reader. `NumericMeasures.takenOf`
reaches through the newtype chain, and the type checker does not apply the same
transparency to a function argument -- so `List.length(value)` where `value : Inner`
is E1317 to one and a readable size comparison to the other. Making the readers
agree is a change to what newtype transparency means per operation, which is not
this PR.

What is in this PR: the walk no longer takes an ill-typed rule as evidence. The
cycle is decided only where every declaration stated rules that checked. Only
declarations have been checked at that point, so a mistake in a body further down
does not silence it.

That has a cost and there is a test naming it: a real cycle beside a wrong
declaration is reported one build later. Against a type refused on the strength of
a clause the author has already been told to fix, one of those is a diagnostic
arriving late and the other is one that should never have been derived.

## Verified

Full suite green: compiler 4350, lsp 202, bench 1. souther-examples unchanged --
`clean compile` against this branch, zero E1013 across all 11 modules, `member`
still failing only on the pre-existing `会員 as _` parse error.

kawasima:
"根本に問題ありそう" was the right call. Both findings reproduce, and they are the
same finding: I was matching invariant syntax to answer a semantic question, so
every way of writing the same rule was another row in a table. Fixed in 1463f444
by asking the rules instead. PR body updated -- it did describe the superseded
design.

## What the table was missing

Reproduced, plus one you predicted:

```
negative ceiling, < 0                  admitted
negative ceiling, <= -1                admitted
record rule via kids.value             admitted
record rule via a sibling floor        admitted     <- your FieldDomains point
control: bare kids >= 1                E1013
```

## The reading

Seeding a declaration already relates every rule that reaches a position -- the
record's clauses and each field's own type's, at whatever depth they name. So the
question goes there rather than to a comparison table: settle the count at none
and see whether anything is left.

```java
seeded.numbers().assume(count, Rel.LE, DISCRETE).isBottom()
```

Measured against the domain before writing it:

| rule | domain |
|---|---|
| `List.length(kids) >= 1` | cannot be empty |
| `List.length(kids.value) >= 1` | cannot be empty |
| `>= least` beside `least >= 1` | cannot be empty |
| `== 1` | cannot be empty |
| `List.length(value) >= 1` on the field's type | cannot be empty |
| `<= 3` | may be empty |
| `< 0` | contradictory, no value at all |
| no rule | no counted atom, may be empty |
| `/= 0` | **may be empty** |

So your suggestion to bring `FieldDomains` back was right, and stronger than
OR-ing it with the direct reader: the seeding subsumes the direct reader
entirely. `DeclaredBounds` goes back to being a range, which is what it was for --
where the rules leave holes between their ends was never a range's question.

`aFloorStatedThroughASiblingIsStillAFloor` is now a test. Neither clause alone
removes the empty list.

## The residue

`/= 0` is the one shape left. The seeding widens a disequality to a floor of none,
so the domain cannot see it -- `entails(count NE 0)` is false there too. That is
the domain's representation and not this reading's, and closing it means teaching
the seeding to keep a hole, which is not this PR. That shape alone is read off the
clause, and the javadoc where it is read says why it is there.

`< 0` needed no case: the domain is already bottom, so nothing is left after
settling the count at none, and it falls out as "cannot be empty". Held as a test
with its own reason -- a ceiling below none leaves no collection at all, not just
no empty one.

## Spec

The sentence you flagged was false for negative ceilings. It no longer enumerates
spellings at all, since the check no longer does:

> A rule stating only how many at most leaves the empty collection where it was,
> unless it puts the count below none, which leaves no collection at all.

## Verified

Full suite green: compiler 4353, lsp 202, bench 1. souther-examples unchanged --
`clean compile` against this branch, zero E1013 across all 11 modules, `member`
still only on the pre-existing parse error.

kawasima:
Closed in 4cc83317, at the domain rather than beside it. No separate issue needed.

## Where it actually was

Your two reproductions both landed. Chasing them led one level down, to the point
where the disequality was being thrown away:

```java
if (rel == Rel.NE) {
    // `f != 0` is a disjunction, and the domain holds conjunctions of bounds. Nothing to
    // record; what settles such a guard is the fact keyed on the comparison itself.
    return f.coefs().isEmpty() ? d : d.losing(Loss.DROPPED_DISEQUALITY, f.coefs().keySet());
}
```

That is true in general and not true here. A disequality is a disjunction only
while both sides are open, and a count is never below none -- so `count != 0` has
one side left and is `count > 0`. The domain was declining to record a bound it
had every right to.

So the fix is in `NumericDomain.assume`: before dropping, ask what is already
known. If the domain proves `f >= 0`, assert `f > 0`; if it proves `f <= 0`,
assert `f < 0`; otherwise drop as before. Asked of the domain's knowledge rather
than the shape of the form, which is why it reaches the cases a reader would have
had to be taught one at a time:

| spelling | |
|---|---|
| `List.length(value) /= 0` on the newtype | E1013 |
| `List.length(kids) /= 0` on the record | E1013 |
| `List.length(kids.value) /= 0` | E1013 |
| `List.length(kids) /= zero` beside `zero == 0` | E1013 |
| `0 /= List.length(value)` | E1013 |
| `Set.size(kids.value) /= 0` | E1013 |
| `List.length(value) /= 3` | admitted |

`aHoleAtNone` is gone. `mayHoldNothingAt` is now one query and reads no clause at
all -- there is no spelling the check recognises, only rules and what they leave.
Your sibling case is a test.

## Two tests were built on the imprecision

Both say so in their own names, which is how I found what they were resting on.

`aMapNothingWasPairedForIsSaidAsARefusalAndNotASearch` had this javadoc:

> `/= 0` says the map is not empty and no range holds it, so no minimum is read
> and no pairing is built

That was a description of the gap. With the minimum now read, pairings are built
and the search for a key and value clearing eight rules apiece is what stops, so
the reason is `SEARCH_LIMIT` and it is that because that is what happens. Renamed
and rewritten, with a note that the refusal/search distinction still matters and
that a map the rules will not let be empty is no longer an example of it.

`aFloorReachedByChoosingASiblingFirstIsOfferedThere` has a `Count` with
`value >= 0 && value <= 5` and `value /= 0`. Its floor is now one rather than
none, so the row offers `n = Count(1), xs = [0]` instead of five of each. What
that test is about -- as many elements as the count it settled on asks for -- still
holds.

## On seedFields being re-run

Agreed it is not a merge blocker, and I have left it. Recording the shape you
suggested: `mayHoldNothingAt` wants the seeded domain, `FieldDomains` already
holds one, so the call belongs on the instance rather than on the class. It would
also let the walk hold one per declaration instead of one per field.

## Verified

Full suite green: compiler 4355, lsp 202, bench 1. souther-examples unchanged --
`clean compile` against this branch, zero E1013 across all 11 modules, `member`
still only on the pre-existing parse error. PR body updated.

Issue #508, opened 2026-08-09
A guard on a computed value does not chain with the input type's invariant


kawasima:
A guard on a computed value does not chain with the input type's invariant.

## What happens

Two steps are needed to settle `withinCap` here: the guard says `t <= bal.outstanding`, and `Bal`'s own invariant says `outstanding <= principal.value`. The chain is made when the guarded value is a parameter or a name bound to one, and is not made when it is a difference.

```
data M = { cap: Decimal, v: Decimal }
    invariant withinCap = v <= cap

data Yen = Decimal
    invariant nonNegative = value >= 0m

data Bal = { principal: Yen, outstanding: Decimal }
    invariant outstandingWithinPrincipal = outstanding <= principal.value

// u1 — settled
let u1 (bal, d) = {
    guard d <= bal.outstanding else No
    M { cap = bal.principal.value, v = d }
}

// u2 — settled (a name bound to a parameter)
let u2 (bal, x) = {
    let d = x
    guard d <= bal.outstanding else No
    M { cap = bal.principal.value, v = d }
}

// u3 — E2011
let u3 (bal, x, y) = {
    let d = x - y
    guard d <= bal.outstanding else No
    M { cap = bal.principal.value, v = d }
}

// u4 — E2011 (the same difference written inline at both places)
let u4 (bal, x, y) = {
    guard x - y <= bal.outstanding else No
    M { cap = bal.principal.value, v = x - y }
}
```

```
-- INVARIANT (WARNING)  E2011 --------------matrix2.sou:35:5
35|     M { cap = bal.principal.value, v = d }
```

A difference is inside the fragment on its own: `guard a >= t` followed by `NN { p = a - t }` is settled, including where `t` is a field or a sum of two fields. What is lost is the second step — carrying a relation the guard established onward into the invariant of the type the other operand came from.

## Why it matters

This is the shape every "take a payment off a balance" behavior has, and it is the shape the specification's own `withdraw` example uses. Because the relation cannot be carried, the rule has to be discharged by an attempted construction whose failure arm is `unreachable` — a runtime check plus a written promise, where the compiler had both halves of the proof.

## Expected

`u3` and `u4` settle, as `u1` and `u2` do. What the guard establishes about a value should not depend on whether that value was written as a name or as the expression the name stands for.

Found in round-5 MCP dogfooding (loan repayment model), narrowed with `matrix2.sou`.

kawasima:
[...]

PR #669, opened 2026-08-11
Enter a binding an expansion writes where it stands

### LINKED ISSUE #667: A field read on a call's result leaves the construction unchecked, so E2011 is silent where it must fire
A construction whose term is read through a field access on a call is not checked at all. E2011 is
raised for the same construction written as a field chain, and is silent when the same value is
reached through a helper call, including where the construction must violate.

## What happens

```souther
module repro

data Amount = Decimal
    invariant value >= 0.0m

data Order = { amount: Amount }
data Refund = { amount: Amount }

let amountOf (order: Order): Amount = order.amount

behavior byChain : (order: Order, fee: Amount) -> Refund
    constructs Refund, Amount

let byChain (order, fee) = {
    Refund { amount = Amount(order.amount.value - fee.value) }        // E2011
}

behavior byCallField : (order: Order, fee: Amount) -> Refund
    constructs Refund, Amount

let byCallField (order, fee) = {
    Refund { amount = Amount(amountOf(order).value - fee.value) }     // silent
}
```

Neither behavior guards anything and `fee` may exceed the order's amount, so both constructions can
violate `Amount`'s invariant at run time. The first is E2011. The second is not diagnosed.

The silence is not a discharge. Nothing was established for anything to be discharged against, and
it holds with a guard added and with a construction that violates on every input.

## What the shape is

Measured on `develop` (eb73165d). No guard anywhere, so every row is one E2011 must be raised for
whatever the discharge machinery can prove. `neg` is `let neg (order: Order): Decimal =
order.amount.value - 100.0m`.

| construction | E2011 |
| --- | --- |
| `Amount(neg(order))` | raised |
| `Amount(neg(order) - fee.value)` | raised |
| `let n = neg(order)` then `Amount(n - fee.value)` | raised |
| `let paid = amountOf(order)` then `Amount(paid.value - fee.value)` | raised |
| `Amount(order.amount.value - fee.value)` | raised |
| `Amount(amountOf(order).value - fee.value)` | **silent** |
| `Amount(orderOf(line).amount.value - fee.value)` | **silent** |
| `Amount(order.amount.value - amountOf(order).value)` | **silent** |
| `Amount(amountOf(order).value - 100.0m)` | **silent** |

A field read applied to a call is the whole of it. Expansion reaches a call written as an operand
(rows 1–3), and binding the call to a name first leaves the term readable (row 4). One operand of
the dropped shape is enough to drop the construction (row 8), and the arithmetic is not involved
(row 9).

## Why this is a soundness hole and not missing proof power

E2011 is what stands between a construction the guards leave open and a run-time abort. Where it is
silent an author reads the construction as established, and the two silences are not distinguishable
in the source: the same file warns when the term is written as a field chain and says nothing when
the same value is reached through a helper. The shape that says nothing is the one an author keeps.

#409 separated "says nothing because nothing could be asked of it" from "said something wrong", and
settled that a function result is not a term. This is neither: the construction is one the check
names in every other spelling of the same value, and a field read is what `<<invariant-discharge-
terms>>` lists as a term.

## Where it comes from, to be confirmed at the fix

A call is expanded by binding each argument, so `amountOf(order).value` becomes a field read whose
target is a `let`. `Terms.composed` reads through a `Core.LetIn` at the head of an expression; the
location behind a `Core.Field` is read without that step, so the affine form comes back null and the
construction is owed no clause. A statement-level `let` enters its binding by the ordinary route,
which is why row 4 is checked.

## Acceptance criteria

- `Amount(amountOf(order).value - fee.value)` and `let paid = amountOf(order)` followed by
  `Amount(paid.value - fee.value)` are the same term to construction-side recognition: the first is
  diagnosed exactly where the second is.
- The rows above are held as a regression test, written with no guard, so each row asserts a
  diagnostic that must be raised rather than one the check happens to raise today.
- A field read of more than one level on a call (`orderOf(line).amount.value`), and a call's field as
  one operand of a larger expression, are covered.

## Provenance

Discovered while investigating #665. The reproduction pasted there does not trigger this, because it
binds the helper's result to `paid` before reading the field; the reporter's separate observation
that wrapping a field read in a helper silenced E2011 is consistent with this failure mode. #665's
stated direction — that routing a read through a helper makes a term easier to discharge — is not
what happens. The construction leaves the check.

The discharge issues so far divide in two, and this is the second kind: a relation that cannot be
recovered as meaning (#508, #648), and a construction expression that cannot be recovered as a term
(#409, and this).

kawasima:
Fixes #667.

A construction whose term is read through a field access on a call was not checked at all:
`Amount(amountOf(order).value - fee.value)` is silent with no guard anywhere and on every input,
while `Amount(order.amount.value - fee.value)` is E2011.

## What was happening

A call the discharge analysis expands is a binding holding what the call was given. Where that
binding stands inside a value — under a field read, under one side of a comparison — the walk does
not step into it, so the name it writes is one nothing has entered, and an unentered binding denotes
nothing. The chain read off it names no location, and a construction over a term nothing can name is
owed no clause. The silence is not a discharge: nothing was established for anything to be
discharged against.

## The change

**Enter the binding where it stands.** The walk already reads a conditional in a value position from
inside the binders it is under; this asks the same of the binder itself. What is left after entering
it is the tree the source would have written with a `let`, which is the tree the rest of the walk
already reads. `<<invariant-discharge-terms>>` already states this — a helper "is expanded by
binding each argument and reading the parameter through that binding, which is why a construction
moved into a helper still reads the terms the caller's guards and input types settled" — so this is
the implementation catching up to the specification, and nothing in the specification changes.

**Read a newtype's `.value` off a computed value.** A helper whose body computes rather than locates
answers arithmetic, and `.value` read off that answer was one opaque value where the same call given
a name was the arithmetic its body wrote. The rule locations are keyed by (`pathKey`) is now read of
a computed value too. Without it, entering the binding names the term but leaves it unreadable — the
form is named and not read, which is how a first attempt at this turned six examples into false
positives (`statutoryDailyHours.value` became an opaque atom rather than the constant it wraps, and
the `if` guarding the construction stopped settling it).

## What it is held to

`CompileInvariantThroughACallTest`, 12 cases. The ones about a construction leaving the check are
written with no guard: each is a diagnostic that must be raised whatever the discharge machinery can
prove, which is what tells a silence that discharged from a construction nobody looked at. The two
that assert silence are each paired with a spelling one step past what the guard establishes.

`mvn clean test`: BUILD SUCCESS, no test changed its answer.

The `souther-examples` corpus (14 projects) goes from 77 diagnostics to 71: **six warnings removed,
none added.** The six are guards whose terms were unreadable and now settle their constructions —
`let gross = totalRemuneration(remuneration) + upliftPay` and the like. Each was checked against the
same source with the addition replaced by a subtraction, which still warns, so the six are
discharges rather than droppings. Compile time on the largest example is unchanged (1.5s, JVM start
included).

## Not in this change

A guard over two newtype values bound by `let` — `guard gross >= subtracted` then
`Yen(gross.value - subtracted.value)` — warns before and after this change alike (`payroll.sou:195`
in the corpus). It looks like #508's family rather than this one.


kawasima:
Both addressed in d8b9ded1.

## P1 — the search crossed control-flow boundaries

Confirmed, and it reproduces at the level you described:

```souther
let same (n: Net): Net = n
let settle (x) = Refund { net = if x >= 0.0m then same(Net(x)) else Net(0.0m) }
```

E2011 on `Net(x)` — the construction the condition establishes. The binding `same` writes was found
from outside the conditional and entered before `x >= 0.0m` was assumed.

`bindingIn` now keeps the same boundary at every level rather than only at the root: a `Block`, a
branch, an arm and a departure each stop it, and it looks only where the value is computed on the
way to this one — a condition, a scrutinee, an attempted construction, and the arithmetic around
them. What stands inside a branch is found again when the walk goes in, under what holds there.

Three regression cases, one per boundary: an expansion in a branch read under the condition that
chose it, one reading what a `match` arm binds, and one reading what an attempted construction
built. Each asserts `PROVED` rather than the absence of a diagnostic, so a binding lifted out of the
scope fails it whether the loss shows up as an unknown or as an unreadable term.

## P2 — the silences are read off the verdicts

You are right that this branch's own defect satisfies `assertFalse(hasWarning(...))`. The tests moved
to `souther.compiler.check` and read `WATCHING`: `UNREPRESENTABLE` is the silence this is about,
`PROVED` is the answer. The unguarded cases assert `UNKNOWN` and, where an author is the point, the
diagnostic as well.

That also made the two rules hold each other honestly. Turning off each one in turn, the failures do
not overlap:

- without entering the binding: the two-level chain (`orderOf(line).amount.value`) is
  `UNREPRESENTABLE`, and a location less itself is `UNKNOWN` rather than `PROVED`
- without reading `.value` off what a call answered: both cases through a helper that computes are
  `UNREPRESENTABLE`

## Unchanged by the fix

`mvn clean test` BUILD SUCCESS. The example corpus is identical to the build you reviewed — 77
diagnostics to 71, six removed and none added — so the boundary fix costs none of what the branch
settles.

Issue #672, opened 2026-08-12
Suppress the diagnostics read off the type an unresolved name leaves behind


kawasima:
A check that reads a value another error produced reports what it finds there, and the author is told about both — the error and the thing that followed from it. `CascadeStopsAtTheCauseTest` holds the rule for the case it covers: what follows from a name that denotes nothing is not reported again.

These were invisible while a compile answered with one error. Reporting every error a compilation found (#670) makes them observable.

The mistake is one: a type name that denotes nothing. It is reported where it is written, and what it names becomes the type that absorbs, so the rest of the module is still checked. Two rules then read that type as though it were a type an author wrote, and each says a second thing about it.

## The type that absorbs is read as a union member

```souther
module up exposing ( Amount )
data Amount = Int
```

```souther
module down

data Out = { n: Int }

behavior f : (a: Int) -> Out | up.Refused constructs Out
let f (a) = Out { n = a }
```

```
5:32 [E1506] `up` declares no `Refused`.
5:26 [E1613] `?` cannot be a union member: a member is a type a `match` arm can name — one of the primitives, or a data type declared here or imported.
```

E1506 is the error. E1613 is what the union rule finds once `up.Refused` denotes the type that absorbs, and there is nothing an author can do about it other than fix the name — the member it names, `?`, is not a spelling they wrote or could write.

`TypeOps.memberName` answers `null` both for a type no arm can name (`List<Int>`, an `Option`, a function, `Raw`) and for a type the compiler could not work out. `TypeOps.successType` reads that one answer and reports the second sentence about the first.

Swept over every position the specification says a type is written in (`a-type-is-written-in-a-type-position`) — thirteen of them, as the sweeps over the positions write them out — this fires in **every position a union can be written in**: nine. The other four (a data field, a newtype base, a type argument, a tuple member) refuse a union at the parse, so there is no name in them to read. Not the behavior output alone.

## The type that absorbs is read as a declared output with no cases

Against the same `up`:

```souther
module down

data A = { a: Int }
data Out = { v: Int }

behavior s1 : (i: Out) -> A constructs A
let s1 (i) = A { a = 1 }

behavior s2 : (i: A) -> A constructs A
let s2 (i) = A { a = i.a }

behavior go = s1 >-> s2
    -> up.Refused
```

```
13:8  [E1506] `up` declares no `Refused`.
12:1  [E1604] behavior go declares its output as , but the pipeline produces A.
```

[...]

PR #673, opened 2026-08-12
Answer with every error a compilation found

### LINKED ISSUE #670: Compiler.compile drops all but the first accumulated error report
A compilation accumulates every error it finds, and then reports one. `Compilation.firstError` picks a single `Db.Found` and rebuilds a single-diagnostic exception from it, so everything else that was found is dropped on the way out.

This is the reason a `souther compile` that has several independent errors sends the author back for another run per error, even where the compiler had already worked all of them out.

## What is already in place

Nothing downstream of that method needs the errors to be one.

- `CompileException` carries a list (`ofAll`, `ofAllInSources`), and `code()` / `getMessage()` answer from the first, so a caller reading either is unaffected by the rest being there.
- The CLI renders every diagnostic an error carries: `Main.report` is handed `e.locatedDiagnostics()` and `DiagnosticRenderer.renderAll` walks it. `--format json` takes the same path, one object per diagnostic.
- The editor path already reports all of them. `Compiler.diagnoseModules` goes through `Compilation.diagnostics()`, which reads `errors(found)` — the same reports, none collapsed.

So the loss is at one method, and only on the way from a compilation to the exception a compile throws.

## Measured

With `firstError` replaced by an aggregate over the error reports (`ofAllInSources`, first report leading):

```
souther-syntax     93 tests  failures=0 errors=0
souther-fmt      2224 tests  failures=0 errors=0
souther-compiler 4382 tests  failures=0 errors=0 (1 skipped)
souther-lsp       202 tests  failures=0 errors=0
BUILD SUCCESS
```

A green suite here says the existing external contract is not broken by the change. It does not say the new behaviour works — nothing in the suite constrains how many errors a compilation reports, so the aggregation arrives with no coverage of its own and needs its own tests.

Positive control, on a module where two behaviors each under-declare `constructs` (`SpecChecker` left as it is):

```
diagnostics = 2
   [E1002] Behavior `start` constructs `InProgress` but does not declare `constructs InProgress`.
   [E1002] Behavior `finish` constructs `Done` but does not declare `constructs Done`.
getMessage = 10:1 E1002: Behavior `start` constructs `InProgress` but does not declare `constructs InProgress`. ...
code = E1002
```

Two diagnostics where there was one, and the leading message and code unchanged.

## What this asks for

Report every error a compilation accumulated, not the first.

The order they come out in is part of it, because an author reads them as a list:

- by source, in the order the sources were given (what `firstError` already decides);
- then by position within a source;
- ties broken by the order the checker produced them, so a stable sort is enough. Several diagnostics at one position is not hypothetical — a checker that reports each of its own violations puts them all at the declaration it is about.

The tests should hold the contract rather than the count: every error report reaches `locatedDiagnostics()`, `code()` and `getMessage()` still answer from the first, and the order is the one above.

Whether the same diagnostic can reach the exception twice is unmeasured — with one returned there was nothing to see. If it does, the producer registering it twice is where it is fixed; `TypeChecker.deduped` is the precedent, and deduplicating at this boundary would be guessing at an identity two independently found violations may not share.

## Relation to #664

#664 is the same loss one level in: `SpecChecker` stops at the first missing `constructs` entry within a behavior, so the later ones never reach a report at all. The two are independent, and both are needed before the case in #664 compiles clean in one run.

This one goes first. On its own it takes a module with N under-declared behaviors from N runs to one run naming the first missing type of each. #664 on its own would change nothing an author sees, because what it adds is dropped here.

kawasima:
Closes #670.

A compilation works out every error it finds and files a report for each. `Compilation.firstError` then picked one of them and rebuilt a single-diagnostic exception, so the rest were dropped on the way out — and the author fixed one error, compiled again, and was told the next.

Nothing downstream needed them to be one. `CompileException` already carries a list, `Main.report` is handed `locatedDiagnostics()` and renders each with its own quoted line, `--format json` writes one object per diagnostic, and the editor path (`Compiler.diagnoseModules`) never went through here at all.

## Before / after

```souther
behavior start : (t: Ticket) -> Ticket
    constructs Ticket
let start (t) = Ticket { lines = [ CookLine { state = InProgress } ] }
-- and the same for `finish` and `reopen`
```

Before, one E1002. Now, on `souther compile`:

```
-- INSUFFICIENT CONSTRUCTION AUTHORITY  E1002 -repro.sou:7:1
7| behavior start : (t: Ticket) -> Ticket
   ^
Behavior `start` constructs `CookLine` but does not declare `constructs CookLine`.
Hint: Add `constructs CookLine` to `start`, or stop constructing CookLine here.

-- INSUFFICIENT CONSTRUCTION AUTHORITY  E1002 -repro.sou:11:1
... `finish` ...

-- INSUFFICIENT CONSTRUCTION AUTHORITY  E1002 -repro.sou:15:1
... `reopen` ...
```

`--format json` gives three lines. The exit code is unchanged.

## The order

Source order, then position within a source, stable where two sit at one place — so a check reporting each of its own violations at one declaration keeps the order it produced them in.

Where a diagnostic is says where to look and nothing about what caused what. A check reading a value another error produced can be written to the left of the error it came from, and then it leads; that is not this order failing to find the cause, because a cause is not recoverable from two positions. Withholding a diagnostic that should not be said at all is the checker's own question — #672, with the two pairs this surfaced.

## Leading

`code()` and `getMessage()` answer from the first of the list, and the error takes on the others through the exception it was raised with rather than being rebuilt from its leading diagnostic. A pass that words its own summary — the one reporting every failing `example` row — words it over what it found, and re-rendering would answer with a sentence about one row.

Deduplication needs nothing here: `Db.allReports` already keeps one report per (module, source, `Diagnostic.identity()`), and the identity carries what the diagnostic says with its arguments.

## Tests

New: `ACompileAnswersWithEveryErrorItFoundTest` (9) and `AnErrorKeepsTheMessageItWasRaisedWithTest` (5) — every error carried, the leading code and message, the ordering including the tie, and the raised message surviving.

Four existing tests were reading the leading diagnostic as though it were the only one. Two turned out to be sitting on a cascade and now say what they are about; one had a fixture carrying a second error of its own, written without it now; one had a disjunction narrower than its sibling's.

Verified: `mvn clean test` across all modules — 6916 tests, no failures — and the CLI run above.

### DIFF
--- a/souther-compiler/src/main/java/souther/compiler/Compiler.java
+++ b/souther-compiler/src/main/java/souther/compiler/Compiler.java
@@ -215,19 +215,19 @@ public final class Compiler {
         compilation.measure(measure);
         Db db = compilation.db();
 
-        CompileException structural = compilation.firstError(compilation.structuralReports());
+        CompileException structural = compilation.failure(compilation.structuralReports());
         if (structural != null) {
             throw structural;
         }
 
         db.ask(new Output.All());
-        CompileException failed = compilation.firstError(db.allReports());
+        CompileException failed = compilation.failure(db.allReports());
         if (failed != null) {
             throw failed;
         }
         for (String module : compilation.modules()) {
             if (!db.ask(new Output.ConstConstructions(module)).present()) {
-                CompileException bad = compilation.firstError(db.allReports());
+                CompileException bad = compilation.failure(db.allReports());
                 if (bad != null) {
                     throw bad;
                 }
@@ -424,13 +424,13 @@ public final class Compiler {
         compilation.measure(measure);
         Db db = compilation.db();
 
-        CompileException structural = compilation.firstError(compilation.structuralReports());
+        CompileException structural = compilation.failure(compilation.structuralReports());
         if (structural != null) {
             throw structural;
         }
 
         db.ask(new Output.All());
-        CompileException failed = compilation.firstError(db.allReports());
+        CompileException failed = compilation.failure(db.allReports());
         if (failed != null) {
             throw failed;
         }
@@ -442,7 +442,7 @@ public final class Compiler {
         List<String> exampleSources = new ArrayList<>();
         for (String module : compilation.modules()) {
             if (!db.ask(new Output.ConstConstructions(module)).present()) {
-                CompileException bad = compilation.firstError(db.allReports());
+                CompileException bad = compilation.failure(db.allReports());
                 if (bad != null) {
                     throw bad;
                 }
--- a/souther-compiler/src/main/java/souther/compiler/query/Compilation.java
+++ b/souther-compiler/src/main/java/souther/compiler/query/Compilation.java
@@ -4,11 +4,14 @@ import souther.compiler.ast.Ast;
 import souther.compiler.check.Sig;
 import souther.compiler.check.Symbols;
 import souther.compiler.diag.CompileException;
+import souther.compiler.diag.Diagnostic;
 import souther.compiler.diag.LabeledRegion;
 import souther.compiler.diag.Located;
+import souther.compiler.diag.SourcePos;
 import souther.compiler.meta.ModulePath;
 
 import java.util.ArrayList;
+import java.util.Comparator;
 import java.util.LinkedHashMap;
 import java.util.LinkedHashSet;
 import java.util.List;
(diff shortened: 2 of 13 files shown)


kawasima:
Both review points are in, along with the optional one.

**The source key of the order was unheld.** Confirmed by collapsing it to a constant: every one of the nine ordering tests stayed green, exactly as the review said it would. The new test hands two sources over as a list, the error far down the first and the other at the top of the second, and holds that the first source leads.

**`ofDocuments` assigns no index.** Left as it is — nothing reaches it, since an editor reads the reports per file rather than asking for the failure. But the sentence about "the order the sources were given" reads as a promise covering both compiles, so `failure` now says which one it is about and what happens in the other.

**The class javadoc** no longer says the batch compiler raises the first.

**A warning among the reports** is now what `nothingAmongThemIsNotAnError` also passes, since that is what its name claims. It bites: removing the `isError()` filter fails it.

`mvn clean test` across all modules — 6917 tests, no failures.

PR #674, opened 2026-08-12
Read a comparison call as the order it decides

### LINKED ISSUE #668: invariant discharge does not recognize an ordered relation written as compare(a, b) against zero
Two guards stating the same order supply different things to the invariant-discharge check. Written
with the operator, the relation is available where a construction needs it; written as the library's
`compare` against zero, nothing is available at all.

```souther
module repro

data Amount = Decimal
    invariant value >= 0.0m

data Order = { amount: Amount }
data Refund = { amount: Amount }
data FeeTooHigh

behavior byOperator : (order: Order, fee: Amount) -> Refund | FeeTooHigh
    constructs Refund, Amount, FeeTooHigh

let byOperator (order, fee) = {
    guard order.amount.value >= fee.value else FeeTooHigh
    Refund { amount = Amount(order.amount.value - fee.value) }        // silent
}

behavior byCompare : (order: Order, fee: Amount) -> Refund | FeeTooHigh
    constructs Refund, Amount, FeeTooHigh

let byCompare (order, fee) = {
    guard Decimal.compare(order.amount.value, fee.value) >= 0 else FeeTooHigh
    Refund { amount = Amount(order.amount.value - fee.value) }        // E2011
}
```

```
-- INVARIANT (WARNING)  E2011 --------------repro.sou:26:23

26|     Refund { amount = Amount(order.amount.value - fee.value) }
                          ^

Constructing `Amount` here may violate its invariant: the guards do not establish it.
```

## The guard's spelling is the whole of it

Every shape the guarded term can take discharges under the operator guard, and every one of them
warns under the `compare` guard. Measured on `develop` (eb73165d), one element moved at a time:

| the term the guard and the construction read | `x >= y` | `Decimal.compare(x, y) >= 0` |
| --- | --- | --- |
| two parameters | silent | E2011 |
| a field chain (`order.amount.value`) | silent | E2011 |
| a three-level chain (`line.order.amount.value`) | silent | E2011 |
| a `let` bound to the field | silent | E2011 |
| a `let` bound to a helper's result | silent | E2011 |

`Int.compare(a, b) >= 0` over two `Int` parameters goes the same way, and so does the mirrored
`0 <= Decimal.compare(a, b)`.

Nothing is being dropped from the check here: after the `compare` guard, a construction over
`x - y - 1.0m` warns as it should. The relation the guard states is what is missing.

## Why this is not the rule #409 settled

#409 settled that a function result is not a term: `Int.max(0, n)` gives its result no bound, and a
construction over it is silent because the check cannot name the value at all. That rule is about
carrying a call's *result* onward as a term, and this does not ask for that.

`compare(a, b) >= 0` asks for the relation the call's own contract states between its two
*arguments*. The sign of `compare`'s result is defined as the order of `a` and `b`, so

```
compare(a, b) >  0   is   a >  b
compare(a, b) >= 0   is   a >= b
compare(a, b) == 0   is   a == b
compare(a, b) <= 0   is   a <= b
compare(a, b) <  0   is   a <  b
```

and the same five with the call on the right of the zero. Reading these widens nothing about what
counts as a term, and `<<invariant-discharge-terms>>` can stay as it is.

## Where it comes from

`Predicates.assumeCond` reads a `Core.Binary` by asking `Terms.affineOf` of each side. A
`Core.PreservedCall` has no affine form, so `compare(x, y) >= 0` contributes no constraint and the
guard leaves `Known` unchanged. The construction is then owed a clause with nothing to establish it.

## What the fix should be, and where it should not go

Not a case in `assumeCond` naming `Int.compare` and `Decimal.compare`. That grows a stdlib catalogue
inside the discharge check, and every ordered type the library gains later adds a row to it.

A comparison against zero over a declared comparison operation should be normalized to the relation
it states, before the affine reading — one rule from which all five relations and both operand orders
follow, keyed on what the library declares a comparison operation to be rather than on a list of
names written into the checker.

The operator-centric shape of the present tables is visible in `Question`: `Int.compare(Int, Int):
Int` falls in `OPERATOR`'s range by an accident of its types and is listed in `NOT_AN_OPERATOR`,
while `Decimal.compare(Decimal, Decimal): Int` is asked no question at all, its result type not
matching its operands'. What a guard needs answered is not *which operator does this call compute*
but *which relation does this expression state*.

## Acceptance criteria

- `guard x >= y` and `guard Decimal.compare(x, y) >= 0` supply the same relation to the discharge
  check, for `Int.compare` and `Decimal.compare`, for each of `>`, `>=`, `==`, `/=`, `<=`, `<`, and
  with the call on either side of the zero.
- A construction the guard does not establish still warns under the `compare` spelling: after
  `guard Decimal.compare(x, y) >= 0`, constructing over `x - y - 1.0m` is E2011.
- The matrix above is held as a regression test: each term shape against each guard spelling.

## Provenance

Found while measuring what the discharge check reads across guard spellings, alongside #667. The two
are separate defects and should stay apart: this one refuses a construction that is in fact
established, #667 leaves a construction unchecked. Fixing them together would let each one's fix
cover part of the other's regression test.

The discharge issues so far divide in two, and this is the first kind: a relation that cannot be
recovered as meaning (#508, #648, and this), and a construction expression that cannot be recovered
as a term (#409, #667).

kawasima:
Fixes #668.

`guard Decimal.compare(paid, fee) >= 0` stated no relation between `paid` and `fee`, so a
construction the same fact settles when written `paid >= fee` was reported. The sign was read as a
number and nothing asked what it was the sign of.

## The rule is the library's

An operation answering an `Int` from two values of one type is now asked whether its sign is their
order, and answers with the relation a positive answer states between its first argument and its
second. One row per operation, and the six relations and both sides of the zero follow from it —
there is no case in the reader naming `Int.compare` or `Decimal.compare`.

`Question` holds every operation in that range to an answer, which is what turned up the two the
library already had beyond `compare`:

- `Date.daysBetween(from, to)` counts forward from its first argument, so it is positive where the
  **second** is the greater — the direction the other way round from `compare`, read off the same
  row.
- `DateTime.minutesBetween` answers no order: it counts whole minutes, so a zero says the two are
  less than a minute apart rather than that they are equal, and a non-negative count does not say
  which is the earlier. Its strict signs do state the order, and a rule holding for four of the six
  relations is not this one.

## Reading a predicate never takes a reading away

The first version rewrote the clause in place, and that **silenced a warning in the corpus**: the
count is itself a number this check names, so `invariant Date.daysBetween(acquiredOn, lostOn) >= 0`
is a clause it can read, while the two dates it orders need not both be nameable — read against a
construction whose `lostOn` is `Date.addDays(1, separatedOn)`, the rewritten clause compares two
dates nothing relates and stops being read at all. A construction that had been reported left the
check, which is the shape #667 was about.

Both readings stand now. A guard states each of them; a clause is owed the order where that can be
read and the bound on the count where it cannot. `theOrderReadingDoesNotCostTheBoundOnTheCount`
holds that case at `UNKNOWN` rather than `UNREPRESENTABLE`.

## What it is held to

14 cases in `AComparisonCallStatesTheOrderItDecidesTest`, read off the verdicts: `PROVED` where the
guard settles the construction, `UNKNOWN` where it does not. Each spelling is paired with one step
past what it establishes, so a silence cannot pass for a discharge — the relations, both operand
orders, `Int` and `Decimal`, a comparison against something other than zero, the two date
operations, a clause written as a call meeting a guard written as an operator, and a cap stated of
every element through `List.all`.

`mvn clean test`: BUILD SUCCESS.

The example corpus goes from 71 diagnostics to 69, none added:

- `forecasting.sou:375` — the gap taken in the branch where `Decimal.compare(got, quota) >= 0` was
  false, so the subtraction is positive there. Reversing that comparison turns the diagnostic into
  **E2010**, a refutation: the difference is then negative wherever it is written.
- `quoting.sou:228` — a discount cap stated of every line as a comparison call. Held here by a pair:
  a cap at the clause's own bound proves the construction, and one that leaves room past it does not.

### DIFF
--- a/souther-compiler/src/main/java/souther/compiler/check/DischargeRules.java
+++ b/souther-compiler/src/main/java/souther/compiler/check/DischargeRules.java
@@ -362,6 +362,40 @@ final class DischargeRules {
             op("Int", "min"), op("Int", "max"), op("Int", "floorMod"), op("Int", "compare"),
             op("Decimal", "min"), op("Decimal", "max"));
 
+    /**
+     * The operations answering the order of their two arguments as the sign of a number, and the
+     * relation between the first argument and the second that a positive answer states. Zero states
+     * the equality, and a negative answer the relation the other way, so one row is the whole of what
+     * such an operation says.
+     *
+     * <p>Which relation a guard writing one states then follows from where the sign stands against
+     * zero and from nothing else, so the six relations and the two operand orders are one account
+     * rather than twelve rows ({@link Predicates#asOrderComparison}). The direction is not the same
+     * for all of them: {@code compare(a, b)} is positive where {@code a} is the greater, and
+     * {@code daysBetween(from, to)} counts forward from its first argument, so it is positive where
+     * the second one is.
+     */
+    private static final Map<ValueName, Ast.BinOp> ORDERS = Map.of(
+            op("Int", "compare"), Ast.BinOp.GT,
+            op("Decimal", "compare"), Ast.BinOp.GT,
+            op("Date", "daysBetween"), Ast.BinOp.LT);
+
+    /**
+     * The operations answering a number from two values of one type whose sign is not their order.
+     *
+     * <p>Arithmetic and a choice between two values are not orders at all: what {@code Int.subtract}
+     * answers has the sign of one and says how far apart they are as well, which is what a term
+     * already reads it as, and {@code min} answers one of the two rather than anything about the
+     * pair. {@code DateTime.minutesBetween} is the one that has to be told apart: it counts whole
+     * minutes, so a zero says the two are less than a minute apart rather than that they are equal,
+     * and a non-negative count does not say the second is not the earlier. Its strict signs do state
+     * the order, and a rule that holds for four of the six relations is not this one.
+     */
+    static final Set<ValueName> DECIDES_NO_ORDER = Set.of(
+            op("Int", "add"), op("Int", "subtract"), op("Int", "multiply"),
+            op("Int", "min"), op("Int", "max"), op("Int", "floorMod"),
+            op("DateTime", "minutesBetween"));
+
     /** Denial, which the analysis representation keeps as the call it is. */
     static final ValueName NOT = op("Bool", "not");
 
@@ -427,6 +461,10 @@ final class DischargeRules {
 
     static Set<ValueName> operatorForms() {
         return OPERATOR_CALLS.keySet();
+    }
+
+    static Set<ValueName> orderings() {
+        return ORDERS.keySet();
     }
 
     /** Those of them the building table has, by name, for the test that holds each to a construction
@@ -592,6 +630,17 @@ final class DischargeRules {
     /** The operator {@code operation} is, where it is one written as a function, else null. */
     static Ast.BinOp operator(ValueName operation) {
         return OPERATOR_CALLS.get(operation);
+    }
+
+    /** The relation a positive answer from {@code operation} states between its first argument and
+     * its second, or null where its sign is not an order. */
+    static Ast.BinOp orderStatedBy(ValueName operation) {
+        return ORDERS.get(operation);
+    }
+
+    /** Whether {@code operation} answers the order of its two arguments as a sign. */
+    static boolean decidesOrder(ValueName operation) {
+        return ORDERS.containsKey(operation);
     }
 
     /** Whether the check has a rule about what a call answers, rather than only about how to render
--- a/souther-compiler/src/main/java/souther/compiler/check/Predicates.java
+++ b/souther-compiler/src/main/java/souther/compiler/check/Predicates.java
@@ -228,9 +228,26 @@ final class Predicates {
         return obligations(inv, k, at, unnamed, true, decidesFalse);
     }
 
+    /**
+     * A clause is read as the comparison it states, and where an order call states one there are two
+     * readings of it: the order the call decides, and the bound on the sign that decides it. Which of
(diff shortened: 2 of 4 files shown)


kawasima:
Added in 682e0f00: the six relations as a table, each row read with the call on both sides of the zero — twelve spellings, against what that relation leaves of `paid - fee`.

| written | mirrored | verdict |
| --- | --- | --- |
| `compare(paid, fee) >= 0` | `0 <= compare(paid, fee)` | PROVED |
| `> 0` | `0 <` | PROVED |
| `== 0` | `0 ==` | PROVED |
| `/= 0` | `0 /=` | UNKNOWN |
| `<= 0` | `0 >=` | UNKNOWN |
| `< 0` | `0 >` | REFUTED_NOT_ALONE |

The last row is read the same way as the rest: a refuted construction is an error rather than a warning, and the verdict that says so is reached before it is raised, so the helper reads the verdict and lets the refusal stand.

The four standalone cases the table now covers (`>=`, `>`, `==`, and the one zero-on-the-right spelling) are gone from beside it. What is left is what the table does not say: one step past what a guard establishes, the same relation over `Int`, a comparison against something other than zero, the two date operations, a clause written as a call meeting a guard written as an operator, a cap stated of every element through `List.all`, and the reading that must not be lost.

`mvn clean test` BUILD SUCCESS.

On the typing nit: `ORDERS` still carries `Ast.BinOp`, and you are right that `asOrderComparison` reads anything that is not `GT` as second-greater. Left as it is per your "だけ" — say the word and it becomes a two-value enum of its own.

Issue #666, opened 2026-08-11
An example fixture can't project a field or call any stdlib function, so a derived collection fixture must be a hand-copied literal


kawasima:
Two everyday shapes for writing an example fixture out of an already-defined value fail, for two different reasons, and together they force every fixture that needs a computed collection to be spelled out as a full literal instead of derived from a named one.

## 1. Field projection of a fixture-evaluable value is not itself fixture-evaluable

```souther
module repro.issue3 exposing (Line, Ticket, allDone)

data Line = { done: Bool }
data Ticket = { lines: List<Line> }

let sample = Ticket { lines = [ Line { done = false }, Line { done = false } ] }

behavior allDone : (t: Ticket) -> Bool
let allDone (t) = List.all(l -> l.done, t.lines)

example allDone
    | "field projection of a fixture-evaluable value" :
        (Ticket { lines = sample.lines })
        -> false
```

```
-- EXAMPLE  E1903 ---------------------------issue3.sou:12:7

12|     | "field projection of a fixture-evaluable value" :
          ^

Input #1 of `allDone` could not be built: an example fixture must be a literal or a construction
```

`sample` on its own is a valid fixture (bare reference to a fixture-evaluable value), and `Ticket { ...sample }` (spread) is too. `sample.lines` — reading one field back out of the same value — is refused. A spread and a field projection read the same information out of a value already proven fixture-evaluable; only one of the two forms is admitted.

## 2. No stdlib function may be applied inside a fixture, closures or not

```souther
module repro.issue3c exposing (isTwo)

behavior isTwo : (n: Int) -> Bool
let isTwo (n) = n == 2

example isTwo
    | "a non-closure intrinsic applied inside a fixture" :
        (List.length([ 1, 2 ]))
        -> true
```

```
-- EXAMPLE  E1903 ---------------------------issue3c.sou:7:7

7|     | "a non-closure intrinsic applied inside a fixture" :
         ^

Input #1 of `isTwo` could not be built: `List.length` cannot be called from an example fixture: it has no executable helper method
```

[...]

Issue #664, opened 2026-08-11
E1002 reports a missing constructs entry one name at a time


kawasima:
E1002 (`constructs` under-declared) names exactly one missing type per compile, even when a behavior is missing several. Closing the gap on a behavior whose call graph reaches a few unit cases and a few records two or three helper-calls down means `compile → add one name → compile` repeated once per name, not once per behavior.

## Reproduction

```souther
module repro.issue1 exposing (Ticket, CookLine, ItemCookState, Pending, InProgress, Done, start)

data ItemCookState = Pending | InProgress | Done
data CookLine = { state: ItemCookState }
data Ticket = { lines: List<CookLine> }

let setState (next: ItemCookState, lines: List<CookLine>): List<CookLine> =
    List.map(l -> CookLine { state = next }, lines)

behavior start : (t: Ticket) -> Ticket
    constructs Ticket

let start (t) = Ticket { lines = setState(InProgress, t.lines) }
```

First compile:

```
Behavior `start` constructs `InProgress` but does not declare `constructs InProgress`.
```

Add `InProgress`, compile again:

```
Behavior `start` constructs `CookLine` but does not declare `constructs CookLine`.
```

Add `CookLine`, compile a third time: clean. Two round trips for one behavior that constructs two things beyond its own return type — `InProgress` is a bare unit case referenced two calls down inside `setState`, and `CookLine` is the record `setState` builds around it.

## Where this showed up

Modeling a ~35-behavior domain (a mobile-order system exercise, not part of this repo) in the `crm.sou`/`pipeline.sou` style — several small shared helpers behind state-transition behaviors, each behavior touching a handful of unit cases and records through them. Getting a clean build took 15+ separate `compile` invocations across roughly 15 behaviors, each invocation adding exactly one name.

## What would help

The checker already knows the full set it is comparing the declared `constructs` clause against — that is what lets it name the *next* missing one correctly each time. Reporting the whole missing set in one E1002 (one diagnostic per behavior, listing every name absent from its `constructs` clause) would turn a chain of round trips into a single fix per behavior.

kawasima:
Reading the checker, the missing names are all in hand when the first one is reported: `SpecChecker` walks `constructed.originated()` (a `LinkedHashMap`, so the walk is in the order the body builds them) and throws at the first entry the clause does not declare. So the set is there and the walk stops.

[...]

PR #679, opened 2026-08-12
A fixture may read a field off a value it can already build

kawasima:
A spread and a field projection read the same information out of a value a
row has already proven fixture-evaluable, and only the spread was admitted.
This states the closure instead: where `e` is fixture-evaluable and its type
declares `f`, `e.f` is one — recursively on its own conclusion, so a field
taken off a field is one too, and a newtype declares `value`, so what a
newtype wraps is reached by the same rule rather than by one of its own.

Refs #666.

## What a taken field supplies

A declaration says it, and the value found there is not read. That is the
answer no value could have given: a field declaring `List<AmountN>` supplies
one while it is empty, and an empty list has no element to name. This is
therefore a third source of evidence rather than a spelling of an existing
one — the syntax says what a construction supplies, a declaration says what
a taken field supplies, and a helper's answer says what a helper supplied.

Where the value a field is taken off is a helper's answer, what the helper
was declared to answer with is not read, so what says what it supplied is the
value it answered with, as it is wherever a helper stands. Which field may be
taken is still the declaration's, read from what that value is.

## Evidence survives a chain

A projection consumes evidence and produces it. A step that turned back
into an ordinary value lost what the next step is held by: the neutral form
reads a newtype as what it wraps, so an answer that has been through it no
longer says what it is.

The distinction is not whether a helper is written as the target. A value
whose body is a helper application is nameable, so the same answer arrives
through a name, and it arrived unadmitted — a projection reads its target as
a frame that states nothing, and there the answer is not held. What is held
now is the value the field was taken off of, at the position the whole chain
stands at, and the helper still runs once.

## Both readers, not the one the report named

Two readings walk the same fixture grammar — one builds the value the model
is given, one re-materialises what a row wrote — and each refused a
projection. Widening only the one the report named would have left the
language rule and the accepted set disagreeing again.

## A standard-library call says why it cannot stand in a fixture

Separate defect, in the same reading. The refusal named the compiler's
structure, that no helper method could be found, where the row needed the
rule and what to do instead. Two different things reached that one sentence,
so it could only say what was true of both. Told apart now, and the second
reason has a row of its own — every existing one landed on the library
branch.

## Measurements

- Every module's tests: 7194, none failing.
- The `example` adequacy of `souther-examples`, read with the CLI over all
  thirteen modules and compared against `develop` at c1c7c338: byte-identical.
  Widening the admitted set moved no measurement.

## Not in this change

Applying a standard-library function inside a fixture, and evaluating a
closure there. Neither is needed for the closure rule, and the second is
what the reported use case actually wants. Both are filed separately.


kawasima:
All three reproduced before being fixed, and each has a row that fails without
the fix.

**A JVM method reachable as a field.** `makeString("").isEmpty` compiled and
passed `true` into a `Bool` position. `readOrNull` is reflection over a no-arg
method and a fixture is not elaborated, so the value's Java surface stood where
the data's fields do. Which field may be taken is now asked of the answer's
nominal identity through `fieldTypeOf` — the same one-step rule the static walk
uses — before anything is read off the value. So the two evidences differ in
what says the *type*, and agree on what says the *field*.

**Optional to the same optional.** `a field declaring `AmountN?` states no value
of `AmountN`` — the position was unwrapped before being asked. Now the whole
position is asked first and the unwrap is the fallback, which is the shape you
wrote.

**Position-dependent representation.** `box.amount` at a `Choice = AmountN |
Missing` position gave `a number does not fit Choice`, while `AmountN(100)`
written out at the same position compiled — the control that makes it a defect
rather than a limit. `Declared` now goes through `NeutralForm.reread`, which
re-reads a newtype the way the position it reaches reads it and descends
collections and optionals, so representation follows assignability. Only the
widening direction is answered, because admission lets no other through: a case
reaches a sum that lists it, never the reverse.

Rows added: the java-method refusal, optional-to-optional, projected newtype
into a sum with its written-out control, and a `List<AmountN>` into a
`List<Choice>`. The sum rows use an identity behavior and name the case in the
expectation, so a wrong representation fails rather than passing on an
unexamined value.

Spec corrected: a helper's answer does have a declaration; what is not read is
the result the helper was declared to answer with.

**Measurements.** Every module's tests: 7091, none failing. The `example`
adequacy of `souther-examples` over all thirteen modules is byte-identical to
`develop` at 5049e95b.

One correction to the earlier claim in this PR's body. The first adequacy
comparison was taken against the merge base, and `develop` has since gained the
invariant work in #664/#668/#670, which silences several E2011 warnings in the
examples. Re-measured against current `develop`, so the comparison is of this
branch alone.

kawasima:
Both fixed in 8a3f788, and both reproduced first — the second one only after
the shape it needed turned out to be different from the one described.

**A unit case keeps a form its position decides.** The shape in the review is
not constructible: a bare unit case cannot be a field type at all, as a record
field, as a newtype's base, or as a list element — each is E2202, `Ready` has
no decoder. So `fieldTypeOf` never answers a bare unit-case type and that route
is closed.

The defect is real one step out, where a field type can hold it: a field
declaring a unit-*only* sum, projected into a sum that also lists a product.

```
data State = Ready | Pending          -- an enumeration: a case is a bare name
data Wider = Ready | Pending | Other  -- lists a product: a case carries its tag
data Box = { state: State }
```

`box.state` at a `Wider` position gave `a string does not fit Wider`, while
`Ready` written out at the same position compiled. `reread` now moves a unit
case's form as it moves a newtype's, and both rows are in the suite with their
written-out controls.

**A `let` in the target no longer skips admission.** Confirmed exactly as
described, with the block form:

```
let basket = {
    let inner = Basket { total = AmountN(100) }
    inner
}
```

`basket.total` at an `Int` position compiled. `declaredTypeOf` now carries a
binding environment of its own and enters a `let` as the reading that takes the
value does, so the two agree about what is in scope. It still runs nothing.
Fixed the way you preferred — in the walk rather than by admitting twice at the
end. Both the refusal and its positive control are in the suite; without the
control, an implementation that refused every let-bound target would have
passed.

**PR body corrected.** The stale `there is no declaration of this module's to
read` is gone from the body as well as the spec.

**Measurements.** Every module's tests: 7095, none failing. The `example`
adequacy of `souther-examples` over all thirteen modules is byte-identical to
`develop` at 5049e95b.

kawasima:
`develop` is merged (c1c7c33), and the identity blocker is fixed in f32fc23 at
both sites.

**The answer's type.** Read off the class it was generated as — a binary name
is a qualified type name — instead of resolving its simple name here. That is
the comparison this reader already makes where it asks whether a value
represents a case; the projection path had reduced the identity to a spelling
and resolved it in the reading module.

**The case a bare name stands for.** Asked of the type the value stands at,
through its leaf cases, rather than of this module's scope. The newtype
overload took a resolved name for exactly this reason and the unit case beside
it did not, which is the same slip one line apart.

**The rows carry a same-named declaration in the reading module**, so they fail
rather than merely answering nothing. Confirmed by reverting both fixes and
re-running: `aForeignAnswerIsReadAsItsOwnModulesType` gets `declares no field
'amount'` — the caller's `Remote { other: Int }` answered for `demo.lib.Remote`
— and `aForeignCaseIsMovedAsItsOwnSumsCase` gets `a string does not fit Wider`.
The third refuses either way, but for the wrong reason without the fix
(reflection's `has no field` rather than the declaration's `declares no
field`), and the assertion holds the reason.

Two notes.

`assertedLive` resolves a simple name in this module's scope the same way, for
the same purpose — re-materialising a helper's answer. It predates this branch
and is on the expected-value side, so I have left it alone rather than widen
the diff. Say the word and I will fix it here, or file it.

The `tagged` observation is right and I have not touched it: a case belonging
to two sums with different discriminator keys picks whichever visible sum
lists it first, in the written path as much as the projected one, so the two
agree — which is all this change owes. Worth its own issue.

**Measurements.** Every module's tests: 7193, none failing. The `example`
adequacy of `souther-examples` over all thirteen modules is byte-identical to
`develop` at c1c7c338.

kawasima:
All four done.

**`assertedLive` fixed here** (04c14a0), and you are right that it belongs in
this PR rather than beside it. The row proves it: an expected value reaching a
field goes through that re-materialisation, so the closure this change states
did not hold on the expected side at all. Written first and confirmed failing —
`The expected value of `same` could not be built: `Remote` declares no field
`amount``, the caller's `Remote { other: Int }` answering for `demo.lib.Remote`
— then green. Both readings now ask the same question of the same thing: the
type a value is, read off the class it was generated as.

**The expected-side regression row is in**, paired with the input-side one and
carrying the same same-named declaration in the reading module, so an
implementation that resolves the spelling is led to the wrong declaration
rather than merely answering nothing.

**`tagged` filed as #683**, with the reasoning for keeping it out of here: every
path shares the rule, so a projected value and the same value written out agree
on whatever it says, which is all this change owes. The issue also notes what a
fix would have to decide — the position knows which sum it is — and that it is
worth checking whether two sums may name one case with different tags at all
before treating the search as a choice.

**PR body brought up to date**: 7194 tests, adequacy compared against `develop`
at c1c7c338.

PR #691, opened 2026-08-12
Read a name given a computed value as what it was given

### LINKED ISSUE #676: A guard between two names holding computed values does not reach the difference of what they hold
A guard comparing two names holding computed values does not reach the construction over the
difference of what they hold. Written inline the same two expressions settle it, and so does writing
the guard on the numbers instead of on the values that wrap them.

## Repro

```souther
module repro

data Yen = Decimal
    invariant value >= 0.0m

data Rem = { base: Yen, commute: Yen }
data TooMuch

let grossOf (r: Rem, uplift: Yen): Yen = r.base + uplift
let takenOf (r: Rem, si: Yen): Yen = r.commute + si

behavior boundFirst : (r: Rem, uplift: Yen, si: Yen) -> Yen | TooMuch
    constructs Yen, TooMuch

let boundFirst (r, uplift, si) = {
    let gross = grossOf(r, uplift)
    let taken = takenOf(r, si)
    guard gross >= taken else TooMuch
    Yen(gross.value - taken.value)          // E2011
}
```

```
-- INVARIANT (WARNING)  E2011 -----------------repro.sou:19:5

19|     Yen(gross.value - taken.value)
        ^

Constructing `Yen` here may violate its invariant: the guards do not establish it.
```

## Where the line is

Measured on `develop` (5049e95b), one element moved at a time. Every row states the same fact and
builds the same difference.

| how the two values are written | how the guard compares them | E2011 |
| --- | --- | --- |
| the calls inline, twice | as newtypes | silent |
| a `let` holding each call | as newtypes | **raised** |
| a `let` holding the arithmetic those calls do | as newtypes | **raised** |
| a `let` holding each call | on their `.value`s | silent |
| the arithmetic over plain `Decimal`s, in a `let` | as numbers | silent |
| a `let` holding a parameter | as newtypes | silent |
| the parameters themselves | as newtypes | silent |

So it is not the `let`, not the newtype comparison, and not the arithmetic on its own. It is a name
holding a *computed* value, compared as the newtype it is: a name given a location is that location
and the comparison reaches it, and a name given a term is a term the newtype-level comparison does
not relate to the number read out of it.

The inline row is a discharge and not a construction dropped from the check — one step past what the
guard establishes (`- 1.0m`) is E2011 there.

## Why it is worth closing

This is the shape a "take the deductions off the gross" behavior has once its two totals are given
names, which is what an author does as soon as either total is used twice. It is in the example
corpus: `hr/src/main/souther/payroll.sou:195` (souther-examples), where `taxableAmount` binds `gross`
and `subtracted`, guards one against the other, and subtracts them — a warning that survived #669 and
#674 and cannot be cleared by the remedies E2011 names. The guard is already written, and there is no
relation between inputs to reify: what the author has to discover is that moving the guard onto
`.value` makes it discharge.

## Related

#508 was the second step of a chain — a guard on a computed value not composing with an input type's
invariant — and was closed by #525. This is the first step, over the newtype rather than over the
number. #409, #667 and #674 are the same boundary from other sides: which spelling a value is written
in decides what the check reads of it.

kawasima:
Closes #676.

## What was wrong

Whether a newtype's `.value` is the value it wraps was asked of how the target was written:

```java
case Core.FieldAccess fa when rootBinding(fa.target()) == null && !Location.isStep(...)
```

`rootBinding(...) == null` is true of a call and false of a name given one. So a name holding a computed total was read through to the arithmetic behind it, and a field taken off that name was an atom of its own — one value under two readings, with nothing between them. Measured on the issue's repro, cut to one name:

```
[bindLet] binder=gross denotes=Term[key=@5] affine=uplift + 1.0
[guard]   GE la=uplift + 1.0        ra=si          <- what the guard put in the domain
[clause]  GE la=@5 - si  >= 0                      <- what the construction asks
```

So a guard an author wrote about the very values being built settled nothing, while the same two expressions written out where they stand settled it. Which reading a body landed in was decided by whether either total had been given a name — that is, by whether it was used twice.

The spec already says a name and the expression it was given are read alike (§invariant-discharge-terms). It is `rootBinding` that was answering a different question.

## What it does

Asks what the target denotes instead. `Location.of` already answers this: a binding that is an alias for a place is that place, and one given a computed value is none. A name given a call is then no more a place than the call it was given, so the field off it reads through exactly as the field off the call does, and `gross`, `gross.value`, `grossOf(r, uplift)` and `grossOf(r, uplift).value` all reach one form.

Two things follow from there rather than being arranged:

- **A guard and a construction meet whichever pair they are written in.** Either may name a total or write it out, and either may compare the newtypes or their `.value`s. There is no spelling they have to agree on, which is what an author cannot be asked to predict.
- **A binding introduces no numeric value of its own.** `bindLet` gave a name's atom the interval bounds of the form it was given, which was the only thing standing between the two readings, and an interval is not a relation — it never carried them. With one reading nothing gives an atom a form, so `Known#assigning` and `NumericDomain#assign` go with it. Measured before taking it out: the suite and the whole corpus answer the same with them and without them, down to the site of every diagnostic.

## What it was held to

`ANamedComputedValueIsReadAsWhatItWasGivenTest`, sixteen routes: the parameters, a name given a parameter, a name given a call, a name given the arithmetic that call does, the calls written out twice, one side a name and the other a place, the construction written over one initializer and over the other, the guard itself written over one of each, and a name standing inside either operand — over the newtypes and over their `.value`s.

Every route answers twice: the difference its guard bounds is `PROVED` and silent, and one step past it is `UNKNOWN` and reported. The verdicts are what these read, so a route the check stopped reading cannot pass as one its guard settled.

`mvn test`: 4248 in souther-compiler, green across every module.

## What it changes in the corpus

Every `.sou` in souther-examples, compiled with the develop CLI and with this one — including `issuetracker` and `account`, which are outside the Maven reactor:

| | E2011 |
| --- | --- |
| develop | 64 |
| this | 63 |

The one that goes is `hr/payroll.sou:195`, the site the issue names, and every other diagnostic is identical to develop's. It is the only one of its shape there: `payroll.sou:194` is the corpus's only guard comparing two bare names. What is reached is wider than that one site — the reading changed at 28 guards across the corpus and the answer at one.

## History

The first two commits state a comparison in both spellings rather than reading in one. That leaves the pairs the two spellings do not cover, which is where the review put it; the third takes the question back to what it was asking.

### DIFF
--- a/souther-compiler/src/main/java/souther/compiler/check/InvariantChecker.java
+++ b/souther-compiler/src/main/java/souther/compiler/check/InvariantChecker.java
@@ -5,7 +5,6 @@ import souther.compiler.check.Combinators.Handed;
 import souther.compiler.numeric.Granularity;
 import souther.compiler.numeric.Count;
 import souther.compiler.numeric.NumericDomain;
-import souther.compiler.numeric.NumericDomain.LinearForm;
 import souther.compiler.core.Core;
 import souther.compiler.diag.CompileException;
 import souther.compiler.diag.Diagnostic;
@@ -1592,31 +1591,23 @@ public final class InvariantChecker {
      * recorded under that denotation and not under the binding. Recording it under the binding is
      * what made a named subexpression a term of its own, answering differently from the very
      * expression it was given.
+     *
+     * <p>Nothing is recorded of the name itself. A name is read as the expression it was given —
+     * {@link Terms#affine} reads through it, and reads through a field taken off it the same way —
+     * so there is no second reading for a fact about the name to be needed by. Recording one meant
+     * giving the name's own atom the bounds of the form it was given, which is what a guard read one
+     * way and a construction read the other had between them, and a bound is not a relation: it left
+     * the guard settling nothing about the construction (#676).
      */
     private Entered bindLet(Core.LetIn li, Known k, Denotations at) {
         // Entering a binding the walk is already inside is not a second binding of it. A branch is
         // read from where its conditional stood, which is inside these, over a tree that still holds
-        // them; running the transition again would assign the name its form a second time, and an
-        // assignment forgets what was known of what it assigns to — including what the branch had
-        // just established.
+        // them.
         if (at.valueOf(li.binder().id()) == li.value()) {
             return new Entered(k, at);
         }
         Denotes what = terms.denotationOf(li.value(), at, k);
-        Known out = k;
-        // A binding that denotes what it was given is an alias and introduces no value, so there is
-        // nothing to record of it: what holds of what it names already holds. Recording it anyway
-        // assigns that name its own form, and an assignment drops what was known of what it assigns
-        // to — the bound on it would be lost to the copy. A location is always this; a term is where
-        // the form is that term's own atom.
-        if (what instanceof Denotes.Term term && terms.affineScalarBase(li.value().type()) != null) {
-            LinearForm vf = terms.affineOf(li.value(), at, k);
-            if (vf != null && !vf.equals(LinearForm.atom(term.key()))) {
-                out = out.assigning(term.key(), vf,
-                        terms.kindsOf(vf, term.key(), li.value().type()));
-            }
-        }
-        return new Entered(out, at.binding(li.binder().id(), li.value(), what));
+        return new Entered(k, at.binding(li.binder().id(), li.value(), what));
     }
 
     /** Where {@code site}'s conditional stands: {@code k} and {@code at} with every binder it is
--- a/souther-compiler/src/main/java/souther/compiler/check/Known.java
+++ b/souther-compiler/src/main/java/souther/compiler/check/Known.java
@@ -40,10 +40,6 @@ record Known(NumericDomain numbers, PredicateFacts facts, List<Quantified> quant
         Unguarded taking(String key, boolean positive) {
             return new Unguarded(numbers, facts.assume(key, positive));
         }
-
-        Unguarded assigning(String atom, LinearForm f, Map<String, Granularity> kinds) {
-            return new Unguarded(numbers.assign(atom, f, kinds), facts);
-        }
     }
 
     /** How far a fact reaches: a value's type and a name's binding say something wherever that value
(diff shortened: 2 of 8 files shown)


kawasima:
Both are real, and the second is the one that decides the shape. Two commits.

## The four corners were the same defect one step over

`b9ea226a`. Rather than record the comparison in more spellings, the reading is now one. Whether a `.value` read is the value it wraps was asked of how the target was spelled:

```java
case Core.FieldAccess fa when rootBinding(fa.target()) == null && !Location.isStep(...)
```

`rootBinding(...) == null` is true of a call and false of a name given one — so `gross` was read through to the arithmetic behind it and `gross.value` was an atom of its own. Asked of what the target *denotes* instead, a name given a computed value is no more a place than the call it was given, and the field off it reads through exactly as the field off the call does. `Terms#namedForm` and the two-spelling statement in `assumeCond` are gone: with one reading there is nothing to state twice.

The five routes you named are in the test, and all five fail on the two-spelling reading:

| | two spellings | one reading |
| --- | --- | --- |
| `guard gross >= taken` / `Net(gross.value - takenOf(r, si).value)` | E2011 | PROVED |
| `guard gross >= taken` / `Net(grossOf(r, uplift).value - taken.value)` | E2011 | PROVED |
| `guard gross >= takenOf(r, si)` / `Net(gross.value - taken.value)` | E2011 | PROVED |
| `guard gross + si >= taken` / `Net(gross.value + si.value - taken.value)` | E2011 | PROVED |
| `guard gross >= taken + si` / `Net(gross.value - taken.value - si.value)` | E2011 | PROVED |

Each is held to one step past its guard as before, so none of them passes by having left the check. Sixteen routes now, and the nested ones are in rather than deferred — the Javadoc sentence you quoted was wrong about them, and the answer was to make it true rather than to write it down.

## And the thing they were meeting through is now unused

`2793124a`. `bindLet` gave a name's atom the interval bounds of the form it was given, which is what a guard read one way and a construction read the other had between them. An interval is not a relation, so it never carried them — that is what the measurement in the issue thread showed. With one reading nothing gives an atom a form at all, so `Known#assigning` and `NumericDomain#assign` go with it. Measured before taking it out: the suite and the whole corpus answer the same with it and without it, down to the site of every diagnostic.

## Unchanged

`mvn test` green across every module (4248 in souther-compiler). Over every `.sou` in souther-examples, still 64 E2011 to 63 with `hr/payroll.sou:195` the one that goes, and every other diagnostic identical to develop's.

Issue #692, opened 2026-08-12
A fixture refuses a polymorphic helper even when the row's arguments determine its type variables


kawasima:
Split out of #680, which found this behind a representation-dependent refusal.

## What is true today

    example echoBool
      | (List.isEmpty([ 1 ])) -> false

    `List.isEmpty` parameter `xs` is List<'a>; what it holds is decided by each
    call, and a fixture is built before a call can decide it

The row wrote `[ 1 ]`. Nothing about `'a` is open at that point.

`FixtureReader.answered` reads `helper.def().params()` and refuses where any
parameter type mentions a `Type.Var`. The applicability decision is therefore
made from the declaration alone, without using the row's argument types to
instantiate that variable.

## The question

Whether a fixture settles a helper's type variables from what the row wrote, and
applies the helper monomorphically, or asks only what the declaration says.

This is not about admitting polymorphic values into fixtures. A fixture holds no
polymorphic value either way: `List.isEmpty([ 1 ])` resolves `'a := Int` from one
written argument and applies one monomorphic instance. What is being asked is
whether the applicability check may read the call.

## Why it is separate from #680

They are independent. #680 is about `String.length` being refused for having no
`$Fns` method while `String.isEmpty` is admitted for having one — both are
monomorphic, so this issue does not reach them. This one is about `List.isEmpty`
and `List.length`, which stay refused after #680 either way.

They were entangled only because `List.length` reached #680's refusal first, and
that hid this one.

## What would need deciding

Which call-site constraints a fixture may use to instantiate a variable, and what
happens where they do not determine one uniquely. `List.isEmpty([])` is the
simple open case; a variable appearing only in the result raises the same
question from the other side. A refusal may still be right there, but it should
say which variable could not be settled rather than restate the declaration's
polymorphic shape.

kawasima:
Fixed in #729 (develop 経由)。

適用可否は宣言ではなく call の問いになりました。行が書いた引数が宣言の型変数を全部決めれば、その1つの instance で適用します。決まらない変数が残れば、その変数を名指して拒否します（`List.isEmpty([ ])`）。行が立っている位置の型は読みません。

証拠として読むのは、宣言が既に言っていること（構築・フィールド・名前・`let` の束縛）と、言語のリテラルの読みだけです。helper の適用は読みません。

[...]

Issue #680, opened 2026-08-12
Fixture callability unnecessarily depends on a function's compilation form


kawasima:
Split out of #666. The refusal this was opened about is not the rule it states,
and measuring the mechanism changed what the issue is about.

## What is true today

A fixture applies a standard-library function, and has for as long as
`exampleHelpers` has existed:

    example echoInt
      | (Int.abs(0 - 2)) -> 2        compiles
      | (Int.clamp(0, 5, 9)) -> 5    compiles

    example echoBool
      | (Bool.not(false)) -> true          compiles
      | (String.isEmpty("")) -> true       compiles

Writing the wrong expectation answers `E1905: It is: 2 But it needs to be: 3`,
so the value is one the row computed by running the function.

What is refused is a smaller set, and it is refused for two unrelated reasons:

    List.length([ 1, 2 ])   `List.length` cannot be called from an example
                            fixture: a standard-library function is not one a
                            fixture may apply

    List.isEmpty([ 1 ])     `List.isEmpty` parameter `xs` is List<'a>; what it
                            holds is decided by each call, and a fixture is
                            built before a call can decide it

The second row never reaches the first refusal. It is already being applied as a
helper when the parameter check stops it.

## Where the boundary actually is

`HelperInliner.exampleHelpers` emits a method for every helper a row applies,
and does not read who declared it — the module's own, one an import publishes,
and a prelude one all pass. The filter is the body's form:

    body instanceof Ast.FnBody.Written
      && !params.isEmpty()
      && !Elaborator.producesFunction(body)

`injectedExampleHelpers` renames what the module did not declare to the name it
reaches it by, `Shapes.Prepared` puts it among `takenOn`, and the backend emits
it on `$Fns`. `FixtureReader` looks a call up in imported + own + taken-on
definitions, so a prelude helper is found there like any other.

An intrinsic has no `Written` body for `exampleHelpers` to emit a `$Fns` method
from. `BodyGen` lowers it at its call site through `Intrinsics.emit`. For 93 of
the 96 intrinsic keys that lowering is itself a call to an existing runtime or
JDK method — 73 an `invokestatic` into `souther.runtime`, 13 a JDK instance
method, 7 the remaining forms — but no fixture-visible `$Fns` entry is written
for any of them.

That is the difference a fixture currently observes: not whether executable code
exists, but whether it has been materialised as a helper method `HelperInvoker`
knows how to reach. `String.length` is refused and `String.isEmpty` is admitted,
[...]

PR #694, opened 2026-08-12
Apply a kernel in a fixture through a method emitted for the row

kawasima:
Addresses #680.

## What was wrong

A fixture applied a standard-library function and was told it may not. Both of
these compile today and always have:

    example echoInt
      | (Int.abs(0 - 2)) -> 2
      | (Int.clamp(0, 5, 9)) -> 5

while `String.length("abc")` in the same position was refused with "a
standard-library function is not one a fixture may apply. Compute the value in
a `let` helper and apply that instead". Both are monomorphic library functions
taking no closure. What decided was whether `HelperInliner.exampleHelpers` had
a `Written` body to emit a `$Fns` method from — a fact about how the backend
lowers a function, stated to the author as a rule about the standard library.

`Prelude.isLibraryFunction` at the head of `FixtureReader.noMethod` stated the
same rule and enforced nothing: a library function with a method is found as an
applied helper before that branch is reached.

## What this does

Emits the method. What a row applies is materialised as a `$Fns` method whose
body is the kernel call applied to the declaration's own parameters, so
`BodyGen` lowers the one call in it exactly as it lowers that call anywhere
else. The descriptors, the argument permutation, the boxing and the `l2i`
narrowing are the ones already written; nothing reaches an intrinsic target a
second way and `HelperInvoker` is unchanged.

It covers all 96 intrinsic keys, not the 93 with a row in `Intrinsics.TABLE`.
`int.divide`, `int.truncatingRemainder` and `decimal.divide` are written out as
inline arithmetic with a branch on a zero divisor, and that branch is written
inside the emitted method like any other lowering.

The method is emitted under `$intrinsic.<name>`, which no source can spell.
Under the kernel's own name it would be a helper the inliner's table reaches,
and a body calling `String.length` would expand into a body whose one call is
`String.length`.

`FixtureReader` now asks the library for a declaration it did not take on to
emit, so which function a fixture may apply is decided by the fixture's own
rules reading that declaration — the arity it takes, whether a call has to
settle one of its types — rather than by a table's silence.

## What still refuses, and why

`List.length([ 1, 2 ])` does not start compiling. It is declared
`(xs: List<'a>): Int`, so it is refused by the parameter check that already
refuses `List.isEmpty([ 1 ])`:

    `List.length` parameter `xs` is List<'a>; what it holds is decided by each
    call, and a fixture is built before a call can decide it

That is a reason about fixtures rather than about bytecode. Whether a fixture
should settle a type variable from the row's own arguments is #692.

Of the 96 intrinsic keys, 61 are monomorphic and 35 mention type variables.

## Tests

- `ExampleCallsHelperTest.anIntrinsicIsRefusedByThatReason` stated the removed
  rule and is replaced by four: a row applies an intrinsic, the value it
  answered is what the row is held to (`E1905`), a kernel the backend writes
  out in place is applied, and a polymorphic intrinsic is refused for its type
  variable and not for how it is compiled.
- `ARowAppliesTheHelperItsCallReachesTest` and `CompileFixtureNamesAValueTest`
  each had a test asserting the removed rule. Both are rewritten to assert what
  they were guarding — that the reach name is what a bare-written intrinsic is
  applied under, and that a value whose body applies an intrinsic is a fixture
  — each with a mismatch case so neither passes without the function running.

Full reactor: 7209 tests, 0 failures.

## Documents

`specification.adoc` and ADR-0077 both stated the removed rule. The spec now
says what the two refusals are; the ADR paragraph says what it used to read and
why that rule was never held, since the decision it records is otherwise
unchanged. Per `docs/adr/README.md` this is not a new ADR: the decision that a
helper with no method is refused by that reason stands, and what changed is
which functions have one.

## Not verified

souther-examples was not built. Its build resolves the compiler from `~/.m2`,
and this worktree installs into an isolated `.m2repo`, so running it means
writing to the shared repository. No example can contain an intrinsic applied
in a fixture today — such a row does not compile — and this change only widens
what is admitted, so nothing there should move.


kawasima:
All four addressed in e57e81ed. The two blocking ones were right, and the
second was a real disagreement rather than only future-proofing.

**1 — the synthetic name reached identity, not just lookup.** Confirmed by
reverting the split and running the new test:

    `$intrinsic.String.slice` did not produce a value: String.slice
    toExclusive out of range: 99

`HelperInvoker.invoke` now takes `(helper, emittedAs, args)`: the lookup takes
`emittedAs`, and `running()` and every report take `helper`. Pinned by
`aKernelThatDoesNotAnswerIsNamedAsTheKernel`, which asserts the report names
`String.slice` and holds no `$intrinsic`. `String.slice` past the end throws
inside `souther.runtime.Strings`, so the test goes through the same
`InvocationTargetException` path a timeout's `running()` reads.

**2 — the two policies now ask one reading.** Measured first: no kernel is
declared with monomorphic parameters and a polymorphic result, so the result
check refused nothing. It was capability leaking into policy exactly as you
describe, and it is gone. Both sides call
`HelperInliner.decidedByEachCall(Type)`, over parameter types only —
`HelperInliner.intrinsicCallsIn` to decide what a method is emitted for, and
`FixtureReader.answered` to decide what a fixture may apply.

Renaming it out of `settled` turned out to matter for its own reason:
`Elaborator.settled` already answers a different question in the same package
(whether a declared type is done saying what a value is, over `Type.Open` and
`Type.Nothing`). Two predicates named alike over `Type` is the next place the
same confusion would have grown. `fixtureIntrinsicName` is
`intrinsicWrapperName` and its javadoc says it is where a method went, not what
a function is.

**3 — the zero-divisor arm.** `theZeroDivisorBranchIsWrittenInsideTheEmittedMethodToo`:

    Int.divide(7, 0)  ->  `Int.divide` answered with a `DivisionByZero`, which
                          is not a value of `Int`

which is the branch running inside the wrapper. One kernel, both arms, as you
suggested.

**4 — the two over-generalisations.** The `helperDef` comment now says a
library function *this module did not take on to emit* is not in the table; one
it did take on is there like any other. The specification counted two functions
and there are three — a parameter that declares no type is refused as well, and
already was before this PR. It now names all three.

**Tests removed.** Three asserted the same fact, that an applied kernel is
evaluated rather than assumed:
`ARowAppliesTheHelperItsCallReachesTest.aBareIntrinsicApplicationIsEvaluatedAndNotAssumed`
and `CompileFixtureNamesAValueTest.theValueTheIntrinsicAnsweredIsWhatTheRowIsHeldTo`
are gone, with the four imports they were the last reader of;
`ExampleCallsHelperTest.aRowApplyingAnIntrinsicStillCatchesAMismatch` keeps it.
The rows that stay assert the correct answer, so a kernel that did not run or
ran wrong is still E1903 or E1905.

Full reactor: 7209 tests, 0 failures.

The invariant you named holds now on the kernel path: what a fixture admits is
decided by the fixture's rules, and a call those rules admit is one there is a
method for. It does not hold for a helper whose body produces a function — that
one is still admitted by `answered` and refused by the absence of a method,
which is ADR-0077's own rule and untouched here. Worth its own issue rather
than widening this PR, if you want the whole policy in one place.

Issue #698, opened 2026-08-12
Untyped fixture representation assumes every sum that lists a case agrees on its discriminator


kawasima:
Where a fixture position has no declared type, `NeutralForm` cannot ask the position
what the case is read under, and the sums that list the case answer instead
(`taggedWithoutADeclaredType`). That is right only while every sum that lists a case
agrees on what it reads the case under.

Today they cannot disagree. Every discriminator is derived by `Deriver` — keyed
`"type"`, tagged with the case's own declared name — there is no surface syntax for a
written one, and a sum's cases are declared in its own module, so one case has one
tag however many sums list it. #683 measured that.

A custom discriminator would end it: two sums could then read one case under
different keys or different tags, and a value at an untyped position would have no
ground for choosing between them.

## What is already known

The fallback has one producer: a behavior answering with several types, where
`FixtureShape.ofWholeAnswer` returns null for `BoundaryOutput.Cases` and admission,
not a decoder, picks the case (`FixtureReader.builtStandIn`). Every other position a
fixture is written at carries a declared type.

So this is a constraint on a feature that does not exist yet rather than a defect:
whoever specifies custom discriminators has to say what an untyped position reads,
and the change lands in one place.

kawasima:
Fixed by #724, merged into develop.

Measured, this was not a constraint on a future custom discriminator: the two searches over `symbols.visible()` already moved a value's form with a declaration nothing reads it through. Against the pre-change compiler, adding `data Revision = Filed | Note` beside an enumeration listing `Filed` moved a `Filed` standing where nothing reads it from `"Filed"` to `{"type":"Filed"}`.

The place a value stands now answers for its form and nothing else: `Position` is `At(Type)` or `Unread`, and `Unread` asks for nothing beside the case, leaving the case's own form. `taggedWithoutADeclaredType` and `onlyEnumerationsList` are gone, and `readsABareName` no longer takes the case.

Issue #697, opened 2026-08-12
Determine whether codec-target selection is duplicated across three type walks


kawasima:
Three walks read a type and answer something close to "which nominal codec does a
value at this position go through":

- `derive/CodecShape.of(Type, ...)` — a data field, for the codec derived from it
- `check/SignatureBoundary` + `BoundaryInput` — a behavior's parameters and answer
- `examples/NeutralForm` — `open` / `elementOf` / `entryTypes`, for the form a
  fixture is written in

Both of the first two stop at the name — `CodecShape.Named(TypeName)`,
`BoundaryInput.Nominal(TypeName)` — and neither says what that named type's
representation is. That part is decided once, by `Deriver`, and carried on the
declaration (`Ast.SumData.decoder()`).

## What this is not yet

Not a report that the three duplicate each other. Being close is not the same as
answering one question, and each of the first two says in its own doc that reading
and refusing are one walk: a shape exists because the walk that would have refused
it did not. So a shared classifier would have to carry the refusals too, or leave
them behind and let the two sides drift — which is the failure those docs describe.

## What to find out

For each walk, write down:

| | input | output | stop condition | refusal condition | diagnostic ownership |
|---|---|---|---|---|---|

Extract only the rows that agree. Where the refusal or the diagnostic differs, the
walks answer different questions however alike the traversal looks.

## Why now

#683 found `NeutralForm` reconstructing a decision `Deriver` had already made, and
fixed it by reading the derived decoder instead. That was the part where an
authoritative answer demonstrably existed. This is the part where it is not yet known
whether one does.

kawasima:
Closed by #719 (merged into develop as 09e29796).

Issue #696, opened 2026-08-12
Compiler-generated `Ast.Apply` drops resolved callee identity and forces FixtureReader to resolve source spelling again


kawasima:
`FixtureTemplate` builds the candidate values a generated row is made of, and it
builds a newtype construction from a spelling:

```java
new Ast.Apply(type.name(), List.of(inner.value()), NOWHERE, NO_SOURCE)   // partition/FixtureTemplate.java:164
```

That constructor writes the name with nothing behind it:

```java
public Apply(String fn, List<Expr> args, SourcePos pos, Region region) {
    this(Var.respelled(fn, null, null, pos, null), args, ConstructionOrigin.own(), pos, region);
}
```

`type` is a resolved `TypeName` at that line. What reaches `FixtureReader` is the
spelling alone.

## What it costs downstream

`FixtureReader` has seven readings of the form `neutral.isNewtype(c.written())` and
`symbols.resolve(c.written())` / `symbols.resolveCase(c.written())`. Both of the
things they lean on say of themselves that this is not what they are for:

- `Ast.Apply#written`: "**Never a lookup key.** … Every question of the form 'which
  declaration is this' asks `reaches()`."
- `Symbols#resolve`: "a name is resolved in the module that wrote it, by that
  module's own `Resolve` pass, so this never has to answer for a spelling written
  somewhere else."

## Measured

Converting those seven readers to the carried identity (`ValueName.OfType`, which
`Resolve` settles and `NewtypeDesugar` already reads) with no upstream change fails
40+ tests across generated rows, partition and candidate proposal. The identity is
absent exactly on the nodes the generator built, so the readers cannot be moved off
the spelling until the generator carries it.

## What to decide

Where the identity is put, which is a question about the API and not about
`FixtureReader`: a constructor that takes the resolved name, a `Var` the caller
builds, or a separate path for compiler-generated expressions. The three place the
responsibility differently.

## Not the same defect as #683

#683 looked like this and was not. There the position's declared sum was already an
authoritative fact, so the re-resolution could be deleted rather than fed: the
question changed from "which sums list this case" to "what does the type at this
position read it under", and no name is resolved at all. Here the rediscovery is
forced, because the identity was dropped upstream.

kawasima:
Fixed by #701 (merged into develop). The generator writes both halves of a fixture — the row's text and the tree a decoder reads — from one `TypeReachName`, which `Symbols.reach` answers for the module the row is offered to.

[...]

Issue #683, opened 2026-08-12
NeutralForm chooses case representation from global sum membership instead of the position's declared type


kawasima:
Reported as a case belonging to two sums taking whichever discriminator the first
visible sum names. That was measured and does not happen. What the search does is
answer a question about the case where the question is about the position.

## What was measured

The reported ambiguity is unreachable. Every `Ast.Discriminate` comes from
`Deriver.deriveSum` — `AstBuilder` builds every `SumData` with an empty decoder and
encoder, and there is no surface syntax for a written one — and derivation keys it
`"type"` and tags each case with its own declared name. A sum's cases are declared in
its own module (`ACaseIsDeclaredInAnotherModule`), so two sums that list one case are
in one module and the case has one name.

Compiling `data Reason = High | LowRole` beside `data Cause = High | LowRole` and
reading `High` through both:

```
Reason <- Ok[High[threshold=Amount[value=5]]]
Cause  <- Ok[High[threshold=Amount[value=5]]]
Reason -> {threshold=5, type=High}
Cause  -> {threshold=5, type=High}
```

Whichever sum the walk stopped at, it wrote the same thing.

## What the defect is

The discriminator a value carries is decided by the type the position declares
(spec §encode-law): a case's own decoder reads none and its sum's reads one, so
`Approved { id = 1 }` is `{"id":1}` where `Approved` is declared and
`{"id":1,"type":"Approved"}` where `Approved | Rejected` is.

`Ast.SumData.decoder()` is where that is decided, once, by `Deriver`. `CodecGen`
lowers it to bytecode. `NeutralForm` did not read it: it walked `symbols.visible()`
for any sum that lists the case, which is a question about the case's membership and
not about where the value stands. So a product case carried the tag at every
position, including its own — invisible until now because both directions of a row
did it alike and a case's own decoder ignores the extra key.

The reachable consequence was elsewhere. The unit-case path reached the search
through a spelled name (`symbols.resolve(written)`), which answers nothing for a case
the module did not import by name, and the envelope came out empty: the value was
refused at construction with `type: is required`.

## What changed

`NeutralForm.tagged` takes the declared type at the position and reads the key and
the tag off that type's own derived decoder. Nothing is decided a second time here.
Where the position has no declared type — a behavior answering with several types,
where admission and not a decoder picks the case — the old walk stays as an explicit
fallback, on its own path and named as one.

Measured over souther-examples: `hr` generates 4 more rows (56 -> 60) and 4
[...]

PR #699, opened 2026-08-12
Read a case's discriminator off the type at the position it stands in

kawasima:
`NeutralForm` decided which discriminator a case carries by walking every visible
declaration for a sum that lists it. That answers a question about the case where the
question is about the position the value stands at.

## What was measured first

The ambiguity #683 reported — two sums naming one case under different keys or tags —
is unreachable. `AstBuilder` builds every sum with an empty decoder and there is no
surface syntax for a written one, so every `Ast.Discriminate` comes from
`Deriver.deriveSum`, which keys it `"type"` and tags each case with its own declared
name. A sum's cases are declared in its own module. `Reason = High | LowRole` beside
`Cause = High | LowRole` compiles, and both codecs read and write
`{"type": "High", ...}`.

## What is fixed

What is written is decided by the type the position declares (spec §encode-law): a
case's own decoder reads no discriminator and its sum's reads one. `tagged` now takes
that type and reads the key and the tag off its own derived decoder —
`Ast.SumData.decoder()`, what `Deriver` decided and `CodecGen` emits — rather than
deciding the form a second time. `readsAsItself` and `adjacentlyTagged` go with it:
at a position that is not a sum listing the case, no envelope is written, which is
what both were saying separately.

Where the position has no declared type the old walk stays, on its own path and named
as the fallback it is. It has one producer: a behavior answering with several types,
where admission and not a decoder picks the case.

## Effect on souther-examples

16 lines of ~20,600 moved. `hr` generates 4 more rows (56 -> 60) and 4
`every value tried was refused at construction` lines are gone: `BusinessClosure` and
`EncouragedResignation`, unit cases `employmentinsurance` does not import by name.
The old path resolved a spelling, got null, left the envelope empty, and the value was
refused with `type: is required`.

## How it is held

Each of a product case, a unit case and a newtype is pinned at its own position and at
a sum that lists it; an enumeration keeps its bare name; the fallback is asserted where
no type is declared.

Both directions are controlled for. Making the typed path write no tag fails 34 tests
across eight classes. Restoring the old walk fails the three that assert what a value's
own position reads. All modules are green.

## Left for later

- #696 — compiler-generated `Ast.Apply` drops the resolved callee identity, which is
  what still forces `FixtureReader` to resolve spellings. Attempted here and reverted:
  the identity is absent on generator-built nodes, so it needs upstream work.
- #697 — whether codec-target selection is duplicated across three type walks.
- #698 — the untyped fallback holds only while every discriminator is derived.

Refs #683

### DIFF
--- a/souther-compiler/src/main/java/souther/compiler/examples/FixtureReader.java
+++ b/souther-compiler/src/main/java/souther/compiler/examples/FixtureReader.java
@@ -1322,15 +1322,24 @@ public final class FixtureReader {
                         : denotes instanceof ValueName.Helper ? valueBody(v.name()) : null;
                 // A name standing for no value is not one a field can be taken off; the ordinary
                 // reading below says what it is instead of this one guessing.
-                yield body == null ? new Projected.Declared(declaredTypeOf(v, new HashSet<>()),
-                                raw(v, null, below))
+                // Read at the type the name is declared with, which is the position it stands at:
+                // the same type this hands on as the evidence of what was projected. Reading it at
+                // no position would leave the form to be worked out from the case rather than from
+                // where the value is (issue #683).
+                yield body == null ? projectedAtItsDeclaredType(v, below)
                         : expanding(denotes, () -> projectTarget(body, admission));
             }
             case Ast.Apply c when appliedHelper(c) instanceof Applied helper ->
                     new Projected.Live(answered(c, helper), c.written());
-            default -> new Projected.Declared(declaredTypeOf(e, new HashSet<>()),
-                    raw(e, null, below));
+            default -> projectedAtItsDeclaredType(e, below);
         };
+    }
+
+    /** What a projection takes a field off, read at the type it is declared with — the one type the
+     *  evidence it carries is stated in, so the value and the type it is read at come from one place. */
+    private Projected projectedAtItsDeclaredType(Ast.Expr e, Admission below) {
+        Type declared = declaredTypeOf(e, new HashSet<>());
+        return new Projected.Declared(declared, raw(e, declared, below));
     }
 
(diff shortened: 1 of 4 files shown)


kawasima:
Review addressed in 312ac69d.

**[P2] cross-module regression test** — added `ACaseIsBuiltAtItsPositionWithoutBeingNamedHereTest`, the shape `hr` has. `domain` declares `Reason = EmployerCause | Resignation` over `EmployerCause = BusinessClosure | Dismissal`; `consumer` matches on `EmployerCause` and never names its leaves, so it does not import them. Filling the behavior's rows builds values of cases spelled nowhere in the reading module, through `FixtureReader`'s unit-case path — the one the corpus broke on.

Controlled: putting the spelling lookup back (`neutral.tagged(expected, symbols.resolve(caseName.name()), unit)`) fails it with

```
[UnresolvedCombination[classes=[reason=BusinessClosure], reason=ALL_CANDIDATES_REJECTED],
 UnresolvedCombination[classes=[reason=Dismissal],       reason=ALL_CANDIDATES_REJECTED]]
```

the same two names and the same reason as the four corpus lines.

**nested sum** — added `aCaseReachedThroughANestedSumIsReadAtTheOuterPosition`. Reading `sum.cases()` instead of the decoder's variants leaves it with no tag, which is the mutant it kills.

**naming** — `eachSumThatListsTheCaseReadsItThroughItsOwnDecoder` is now `twoDerivedSumsThatListOneCaseCurrentlyAgreeOnItsDiscriminator`, and its doc says the agreement is a fact about derivation rather than about either position, so #698 has to come past it.

All modules green. The compiler sources are unchanged from 43968909 — this commit is tests only, so the corpus measurement in the description still stands.

Issue #700, opened 2026-08-12
A behavior's answer is identified by resolving its class's simple name, so a module reaching the type through an alias observes none of its own cases


kawasima:
`ExampleVerifier` records which case a row's behavior answered with by resolving the
runtime class's *simple* name in the reading module's scope:

```java
state.resultArm = symbols.resolve(NeutralForm.simpleName(result));   // ExampleVerifier.java:721
```

`NeutralForm.simpleName` drops the module the class names, and `Symbols.resolve` answers
for the module doing the reading. A module that reaches the type through an alias has no
bare name for it, so the lookup misses. A module that declares something else of that
spelling gets that instead, and nothing says so.

## Measured

Two modules, two rows that pass, and the only difference between the two runs is how the
importing module reaches the type.

```souther
module lib2 exposing ( Yes, No, Answer )

data Yes
data No
data Answer = Yes | No
```

```souther
module app3 exposing ( In, f )

import lib2 as up

data In = { n: Int }

behavior f : (i: In) -> up.Answer constructs up.Yes, up.No
let f (i) = if i.n > 0 then up.Yes else up.No

example f
    | "yes" : (In { n = 1 }) -> up.Yes
    | "no"  : (In { n = 0 }) -> up.No
```

```
  f                        implemented   rows 2    pending 0
    signature   out specified 2/2  observed 0/2  verified 0/2
      · no row confirms `Yes`
      · no row confirms `No`
```

The same model written `import lib2 ( Answer, Yes, No )`, with the boundary and the rows
spelled bare:

```
  g                        implemented   rows 2    pending 0
    signature   out specified 2/2  observed 2/2  verified 2/2
```

Both rows pass in both runs. `observed 0/2` is the report telling the author that no row
confirms two cases that two green rows confirm.

It is not only a number. `resultArm` is what `Adequacy` counts as the case a behavior was
seen to answer with (`Adequacy.java:1820`), and `RowOutcome.witnesses()` is false while it
is null, so nothing downstream recovers it.

## The other failure mode is silent

Where the reading module happens to declare a type of that spelling, the lookup answers —
with the wrong declaration. A module declaring `data Amount = String` beside another
module's `Amount` reached through an alias gets:

```
This example does not hold.
Hint: At $ the row states Amount, but the behavior answered with Amount.
```

Two sides of one comparison, one of which came from the fixture's carried identity and the
other from this lookup. They are the same type. Without the local declaration, the same
model reports the value as unreadable instead:

```
It is:
[...]

PR #701, opened 2026-08-12
Write a reference the way the module reading it reaches the type

kawasima:
Reported as #696: a compiler-generated `Ast.Apply` drops the resolved callee identity, so
`FixtureReader` resolves the source spelling again. The identity is half of it. The
generator writes two forms of one value — the text a row carries and the tree a decoder
builds — and it wrote both from the declaration's own name, which is not what any module
necessarily calls the type.

## The defect, measured

`lib` declares a newtype; `app` takes it at a boundary. The only difference between the
runs is how `app` reaches it.

| how `app` reaches it | `symbols.resolve("Amount")` | before | after |
| --- | --- | --- | --- |
| `import lib as up` | `null` | no row: every candidate refused | `up.Amount(500)` |
| `import lib ( Amount )` | `lib.Amount` | `Amount(500)` | unchanged |
| alias, plus `data Amount = String` of its own | `app.Amount` | no row; read as the wrong type | `up.Amount(500)` |

The text half fails on its own. An imported unit case carried its identity already, so the
candidate built and the row was offered — `Req { kind = Domestic }`, pasted into a module
that cannot write `Domestic`. Fixing only the tree moves the newtype case onto the same
footing rather than off it.

## What is in it

`types/TypeReachName` — `Bare` / `ViaAlias` / `ViaModule`, and `Unnameable` for a type no
spelling here reaches. `rendered()` lives on `Written` alone, so a caller has to say what
it does where there is no name. `Symbols.reach` decides it once, and asks whether the bare
spelling means *this* declaration rather than whether the name is in scope.

`FixtureTemplate` writes text and tree from that one answer. `FixtureReader`'s seven
readers ask what the callee denotes. Three shapes that let a synthesized node stay silent
are removed: `Ast.Apply(String fn, …)`, `Var.reaches()`'s fall back to the spelling, and
`ReachName.of`'s `null -> Bare(written)`. ADR-0067 stated the rule and nothing held it.

A case whose module does not expose it has no spelling here at all — E1611 holds a
signature to exposed types, and a sum's cases come through the sum. Those classes are
still counted and each says why no row can be offered for it, instead of being offered a
name that resolves to nothing wherever the row is pasted. `ACaseIsBuiltAtItsPosition…Test`
was pinning one of those rows as expected output.

## Effect on souther-examples

Both jars over all fourteen modules:

- generated rows 953 → 1100
- `every value tried … was refused` 499 → 84
- no module lost a row (crm 377 → 499, realworld 74 → 97, hr 404 → 406)

## Tests

`AGeneratedRowIsWrittenTheWayTheModuleReachesItTest` pastes the offered row back into the
module it was offered to and compiles it, beside the same row under the declarations' own
names — which is two names out of scope, so the check is one the model can fail.
`AGeneratedFixtureSaysWhatItsNamesDenoteTest` holds the tree invariant on the side that
writes: a reader can always fall back to a spelling, so a reader's tests would pass
without it. `ANameAnsweredHalfwayIsRefusedTest` loses a test that had become half of
another one.

All modules green (souther-compiler 4276).

## Not closed by this

#700 — the same class, one layer over: four readers identify a live value's declaration by
resolving its class's *simple* name, so a module reaching the type through an alias
observes none of its own cases. Nothing upstream has to change for it; the value carries
its class and the readers throw it away.

`Type.show` renders a type's bare name and qualifies only when two shown types collide. It
is diagnostic presentation rather than a reference something is pasted back as, so it is
left alone here.

Refs #696

### DIFF
--- a/souther-compiler/src/main/java/souther/compiler/ast/Ast.java
+++ b/souther-compiler/src/main/java/souther/compiler/ast/Ast.java
@@ -16,6 +16,7 @@ import souther.compiler.types.ValueName;
 
 import java.util.ArrayList;
 import java.util.List;
+import java.util.Objects;
 import java.util.Map;
 import java.util.Optional;
 import java.util.function.UnaryOperator;
@@ -1418,11 +1419,17 @@ public interface Ast {
         /**
          * A name that has been resolved has both answers or neither.
          *
+         * <p>Both, or neither: a name the parser read is unanswered on both counts, and a name
+         * resolution answered is answered on both. Nothing in between is a state — the pair comes
+         * from one pass, which has both or has not run.
+         *
          * <p>Refused rather than allowed to stand, because what a name reaches is asked of a table
          * and a table answers a key it has not got with silence. A rewrite that carried the
          * denotation across and dropped the reach name would leave a reference that resolves to a
          * declaration and reaches nothing, and the first reader of it would read the spelling
-         * instead — which is what this pair was separated out to stop.
+         * instead — which is what this pair was separated out to stop. The other half is the same
+         * defect facing the other way: a reach name beside no denotation is a key nothing says the
+         * meaning of, and {@link ReachName#of} is where one would have to come from.
          *
          * <p>The region is the expression's and the name's is {@code written}'s, and they part
          * company only where the author wrapped the name in something the tree does not keep:
@@ -1431,9 +1438,10 @@ public interface Ast {
          * name is written somewhere this expression is not, which no source can produce.
          */
         public Var {
-            if (denotes != null && reachedAs == null) {
-                throw new IllegalArgumentException("`" + written.canonical() + "` denotes " + denotes
-                        + " and says nothing about how this module reaches it");
+            if ((denotes == null) != (reachedAs == null)) {
+                throw new IllegalArgumentException("`" + written.canonical() + "` denotes "
+                        + denotes + " and is reached as " + reachedAs
+                        + "; a name has both answers or neither");
             }
             if (written.region() != null && region == null) {
                 throw new IllegalArgumentException("`" + written.canonical()
@@ -1662,7 +1670,16 @@ public interface Ast {
          */
         public Apply(String fn, ValueName denotes, ReachName reachedAs, List<Expr> args,
                      ConstructionOrigin origin, SourcePos pos, Region region) {
-            this(Var.respelled(fn, denotes, reachedAs, pos, null), args, origin, pos, region);
+            this(Var.respelled(fn, Objects.requireNonNull(denotes, unanswered(fn, "denotes")),
+                            Objects.requireNonNull(reachedAs, unanswered(fn, "is reached as")),
+                            pos, null),
+                    args, origin, pos, region);
+        }
+
+        /** Why a pass may not apply a name it has not answered for. */
+        private static String unanswered(String fn, String half) {
+            return "a pass applying `" + fn + "` says what it means: nothing here " + half
+                    + " it, and the spelling would be resolved again wherever this is read";
         }
 
         /** Whether what this applies is a name. A reader that wants the name itself matches on
--- a/souther-compiler/src/main/java/souther/compiler/ast/Ast.java
+++ b/souther-compiler/src/main/java/souther/compiler/ast/Ast.java
@@ -11,6 +11,7 @@ import souther.compiler.types.CoverageOrigin;
 import souther.compiler.types.ReachName;
 import souther.compiler.types.Type;
 import souther.compiler.types.TypeName;
+import souther.compiler.types.TypeReachName;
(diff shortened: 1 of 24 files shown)


kawasima:
All three fixed in e451000f. Each was reproduced first.

## 1 — the synthesized `Apply` still took `(null, null)`

Confirmed, and the claim in the description was wrong: two of the three shapes were
removed, not three. `Var` refused only `denotes != null && reachedAs == null`, so the
other way round and the pair being absent were both legal, and the constructor that
replaced `Apply(String fn, …)` passed them straight through.

`Var` now reads its own comment:

```java
if ((denotes == null) != (reachedAs == null)) { throw … }
```

and `Apply(String, ValueName, ReachName, …)` requires both. Two states are left: the
parser's, which has neither and builds its own callee, and resolution's, which has both.

Checked before changing it that nothing produces `(null, non-null)`: the only path that
could is `HelperInliner.renameVar`, whose `Renaming.Copy.of` returns its argument unless
it is a `Local`, and a `Local` for a `Local`.

`CallElaboratorNoCalleeTest.anUnresolvedCallIsAnInternalError` was the one test that
broke, and it was pinning `noCallee`'s absent-denotation arm with a node the AST no longer
builds. That arm is reachable and always was — an application of something that is not a
name has no name to denote anything — so it is written as that now.

`ANameAnsweredHalfwayIsRefusedTest` holds both halves and the application.

## 2 — `reach` did not compose with `resolve`

Confirmed and reachable. A module declaring `data RoundingMode = { tag: String }` compiles,
and:

```
reach(souther.runtime.RoundingMode)  = Bare("RoundingMode")
resolve("RoundingMode")              = app.RoundingMode
```

The runtime namespace now goes through the same question as everything else — bare only
while nothing here has taken the spelling, and `Unnameable` when something has, since a
qualifier does not name that namespace. The primitive branch stays: a module cannot declare
`Int` (E1502), which the javadoc now says rather than leaves to be worked out.

`AReachedNameResolvesBackToWhatItWasAskedAboutTest` is the law, over every declaration a
module can meet rather than over cases picked out by hand. Stashing the fix, three of its
four fail, the first with

```
RoundingMode is written for souther.runtime.RoundingMode and resolves to app.RoundingMode
```

## 3 — the paste-back test's name and its assertion

Right, and the filter would have called a syntax error a row. It now holds everything the
compile added over the same model without the row, so the offered row adds `E1905` — the
expectation pasted in is deliberately the wrong one, and that it is *read* is the point —
and the declarations' own spellings add `E1023` twice. The difference rather than the whole
because one fixture leaves an import it does not use, which is about the fixture and not
about the row.

## Effect

souther-examples is byte-identical to the run on 954eaa42 (1100 generated rows, 84
`every value tried … was refused`). `RoundingMode` has no codec and reaches no fixture, so
finding 2 was a contract hole rather than a live defect — which is why it is worth closing
now, while the answer has one home.

All modules green (souther-compiler 4281).

kawasima:
Both corrected in the commit above. Documentation and one test; nothing in the compiler
changed, and the `souther-compiler/src/main` diff is comment-only.

## `inverse` → section

Right, and the same holds one step further than the finding says. `resolve` answers for
declarations, and the language's own vocabulary is not among them — the runtime's error
cases are outside its domain too, for the same reason:

```
souther.Int                     reach=Int             resolve=null              resolveCase=souther.Int
souther.runtime.DivisionByZero  reach=DivisionByZero  resolve=null              resolveCase=souther.runtime.DivisionByZero
souther.runtime.RoundingMode    reach=Unnameable      resolve=app.RoundingMode  resolveCase=app.RoundingMode
```

`Symbols.reach` now says it is a section of `resolve` rather than its inverse, and says
which reader reads a name back where that reader is not `resolve`.

Rather than only narrowing the wording, the edge is now a test. Both of those reach a
position a writer meets — a primitive heads a union a behavior answers with, and the
runtime's failures are its cases — so leaving them out of `everyDeclaration()` is how a
domain nothing states becomes one a later reader takes for everything.
`theLanguagesOwnVocabularyIsWrittenAsItselfAndReadBackAsACase` holds all three of: written
as itself, read back by `resolveCase`, and `resolve` answering nothing.

The third line is worth keeping in view. `reach` answers `Unnameable` for a shadowed
`RoundingMode` while `resolveCase` would still hand back `app.RoundingMode` — the two agree,
because at a *value* position, which is where a generated row writes one, `Resolve` asks
`resolve`. If a case a row writes ever comes to be read through `resolveCase` instead, this
is the pair to look at.

## "the AST refuses to build"

Also right. The constructor a pass writes an application with refuses; the shape stays
writable, since the parser holds a tree resolution has not seen. Splitting pre- and
post-resolution trees by type is the only thing that would close the shape itself, and that
is not this. The comment now says what was actually closed.

All modules green (souther-compiler 4282).

Issue #702, opened 2026-08-13
Where the identity of every JVM class the compiler invents is defined, and how producers and consumers refer to it without reconstructing it


kawasima:
`CodegenContext` owns how this backend names the classes it makes up, and hands out fragments:

```java
static String bridgeCaseName(TypeName member) {
    return member.name() + "Case";
}
```

so every reader finishes the name itself. `ExampleVerifier` writes the join out —

```java
if (result.getClass().getName().equals(module.name() + "." + member.name() + "Case")) {
```

— and the module it adds is the one doing the reading, which is right only when that module also
happens to be the one that emitted the case. A bridge case is named by whoever emits it.

Measured across the compiler: the three public APIs that return a fragment
(`behaviorClass`, `behaviorImplClass`, `behaviorResultClass`) have six consumers and all six
assemble the rest by hand, one of them by copying the capitalisation itself.
`moduleClassName`, the one that returns a complete binary name, has two consumers and neither
assembles anything. What is restated is what was left unfinished.

The same shape holds inside the backend. `generated(simpleName)` is asked with `name + "$Enc"`,
`name + "$Dec"`, `name + srcSuffix(src)`, `name + "$Ctfe"`, `behaviorClass(name) + "Result"` at a
dozen emission sites, each beside a second spelling of the same class for the output map's key.

## What to decide

Where the identity of every JVM class the compiler invents is defined, and how producers and
consumers refer to it without reconstructing it.

Not only a type's classes: a behavior's interface and implementation, a result union, a bridge case,
the codecs, the compile-time checks, `$Fns`, `$Module`, the lambda classes, and the stand-in an
example defines for an injected behavior. What identifies each of them differs — a value class is a
declared type, a bridge case is a member *and* the module emitting it, a lambda is a number — so a
name is not what any of them should be passed around as.

Class identity only. `helperMethod` returns a complete method name and its consumer reconstructs
nothing, so the defect that was measured does not reach JVM members.

Not the same defect as the resolution work. Nothing here reads a spelling in a module's scope; the
question is which name the backend gives a class it invents, and that has an owner already — it just
does not reach the people asking.

kawasima:
Settled by #705 and #706.

**#705** — the identity of every JVM class this compiler invents is a `GeneratedClass`, and
`SoutherJvmAbi.nameOf` is the one thing that turns one into a `JvmClassName`. Nothing outside that
package can make one: the constructor is package-private, and the switch has no default, so a kind of
[...]

PR #705, opened 2026-08-13
Give a generated class an identity, and leave the spelling to the one thing that decides it

kawasima:
Closes the first half of #702: the identity of every JVM class this compiler invents is defined in
one place, and nothing outside that place can produce one.

## What was wrong

The rule had an owner and handed out fragments, so every reader finished the name itself. Measured
across the compiler: the three public APIs returning a fragment (`behaviorClass`,
`behaviorImplClass`, `behaviorResultClass`) have six consumers and all six assemble the rest by
hand, one by copying the capitalisation into a private method of its own. `moduleClassName`, the one
returning a complete binary name, has two consumers and neither assembles anything. `GeneratedBehavior`
carries a paragraph saying not to restate the name, and the class beside it restated it anyway — so
the fix cannot be another paragraph.

## The shape

```
Souther identity          GeneratedClass.BridgeCase("ship", inv.Shortage)
      |                        (what identifies it, and nothing else)
      v  SoutherJvmAbi.nameOf  (the only mapping; a switch expression, no default)
JVM identity              JvmClassName  ->  binary name / ClassDesc / registry key
```

`JvmClassName` has a package-private constructor, so no reader can mint one. That is the guarantee;
a record could not hold it, since a public record's canonical constructor is public. What is public
is the whole name — never a simple name, a package, or a suffix, which are what a caller would
finish by hand.

What a module emits is now a registry that takes the identity and keys on the name. The mapping is
deliberately not injective — a data `Quote` and a behavior `quote` land on `demo.Quote` — so the
collision is detected where it exists and reported with the two identities that wanted it. The three
declaration-time refusals compare what the ABI answers rather than spellings worked out beside them.

## Coverage

- Eleven consumer-side reconstructions in main are gone, and the descriptor/output-key pair that
  spelled each emitted class twice on neighbouring lines is one expression.
- 356 spellings across 150 test files go through a helper that forwards to the ABI. The spellings are
  written out in exactly one contract test, stated independently of the mapping, so a deliberate
  change to a name is one failure that says what changed.
- A constant-pool scan over every module's `classes` and `test-classes` fails on a name assembled out
  of a spelling this ABI writes. It derives what to look for from the ABI rather than listing
  suffixes, and reads the module list off the reactor. What it cannot see is a behavior's interface,
  whose rule adds no suffix — that leaves a call instead, so a second check asserts that nothing
  outside the ABI package calls `Character.toUpperCase` in main classes, and there is one caller.
  Both were verified by planting violations and watching each go red. Between them they found two
  real sites in the backend (`$Fns` reached by concatenation), one test picking lambda classes out by
  shape, and four tests still capitalising a behavior's first letter.
- No emitted name changed, so `BOUNDARY_VERSION` stays where it is. All 4297 compiler tests pass,
  along with the other modules.

Separately, 107 unused imports across 75 files.

## Not in this change

The repository-wide audit for the same shape in other identities, and whether `ModuleName` should be
a value type — held back so that this change's hypothesis stays measurable on its own. JVM member
identity is out of scope: `helperMethod` returns a complete method name and its consumer
reconstructs nothing, so the defect that was measured does not reach there.


kawasima:
Pushed `b7fccc6f`. The full suite is green on a clean tree (`mvn clean install -DskipTests` then `mvn test`): 4297 in souther-compiler, 4 in souther-bench, the rest unchanged.

**1 — identities this compiler would never emit.** Fixed by narrowing, not by checking. `Decoder(Value, kind)`, `Ctfe(Value)`, `ExampleFake(BehaviorInterface)`, and a `sealed interface Encodable permits Value, BehaviorResult` for `Encoder`. `Encoder(Encoder(…))`, `Ctfe(Helpers)` and the rest no longer compile. You were right that the tripwire was living off the hole — it built `ExampleFake(value)` to collect the suffix, and now builds it from a behavior interface.

The two tests that counted `getPermittedSubclasses()` now flatten the sealed intermediate, so `Encodable` narrows what can be built without becoming a kind of class.

**2 — `$Module` bypassing the registry.** `Backend.generate` returns the `Emissions`, `ModuleMetadata.stamp` takes it, and the projection to `Map<String, byte[]>` happens once at the end in `Output`. Stamping a declaration onto a class is `Emissions.rewrite(GeneratedClass, …)`, which refuses a rewrite of something nothing emitted rather than quietly becoming the emission of it. `$Module` is a `put` like everything else.

`ExampleFake` is documented as what it is instead: defined into the run's loader rather than emitted, where a defined class shadows a module's own rather than colliding with it — which is why its base is a behavior interface, a shape no declaration can produce.

**3 — the module list.** Read off the reactor's `<modules>` now. `souther-fmt`, `souther-lsp` and `souther-cli` are declared as test-scoped dependencies of `souther-bench` with a comment saying why, so `-am` builds what the check reads.

**4 — what the scan catches.** A spelling containing `$` is now read wherever it falls in an assembled string, so `m + ".Foo$Impl"` is caught; `Case` and `Result` still need the joint against them so prose is not counted. Controls are built out of spellings the ABI gave, so this class holds no name of its own.

For the behavior interface, which has no suffix to look for: a second test asserts that nothing outside `souther.compiler.jvm` calls `Character.toUpperCase` in main classes. There is exactly one caller. The control test states in a `assertFalse` that a hand-rebuilt interface is not a shape the scan can name, so what it does and does not catch is measured rather than claimed.

Both were verified by planting violations in `souther-cli` — a constant-tail name and a capitalisation copy — and watching each test go red, then reverting.

**5 —** `Emissions`'s example is `BridgeCase("m", a.Foo)` / `BridgeCase("m", b.Foo)`, and `ModuleMetadata`'s javadoc names the registry rather than the deleted `Backend` methods. Four tests that still capitalised a behavior's first letter ask for the class instead.

The PR body's claim about the scan has been overstated in the same way you flagged; I have kept the wording here to what the controls actually measure and will fix the body text too.

kawasima:
Pushed `461e8e12`.

`rewrite` now refuses unless the class held under the name is the identity that asked, and keeps `held.generated()` on the way out, so a rewrite changes what a class holds and never what it is. You were right that this was the pair the registry was built for, tested on one side only.

Three rows in `TwoClassesUnderOneNameAreNotOneClassTest`:

- `oneIdentityCannotRewriteAnotherThatHasTheSameName` — `Value[demo.Quote]` emitted, `BehaviorInterface[demo, quote]` rewriting it, refused, and the class is not rewritten.
- `andTheIdentityThatWasEmittedMayRewriteIt` — the control, which also holds that a rewrite is not a second emission.
- `andNothingCanBeRewrittenThatWasNeverEmitted` — the `held == null` branch, which had no row either.

Verified the first one is not vacuous by removing the check and watching it fail, then putting it back. Full suite green on a clean tree: 4300 in souther-compiler now, the rest unchanged.

On the `Character.toUpperCase` guard: agreed, leaving it as it is. It is a tripwire, it has no callers to accommodate, and the moment an unrelated capitalisation arrives is the moment to decide between an allowlist and a narrower check — deciding now would be guessing at which.

PR #706, opened 2026-08-13
Ask the ABI where a class is written, and what type a class's name says it is

kawasima:
The second half of #702: the repository swept for the shape the naming work was about — a consumer
rebuilding a semantic identity out of primitives — and the two that came back fixed.

## What was swept

Every place in main that joins on `.`, splits on it, walks back from `lastIndexOf('.')`, turns a name
into a path, or compares an assembled string. Each was read for the three questions: who is the
authority for this identity, does it answer completely, and is the consumer re-deriving something.

## What came back

**A value class read back out of a runtime class name.** `NeutralForm.typeOf` split a binary name at
the last dot to rebuild a `TypeName`. That is the inverse of `nameOf(Value(type))`, written in the
package the last two defects of this kind were found in. It is now `SoutherJvmAbi.declaredTypeOf`,
and it is the only inverse there can be: a value class is its type, so the two directions are one
rule. Every other kind either adds something to a name or shares one with a kind that does —
`demo.Quote` is a data `Quote` and it is also the interface of a behavior `quote` — so a name alone
does not say which. It answers for the name and claims nothing about what was declared, which is what
the caller's own scope answers and still does.

**Where a class of a given name is written**, stated at four readers: the two that write a
compilation out, the one that reads a module off the class path, and the one that reads a jar. The
rule is the JVM's rather than this compiler's, so it is a function of a name rather than a kind of
generated class — but it is still a rule, and a compiler that ever writes a class somewhere else now
has one place to look.

Both are stated in the contract test beside the spellings, including what the inverse does not claim.

## What was left, and why

`SoutherProcessor` asks a `Filer`, whose API takes a package and a file name rather than a path, so
the same rule appears there in the shape that API forces. Expressing it through the new one would be
more restatement rather than less.

The joins in `check` and `query` that walk back from a dot are about a name as a module's source
wrote it — `A.member` — which is the resolution question, not this one. It has its own owner and its
own issues behind it.

`HelperNames.qualified` is the shape done right already: one authority, a complete answer, and both
consumers ask.

The three derivations that produce a `.sou` name — a bundled module's resource, the standard
library's, and the `SourceFile` attribute — are three questions that happen to end in the same four
characters, not one identity stated three times.

## `ModuleName`

Not now, on the evidence. The sweep found nothing a module-name value type would have prevented: what
was wrong was deriving other things from a module name, not confusing a module name with another
string. Held open rather than decided against — it would be worth revisiting if a defect of that
second kind ever shows up.

All 4302 compiler tests pass, along with the other modules.

### DIFF
--- a/souther-cli/src/main/java/souther/cli/Main.java
+++ b/souther-cli/src/main/java/souther/cli/Main.java
@@ -1,5 +1,6 @@
 package souther.cli;
 
+import souther.compiler.jvm.JvmClassName;
 import souther.compiler.Compiler;
 import souther.compiler.Reserved;
 import souther.compiler.cst.CstError;
@@ -1041,7 +1042,7 @@ public final class Main {
     static List<Path> writeClasses(Map<String, byte[]> classes, Path outDir) throws IOException {
         List<Path> written = new ArrayList<>();
         for (Map.Entry<String, byte[]> entry : classes.entrySet()) {
-            Path file = outDir.resolve(entry.getKey().replace('.', '/') + ".class");
+            Path file = outDir.resolve(JvmClassName.classFile(entry.getKey()));
             Files.createDirectories(file.getParent());
             Files.write(file, entry.getValue());
             written.add(file);
--- a/souther-compiler/src/main/java/souther/compiler/Compiler.java
+++ b/souther-compiler/src/main/java/souther/compiler/Compiler.java
@@ -1,5 +1,6 @@
 package souther.compiler;
 
+import souther.compiler.jvm.JvmClassName;
 import souther.compiler.diag.CompileException;
 import souther.compiler.diag.Diagnostic;
 import souther.compiler.diag.msg.DeclarationMessage;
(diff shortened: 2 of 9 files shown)


kawasima:
Pushed `65100e19`. `SoutherJvmAbi.valueTypeCandidate` now.

You are right that the old name promised the second half. The rows said so themselves and I read them as fine: `shop.FindOrder$Impl` is not a declaration any source could write, and `souther.Int` is a primitive that reaches codegen as a boxed class rather than a value class of its own — both read back all the same, because reading back is all it does. The javadoc says that, and says whose answer the other half is.

The two steps are now what they were doing anyway, with the names to match:

```
ABI      valueTypeCandidate(binaryName)  — the identity that spelling would be
Symbols  contains(candidate)             — whether it is there
```

Left as it was, a later reader takes the answer as the declaration it is named for. That is an authority claiming something it has no way to see, which is this defect pointed the other way — worth the rename on its own.

Full suite green on a clean tree: 4302 in souther-compiler, the rest unchanged.

Issue #703, opened 2026-08-13
One AST type holds both a tree Resolve has read and one it has not, so the pass boundary is a convention every consumer keeps for itself


kawasima:
`Resolve` states the invariant:

> After it, every `Ast.Name` in the tree carries the `TypeName` it denotes, and a check that wants to
> know whether two names are the same type compares what they denote. There is no spelling left to
> compare.

The sentence is true of what the pass produces and is not a property any type holds. One `Ast.TypeRef`
stands for a reference nobody has read yet and for one that denotes a type, told apart by
`denotes() != null`; one `Ast.Name` likewise. A consumer downstream of the pass is handed the same
type either way, so "has this been resolved" is a question it answers by looking, and answering it
wrong is what #464, #696 and #700 were.

Making the state a type would put the invariant in the signatures:

```
ParsedModule   — carries WrittenName
      ↓ Resolve
ResolvedModule — carries TypeName
```

with `ExampleVerifier`, the backend and `Deriver` taking the second. A consumer would then not be able
to ask for a resolution, because it would hold nothing to resolve.

## Why this is not small

Every pass between the two ends reads and writes these trees, and several build nodes of their own:
`HelperInliner`, `NewtypeDesugar`, `Deriver` and `FixtureTemplate` all mint nodes after Resolve has
run. Each has to mint them resolved — `FixtureTemplate` now does, and it is the smallest example of
what the split would require everywhere.

Worth measuring before committing to it: how many nodes are built post-Resolve, and how many of them
already carry what they denote.

kawasima:
## Measurement

This issue asks for a measurement before committing to the split — how many nodes are built after
`Resolve` has run, and how many of them already carry what they denote. Here it is, with what the
split turned up once it was done. Four of the issue's own premises come out differently.

### 1. Post-Resolve producers were not the problem

Every producer that mints a node after `Resolve` already mints it answered. Measured on `develop`
before any change, by construction site:

| producer | how it builds one |
| --- | --- |
| `Deriver` (4), `HelperInliner` (3), `HelperNames` (2), `Names` (1), `FixtureTemplate` (1), `HelperParams` (1) | `TypeRef.of`, `Name.resolved`, `Name.reached`, `Var(written, denotes, reachedAs)`, `.denoting` |
| `AstBuilder` (20) | the only producer of an unanswered form |

The worry in "Why this is not small" — that `HelperInliner`, `NewtypeDesugar`, `Deriver` and
[...]

Issue #464, opened 2026-08-08
A published invariant cannot be read again by the module that imports it


kawasima:
A newtype's invariant cannot be read again once its declaration has crossed a module boundary.

The rule is there and the decoder runs it. What cannot be done is re-elaborating it in a module that
imports the type, and everything that reads an invariant by elaborating it therefore gets nothing.

```souther
// module probe.money exposing ( Amount )
data Amount = Int
    invariant within = value >= 0 && value <= 10
```

Reading that declaration from another module and asking for its clause as the checks read it:

```
E1023: unknown identifier `value`
  | fieldTypes=[value] | fieldBindings=[value] | newtype=true
```

The field is there and so is its binding, so the scope offered to the elaborator has `value` in it.
What fails is the resolution: a `Var` of the published declaration carries the `denotes` and
`reachedAs` its own module settled (§`Ast.Var`, "a name that has been resolved has both answers or
neither"), and re-elaborating it in a reader's scope looks up a binding that scope does not have. The
same declaration written in the reading module types without complaint.

`Clauses.typed` answers `null` and every caller reads that as *a clause outside the fragment*, which
is the answer for a rule the language cannot express rather than for one that was not looked up.

## What it costs

Anything that reasons over an imported type's own rules is reasoning without them. The one measured so
far is the projection a boundary is derived from (#427): a rule relating two positions of an imported
numeric type reaches the domain while their ranges do not, so nothing narrows, and the derivation puts
the types' own edges back.

```souther
// another module
data Pair = { a: Amount, b: Amount }
    invariant ordered = a < b
```

```
imported     · no row is at f/p.a = 10      · no row is at f/p.b = 0
same module  · no row is at f/p.a = 9       · no row is at f/p.b = 1
```

10 and 0 are values no `Pair` holds, and the generator refuses every attempt at them. PR #459 reports
such an edge as *not known to be writable* rather than as a gap, which is true and is the answer for a
rule that could not be read — it stops being reached the moment this is fixed, so it is not in the way
of fixing it.

The invariant-discharge check reads clauses the same way, so a construction of an imported type is
presumably discharged by nothing and reported by nothing. Not measured.

## Related

#427 is the shape one module in: a rule that is there, a reading that does not reach it, and a report
naming an edge nobody can write.

kawasima:
Fixed by #466.

[...]

PR #708, opened 2026-08-13
Own a failure where it happened, so a mistake in one definition is not one in the next

kawasima:
`Resolve` states an invariant its consumers keep for themselves, and #464, #696 and #700 were
consumers keeping it wrong. Measuring that turned up a second thing of the same shape, in a place
the issue does not mention: what a pass could not do to one definition was recorded as the absence
of the whole module, and every stage below read that absence as a fact about every definition in
the file.

Three of them, each measured:

- `data Order = { total: Nowhere }` beside `data Line = { order: Order }` reported the unknown name
  and then, from further down, that `Order` has no decoder — one mistake, two reports, the second
  against a line the author has no reason to look at.
- One invariant writing `Amount(value, value)` left `Note`, which has nothing to do with it, with no
  derived form for anything downstream to read.
- One body writing the same thing left the module undesugared, so nothing was checked and a
  behavior written about something else lost the type error the compiler had found in it.

## What changed

**What resolution worked out is its own artifact.** `Resolve.ResolutionIndex` holds the type
occurrences, the value occurrences and the binders, and it is partial: an occurrence the pass
answered is in it, one it did not is in none of them. It used to record the identity the traversal
carries past a mistake — a cursor on `Nowhere` was answered with a declaration called
`souther.unresolved.Nowhere`, and one on a name nothing binds with a binding of that spelling, so
go-to-definition, find-references and rename each acted on a declaration that is not there.
`Names.Facts` is where it is asked for, and every reader outside the compiler asks that rather than
the type that carries the module.

**Whether a declaration came out is settled per declaration.** `Names.Unbuilt` settles the two ways
one has no meaning — a name in it was not answered, or what it reaches has none — and neither is
"something was reported upstream" nor "the module will not be emitted", either of which would put
what a name means back in the hands of a later check. `Names.Definition` hands over the ones that
came out. The check over a declaration is handed those and asks nothing about the others.

**A stage answers for the unit it is about.** Deriving, settling and normalizing ran over a module
inside one `try`; normalizing is now a declaration at a time and desugaring a definition at a time,
each owning what it has to say. `Shapes.Derived` and `Shapes.Desugared` are assemblies of those
answers: every one asked for, and a module handed over when each is there.

## What this is not

Removing the failed declaration from the module was tried first, and is what the third commit
records. Read as a module that does not declare it, `exposing ( Order )` became an export of an
imported name and said so, and the definitions beside it lost the answers the compiler had about
them. What a module declares is a question about what is written in it; whether a declaration came
out is another question, and the two are asked of different things throughout.

## Not in scope

The middle of the pipeline — `Prepared` through `CheckedBehavior` — still funnels through a complete
module, so a definition that does not desugar still costs its siblings their check. Moving it needs a
behavior-local environment built from what each behavior reaches, which is one indivisible change
across eleven queries and rests on six equivalences that have to be measured rather than assumed:
`HelperInliner.recursiveHelpers()` mixes the graph's answer with `takenOnRecursive`, which is read
off every body in the module, so handing that band the definitions that came out would silently
change which helpers are recursive. The sentinels themselves — `TypeName.unresolved`,
`ValueName.Unresolved`, `Names.Nameless` — are removed once nothing reads them.

Refs #703

### DIFF
--- a/souther-compiler/src/main/java/souther/compiler/check/NewtypeDesugar.java
+++ b/souther-compiler/src/main/java/souther/compiler/check/NewtypeDesugar.java
@@ -32,15 +32,28 @@ public final class NewtypeDesugar {
     public static Ast.Module rewrite(Ast.Module m, Symbols symbols) {
         List<Ast.FnDef> fns = new ArrayList<>();
         for (Ast.FnDef fn : m.fns()) {
-            Ast.FnBody body = switch (fn.body()) {
-                case Ast.FnBody.Written w -> new Ast.FnBody.Written(go(w.expr(), symbols));
-                case Ast.FnBody.Intrinsic i -> i;
-            };
-            fns.add(fn.withBody(body));
+            fns.add(rewriteOf(fn, symbols));
         }
         return new Ast.Module(m.name(), m.exposing(), m.exposedOutputs(), m.imports(),
                 m.defs(), m.behaviors(), fns, m.takenOn(), m.examples(), m.fakes(),
                 m.exampleFileTarget(), m.pos());
+    }
+
+    /**
+     * One definition's body, with each newtype construction written in it rewritten to the
+     * construction it is.
+     *
+     * <p>A definition at a time, because an application that wraps no single value is wrong in the
+     * body that wrote it and in no other. Rewriting them together answers for a whole module, and
+     * one such application would leave every other definition without the form the check and the
+     * backend read.
+     */
+    public static Ast.FnDef rewriteOf(Ast.FnDef fn, Symbols symbols) {
+        Ast.FnBody body = switch (fn.body()) {
+            case Ast.FnBody.Written w -> new Ast.FnBody.Written(go(w.expr(), symbols));
+            case Ast.FnBody.Intrinsic i -> i;
+        };
+        return fn.withBody(body);
     }
 
     /**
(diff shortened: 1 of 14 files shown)


kawasima:
Both blockers reproduced and fixed in cd2e394b. Each was written as a failing test first.

**The sentinel path.** `answered` was read off how many diagnostics resolution added, and a name
left unanswered on purpose adds none — `Imports.nameless` puts the exported names of a module this
compilation has and cannot use into the importer's scope as identities nothing declares, so the
importer is not told a second time about a file that is fine. `data A = { x: X }` came out as a
declaration with a meaning:

```
ANameLeftUnansweredCountsWhetherOrNotItWasReportedTest
  `X` was not answered, whatever was or was not reported about it ==> expected: <false> but was: <true>
```

It is counted now, at the two places a name comes back unanswered, and what was reported is a
separate matter. The first attempt replaced the diagnostic delta with the count and lost the
misspelling case, which the existing tests caught; it is the conjunction of the two.

While reproducing it, `Ast.TypeRef` turned out to record the standing-in identity as a denotation
as well, so a cursor on such a name was answered with `souther.unresolved.X` — the same leak this
branch takes out of the editor's answers, by a path the first commit did not cover. That is not
recorded either now.

**The duplicate declaration.** Reproduced as described:

```
ADeclarationWrittenTwiceIsAnsweredForByTheOneTheModuleHasTest
  what `m.a` declares as `A` is made of `Int` ==> expected: <true> but was: <false>
```

`putIfAbsent`, so the one `TypeChecker.declared` accepts is the one answered for. The wider split —
resolution facts for every written definition, `OfDeclaration` for the accepted ones — is the right
shape and is not in this commit: the facts an editor gets for the rejected copy are a question about
what `ResolutionIndex` holds, and it holds them today. If you want that separated here rather than
with the rest of the definition-unit work, say so and I will.

**The comment.** Corrected on both `DerivedDef` and `DesugaredFn` to say what is guaranteed: the
failure is the named declaration's, while the dependency is still the module's, because both are
read through an answer worked out for the module.

4315 tests green in `souther-compiler`, whole reactor green.

PR #709, opened 2026-08-13
Make the pass boundary a type, so a consumer cannot ask whether a name has been resolved

kawasima:
`Resolve` states an invariant its consumers keep by looking, and #464, #696 and #700 were consumers
keeping it wrong. This puts it in the types.

#703 asked for a measurement before committing to the split. It is recorded on the issue:
https://github.com/souther-lang/souther/issues/703#issuecomment-5281228149 — including the three of
the issue's own premises it changes, and what the split turned out to be about, which was not
producer migration. This describes what changed.

## Three occurrences, and the module

Each of the three node types that carried a nullable answer becomes a sealed set of the states it is
actually in, as `Ast.Binder` already was:

```
TypeRef   Written | Denoting(Type)
Name      Written | Denoting(TypeName) | Unanswered
Var       Written | Denoting(ValueName, ReachName) | Unanswered
```

`denotes()` is one default per type that refuses the states that have no answer and says which they
are — a `Written` one is a fault in this compiler, an `Unanswered` one a mistake in the source
already reported where it is written. `Binder.id()` had written that shape and the comment that goes
with it: this is the one place that says so.

Three states rather than two, because two would put back the thing being removed. A name nothing
declares has been read; it is not waiting to be. `Denoting` refuses a stand-in identity at
construction, so there is no value that says it has been resolved and names nothing.

`Var.Denoting` holding both the denotation and the reach name is the pair invariant those two used
to enforce by hand.

`ResolvedModule` says of the whole module what the three say at each occurrence, which is the part a
signature can carry. It says the pass has been over the tree — not that every name found a
declaration, which is the distinction `Unanswered` exists for. Its constructor is package-private, so
`Resolve` is what mints one, and a pass that rewrites a resolved tree goes through
`with(Ast.Module)`, asked of the module it came from.

## What went with them

**`ValueName.Unresolved` is gone.** Nothing produced it but `Resolve` and `Names`, every reader of it
was a walk asking what a reference means, and there was no scope it stood in — which is the whole
reason `TypeName.unresolved` stays. That one is a fact about the scope resolution reads
(`Names.nameless`), and every reader of it now takes its value from `symbols.resolve(...)`; none
reads one off a tree.

**`ReachName.of` no longer answers a stand-in with `Bare(written)`.** That was the spelling handed
back as an identity-like value, which is the fallback ADR-0067 exists to remove.

**Second reports about one mistake.** Reading the recovery representation as a fact is how an unknown
name came to be told twice: an unknown spread reported as "not a product data", two stand-ins
reported as a duplicate, an unknown match arm reported as "not a case of an optional" and as
"wraps another type". Each is now what it was: nothing to build a further diagnosis on, so the
definition is `Unanswerable` and the ones beside it are still checked.

**`TypeOps.walkFields` carried a `resolving` boolean beside a null test** to say which callers may
meet an unanswered include. The states say it.

**`CallElaborator.typeOfCall` opened with an `IllegalStateException`** for a null denotation — the
pass boundary written as a runtime assertion in one consumer. `denotes()` says it now, once, for all
of them.

The graph and walk readers — `HelperInliner`, `ValueCycles`, `PartialReachability`,
`TotalityChecker`, `Requirements`, `SpecChecker`, `PipelineSigs`, `Bodies`, `HelperNames`,
`NewtypeDesugar`, `HelperParams` — ask which state a reference is in rather than what the stand-in
denoted. A name that answers to nothing reaches no declaration, so it is no edge and no rewrite.

## What this does not do

It does not make the state unrepresentable at each occurrence. A switch over `Ast.Var` is total over
three records, so `Elaborator` carries an arm for the one that cannot be there. Java cannot be told
that a particular module has no `Written` node without a phase parameter on every AST type. The
invariant moved from a convention each consumer keeps to a boundary held by `ResolvedModule` and the
three sealed types; it did not become impossible to express.

`ResolvedModule` stops at `Shapes.Settling`. Past there it reaches `Derived`, `Desugared` and
`Prepared` — the whole-module band #708 left out of scope, and the reason a declaration with no
meaning reaches `Deriver` at all.

## Verification

Each commit, on its own: the full suite green (7297 tests, the compiler module's count unchanged but
for the tests added), the examples corpus compiling, and `souther examples` over crm / invoicing /
issuetracker / ordering — 1769 lines — byte-identical to `develop`.

Refs #703

### DIFF
--- a/souther-compiler/src/main/java/souther/compiler/ast/Ast.java
+++ b/souther-compiler/src/main/java/souther/compiler/ast/Ast.java
@@ -175,21 +175,29 @@ public interface Ast {
      * one name too, so neither says what the author typed. A report asks {@link WrittenName#quoted()}
      * and underlines {@link WrittenName#region()}.
      */
-    record Name(WrittenName name, TypeName denotes) implements Ast {
+    sealed interface Name extends Ast permits Name.Written, Name.Denoting, Name.Unanswered {
+
+        /** The name, and the occurrence of it that was read. */
+        WrittenName name();
 
         /** A name as the parser read it, before the resolve pass has said what it denotes. Only the
          * parser writes one: every other producer knows what it means and says so. The spelling is
          * the source's, not one a caller canonicalized on the way in — {@link WrittenName} is what
          * canonicalizes, so the name and the characters it was written with stay one value. */
-        public static Name written(String spelling, SourcePos pos) {
-            return new Name(WrittenName.of(spelling, pos), null);
+        static Name written(String spelling, SourcePos pos) {
+            return new Written(WrittenName.of(spelling, pos));
+        }
+
+        /** The same, off an occurrence the parser has already read. */
+        static Name written(WrittenName name) {
+            return new Written(name);
         }
 
         /** A name a pass synthesized, already knowing what it denotes. It is written nowhere;
          * {@code pos} is what a complaint about it points at. The spelling is the declaration's own,
          * which is what a reference internal to a module reaches it by. */
-        public static Name resolved(TypeName denotes, SourcePos pos) {
-            return new Name(WrittenName.synthetic(denotes.name(), pos), denotes);
+        static Name resolved(TypeName denotes, SourcePos pos) {
+            return new Denoting(WrittenName.synthetic(denotes.name(), pos), denotes);
         }
 
         /**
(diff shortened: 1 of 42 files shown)


kawasima:
Fixed in 299accec.

**Blocker — the state collapse.** You are right, and it is the issue's own shape one step along:
the question moved from `denotes() == null` to `!(instanceof Denoting)` and was still being
answered again at each consumer. `NewtypeDesugar` was the one that mattered most — an unread
callee would have been read as "not a newtype application" and left standing.

`Ast.Var.answered()` is where the three part company, and it is one place:

```java
default Denoting answered() {
    if (this instanceof Denoting denoting) return denoting;
    if (this instanceof Unanswered) return null;
    throw new IllegalStateException("... was read as a declaration before it was resolved");
}
```

`denotes()` and `reachedAs()` are written over it, so the refusal is stated once rather than in
each of them. `Ast.Apply.answered()` is the same of the name an application applies. Twenty-eight
readers go through it now — `HelperInliner`, `NewtypeDesugar`, `PartialReachability`, `ValueCycles`,
`HelperNames`, `HelperParams`, `SpecChecker`, `TotalityChecker`.

Checking the type side for the same shape turned up one you did not name:
`TypeOps.holdsAnErroneousType` skipped an unread `Ast.Name` in silence. It is a switch over the
three now.

The tests hold both halves against each other **at a walk**, not only at the type — an unanswered
name adds no edge, an unread one is refused. Planting the `!(instanceof Denoting)` form back into
`ValueCycles` turns the second red, so the green is not an empty probe.

**Non-blocking — the stale comments.** All of them, and two more found while looking:

- `Ast.Name`: "the compile stops, so nothing downstream sees an unresolved one"
- `Name.denotes()`: "Either is a fault in the compiler rather than in the source" — the very
  distinction the blocker is about
- `ResolvedModule`: "a reader below has no such state to have a branch for" — now says what the PR
  body says
- `Resolve` and `Resolve.Resolved`: still described `TypeName.unresolved` as what a name that
  denotes nothing carries
- `ReachName.of`: still described answering a stand-in with its spelling, which is the fallback
  this PR removed

Agreed that on a change moving an invariant out of prose and into types, prose still stating the
old one is how the next change comes to be written against it.

**Verification unchanged in kind:** full suite green (7301 tests, +4), examples corpus compiling,
and `souther examples` over crm / invoicing / issuetracker / ordering — 1769 lines — byte-identical
to `develop`.

Issue #137, opened 2026-07-27
The build wiring a Souther project needs is written by hand, and getting it wrong reports nothing


kawasima:
## What happens

`SoutherProcessor` runs as a javac annotation processor (spec 4, 20), and javac does not start a
processing round when a compilation has no Java source. A project whose `src/main` holds only `.sou`
therefore compiles nothing:

```
[INFO] --- compiler:3.15.0:compile (default-compile) @ sharedmoney ---
[INFO] No sources to compile
...
[WARNING] JAR will be empty - no content was marked for inclusion!
```

Nothing about the `.sou` is read, the jar is empty, and no error is reported. It looks like a build
that succeeded. The way out is a file that says nothing, in a language the project chose not to use:

```java
/**
 * ...
 * <p>このファイルがあるのは、アノテーションプロセッサが javac の一部として動くからで、javac は
 * コンパイルする Java ソースが1つも無いと何もしない（"No sources to compile"）。
 */
package shared.money;
```

Ten of the examples carry one. A library published for other projects to import has no reason to
contain Java at all, which is what cross-project import (ADR-0063) is for.

## The same thing on Gradle, and worse

`examples/issuetracker` is the Gradle build, and it needs about forty lines to say what the Maven
build says in twenty-five. Three of them are load-bearing in a way nothing reports:

```kotlin
val souther by sourceSets.creating {
    java.setSrcDirs(listOf("src/main/java"))
    resources.setSrcDirs(emptyList<String>())
}

tasks.named<JavaCompile>(souther.compileJavaTaskName) {
    options.compilerArgs.add("-Asouther.source=$southerSource")
    inputs.dir(southerSource)          // without this, editing a .sou leaves the task UP-TO-DATE
        .withPropertyName("southerSource")
        .withPathSensitivity(PathSensitivity.RELATIVE)
}

dependencies {
    implementation(souther.output)     // this is what orders compileSoutherJava before compileKotlin
    "southerAnnotationProcessor"("org.souther-lang:souther-compiler:$southerVersion")
    "southerAnnotationProcessor"("org.souther-lang:souther-runtime:$southerVersion")
    "southerImplementation"("org.souther-lang:souther-runtime:$southerVersion")
}

tasks.jar { from(souther.output) }     // without this, the jar has no generated classes in it
```

Omit `inputs.dir` and a source edit produces the previously generated classes. Omit `jar { from }`
and the jar is empty — the same failure this issue opens with, arrived at a second way. Both builds
succeed.

[...]

PR #713, opened 2026-08-14
Run a build's compile without that build linking the compiler

kawasima:
The driver half of #137: `souther-build-driver`, a module that runs a compile on behalf of a build
plugin, and the reactor wiring for it.

## What it is

A build plugin that calls `Compiler` directly has to be released whenever the compiler's entry
points move, and it decides for a project which Souther it may compile with. This is the other
side of that split:

```
souther-maven-plugin ──┐                       ┌── souther-build-driver:${southerVersion}
                       ├── souther-build-api ──┤        souther-compiler:${southerVersion}
souther-gradle-plugin ─┘   (linked)            │        souther-runtime:${southerVersion}
                                               └── resolved at build time, loaded in isolation
```

`souther-build-driver` ships with the compiler and at its version. Its dependencies are what a
Souther release needs to run a compile, so resolving this one artifact resolves the toolchain and
no plugin holds a list that could fall behind it.

`souther-runtime` is compile scope here, unlike in `souther-compiler` where it is provided: a
compile evaluates the module's examples, which runs generated bytecode, and that bytecode calls the
runtime.

## What crosses the boundary

Text and a verdict. The classes are written by the driver, and `CompileException` is caught and put
through `DiagnosticRenderer` here, the way `SoutherProcessor` does it. No compiler type reaches a
plugin, so a change to one is not a change a plugin release has to follow.

`META-INF/souther-build-protocol` states the protocol this was built against. A plugin reads it
before loading a class, so a Souther it cannot drive is refused by number rather than by an
`AbstractMethodError` at the first call.

## Versions

`souther.build.api.version` is named in the root pom like any other outside dependency, because it
is not this project's version and is not tracked to it. `souther-build-api` is released from
`souther-lang/souther-build-api`, and the two plugins from their own repositories: four things that
release for four reasons.

`SoutherProcessor` stays. It is the integration that needs no plugin.

## Verified

Eight tests over the driver: a module compiles to the output directory, an error comes back
rendered rather than raised, a warning comes back and the build goes on, the request's language is
the language of the diagnostic, a lone header-less source is compiled the way the processor
compiles it, an import resolves against the request's class path, the stated protocol matches the
API compiled against, and the driver is found over the interface by `ServiceLoader`.

The plugins in the other three repositories build real projects against this. On both, a project
whose `src/main` holds only `.sou` builds a jar with its generated classes in it, and editing a
`.sou` and rebuilding produces the edited model.

Part of #137. The plugin repositories carry the rest, so this does not close it.


kawasima:
Both implementation findings are fixed, and both were reproduced first.

**Stale output.** Confirmed in a real Maven build before touching anything: renaming `shared.money`
to `shared.wallet` and rebuilding without a clean left all eleven of the old module's classes in the
jar, `shared/money/$Module.class` among them. A downstream `import shared.money` would have resolved
against a module no longer written anywhere.

The output directory cannot be emptied first — on Maven it is javac's too — so the driver now
records what it generated in a state directory the request names, and removes what it no longer
generates, one file at a time, with the directories left empty. Anything the compile did not write
is left alone; a test puts a hand-written class beside the model and checks it survives. Verified
again in the same real build: after the rename, only `shared/wallet` is there.

This is a protocol change (`BuildRequest` gains `stateDirectory`), which is free only because
protocol 1 has not been released. Both plugins pass one — `target/souther` on Maven, a declared task
output on Gradle, since a task's output directory is not emptied between runs either.

**Header-less source and the class path.** Also reproduced: `import shared.money` in a header-less
source failed with E1504 even with the module on the request's class path. You are right about
ADR-0043 — what it withholds from a header-less source is a name others can import, not the ability
to import. `Compiler.compiled(source, name, warnings, path)` is the entry that was wanted. The test
comment claiming such a source "can import nothing" was reading the ADR backwards and is gone.

**API 1.0.0.** Agreed, and not something I can do from here: it needs a real deploy to Central. The
property is still `1.0.0-SNAPSHOT` and this should not merge until the API is released and the
property points at it.

**The test gap you note is real.** `DriverIsFoundOverTheProtocolTest` puts the API and the driver in
one loader, so it does not exercise the production arrangement. What covers that today is
`DriverLoaderTest` in souther-build-api, which loads a stand-in driver from a package beneath the
API's with a copy of the API beside it, and the plugins' own tests against the real driver jar. The
CI job running the plugin integration tests against this repository's SNAPSHOT is still not built.

kawasima:
CI is green on `4327302`, and the earlier red was not the Central timing.

`EverySurfaceThatAnswersAReaderNamesItsLanguageTest` was failing: it walks the tree for every place
that picks a diagnostic language and holds the result against a named set, and the driver is a
fourth surface that answers a reader — a build log, resolving from what the plugin was asked for.
That test says a new one is "a line here rather than a fact nobody wrote down", so it is a line
there now, and the count in its javadoc moved with it.

It had been failing since the driver landed, on both pushes. I had run `-pl souther-build-driver`
and not the reactor, which is what a new module in this tree needs: its own tests stay green while
a registry test in another module goes red. `mvn -B verify` over the whole reactor passes now, and
so does the CI run.

Both fixes are folded into the commit that introduced the surface, so each commit on the branch
builds on its own.

Issue #718, opened 2026-08-14
A row is identified by where it is written, so nothing outside the file can key on it


kawasima:
A row is identified by where it is written. `RowOutcome.at` is a `SourceRef` — a source id and a
position — and the row's own name is optional:

    @param description    the row's business-case name, or null

Nothing requires a row to carry one, and nothing requires two rows of one behavior to carry different
ones. That is enough for what reads it today and not enough for anything that keys on a row.

## What reads it today

Two places, and both are fine with null and with repetition.

    // examples/ExampleVerifier.java:1053
    if (row.description() != null) {
        b.hint(new ExampleMessage.WhatTheRowSaid(row.description()));
    }

A hint on a diagnostic that is already anchored at the row's position, so the description is
decoration on something already located.

    // examples/Deadline.java:44
    record Row(String target, String sourceId, SourcePos pos, String description)

A label for the work a deadline is measuring, carried beside the position that identifies it.

In neither case is the description the key. The position is.

## Where the position stops working

A position identifies a row within one revision of one file. Add a line above a row and the row is
the same row, written the same way, meaning the same thing, under a different key.

That is fine while nothing outlives a revision. It stops being fine as soon as anything is attached
to a row from outside the file:

- A report line that names a row, so that two runs can be compared. ADR-0089's evidence accumulates,
  and evidence about "the row at 42:5" is evidence about a different row after an edit.
- A generated test named after a row.
- Environment setup attached to a row — the case #695 runs into, where a test has to prepare a
  database for the row that expects a stored todo and not for the one that expects nothing stored.

In every one of those the description is what an author would reach for, and today an author can
write two rows with the same description, or none, and nothing says anything.

## The shape of the fix

The options are not equivalent and the choice is the substance of this issue.

**Require a description on every example row.** A row then always has a name, and the name is what
reports and keys use. It is the strongest and the most disruptive: every row in the repository and
in souther-examples grows one, and rows written to cover a mechanical boundary get names that say
nothing.

**Require uniqueness where one is written.** Cheaper, and it leaves a row that legitimately has
nothing to say without a name — but then a keyed reader still has to handle the unnamed case, and
the failure it must not have is the silent one: keying on null and finding nothing, which reads as
"no environment was needed here".

**Give a row an ordinal within its behavior.** Stable across edits elsewhere in the file, derived
rather than written, and unreadable in a report — `findTodo#2` names nothing a human recognises.
It could stand beside a description rather than instead of one.

[...]

Issue #695, opened 2026-08-12
An injected behavior's recorded rows and stand-ins are run against a bound implementation


kawasima:
A model declares the behaviors it does not implement. A repository behavior is
injected, its answer comes from SQL, and the SQL is written in Java long after
the model is. Everything else in the model is held to something — a `let` body
is run by its rows, an invariant is checked where a value is built, a
composition is applied as the class it emits. The implementation of an injected
behavior is the one part that is never held to anything.

ADR-0088 already put the statement in place. A row on an injected behavior is
recorded rather than refused: its arity is checked, its inputs are built through
their derived decoders and held to their invariants, its expectation is held to
the output's cases. Only the running is missing, and the ADR's title says why —
*until it can be run*. This is the other half of that sentence.

What it buys is a way to hold generated SQL to the model. A coding agent writes
`FindTodoImpl`, and the rows already written for `findTodo` decide whether it
answers what the model says it owes. The reader is not a human reading SQL.

## The thing being tested is the SQL

A model states what it needs and does not say where it comes from:

    module todo
    import String ( length )

    data TodoId = Int
    data Title  = String
        invariant length(value) > 0

    data Todo     = { id: TodoId, title: Title, done: Bool }
    data NotFound = { id: TodoId }

    behavior findTodo : (id: TodoId) -> Todo | NotFound

    example findTodo
        | "a todo that is stored"        : (TodoId(1))
            -> Todo { id = TodoId(1), title = Title("write the SQL"), done = false }
        | "an id nothing is stored under" : (TodoId(99))
            -> NotFound { id = TodoId(99) }

That compiles today, and both rows end `PENDING`. `findTodo` emits an abstract
base and a sealed `FindTodoResult` permitting `Todo` and `NotFound`, and the SQL
is written against it:

    package todo;

    import org.jooq.DSLContext;
    import org.jooq.Record3;

    import static org.jooq.impl.DSL.field;
    import static org.jooq.impl.DSL.name;
    import static org.jooq.impl.DSL.table;

    public final class FindTodoImpl extends FindTodo {

        private final DSLContext db;

        public FindTodoImpl(DSLContext db) {
            this.db = db;
        }

        @Override
        public FindTodoResult apply(TodoId id) {
            Record3<Long, String, Boolean> r = db
                    .select(field(name("id"), Long.class),
                            field(name("title"), String.class),
                            field(name("done"), Boolean.class))
                    .from(table(name("todos")))
[...]

PR #721, opened 2026-08-14
Hold an example row's name to naming one row of its behavior

kawasima:
A row was identified by where it is written, so nothing outside the file could key on it: `RowOutcome.at` is a position, and the row's own description was documented as "a business-case name, or null" with nothing requiring one and nothing requiring two rows of one behavior to differ. This makes the written description a name, holds it to naming one row, and separates what a row names itself from where it sits. #718

## The rule

The string a row is written with is its name. A name is unique among the rows one behavior has, over the module's own source and every `examples for` file attached to it. A string with nothing in it is not a name and is refused where it is written. A row written without a name is held to none of this: it can be shown, and it cannot be named from outside. Editing a name renames the row — stated in the specification beside the rule, since something written elsewhere against the old name stops saying anything about this row.

Every row carrying a name more than one row carries is reported, each in the source it is written in (`E1925`). Neither of two colliding rows is the duplicate: which source came first is what a caller hands the compiler — the build driver walks its files sorted, a command line takes them as typed — so a rule naming one of the two would say different things about one model depending on how the compile was started. Reporting both needs no order at all, and it is what "checked where the row is written" means when the two rows are in different files.

## The type

`RowIdentity` is `Named(name)` or `Unnamed(ordinal)`. It carries no behavior: `RowOutcome` already keeps that, and an identity carrying it too would make a row and its behavior recordable as disagreeing about each other. `Named` cannot hold a blank, so no reader is handed a name that names nothing. `Unnamed.ordinal` is for showing and counts within one source. A namespace-qualified key (`RowKey(behavior, name)`, built only from a `Named`) is left to the reader that needs one — #695 — rather than shipped without a caller.

`Ast`/`Hir.ExampleRow`, `RowOutcome` and `Deadline.Work` carry the identity, so the nullable string is gone from every path. `Deadline.Work`'s note that what makes a row unique is where the writing is has been rewritten.

## The generator

`--generate` writes rows, so it is a producer of names. It named a row for everything the run handed it — candidates sharing an input spelling are one row, and their labels were joined — so writing one unrelated row renamed a row nothing about had changed:

```
before:  | "request.cost=0 <= x <= 100 x request.tier=Gold x request.cost = 0" : (Draft { cost = Amount(0), tier = Gold })
after:   | "request.cost=0 <= x <= 100 x request.tier=Gold"                    : (Draft { cost = Amount(0), tier = Gold })
```

What was written was a row for the *other* class, which met the line at the bottom of the range; the line stopped being offered and the label left the name. `--boundaries` did the same to the rows offered without it. A row is now named for what it was composed for — a cell of the partition, or a line — and what it settles beside that stays the generation's to change.

## Measured

- 498 rows across this repository and `souther-examples`: 487 named, no collision anywhere, and the 11 unnamed are one bench corpus file of mechanical rows. Requiring a name everywhere would have cost 380-odd rows in the `.sou` sources inside Java tests, which is why unnamed rows stay legal.
- `souther-examples` compiled with this build: no `E1925` and no `E2304`. Duplicating one row in `issuetracker` reports both rows at their own positions, so the zero is a zero and not a check that never fires.
- The whole red set was run against the implementation removed: 5 of the 7 fail, and the two negative controls (a name shared by two behaviors, several unnamed rows) pass either way. The two generator tests fail with the old union naming restored.
- Full suite green; each commit compiles on its own.

## What is not here

`fake` rows, which are identified by their inputs. Report lines named by a row rather than a position, and an API that resolves a name to a row: both wait for a reader that compares runs (#695), rather than shipping a field nobody reads. No ADR — this is not an architecturally significant decision, and the rule and its rename semantics live in the specification and in `RowIdentity`'s own javadoc.


kawasima:
Both findings fixed in 0b2844f6.

**P1.** Reproduced first: the POLICY-shaped model offers one row for the two minimum edges, and writing a row that meets `policy.cap = 0` renamed it.

```
before:  | "policy.cap = 0"  : (0, Policy { rate = Rate(0), cap = Cap(0) }) -> <?>
after:   | "policy.rate = 0" : (0, Policy { rate = Rate(0), cap = Cap(0) }) -> <?>
```

A met line is `NotAttempted(A_ROW_IS_ALREADY_THERE)` and composes no candidate, so the set of line candidates at this layer *is* a function of what the rows already cover — no aggregate over it can be run-invariant, and taking the first by name only wrote the choice down. A line row is now offered without a name. A cell still names the row it composes, which is invariant: two cells never compose one row, since a candidate's values follow from the classes it was composed for.

The regression test is the one you specified, with the row identified by its values rather than by a name: the row two lines compose is offered unnamed, and stays unnamed and identical when a row meets one of the two. It fails against the shipped min-over-lines rule (measured, not assumed).

**P2.** `RowIdentity.of` now takes `null` to `Unnamed` and hands anything else to `Named`, which refuses a blank. Pinned by a test on the factory. Reading a blank as "written without a name" would have carried a row the parser refused past its refusal.

Three CLI tests were reading which edge got a row out of that row's name; they ask the values now. Full suite green (4372 in souther-compiler).

Issue #717, opened 2026-08-14
A row's outcome does not say what applied the behavior


kawasima:
`RowOutcome` records what a row turned out to be and does not record what produced the answer. Today
that is not visible, because there is one answerer: bytecode this compile generated, applied through
the `MemoryClassLoader` the run builds. The field would be a constant. That is the reason to add it
now rather than the reason not to.

## What the outcome says and does not say

    public record RowOutcome(SourceRef at,
                             String target,
                             String description,
                             Stage stage,
                             Disposition disposition,
                             FailurePhase failurePhase,
                             TypeSymbol expectedArm,
                             TypeSymbol resultArm,
                             List<TypeSymbol> inputCases,
                             List<ObservedValue> inputs,
                             Set<Integer> hits,
                             long stepsSpent)

`resultArm` is "the case the behavior answered with". Which behavior, and written how, is in
`target`. What ran it is nowhere.

## Three rules are already written around a distinction the outcome does not carry

`EvaluationPolicy` splits its four numbers on exactly this line, and says so:

> The first two are what decide a row ... The last two are not about the model at all.
> `outerTimeout` is the wait after which the evaluation is given up on, and it exists for what a
> counter cannot reach — a call into code this compile did not generate, so has no counted points in.

So the policy already distinguishes counted code this compile produced from uncounted code it did
not. An outcome produced under the second has `stepsSpent` of zero and `hits` empty for a reason
that has nothing to do with what the row did, and the record's own javadoc says the disposition is
what tells a zero-cost run from a run that never happened. With a second answerer it will not.

ADR-0029 says a platform failure is an exception and Souther passes it through. `FailurePhase` has
two arms that would receive one:

    /** Applying the behavior did not produce an answer. */
    INVOCATION,

    /** Something the host was supposed to provide was not there. */
    INFRASTRUCTURE

Deciding between them for a throw means knowing whose code threw. A row that stopped because a
database was unreachable is not a row that stopped because the model is wrong, and today the outcome
has no way to have been told which it was.

ADR-0089 accumulates `specified`, `observed` and `verified` rather than switching measures when an
implementation appears. A row that held against a `let` body and a row that held against an
implementation supplied from outside the compile are both `COMPARED`/`HELD`. They are not the same
evidence, and a report reading the disposition alone would say they are.

## What this is not

Not a source of truth about how a behavior is written — `ExampleVerifier.isPending` answers that from
the module, and adding a second answer to it is what this must not become. This says what the run
[...]

PR #725, opened 2026-08-14
Say what applied a behavior, and keep its counted work distinct

kawasima:
A row's outcome recorded what the row turned out to be and not what produced the answer. With one answerer — the bytecode this compile generated, applied through the loader the run builds — that was invisible, and the numbers beside the outcome read as facts about the row. They are not: how many counted points an application passed is a question about code this compile emitted the counting into, and there is no answer to it for code this compile did not write. #717

## What is recorded

`RowOutcome` carries a `Run` in place of its `hits` and `stepsSpent`, and a run is two axes of one evaluation:

```java
record Run(Applied applied, Counting counting) {}

sealed interface Applied {
    record Nothing() implements Applied {}
    record GeneratedHere() implements Applied {}
}

sealed interface Counting {
    record Read(long steps, Set<Integer> hits) implements Counting {}
    record Unread() implements Counting {}
}
```

They are apart because a row's evaluation is not only its application. The counting starts with the evaluation and a fixture applies the helpers it names before the behavior is reached, so a row whose input fixture costs eleven counted points and then breaks its invariant applied nothing and did eleven points of counted work — measured, and the reason the first version of this PR was wrong to put the numbers inside the applier.

What applied the behavior follows from the stage it reached, and the outcome refuses to be built saying otherwise: the two are written from one evaluation and read apart, so a row saying it applied the behavior and saying nothing applied it is a state no reader should have to resolve. `Stage.INVOKED` is set before control passes to the application, so a throwing invocation records `GeneratedHere` as it should. What was counted is not held to the stage.

A counting is a reading or no reading, for the same reason the applier is its own axis. An evaluation given up on is still running when the outcome is written, so nothing is taken from it — recording zero there would say what a row that passed no counted point says, and the number would need a second question again. Neither the disposition nor the phase settles it: a row that spent its step budget is incomplete too, and its count was read.

The four readers of the arms (`Adequacy` twice, `Coverages` twice) ask for the sites a row is known to have gone through, through a `switch` over `Counting` in each file, so an arm neither knows about is a compile error rather than a row silently counted as having lit nothing. None of them consults `Applied`, which is the right shape — they are about counted work. No compatibility accessor is left on the outcome.

## What is not recorded

`Applied` has no arm for an implementation supplied from outside the compile. Nothing outside a compile applies a behavior today, and an arm nothing writes is a statement about a run that does not happen — the same reason `FailurePhase.INFRASTRUCTURE` is not given a producer here. There is no reader of `Applied` in main today either, so nothing enumerates its arms yet; its javadoc states the contract for when one appears, which is a `switch` over the type rather than a test for one arm, so that the arm a reader did not consider is a compile error.

`FailurePhase.INFRASTRUCTURE` is untouched. Beyond having no producer, what applied a behavior does not say whose fault a throw was: an implementation supplied from outside can throw because the host is missing or because it is wrong, so a classifier reading this instead of the exception's type would have swapped one guess for another. That belongs with the invocation boundary #695 introduces.

## Acceptance

The issue's acceptance was corrected before this PR — 2 rewritten, 3 removed — with the reasoning recorded [in a comment](https://github.com/souther-lang/souther/issues/717#issuecomment-5292509238).

## Measured

- `hits` has four readers in main sources; `stepsSpent` and `failurePhase` have none outside tests. That is why the numbers could move into the arm that defines them instead of being kept beside the outcome.
- `FailurePhase.INFRASTRUCTURE` has zero producers; every other arm has at least one.
- A row whose fixture applies a helper and then breaks its invariant records `stage=NONE`, `phase=INPUT_FIXTURE` and eleven counted steps. That is the case that decided the shape.
- The new tests were run against builds that never name an applier, that record the count only when something applied, and that write zero where nothing was read: each fails the assertions it exists for, so they measure the change rather than the model.
- Full suite green (4393 in souther-compiler).


kawasima:
All four addressed in eb1234b5. 1 was right and it changed the shape.

**1.** Measured before changing anything. A model whose input fixture applies a helper and then breaks the invariant it is built against:

```
stage=NONE  disposition=FAILED  phase=INPUT_FIXTURE  counted steps=11
```

Eleven counted points spent, nothing applied. The shipped version recorded `NotRun` and discarded them, and would go on discarding a bound implementation's fixture cost for the same reason. So the two are axes of one record now:

```java
record Run(Applied applied, Counted counted) {}
sealed interface Applied { record Nothing() ...; record GeneratedHere() ...; }
record Counted(long steps, Set<Integer> hits) {}
```

The stage is held to `applied` and says nothing about `counted` — a row that stopped at `NONE` spent whatever its fixtures spent. The regression test you asked for is `aRowThatAppliedNothingStillSaysWhatItsFixturesSpent`, and it fails against a build that records the count only when something applied (measured, not assumed).

**2.** The premise was mine and it was wrong, but the corrected model removes the site rather than the enforcement: those four readers ask `run().counted().hits()` and never consult `Applied` at all — they are about counted work, which is exactly what they were reading. There is no reader of `Applied` in main today, so there is nothing to enumerate exhaustively yet. `Applied`'s javadoc states the contract for when one appears: a reader that has to tell the arms apart writes a `switch` over the type rather than a test for one arm, so the arm it did not consider is a compile error. I have corrected the claim in the PR body.

**3.** Closed. `Objects.requireNonNull(run)` runs before the check, and the test asserts it.

**4.** The `INFRASTRUCTURE` line is out of the table of what arises, with a paragraph saying nothing writes it and that the host-runtime fact is module-level `Incompleteness`. The phase itself is untouched.

Full suite green (4393 in souther-compiler).

kawasima:
Fixed, and the title now says what the change is.

`Counting` is a reading or no reading:

```java
sealed interface Counting {
    record Read(long steps, Set<Integer> hits) implements Counting {}
    record Unread() implements Counting {}
}
```

The row given up on records `Unread`. You are right that the disposition does not settle it — `STEP_LIMIT` and `DEPTH_LIMIT` are incomplete too and their counts were read — so the javadoc's claim was false as well as the value being ambiguous; both are gone.

Both readers of the arms now take the sites a row is known to have gone through through a `switch` over `Counting`, one helper per file, so a third arm is a compile error there. That is also the shape your earlier point asked for, in the one place that has readers today.

Measured: with the timeout path writing `Read(0, {})` again, `CompileExampleCoverageTest.aRowThatRanOutOfTimeCountsForNothing` fails. Full suite green (4393 in souther-compiler).

PR #726, opened 2026-08-14
Read nothing off the type a name that denotes nothing leaves behind

kawasima:
A type name that denotes nothing is reported where it is written, and what it names becomes the type that absorbs so the rest of the module is still checked. Two rules then read that type as though it were one an author had written, and each said a second thing about the one mistake. This is #672.

`TypeOps.memberName` answered `null` both for a type no arm can name — a list, an option, a function, `Raw` — and for a type nothing could be worked out for. The union rule read the one answer and reported the second sentence about the first: that `?` cannot be a union member, naming a spelling the author never wrote and could not write.

```
5:32 [E1506] `up` declares no `Refused`.
5:26 [E1613] `?` cannot be a union member: a member is a type a `match` arm can name — …
```

The other rule is quieter. The leaf cases of the type that absorbs are the empty set, so a declared composition output resting on such a name was held against what the composition produces as a declaration naming no case at all, and the sentence saying so named nothing:

```
13:8  [E1506] `up` declares no `Refused`.
12:1  [E1604] behavior go declares its output as , but the pipeline produces A.
```

## What changed

`memberName` gives three answers — the case name, a type no arm can name, a member resting on a name that denotes nothing — and a reader has to say which it is acting on. `successType` reads the members: the first no arm can name is reported as it always was, and finding one that rests on an unresolved name does not end the reading, because a member the author owns may be written after it. Such a member makes the whole output the type that absorbs, which is the same answer a lone such case already gave.

The two checks that hold a declared output against what is produced — `SpecChecker.checkExposedPipeOutputs` and `PipelineSigs.signatures` — read the declaration first, so that what is wrong with the union itself is still said, and then ask `restsOnAnUnresolvedName` whether there is a case set to compare — abandoning the definition if there is not, which is what is already done for a name a body rests on. Both sit inside an `Unanswerable` boundary already, so no second recovery is introduced.

`leafCases` keeps the contract it had: of its callers only those two read a declaration an author wrote, and the rest read a sum this compilation already knows. `namesOf` keeps it too — it asks which names are on a type, not why a type has none.

## What it is measured against

The sweep over the positions a type is written in is a test of its own. It crosses them with the name written alone, written as one member of a union, and written beside a member no arm can name — that one in both orders, because a reading rewritten to stop at the unresolved member passes one order and fails the other. What each cell must answer is the codes and not the count, because a cell answering with one diagnostic that is no longer the name has stopped asking the question. Every cell is also asked whether anything it says names the type that absorbs, looked for where a type begins rather than anywhere in a value, so that `A | ?` and `List<?>` are caught and `Int?` is not.

Against the commit before the fix, eleven cells report a second thing: nine `E1613`, and two `E1604` with an empty declaration. Against the commit before the ordering was corrected, exactly the two composition cells lose the `E1613` they should keep. The cells where a union cannot be written are derived from what the same form naming something answers, so a position that comes to admit one is swept without being added anywhere.

One case is held beside the sweep: a composition over a stage whose output rests on such a name, where what would read the type left behind is the routing beside it. That one says only the name, because a signature holding the type that absorbs is abandoned where it is built, and it is measured here rather than reasoned about.

The list of positions moves to `TypePositions` in the first commit, so the two sweeps over them read one list rather than a copy each.

Every module's tests pass. `souther-examples`, all fifteen projects through `examples --generate --boundaries`, is 21,433 lines identical to the same run on a jar built from `develop`.

## Not in this change

The issue's second pair — `E1818` on a helper that calls a behavior, then `E1006` on the behavior that calls the helper — was investigated and is not a cascade. A behavior's `let` may call a behavior with an empty requirement set, so the direct call compiles with nothing reported; and a behavior's `constructs` is its own and adds nothing to its caller's, so `f` does not build `Y` whether or not the helper is in the way. Both spellings are measured. `E1006` is true on its own and its hint is the right one to follow. The issue body records this so it is not opened again.

### DIFF
--- a/souther-compiler/src/main/java/souther/compiler/check/PipelineSigs.java
+++ b/souther-compiler/src/main/java/souther/compiler/check/PipelineSigs.java
@@ -170,11 +170,15 @@ public final class PipelineSigs {
         // §declared-composition-output): neither a missing case (too narrow) nor an extra one (too wide) is
         // accepted.
         if (pipe.declaredOut() != null) {
+            // What was written is read first, and whether it can be compared with what is produced
+            // is asked of the reading. A member no arm can name is a mistake in the declaration
+            // itself, and it is the author's whether or not something beside it went unresolved.
+            Type declaredOut = TypeOps.successType(pipe.declaredOut());
             if (TypeOps.restsOnAnUnresolvedName(pipe.declaredOut())) {
                 throw new Unanswerable(pipe.declaredOut().pos());
             }
             Set<TypeSymbol> inferred = TypeOps.leafCases(out, symbols);
-            Set<TypeSymbol> declared = TypeOps.leafCases(TypeOps.successType(pipe.declaredOut()), symbols);
+            Set<TypeSymbol> declared = TypeOps.leafCases(declaredOut, symbols);
             if (!inferred.equals(declared)) {
                 throw CompileException.of(Diagnostic.at(pipe.pos())
                                 
--- a/souther-compiler/src/main/java/souther/compiler/check/PipelineSigs.java
+++ b/souther-compiler/src/main/java/souther/compiler/check/PipelineSigs.java
@@ -170,6 +170,9 @@ public final class PipelineSigs {
         // §declared-composition-output): neither a missing case (too narrow) nor an extra one (too wide) is
         // accepted.
         if (pipe.declaredOut() != null) {
+            if (TypeOps.restsOnAnUnresolvedName(pipe.declaredOut())) {
+                throw new Unanswerable(pipe.declaredOut().pos());
+            }
             Set<TypeSymbol> inferred = TypeOps.leafCases(out, symbols);
             Set<TypeSymbol> declared = TypeOps.leafCases(TypeOps.successType(pipe.declaredOut()), symbols);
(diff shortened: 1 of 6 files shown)


kawasima:
The blocking finding reproduced. `List<Int> | up.Nope` written as a composition's declared output said only the unresolved name, while the same union written as a behavior's own output said both — so the rule this change states held everywhere except the two positions a composition's output is written in.

Fixed by reading the declaration before asking whether it can be compared: `successType` raises what is wrong with the union itself, and `restsOnAnUnresolvedName` then decides whether there is a case set to hold against what is produced. Both checks, same order.

The sweep carries the mixed union rather than the single case that held it at one position, in both member orders — `List<Int> | up.Nope` and `up.Nope | List<Int>` — because a reading rewritten to stop at the unresolved member passes one order and fails the other. Against the previous commit it fails exactly the two composition cells and no others. The standalone case is gone, since every position holds it now.

The comment on `successType` said every member is read before anything is said, which overstated it: the first member no arm can name is still reported at once, as it was before this change. It now says that, and says why finding an unresolved member does not end the reading.

Every module's tests pass, and `souther-examples` is still 21,433 lines identical to the run on a `develop` jar.

Issue #716, opened 2026-08-14
A fake table may contain a row its dispatch can never answer with


kawasima:
A `fake`'s table dispatches: the first explicit row stating the arguments, and otherwise the `_`
row. Two explicit rows may state the same arguments, and when they do the second one can never be
what the table answers with. Nothing says so.

## What is true today

`ExampleStatements.Standins` holds the rows and owns the rule
(`examples/ExampleStatements.java:762`):

    Standin answering(Object[] arguments) {
        for (Standin standin : explicit) {
            if (sameArguments(standin.arguments(), arguments)) {
                return standin;
            }
        }
        return fallback;
    }

`standins(...)` builds `explicit` in written order (`:812`). It walks the rows once to refuse a wrong
arity, then builds each row's answer and its arguments, and appends. Nothing compares one row's
arguments with another's, though `sameArguments` — the equality that would decide it — is already
there and is the one the dispatch uses.

So this builds, and the second row is dead:

    fake findMember
        | ("m-1") -> Found { id = "m-1" }
        | ("m-1") -> Missing { why = "none" }

`answering(["m-1"])` returns the first row. The second answers nothing, for any input, ever.

## A `match` already refuses this

    @Code(DiagnosticCode.E1204)
    record MatchedByMoreThanOneCase(String caseName) implements MatchMessage, Reported {}

Two arms matching one case is an error, and E1204 is not in the warning set. A fake's table is the
same shape of dispatch over a different key — arguments rather than cases — and says nothing.

## The listing and the resolver disagree

`explicit` is a list a reader can walk. `answering` is the rule that decides which of it is used.
Today they disagree whenever a row is shadowed: the list holds a row the rule can never return.

ADR-0093 is where this matters, because it built the stand-in/recorded-row comparison on the table's
own dispatch rather than row-for-row, and said why:

> A check that compared row-for-row instead would report a contradiction where the fake would have
> picked an explicit row, and stay silent where the fake would have fallen through to its default —
> which is the same class of mistake this diagnostic exists to catch.

A shadowed row is that mistake sitting inside the table. The comparison itself is safe — it goes
through `answering` — but any reader that enumerates `explicit` is not, and #695 wants to enumerate
it: running a bound implementation against each entry of a fake's table would report a disagreement
about a value the fake would never have answered with.

It is also an authoring mistake on its own terms. A row written and never reached is a statement the
author believes is being made and that nothing reads.

## The choice this needs

[...]

PR #727, opened 2026-08-14
Refuse a row a fake's table can never answer with

kawasima:
A `fake`'s table dispatches to the first explicit row stating what it is asked, and otherwise to the `_` row. Two rows can be written and never reached, and the table said nothing about either. #716

## What was measured

Six orderings, each read by which answer the fake gave:

| table | what answers |
| --- | --- |
| `_` then an explicit row | the explicit row |
| an explicit row then `_` | the explicit row |
| two explicit rows of one input | the first |
| two `_` rows | the last |
| three `_` rows | the last |
| an explicit row for another input, and `_` | the `_` |

The issue set the `_` row aside — it is held apart in `Standins(explicit, fallback)`, so it looked out of scope. It is not: `standins(...)` overwrites `fallback` as it walks, so the earlier `_` rows answer nothing, with the opposite precedence to the explicit case. Closing only the explicit half would have left one silent unreachable row still legal, so the issue's title and acceptance were generalised first and the note excluding `_` retracted, with the measurements [recorded there](https://github.com/souther-lang/souther/issues/716#issuecomment-5292978191).

## What decides it

The dispatch, asked of the table the row is in — whether the table answers this row's own arguments with this row:

```java
List<Standin> with = new ArrayList<>(explicit);
with.add(written);
Standin answers = new Standins(with, latestDefault).answering(arguments);
```

Asking the rows accepted *before* the row would be wrong, and the first measurement is what catches it: a `_` written above an explicit row answers where that row is absent, not where it is, so any answer from the rows-so-far is not a shadow. There is no second reading of "these arguments are those arguments" and none of which of two rows wins.

For `_` rows there are no arguments to ask with, and the winner is the last one written, so every earlier `_` is reported against it rather than against the row written after it — which for three of them would name a row that answers nothing either.

## What is reported

`E1926`, at the row that can never answer, quoting the row that answers instead. The two shapes point opposite ways, so which of a pair is dead is not something an author can read off the order, and the report says it.

An error. Which row answers is decided by the table's own rule rather than by anything a run finds out, so it belongs with `E1204` (an arm no `match` can reach) and `E1908` (a table that will not build) rather than with the `E19xx` warnings. The issue's precedent was slightly off — the refusal of an unbuildable table is `E1908`; `E1921` is a warning about a table that could not be read in time. Across this repository and `souther-examples` there are 36 fake tables and none has a repeated row, so refusing costs nothing that exists.

## What the table holds

`standins(...)` now answers with the dispatch and the rows the dispatch cannot reach, apart:

```java
record BuiltTable(Standins standins, List<Shadowed> shadowed) {}
record Shadowed(Standin row, Standin answeredBy) {}
```

`Standins` is the fake's semantics, and every row in it is one `answering` can return — so a reader that enumerates `explicit` is enumerating answers the fake can give, which is what #695 will do. What was written and cannot be reached is a fact about the source and is the reporting's alone.

## Measured

- The six orderings above, before and after: what answers is unchanged, and the two dead shapes are now reported at the row that cannot answer.
- With the detection removed, four of the seven new tests fail — they measure the change rather than the model.
- Full suite green (4401 in souther-compiler).


kawasima:
Both fixed in ac15b6a2.

**P1.** Reproduced first — a dead row whose output breaks its invariant reports `E1908` and never reaches `E1926`. The table is now built in two passes: what each row states is read and the dispatch decides which rows it can return, then the answers are built for those rows and no others. A row whose *arguments* will not build is still `E1908`, since there is nothing to decide the dispatch with.

The rule is one method over what a row states:

```java
static <T extends Stated> T answering(List<T> explicit, T fallback, Object[] arguments)
```

so the pass that decides which row answers and the dispatch a fake runs on are the same walk over the same equality. `Standin` is what states its arguments and the answer it was built into; `Written` states them and nothing else. `Shadowed` carries `Hir.FakeRow` on both sides, as you said — nothing of an unreachable row is built, so there is nothing else of it to carry.

The test is `whatAnUnreachableRowAnswersIsNotBuilt`, with a positive control: the same output in a row the table *can* reach is `E1908`, so it is not passing because the output happens to build. Building the answers up front again turns it red.

**P2.** Rather than assert the invariant, `Standins` holds itself to it: a table whose listing contains a row its rule can never return refuses to be built.

```java
public Standins {
    for (int i = 0; i < explicit.size(); i++) {
        if (answering(explicit.subList(0, i), null, explicit.get(i).arguments()) != null) {
            throw new IllegalArgumentException(...);
        }
    }
}
```

Your regression — `explicit = [first, deadSecond]` with the dispatch and the diagnostics unchanged — is now unconstructable rather than untested. `ATableHoldsNoRowItCannotAnswerWithTest` in the `examples` package pins it directly, building the table by hand from two rows stating one input.

Full suite green (4404 in souther-compiler).

kawasima:
Fixed in 6cc23ad6, and the phase separation is gone.

Which `_` the table falls through to is the last one written, which is read off the rows without building anything. With that known, the rows are read in the order they are written and each as far as it is reached: what it states, then — for a row the table can return — what it answers. So nothing of an unreachable row is built and a row written first is still read first.

Measured both ways, with a first row whose answer breaks an invariant and a second whose arguments will not build:

```
source order:  E1908 why=id: must be at least 1 characters   (the first row's answer)
all arguments first, answers after:  E1908 why=expected long (the second row's arguments)
```

`aRowIsReadWhereItIsWritten` pins the first and turns red under the second.

On the `Object[]` in `Standin`: you are right that the invariant is enforced at construction rather than owned by the value, and that mutating the array afterwards would break it. Both types are package-private and the arrays are built inside `standins(...)`, so nothing today can reach one — I have left it as it is rather than copy on the hot dispatch path, and it is worth doing when `explicit` is handed to a reader outside this package, which is #695's move to make.

Full suite green (4405 in souther-compiler).

PR #729, opened 2026-08-14
Apply the one instance a fixture's call settles

kawasima:
A fixture refused a helper for being written with type variables, which reads the declaration and not the call: `List.length([ 1, 2 ])` was refused although the row had written what `'a` is. #692

## The rule

The written arguments settle the variables the declaration carries, and the row applies the one instance that comes out. Every variable closes, or the row is refused for the one that stayed open — and the refusal names it.

| written | what happens |
| --- | --- |
| `List.length([ 1, 2 ])` | `'a := Int`, applied at that instance |
| `List.isEmpty([ 1 ])` | `'a := Int`, applied |
| `List.isEmpty([ ])` | refused: `'a` is left open |
| `count([ 1, 2 ])`, `let count (xs) = List.length(xs)` | settled the same way; a Souther-written helper is not a different rule |

Nothing is read from the position the call stands at. A call is applicable where it is written or nowhere, so the same call does not become admissible by being moved into a row with a different expected type. Reading the position too is a coherent design — it is bidirectional checking — but it needs a fixture expression that goes through elaboration, and today one never does. That is the condition under which the question comes back, and it is written where the rule is.

## The layers

```
FixtureArgumentTypes   the type each written argument is already known to be
FixtureApplication     the variables it settles, and whether they all close
SettledCall            the witness that this call determines one instance
FixtureCallable        where that instance's method is
```

`FixtureArgumentTypes` reads no application: what a helper answers with is its answer's to say, and deriving it from the declaration here would be a second reading of a call the language already elaborates one way. Where a writing states no type it answers null, and the caller asks nothing more of that — a variable nothing settled is refused by name either way.

`FixtureApplication` runs `TypeOps.bindVars`, which is what settles a polymorphic call anywhere else, and holds no inference of its own. `SettledCall`'s constructor is private to it, so the witness cannot be minted anywhere else and nothing downstream re-checks what it stands for.

`FixtureCallable` takes only the witness. It chooses a representation and answers no question about admissibility, which is what keeps a fact about the backend from deciding which rows compile — the shape #680 removed once already.

## What was measured

**A kernel's method has to be written at the settled instance.** `List.sum` is declared `(List<'a>) -> 'a` and its lowering chooses between two runtime methods by what it is summing. Emitting the wrapper at the declaration instead was tried: the wrapper's own body is then `List.sum` over `'a`, which is `E1817` where it is checked. So the one-erased-method design fails, and `aRowAppliesAKernelWhoseLoweringReadsTheTypeItIsAppliedAt` is the control that says so.

**A helper written in Souther is reached through the method it already has.** The other half of the same measurement: `let total (xs) = List.sum(xs)` is refused with `E1817` where it is written, so a Souther body cannot carry an open variable into a kernel that chooses by type, and the one method such a helper has never has that choice to make. This is a representation decision resting on a compiler invariant rather than on a preference, and the test holds the invariant rather than the decision.

**Identity is the signature, not the substitution.** Two rows settling `List<Int>` share one method; summing `Int` and summing `Decimal` are two. Both are read off the emitted bytecode.

**souther-examples is unchanged.** 21554 lines of `examples --generate --boundaries` across every module, diffed against a jar built at this branch's base — zero difference. No example there exercises the new rule, since a row that did would not have compiled before.

## What else changed

`HelperInliner.decidedByEachCall` is gone, with the emission-side reading that shared it. Emission now settles the same calls the fixture does, so `Bodies.ExampleHelpers` and `Prepared.prepare` take the scope that needs, and a call the fixture admits is one there is a method for by construction rather than by two readings agreeing.

`FixtureShape` no longer refuses a type variable with a sentence about fixtures. Three kinds of type reach that walk — a name a row wrote, a parameter of a settled call, and a position inside one of those — and a boundary does not come through it at all, so a variable arriving there is this compiler disagreeing with itself. It says that instead, with the reason written down: putting the diagnostic back would put back the rule this removes.

One ordering changed as a consequence. Applicability is decided before any argument is built, so a call with both an argument that will not build and a later parameter that declares no type reports the second now. This is implementation order and is not stated in the specification.


kawasima:
Fixed the P1. The evidence a call settles from was the literal rules alone, so `List.length([ Amount(1) ])` was refused for leaving `'a` open — reproduced before changing anything.

What a declaration already says is now read for arguments too: a newtype's construction, a field, a name that stands for a body, and a `let`-bound name. An application of a helper stays unread, which was the exclusion that was meant.

The walk is the one the reading of a row already had, moved to `FixtureEvidence` in `check` so the pass that emits methods reads the same one, and `NeutralForm` asks it rather than keeping a second copy of what a data's fields are.

Three regression tests, each red before the fix:

| written | settles from |
| --- | --- |
| `List.length([ Amount(1) ])` | a construction |
| `List.length(sample.items)` | a named value and its declared field |
| `{ let one = Amount(1) List.length([ one ]) }` | a name a `let` binds |

The third is the control for the two walks agreeing: with the emitting walk not carrying what a `let` binds, the row settles and is then told no method was emitted for the call — the shape #680 removed. Measured by taking the binding out and watching that message come back.

souther-compiler 4442 green, `CI=1 mvn -Plint verify` green, souther-examples 21554 lines against the same base — no difference.

Issue #681, opened 2026-08-12
A closure cannot be evaluated in a fixture, which is what the derived-collection case actually wants


kawasima:
Split out of #666. Of the three things standing between a row and a derived
collection fixture, this is the one the reported use case actually needs, and
it is the one #666 did not name as a defect.

## The case

```
Ticket { lines = List.map(l -> Line { ...l, done = true }, sample.lines) }
```

Field projection is now admitted (#679), so `sample.lines` stands. Applying
`List.map` is refused as a standard-library call (#680). Even with that
opened, the lambda would be refused on its own terms:

```
`List.map` parameter `f` is ('a) -> 'b; what it holds is decided by each
call, and a fixture is built before a call can decide it
```

## Why this is the one that matters

The value that "the same ticket as `X`, but done" needs is a function applied
across a collection. Neither of the other two reaches it. Opening standard-
library calls without this admits `List.length` and stops.

## What is already possible

A `let` helper's body is Souther the module compiles, so a closure inside it
runs, and a row may apply the helper:

```
let markDone (t: Ticket) = Ticket { lines = List.map(l -> Line { ...l, done = true }, t.lines) }

example allDone
    | "a helper derives the ticket from a named value" : (markDone(sample)) -> true
```

That compiles today, and it was already the answer #666 was looking for. The
gap is one named definition, not an inexpressible value — worth saying plainly,
because #666 reported it as the latter.

## The decision this asks for

Whether a fixture holds a function value at all. Today a fixture is written
values and applications of rules this module compiled; a lambda in a fixture
would be neither. That is a change to what a fixture is, and it needs an
argument about what rows become easier to read and what stops being checkable,
not only about what stops being refused.

kawasima:
Closed by #737 (merged as `9ba01cb5`).

The title is not what this turned out to be. Closures were never the axis: a
named helper written in place of the lambda was refused the same way, and the
same refusal split into two different sentences depending on the callee's
signature. What was actually there is that a row's operands were read by a
second front end — a walk beside the one every definition of the module goes
through. `TypeChecker` never saw `module.examples()`; only `Resolve` did.

Each operand is now emitted as a definition of the module, into the same
program the behavior is applied in, so a row runs its own operand the way the
module runs its own body. The example in the issue holds, and so do the other
five shapes measured alongside it.

Two of those measurements are worth keeping: `howMany([1, 2, 2])` was admitted
[...]

PR #737, opened 2026-08-15
Run a row's operand as the module's own code

kawasima:
Fixes the reading behind #681.

## What was wrong

The issue is filed as "a fixture cannot evaluate a closure", but that is not
what the measurements found. A row's operands were read by a **second front
end**: a walk that decided what a fixture meant, running beside the one every
definition of the module goes through. `TypeChecker` never saw
`module.examples()` at all — only `Resolve` did.

Two readings of one language do not agree for long. Measured on develop:

- the same refusal split into two sentences depending on the callee's
  signature (`List.map` said `'b` was open, `List.filter` said a function type
  has no external representation)
- `Decimal.round(1, HALF_UP, …)` was refused for a codec rule, at a position
  where nothing crosses a boundary
- `howMany([1, 2, 2])` was **admitted in a row and refused in a body**
- a value the module can compute was rebuilt from the body's syntax rather
  than used, and failed

Passing a named helper instead of a lambda was refused too, so closures were
never the axis.

## What it does now

Each operand is emitted as a definition of the module — taking nothing,
answering with the value that operand is — into the same program the behavior
is applied in. A row runs its own operand the way the module runs its own
body. Nothing carries a value between the two sides and nothing converts one
representation into another, so what a row is held to is the language's own
rules: what a position takes is the type its wrapper declares (E1812), a field
is what its record declares (E1317).

Four things this needed:

- **The correspondence is constructed, not recomputed.** Which method belongs
  to which operand is built where the methods are emitted and read from there.
  Counting the rows a second time is what put a row on the operand beside the
  one it wrote — a run reads one file's rows out of a module that holds every
  file's, and a count over the subset numbers them from zero.
- **What a position contributes is not what it requires** (`RowPosition`).
  Both reach the operand as the type its wrapper answers with, so a row's
  brackets take their collection from the position as an empty collection in
  any body takes its element type. Only a supplied value is also required to
  be of that type: an expectation contributes and requires nothing, because
  stating what the behavior does not answer with is what a row is for.
- **Where a value stands comes from the signature, not from the syntax.**
  Positions are read from `Bodies.Signatures` — the answer of the rung that
  settles them — so a composition, which has no parameter list, and a
  dependency another module declares are answered for like anything else.
- **An operand is not a declaration of totality.** What stops it is the budget
  the row is evaluated under, which is why a row may apply a `partial` helper.
  Claiming termination would report it as a declaration under a name the author
  never wrote and cannot mark.

The expected value is computed by that method too and compared as the value it
answered with. The walk that built an expectation from its text is gone with
the reading it served.

A row reaches what a row can reach. It supplies the values a behavior is
applied to and stands in for what that behavior depends on; it is not inside
the application, and the definition it becomes is nullary and static with
nothing injected into it. So those definitions are elaborated against an empty
requirement table, and a call to a required behavior is refused (E1818) — by a
helper the row applies as much as by the row.

## Holes this reached

Four were there before this change, each reachable from a body:

- a spread naming something that is not a value raised an internal error
  rather than being refused as the name it is (E1023, E1024)
- three walks read what a name denotes without asking whether it denotes
  anything
- a helper whose body produces a function skipped its own declared-return
  check, leaving the backend a method to emit and no elaborated body for it
- `unreachable` where nothing says what a position holds was refused only
  where the backend reached it

One this change made reachable: a module whose classes could not be made read
as a module whose rows were all seen, so `--generate` offered work for a
behavior whose rows nothing had read. Whether anything was read is each
incompleteness code's own answer now rather than a list of the codes that mean
it, and the test that states the rule has a third model and a measured
positive control.

## What moved, and what did not

Reactor-wide `BUILD SUCCESS`: compiler **4514 green**, lsp 202, cli 333, and
the rest. souther-examples, all 14 projects: compile diagnostics and
`examples --generate --boundaries` **identical to the base jar**.

Semantics moved in five places in the suite, all of them towards what a body
already did:

| | before | now |
|---|---|---|
| two entries under one key | refused | `Map.fromList` resolves it, and the map is compared |
| `Int.divide(7, 2)` at an `Int` | ran | E1812, as in a body |
| `List.isEmpty([ ])` | refused for an open variable | settles, as in a body |
| bare `unreachable` at an expectation | refused | runs and aborts (E1903) |
| a binding shadowing a helper | refused | the binding answers, and the row reports the difference |

## Notes for review

- `FixtureReader` loses the `asserted*` walk and the shim that fed it; both
  were removed only after a mutant showed them unreachable across the suite.
- Row methods are stripped where classes are written out, so the jar footprint
  of an `example` is what it was.
- The spec's fixture rules are unchanged except where the old reading was the
  only thing holding them; their reasons and their codes are updated, and the
  collection section now says that a row is evaluated as this module's code
  with its brackets the one notation a position answers for.


kawasima:
Reviewed all three, reproduced each, and fixed them in `2dc88557` (spec sentence in `4a34f8d0`).

**Each was real, and two were worse than described.**

| | measured before the fix |
|---|---|
| 2. positions rebuilt from syntax | a wrong-typed operand at a composition input was **not refused at all** — it reached the JVM as `class Mid cannot be cast to class In`, surfaced as E1905. The same operand at a `SpecBehavior` target is E1812, which is the contrast that pins it |
| 3. a row reaching a model dependency | calling it directly **compiled with no diagnostic**; through a helper it was an **internal compiler error**, `unknown function lookup` |
| 1. supplied bare type name | `with lookup = Missing` gives **E1023**, the language's own refusal at the name |

On (1) I owe a correction. I had measured this earlier and reported that restricting `computesNothing` to the expectation "lets `with lookup = Missing` through in silence", and left the second reading in place on that basis — it is in the PR body. That measurement was wrong twice over: I read an empty list from `allCodesOf`, which filters to `E19*` and so cannot see E1023, and I confirmed it against a CLI jar built with `-pl souther-compiler`, which had not picked up the change. Rebuilt whole and run through the CLI, the refusal is there. Your reading of it was right and the PR body's paragraph was wrong; it is gone.

**What the fixes are**

- `computesNothing` is `RowPosition.Asserts` only. A unit case still stands at a supplied position as the value it is; a record's name is refused where it is written.
- `RowFixtures.placed` takes `Map<String, Sig>` and reads `inputTypes()` / `outputType()`. `Shapes.Prepared` asks `Bodies.Signatures`, which does not read `Shapes.Prepared` — no cycle. Compositions and imported dependencies come out of the same answer as everything else.
- `HelperTyping` is handed `Prepared.operandMethods().values()` and gives exactly those definitions an empty `reqSigs`. Names are not guessed at. A helper the row applies is elaborated in the same empty context, so `through(...)` closes too.

**Witnesses** are in `ARowOperandIsHeldToWhatItsPositionIsTest` (5): the composition input (with the holding row beside it), the record name, the unit case that must still be admitted, the direct dependency call, the call through a helper.

Two existing tests moved with the change. `aWithThatCannotBeBuiltStandsInForNothing` asserted `[E1908]` through that `E19*` filter and now asserts the E1023 the language gives. `AStateIsReachedOnlyThroughWhatEstablishesItTest` pins the way into `Prepared`, which now takes the signatures; by that test's own rule — the way in performs what the state claims, whatever it is handed — the pin is updated rather than the argument dropped.

**Verified**: reactor-wide `BUILD SUCCESS` (compiler 4514, lsp 202, cli 333, and the rest), `CI=1 mvn -Plint verify` clean, and souther-examples all 14 projects identical to the `a706c3f9` base jar in both compile diagnostics and `examples --generate --boundaries`.

The collection-section sentence is fixed too: a row is evaluated as this module's code, and its brackets are the one notation a position answers for.

Issue #732, opened 2026-08-14
An example run decides for itself what applies a behavior, so nothing else can be given the job


kawasima:
#717 gave a row's outcome a place to say what applied the behavior, and said what is still missing:

> There is no arm for an implementation supplied from outside the compile. Nothing outside a compile
> applies a behavior today, and an arm nothing writes is a statement about a run that does not happen.

Recording which answerer ran is one half. The other half is being able to supply one, and there is
nowhere to supply it: an example run decides both what applies the behavior and which classes the
row's values live in, inside the function that runs the rows.

## The two decisions, and where they are made

    // examples/ExampleVerifier.java:97
    MemoryClassLoader loader = new MemoryClassLoader(classes, parent);

`check` is handed the bytes and a parent, and builds the loader itself. Every value a row constructs
is an instance of a class that loader defined.

    // examples/ExampleVerifier.java:1125
    private Object invokeNow(ExampleTarget target, Object[] args, Object[] fakes) {
        GeneratedClass impl = new GeneratedClass.BehaviorImpl(module.name(), target.name());
        Class<?> c = GeneratedClasses.load(loader, impl);
        ... openCtor(c, ctorParams).newInstance(fakes)
        ... c.getDeclaredMethod("apply", paramTypes).invoke(instance, args)

And applying the behavior is loading that module's `$Impl` through the same loader, constructing it
with the fakes in requirement order, and calling `apply` reflectively.

With one answerer the two are one decision, so nothing has had to tell them apart. They are not one
decision, and the difference is what the next answerer runs into.

## A second answerer is not in that loader, and cannot be put in it

This is not a detail of how #695 would be written. It is a rule the compiler already holds:

    // query/Output.java, evaluationLoader
    a class defined here and one defined by the parent are different types under one binary name, so
    a module split between them hands its own types to its own implementation and the cast fails —
    which is reported as an example that does not hold, about a model that is fine.

`evaluationLoader` therefore defines the path's classes whole rather than sharing the parent's, and
`MemoryClassLoader` delegates child-first for the names it holds, for the reason written on it. Both
are the same rule: one loader, one version of every type.

An implementation supplied from outside is compiled against the module's classes as some other
loader has them. It cannot be moved into this run's loader without the caller giving up
constructing it — and constructing it is the whole of what the caller supplies. So the two ends of
the application are in different loaders by construction, and the values crossing between them
cannot be shared instances.

## What crosses already exists

[...]

PR #741, opened 2026-08-15
Give an example run what applies a behavior rather than building it

kawasima:
Refs #732.

An example run decided both what applied the behavior and which classes the row's values live in,
inside the function that runs the rows. With one answerer the two are one decision, so nothing had
to tell them apart.

## The third decision

The issue names two. There is a third on the same line, and it is the one that would have been left
behind: a stand-in is made into a `Behavior` proxy or a generated subclass of the dependency's base,
and both are classes of the loader the implementation being constructed comes from. Moving the
`$Impl` out and leaving the stand-ins would put a second answerer back where it started, handed
instances its constructor cannot take.

It does not bite an injected behavior, which has no requirements (`check/Requirements.java` answers
an injected name with none), but a seam whose types do not say so would have that hold by accident.

## The shape

```java
public interface Answerer {
    Applying applying(String behavior, List<DependencyStandin> standins);  // StandinNotBuilt only here

    interface Applying {
        Applied applied();                    // asked per behavior, not per compile
        Object to(List<Handed> arguments);
    }
}

@FunctionalInterface
public interface Answering { Answerer over(String module, MemoryClassLoader compiled); }

public final class Handed       { Object built(); NeutralValue neutral(); }   // built in this package only
public final class NeutralValue { Object read(); }                            // and this one
public record DependencyStandin(String dependency, int inputs, Function<Object[], Object> answers) {}
```

The run owns the loader — a row's fixtures are decoded and held to their invariants whether or not
there is anything to run them against — and hands it to `Answering`. That is what makes an answerer
over this compile's own classes one that can only exist over them: a compile's implementation cannot
be paired with a run whose values are some other loader's, and there is no way to state such a pair.
`apply` is erased, so such a pair would not disagree at the boundary but inside the behavior at the
first cast.

`Answering` also settles the stand-ins. The implementation and its stand-ins are in one loader, or
the behavior has none.

`applying` and `to` are two steps because which of them a row got through is what its outcome says
about it. A stand-in that could not be made stops the row where a dependency with no fake stops it,
at `FIXTURES_VALIDATED` / `FAKE_RESOLUTION`, and nothing applied it.

## The crossing is asymmetric

Reading a value back is already loader-free: a class is identified by its binary name
(`SoutherJvmAbi.valueTypeCandidate`, `JvmClassName.is`) and a field by the accessor every data has.
Handing one over is not — a typed `apply` and an injecting constructor both check identity. So only
that direction is written: `Handed` carries the value as this compile built it and the form a derived
decoder reads, worked out only if it is asked for. The rule that decides the form stays in
`NeutralForm`, so an answerer in another loader needs no second reading of what a value is.

`Applied` gains no arm. What applied a row is what entered it said of itself, and it is asked of the
application rather than of the answerer: a module declares behaviors with a body and behaviors
without one, and only the second can have an implementation supplied for it, so which of them applies
is settled one behavior at a time. How far a row got and what entered it are published as one value,
because a row given up on is read from another thread and half of that pair is a row that never
happened.

## Failure

The split #681 drew is kept. The answerer carries out what the applied code came back with and says
nothing about it (`InvocationFailure`), and the row makes the one reading of it — which is what stops
a row's meaning from depending on which answerer it had. Failing to reach the implementation at all
is a third thing and is raised as `ImplementationNotReached`; what a row is told about it does not
change.

## What was measured

- An answerer that defines this compile's classes a second time and applies those: same bytes, two
  loaders, and the row holds with the right `resultArm`. Handed the values as this compile built them
  instead, it fails with E1905 — so the crossing is what makes it hold.
- Where a row stops: `StandinNotBuilt` from `applying()` lands at `FIXTURES_VALIDATED` /
  `FAKE_RESOLUTION` / `Applied.Nothing`; a failure from `to()` lands at `INVOKED`.
- Two behaviors of one module resolved separately: each row records the answer for the behavior it is
  about, and the two do not carry one answer between them. Without `Applied` on the application that
  test cannot be written.
- How far a row got and what entered it are one publication: a reader spinning against a row that
  enters and takes it back never sees a stage saying the behavior was entered with nothing that
  entered it. Split into two volatiles it fails at once.
- souther-compiler 4523 tests (9 new) and souther-cli 333 tests green.
- souther-examples, 14 modules, `examples --generate --boundaries`: 21,924 lines, identical between
  this branch and develop.

## Deliberately not settled

A row that meets a failure from `Handed.neutral()` is recorded as undecided, having applied nothing
(`FIXTURES_VALIDATED` / `INCOMPLETE` / `INFRASTRUCTURE`), with no diagnostic — nothing about it was
established and the model may be right. Whether there is more to say than that is settled by whoever
first supplies an answerer that crosses; no answerer a compile has does, and what `neutral()` throws
is not a type this package publishes, so an answerer cannot catch it by name.

`DependencyStandin` carries the trigger for the other one: when a supplied implementation has to run
a behavior that does have requirements, the values going into and out of `answers` would be crossing
loaders and that boundary has to be designed then.


kawasima:
Review findings addressed in 499eebe2. All three were real; the first two are at the seam's own edges.

**A row that never entered the behavior said something applied it** (`ExampleVerifier.java:766`,
graded low). A row is marked as having entered before the answerer is asked, because that is the one
thing a reader of a row that never comes back can be given and it can only be published beforehand.
`ImplementationNotReached` left the mark standing, and everything at `INVOKED` says what applied it
(`Run.applied()`) — so a row whose `$Impl` could not be loaded recorded `Applied.GeneratedHere`. It
now takes the mark back. This is the same invariant the `applying`/`to` split exists for, so the two
halves agree.

**A crossing failure took the module with it** (`ExampleVerifier.java:762`, medium). Right: the
shipped contract meant "crash the compile", not "undecided". `Handed.neutral()`'s failure is now
caught at the row, which is recorded as `INCOMPLETE` / `INFRASTRUCTURE` at `FIXTURES_VALIDATED`
having applied nothing, with no diagnostic — nothing about the row was established and the model may
be right. `check` already turns an `INCOMPLETE` row into `ROW_UNDECIDED`, so no new reporting was
needed. `FailurePhase.INFRASTRUCTURE`'s javadoc and `RowOutcome`'s table said nothing writes that
phase; they now say what does.

**`answeredBy` at a projection** (`FixtureReader.java:1198`, low). Correct — by the time it is read,
`written` has a field appended for every step `takeField` took, so it names the value rather than the
application that produced the one above it. It is named by its path there; line 1757 keeps the
wrapping, where `c.written()` really is a helper application.

Both new paths were held by taking the handling away and watching them go red: without the first the
stage stays `INVOKED`, without the second the failure leaves `ExampleVerifier.check` entirely.

Re-measured: souther-compiler 4521 tests (2 new) and souther-cli 333 green; souther-examples
21,924 lines still identical to develop.

kawasima:
`Applied` moved to `Applying` in 0471cac3. The argument holds: what applies a behavior is settled per
behavior, and a module that declares one with a body and one without has two answers in one
evaluation, so an answerer resolving between them would have had to break the one-per-answerer
contract the day it was written. That is not a future shape, so the question moves now.

```java
public interface Answerer {
    Applying applying(String behavior, List<DependencyStandin> standins);

    interface Applying {
        Applied applied();
        Object to(List<Handed> arguments);
    }
}
```

One thing came out of doing it that is worth naming. The row's state now carries what entered it, and
a row given up on has that state read from the other thread. Read as two fields, its stage can be
taken before the worker moved and what applied it after, and the pair recorded is a row that never
happened. So both are read together, stage first — both volatile, and the one that says a behavior
was entered is written last:

```java
Reached reached() {
    Stage at = stage;
    return new Reached(at, at.reached(Stage.INVOKED) ? applied : null);
}
```

`neverEntered` takes both back, so `ImplementationNotReached` records `Applied.Nothing` directly
rather than through the stage gate — which is the state machine you described.

Held by a new test: an answerer that answers per behavior, over a module with two of them, and each
row records the answer for the behavior it is about. Without the move that test cannot be written at
all, which is the point.

The non-blocking note is taken too: `aRowThatCouldNotBeHandedOverIsUndecidedRatherThanFailed` now
says it fixes what a row makes of a crossing that failed, not the path that produces one — a row's
input is built by this compile's own decoders and is always readable back, so nothing there can make
`Handed.neutral` fail for real.

Re-measured: souther-compiler 4522 tests (1 new) and souther-cli 333 green; souther-examples 21,924
lines still identical to develop.

kawasima:
Fixed in cfabce99, structurally rather than by retrying.

The race is real and it is worse than a wrong record: `RowOutcome` refuses to be built from a stage
that says the behavior was entered beside a run that says nothing applied it, so the torn pair
reaches whoever is compiling as a build that fails on a timing window.

Ordering the writes was right for entering and not for taking it back, which is the half I had not
thought through. Rather than re-read until two reads agree, the two are one field now — one value,
replaced whole at every step, so a reader takes a pair some moment of the evaluation actually held:

```java
/** How far the row got and what entered the behavior, as one value. */
private volatile Reached reached = new Reached(Stage.NONE, null);

void got(Stage next)          { reached = reached.at(next); }          // keeps what entered
void entered(Applied applied) { reached = new Reached(Stage.INVOKED, applied); }
void neverEntered()           { reached = new Reached(Stage.FIXTURES_VALIDATED, null); }
```

Only the row's own worker writes it, so reading it to write the next one races with nothing.

Held by a test, and the test was held by taking the fix away: a reader spinning while the row enters
and takes it back sees a stage saying the behavior was entered with nothing that entered it within
the first few rounds when the two are published separately, and cannot when they are one.

Re-measured: souther-compiler 4523 tests and souther-cli 333 green; souther-examples 21,924 lines
still identical to develop.

The PR description is updated too — the API sketch showed the first version's `Answerer.applied()`,
and what a row makes of a `Handed.neutral()` failure is no longer open.

PR #744, opened 2026-08-15
Ask what a name names of the form that has an answer

### LINKED ISSUE #740: A row applying an intrinsic ends in an internal error when an argument names nothing
A name nothing declares reaches the walk that settles a kernel a row applies, and the compile ends with an internal error instead of saying the name is unknown.

## Reproducing

```souther
module demo

data Ok = { n: Int }

behavior take : (d: Decimal) -> Ok
    constructs Ok
let take (d) = Ok { n = 1 }

example take
    | "a misspelt mode" : (Decimal.round(2, HalfUp, 1.005m)) -> Ok { n = 1 }
```

```
internal compiler error: IllegalStateException: `HalfUp` at 11:50 denotes nothing and was read as though it did
```

`HALF_UP` is the case core declares; `HalfUp` is a misspelling, which is the
ordinary way to arrive here.

## Where it happens

```
Hir$Var.denotes(Hir.java:1649)
FixtureEvidence.declaredTypeOf(FixtureEvidence.java:76)
FixtureArgumentTypes.stated(FixtureArgumentTypes.java:49)
FixtureArgumentTypes.of(FixtureArgumentTypes.java:42)
HelperInliner.kernelCallsIn(HelperInliner.java:475)
HelperInliner.fixtureKernels(HelperInliner.java:444)
Prepared.prepare(Prepared.java:119)
```

`kernelCallsIn` guards the call it is looking at — `call.answered() != null` —
and then hands the call's *arguments* to the walk that works out what each of
them is, which asks `denotes()` of a name without asking whether the name
denotes anything. `Hir.Var.denotes()` is the reader for someone who would
otherwise have to make something up, so it throws.

## What decides whether it is hit

The callee, not the argument position. Measured:

| written in a row | result |
|---|---|
| `Int.max(1, nope)` | E1023, at `nope` |
| `Decimal.round(nope, HALF_UP, 1.005m)` | internal error |
| `Decimal.round(2, HalfUp, 1.005m)` | internal error |
| `Decimal.round(2, HALF_UP, nope)` | internal error |
| `Decimal.divide(1.0m, 2.0m, 2, HalfUp)` | internal error |
| `(nope)` on its own, or at a field, or on the expected side | E1023 |

`Decimal.round` and `Decimal.divide` are intrinsics a row applies, so they go
through the walk that settles which instance to emit a method for. `Int.max`
does not, and a plain unresolved name never gets there.

A body is unaffected: `let round2 (r) = Rate(Decimal.round(2, HalfUp, r.value))`
is E1023. This walk exists for rows and is reached from `Prepared.prepare`.

## Notes

- Not introduced by #737, and not fixed by it: it reproduces identically on
  `a706c3f9` (before) and `9ba01cb5` (after).
- The same shape as the three unguarded reads closed in #737
  (`HelperNames.publishedBy` / `carriedByValue` / `collectHelpersOf`): a walk
  asks what a name denotes without asking whether it denotes anything. Those
  were found by walking the readers of `denotes()`; this one is in a walk that
  runs before them, so it was not in that set. Whether any reader is left is
  worth asking once rather than per report — `Hir.Var.denotes()` has readers
  that a `Var.Unanswered` can reach, and nothing counts them.
- The name is already reported: resolution answered it with `Unanswered` and
  said so where it is written. What is wrong is only that a later walk reads it
  as though it had an answer, so the compile ends with an internal error
  instead of the report.

kawasima:
Closes #740 — a row applying an intrinsic ended a compile with an internal error when an argument named nothing.

## What was wrong

`Hir.Var` and `Hir.Name` are each split into the two answers resolution has: a form that names a declaration and a form that names nothing. The observations that read a declaration were not. `denotes()`, `reachedAs()`, `bare()` and `reaches()` stood on the interface, so they could be asked of a name that answers nothing, and answered by throwing; `Hir.Apply` carried three of them across from the name it applies. Keeping that safe was a convention every reader had to remember — 213 call sites for the value names, 66 more for the type names — and a reader that forgot compiled, then threw where the name was read rather than where it was reported.

That is what #740 is. `FixtureEvidence.declaredTypeOf` reads a row's arguments to settle which instance of an intrinsic to emit a method for, and it read `denotes()` of a name without asking whether the name denotes anything. Its own `denotes == null` test beside it was dead code — the reader it guarded does not return null, it throws — which says plainly enough what the author took the API to be.

#737 closed three of these by walking the readers of `denotes()`. This one was in a walk that runs before them, so it was not in that set. Answering "is any reader left" one report at a time is what this stops.

## What changed

Those observations are `Denoting`'s now, on all three types, and `Apply` answers only `answered()`. A reader arrives at what a name names through that projection or by narrowing to the two forms, which is where it says what it does with a name that answers nothing — so the question is no longer whether every reader remembered, but whether such a read can be written at all.

`unresolved()` is derived from `answered()` rather than implemented per form, so "did it answer" and "what did it answer" cannot come apart.

`Unanswered` carries `unexpectedHere()`: not a reading — there is nothing to read — but the claim that a reader's input was built from answered names. Three readers make it, and each names the producer that guarantees it beside the throw:

- `Backend.reachedBy` / `Backend.names` (and `Runner.named` in the CLI): `Bodies.Checked` hands an elaboration over only where `Names.Sound` holds of the module, which resolution makes false as soon as it reports a name denoting nothing; `Output.Classes` builds what it generates from that answer and nothing else.
- `DataChecker.names`: `Names.Unbuilt` collects a declaration holding a name that names nothing, `Names.Definition` hands it to nobody, and `TypeChecker` checks only the declarations it was handed.

The order the passes run in is not cited anywhere, and cannot be: a compilation goes on answering after an error. Where a producer could not be named, the reader recovers instead.

Two readers stopped taking one fact as two arguments on the way (`PartialReachability.keyOf`/`declarationOf` and `HelperInliner.expands` took a denotation and a reach name separately, and now take the answered name).

## Measured

- The matrix from the issue is a test: every argument position and both intrinsics answer E1023 exactly once, `Int.max(1, nope)` answers the same, and the correctly spelled call still settles. It fails with the internal error on the parent commit.
- `CI=1 mvn -Plint clean verify` BUILD SUCCESS — compiler 4519, formatter 2224, LSP 202, CLI 333, all modules green.
- souther-examples, all 14 projects, against a base jar built from `develop`: `examples --generate --boundaries` identical over 21,924 lines, and `compile` output identical once the temporary output directory is normalised out of it. No behaviour changed anywhere the corpus reaches.

## One change outside the issue

`pom.xml` raises `-Xmaxerrs` beside the `-Xmaxwarns` already there, with the same reason: javac stops printing at a hundred, and a report that stops at a round number reads as the whole of what was found. This work hit it — 213 call sites were counted as 100 — but it is a build-report change, not part of the fix, and it is its own commit so it can be dropped.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -208,10 +208,14 @@
                 <version>3.15.0</version>
                 <configuration>
                     <compilerArgs>
-                        <!-- javac prints a hundred warnings per compilation and stops counting.
-                             A report that stops at a round number reads as the whole of what was
-                             found, so the limit is raised past anything this build produces. -->
+                        <!-- javac prints a hundred warnings per compilation and stops counting, and
+                             a hundred errors likewise. A report that stops at a round number reads
+                             as the whole of what was found — a change that breaks two hundred call
+                             sites is read as breaking a hundred — so both limits are raised past
+                             anything this build produces. -->
                         <arg>-Xmaxwarns</arg>
+                        <arg>100000</arg>
+                        <arg>-Xmaxerrs</arg>
                         <arg>100000</arg>
                         <arg>-XDcompilePolicy=simple</arg>
                         <arg>--should-stop=ifError=FLOW</arg>
--- a/souther-cli/src/main/java/souther/cli/Runner.java
+++ b/souther-cli/src/main/java/souther/cli/Runner.java
@@ -231,9 +231,23 @@ public final class Runner {
     private static String dependencyNames(Hir.SpecBehavior spec) {
         List<String> names = new java.util.ArrayList<>();
         for (Hir.Var dep : spec.dependsOn()) {
-            names.add(dep.bare());
+            names.add(named(dep));
         }
         return String.join(", ", names);
+    }
+
+    /**
+     * The name a declared dependency or a pipeline stage is reached by.
+     *
+     * <p>Everything below the compile above names something: {@code Compiler.compiled} raises on the
+     * first error and a name that denotes nothing is one, so this reads a module that got past it.
+     * One arriving here would be `run` working from a module the compile refused.
+     */
+    private static String named(Hir.Var name) {
+        return switch (name) {
+            case Hir.Var.Denoting d -> d.bare();
+            case Hir.Var.Unanswered u -> throw u.unexpectedHere();
+        };
     }
 
     /**
@@ -316,8 +330,8 @@ public final class Runner {
                 specs.put(spec.name(), spec);
             }
         }
-        for (Hir.Var named : PipelineSigs.flattenStages(pipe.stages(), pipeStages, pipe.pos())) {
-            String stage = named.bare();
+        for (Hir.Var each : PipelineSigs.flattenStages(pipe.stages(), pipeStages, pipe.pos())) {
+            String stage = named(each);
             if (!implemented.contains(stage)) {
                 return new Blocker("run.pipeline.noimpl",
                         "`" + pipe.name() + "` is a pipeline whose stage `" + stage
--- a/souther-compiler/src/main/java/souther/compiler/ast/Hir.java
+++ b/souther-compiler/src/main/java/souther/compiler/ast/Hir.java
@@ -228,19 +228,15 @@ public interface Hir {
         }
 
         /**
-         * The declaration this names, for a reader that is asking which one it is.
+         * This name where it names a declaration, and null where resolution read it and found none.
          *
-         * <p>Refused for an {@link Unanswered} one, which is a mistake in the source reported where
-         * the name is written: there is no declaration to hand back, and a reader that wants to go
-         * on without one asks which of the two it is rather than taking the spelling. This is the
-         * one place that says so.
+         * <p>The one way from here to the declaration. What names it — {@link Denoting#type()} — is
+         * the answered form's, so a reader arrives at it through this projection or through
+         * narrowing to the two forms, and says what it does with an {@link Unanswered} name where it
+         * makes that choice.
          */
-        default TypeSymbol denotes() {
-            if (this instanceof Denoting denoting) {
-                return denoting.type();
-            }
-            throw new IllegalStateException("`" + written() + "` at " + pos()
-                    + " denotes nothing and was read as though it did");
+        default Denoting answered() {
+            return this instanceof Denoting denoting ? denoting : null;
         }
 
         /** The same name, read and found to name nothing. */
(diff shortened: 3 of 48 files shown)


kawasima:
The blocking finding is real, reproduced, and fixed in `aee51e8e`.

## What I could reproduce

```souther
module demo
import List ( length )
data Tags = List<String>

behavior countTags : (t: Tags) -> Int
let countTags (t) = {
    let Tgas(xs) = t
    length(xs)
}
```

- this PR's head: `internal compiler error: NullPointerException: Cannot invoke "TypeSymbol.key()" because "name" is null`
- `develop` (85905847): `internal compiler error: IllegalStateException: `Tgas` at 7:9 denotes nothing and was read as though it did`

So the hole is the one #740 is about, it was there before this branch, and what this branch did to it was change which internal error comes out. The worse half is that the comment beside it said the name was left to the report it already had. It was not, so the comment was a restatement of a rule that does not hold.

## Abandoning the definition rather than passing the name over

I measured the other reading before choosing. Skipping leaves the binding unopened, so `bindType` stays the value's own type and the body is checked with `xs` as `Tags`: `length(xs)` then answers E1317, which is the second sentence about the one mistake that the comment was written to avoid.

The arms beside this one in `Elaborator` — a `Hir.Var.Unanswered` standing as a value, a `NewData` whose type name is unanswered — all abandon the definition, so this does too:

```java
if (!(li.opens().answered() instanceof Hir.Name.Denoting opens)) {
    throw new Unanswerable(li.opens().pos());
}
```

What comes out is E1023 exactly once, with the spelling hint.

## Whether the same shape is left anywhere else

This branch introduced seven `answered() == null ? null : …` projections. Six hand their null to a reader that takes one: `Combinators.of`'s caller branches on `rule == null` immediately, `DataChecker`'s `viaHelper` and `Elaborator.handedOver`'s `kept` are tested before use, `FixtureEvidence.declaredTypeOf` and `FixtureReader.constructedCase` both document null as absence, and `FixtureReader`'s `Type.Prim written` is used under `written != null &&`. The one that handed absence to a reader that does not take one is the one you found.

`TypeOps.newtypeInner` is left as it is. It takes a symbol and reads the declaration under it; answering about absence as well would put "there is no name" and "the name is not a newtype" in one predicate.

## Tests

Both of the ones you asked for.

- `CompileDestructuringTest.openingSomethingNothingDeclaresIsTheNamingErrorOnce` — an unresolved constructor pattern name, held to `List.of("E1023")` so it fixes exactly-once and not only the absence of an internal error. Putting the previous reading back fails it with the NullPointerException, which I ran.
- `ARowApplyingAnIntrinsicRefusesANameItCannotAnswerTest.aBodyWritingTheSameCallIsTheSameNamingError` — the body-side control, beside the row matrix: `let round2 (r: Decimal) : Decimal = Decimal.round(2, HalfUp, r)` is E1023 exactly once.

## Measured again, against develop 85905847

- `CI=1 mvn -Plint clean verify` BUILD SUCCESS: compiler 4530 (+2), formatter 2224, LSP 202, CLI 333, every module green.
- souther-examples, all 14 projects: `examples --generate --boundaries` identical over 21,924 lines, `compile` output identical.

The `-Xmaxerrs` change is still its own commit (`8d585998`), so it can be dropped or moved to a PR of its own without touching the rest.

Issue #750, opened 2026-08-15
A product of two values the path bounds is read as a value nothing is known of


kawasima:
A quantity times a unit price is non-negative where both of them are, and the invariant-discharge
check says nothing about it. `Terms.scale` answers a linear form only where one side is a constant,
so a product of two values becomes an atom of its own, and nothing relates that atom to the factors
the path has bounds for. The same holds of `x / k`: `/` is truncating division, and the quotient is
an atom no bound reaches.

```souther
data NonNeg = Int   invariant value >= 0

behavior p1 : (a: Int, b: Int) -> NonNeg | Bad
let p1 (a, b) = {
    guard a >= 0 else Bad
    guard b >= 0 else Bad
    NonNeg(a * b)          // E2011
}

behavior p7 : (x: Int) -> NonNeg | Bad
let p7 (x) = {
    guard x >= 0 else Bad
    NonNeg(x * 30 / 100)   // E2011
}
```

Both guards say exactly what the construction needs, and there is no third guard that would say it
about the product.

## Why this is not the table

The rule is not a fact about `*`. Nothing holds of a product wherever it is written — it is what the
path knows of the two factors that decides the sign, and that is known at the construction and not
where the expression is walked. So this cannot be a row read off the operation and asserted like "a
size is never negative": the bound has to be asked of the domain at the point the term is read, and
computed from the bounds it answers.

For a product, that is interval multiplication: the four corner products of the factors' bounds,
their least and greatest. For `x / k` with `k` a written constant above zero, truncation toward zero
puts the quotient between `0` and `x` inclusive, which is a bound in each direction once the sign of
`x` is known. Neither is a linear form, so neither is `Terms.affine`'s to answer; what they produce
is a bound on the atom the expression already has.

## What it is worth

17 of the 64 E2011 warnings souther-examples raises today stand on a product or a divide:

```
joboffer/src/main/souther/joboffer.sou:169,170  総額(上限.value * 人数.value)
joboffer/src/main/souther/joboffer.sou:179      総額(単価.value * 件数.value)
ordering/src/main/souther/inventory.sou:247     Eaches(cases.value * pack.value)
crm/src/main/souther/crm.sou:255                Amount(money.amount.value * rate.value)
ordering/src/main/souther/inventory.sou:260     Cases(eaches.value / pack.value)
hr/src/main/souther/yearendadjustment.sou:165-167,272-278   Yen(salary.value * 30 / 100 + 80000) and its siblings
[...]

PR #753, opened 2026-08-15
Read what an operation guarantees of what it answers

### LINKED ISSUE #749: An operation whose own arguments bound its result establishes nothing about it
`Int.clamp(0, 100, x)` is between 0 and 100 whatever `x` is, and the invariant-discharge check reads
none of it. The call is named — it is a pure total library call, so it is a term, and a construction
over it is read rather than left unsaid — but no table says what the operation does to a number, so
the construction is reported as unproven. The guard the report asks for would be
`guard Int.clamp(0, 100, x) >= 0`, which restates what the operation already guarantees.

## The probe

Seven constructions, each owed a clause the operation standing under it already establishes. Every
one of them warns today — once each is written in a behavior. A construction in a plain value
definition is read by nothing: `let p (x: Int): NonNeg = NonNeg(x)` reports nothing either, so the
probe below has to be rewritten with a `behavior` line and a `constructs` clause before it says
anything about this.

```souther
module probe

data NonNeg = Int   invariant value >= 0
data NonNegD = Decimal   invariant value >= 0.0m
data Pct = Int   invariant value >= 0 && value <= 100
data Span = { from: Date, to: Date }   invariant Date.daysBetween(from, to) >= 0

let p2 (x: Int): NonNeg = NonNeg(Int.abs(x))
let p3 (x: Int): NonNeg = NonNeg(Int.max(0, x))
let p5 (x: Int): Pct = Pct(Int.clamp(0, 100, x))
let p6 (x: Int): NonNeg = NonNeg(Int.floorMod(x, 100))
let p11 (d: Date): Span = Span { from = d, to = Date.addDays(1, d) }

behavior p4 : (a: Int, b: Int) -> NonNeg | Bad
let p4 (a, b) = {
    guard a >= 0 else Bad
    guard b >= 0 else Bad
    NonNeg(Int.min(a, b))
}

behavior p8 : (n: Int) -> NonNegD | Bad
let p8 (n) = {
    guard n >= 0 else Bad
    NonNegD(Decimal.fromInt(n))
}
```

```
E2011  NonNeg(Int.abs(x))
E2011  NonNeg(Int.max(0, x))
E2011  Pct(Int.clamp(0, 100, x))
E2011  NonNeg(Int.floorMod(x, 100))
E2011  Span { from = d, to = Date.addDays(1, d) }
E2011  NonNeg(Int.min(a, b))
E2011  NonNegD(Decimal.fromInt(n))
```

## What the rules would say

| operation | what holds of the result | shape |
| --- | --- | --- |
| `Int.abs(n)`, `Decimal.abs(d)` | `>= 0` | an interval |
| `Int.min(a, b)`, `Decimal.min(a, b)` | `<= a`, `<= b` | a difference |
| `Int.max(a, b)`, `Decimal.max(a, b)` | `>= a`, `>= b` | a difference |
| `Int.clamp(lo, hi, x)`, `Decimal.clamp(lo, hi, d)` | `>= lo`, `<= hi` | a difference |

**This row is false, and the implementation replaced it — see below.**

| `Int.floorMod(x, k)`, `k` a written constant above zero | `>= 0`, `< k` | an interval |
| `Decimal.fromInt(n)` | the result **is** `n` | the form `n` itself |
| `Decimal.toInt(mode, d)` | `d - 1 < r < d + 1` | a difference |
| `Date.addDays(n, d)` | `Date.daysBetween(d, r) = n` | a difference |

Each is a fact about the operation and holds wherever the call is written, which is the kind of fact
`Predicates.sizeFacts` already asserts of every size call it walks past — a size is never negative,
and is no greater than the size of what it was built from. Every row above lands in a shape
`NumericDomain` already holds, an interval bound or a bound on the difference of two atoms, so the
domain itself does not change: what is added is a table in `DischargeRules` and a sibling of
`sizeFacts` that reads it.

Two of the rows are not bounds. `Decimal.fromInt(n)` answers the number it was given, so it is read
into the same linear form rather than given an atom of its own. `Date.addDays` states its fact
through `Date.daysBetween`, which the check already reads as the order it decides
(`DischargeRules.ORDERS`).

## What it is worth

18 of the 64 E2011 warnings souther-examples raises today stand on one of these operations:

```
crm/src/main/souther/crm.sou:635              LeadScore(Int.clamp(0, 100, rawScore))
crm/src/main/souther/forecasting.sou:362      Attainment(Int.clamp(0, 999, ...))
ordering/src/main/souther/billing.sou:388     Amount(Decimal.clamp(0.0m, creditLimit.value, outstanding))
ordering/src/main/souther/billing.sou:323     Amount(Decimal.abs(diff))
ordering/src/main/souther/billing.sou:469     Amount(Decimal.max(0.0m, remaining))
ordering/src/main/souther/billing.sou:481     Amount(Decimal.max(0.0m, Decimal.subtract(...)))
ordering/src/main/souther/returns.sou:202     Refund(Decimal.min(fee, gross))
ordering/src/main/souther/billing.sou:221-223 Amount(Decimal.fromInt(breakdown.net | tax | gross))
hr/src/main/souther/payroll.sou:122           Yen(Decimal.toInt(HALF_UP, ...))
hr/src/main/souther/yearendadjustment.sou:263 TaxableIncome(Int.max(0, ...) / 1000 * 1000)
hr/src/main/souther/socialinsurance.sou:125,370,431,481  ContributionMonth / CoverageLost over Date.add*
hr/src/main/souther/payroll.sou:86            ContributionMonth(Date.addMonths(0 - 1, pay.value))
```

`LeadScore` is `invariant value >= 0 && value <= 100` built by `Int.clamp(0, 100, rawScore)`, and
`Attainment` is `>= 0 && <= 999` built by `Int.clamp(0, 999, ...)`: the clamp decides both clauses of
each on its own. The rest depend on the expression around the call as well.

`Date.addMonths` is not in the table. Months have different lengths, so it states no fixed count of
days; whether it states anything the check can hold is part of what this has to decide.

## What has to be decided

A rule stated too strongly is unsound in the direction that matters — a false E2010 rather than a
missed discharge — so each row is held to the declaration it is written for and to a program that
fires it, as the tables in `DischargeRules` already are.

Per ADR-0097, an operation this bounds nothing about has that absence written down. `Int.compare`
bounds nothing, `Int.floorMod` bounds nothing where its divisor is not a written constant, and
`Date.addMonths` may join them; the set that says so is part of the work rather than a silence left
where nobody looked.

`Decimal.toInt` rounds by the mode it is handed. `|r - d| < 1` holds for every mode, and a per-mode
rule — `HALF_UP` is within a half — is a second question.

A bound can refute a clause the check could previously only warn about, so a construction that warns
today may become an E2010 error. That is the right answer where it fires, and it is a change to what
souther-examples compiles to.

The spec states what the procedure can name (§invariant-discharge-terms) and what an operation keeps
of what it was built from (§invariant-discharge-preservation). What an operation bounds is a third
table and belongs beside them: an author predicts which constructions are judged safe by reading it.

## What the implementation measured

The rows above were a hypothesis. Reading the library's own definitions and running the probe
refuted three parts of it.

**`clamp` bounds nothing.** `Int.clamp` is `if n < lo then lo else if n > hi then hi else n` and is
total: it does not ask that `lo` be below `hi`, so `clamp(50, 0, x)` answers `0` for a high `x` and
`r >= lo` is false. `min` and `max` are the same shape — what they answer depends on the arguments —
so all three moved out of the bounds table and are stated as the cases their definitions are written
in, with the condition each case is reached under. That reads more, not less: `min(a, b) >= 0` from
guards on both arguments follows from the cases, and no difference row states it.

**`Int.abs` is safe to state unconditionally.** `abs` is `if n < 0 then 0 - n else n` and the one
value with no positive counterpart aborts on overflow rather than staying negative, so `r >= 0`
holds wherever the call answers at all.

**The probe does not fire as written.** Five of its seven constructions are plain value definitions,
which the discharge check does not read. Rewritten as behaviors, all seven warn, and all seven
discharge after the change.

**10 warnings went, not 18.** souther-examples goes from 80 E2011 to 70, with none added and no new
E2010:

```
crm.sou:556, crm.sou:635        LeadScore(Int.clamp(0, 100, ...))
forecasting.sou:362             Attainment(Int.clamp(0, 999, ...))
billing.sou:388                 Amount(Decimal.clamp(0.0m, creditLimit.value, outstanding))
billing.sou:323                 Amount(Decimal.abs(diff))
billing.sou:469, billing.sou:481  Amount(Decimal.max(0.0m, ...))
returns.sou:201                 Refund(Decimal.min(fee, gross))
quoting.sou:255, quoting.sou:676  Quote { expiresOn = Date.addDays(days.value, issuedOn) }
```

Two of these — the `Date.addDays` sites in `quoting.sou` — were not in the list above.
`inventory.sou:259` is not among them: its divisor is above zero by `PackSize`'s invariant rather
than as a constant, and what a type guarantees of an argument is what the caller has rather than
what the operation answers. Of the sites
that were, `billing.sou:221-223` (`Decimal.fromInt(breakdown.net)`), `payroll.sou:122`
(`Decimal.toInt` of a product), `yearendadjustment.sou:263` and the `socialinsurance` sites all need
what the path knows of the arguments, which is #750, or rest on `Date.addMonths`, which states
nothing here.

## What was built

Four questions rather than one table, each with its range read off the declarations and its negative
answers written down (ADR-0097):

| question | what it asks | the operations that answer |
| --- | --- | --- |
| bounds | what holds of the result wherever it is called | `Int.abs`, `Decimal.abs`, `Int.floorMod`, `Decimal.toInt` |
| choice | which argument it answers, and in which cases | `min`, `max`, `clamp`, for `Int` and `Decimal` |
| measure | what it states through the measure counting two values apart | `Date.addDays`, `DateTime.addMinutes`, `DateTime.addHours`, `DateTime.addDays` |
| form | whether the number it answers is one it was given | `Decimal.fromInt` |

`Decimal.fromInt` is not a choice with one candidate: selection and a value-preserving conversion
are two statements, and joining them would put every future conversion under a table about cases.
A clause is read case by case where a choice stands under it, and a case is reached where its
condition, the equality saying the call answered that argument, and what the path establishes can
hold at once. Over the cases that are reached, all proving is the clause established and all
refuting is the clause refused. No case reached is neither: it says the guards cannot all hold,
and a guard about the call itself is what answers there. Without that equality a guard naming the
call stands outside every case, and one clause came out established by the cases and refused by the
reading that takes the call as an unknown.

`Int.floorMod` states neither end without a divisor that reads as a constant above zero: the result
takes the sign of the divisor, so `floorMod(1, -3)` is `-2` and the lower end is as much the
divisor's to decide as the upper one. Read as a constant rather than required to be written as one —
a name given `100` is that divisor, as everywhere else the check reads a value.

A guard is read as the cases as well: a condition every reachable case refutes is one its branch is
never entered under, and a construction under such a guard is not reached. Without that the cases
were read at the construction and not at the guard, and a path they show cannot be taken was
reported against. `Date.addMonths` and
`Date.addYears` state nothing, and say so. `NumericDomain` did not change.

kawasima:
Closes #749.

An invariant clause standing over a library call was owed a guard that restated the operation. `Pct(Int.clamp(0, 100, x))` warned, and the guard the report asked for was `guard Int.clamp(0, 100, x) >= 0`.

## What an operation guarantees, as four questions

Each has its range read off the declarations and its negative answers written down, so an operation the library gains is unanswered until someone answers it (`AnOperationTheLibraryGainsIsAnsweredForTest`).

| question | what it asks | answered for |
| --- | --- | --- |
| bounds | what holds of the result wherever it is called | `Int.abs`, `Decimal.abs`, `Int.floorMod`, `Decimal.toInt` |
| choice | which argument it answers, and in which cases | `min`, `max`, `clamp`, for `Int` and `Decimal` |
| measure | what it states through the measure counting two values apart | `Date.addDays`, `DateTime.addMinutes`, `DateTime.addHours`, `DateTime.addDays` |
| form | whether the number it answers is one it was given | `Decimal.fromInt` |

## `clamp` is cases, not bounds

The issue proposed `r >= lo`, `r <= hi` as unconditional rows. Reading the library refutes that: `Int.clamp` is `if n < lo then lo else if n > hi then hi else n` and is total, so `clamp(50, 0, x)` answers `0` for a high `x`. `min` and `max` are the same shape.

So they are stated as the cases their definitions are written in, with the condition each case is reached under. That reads more than a bound would — `min(a, b) >= 0` follows from guards on both arguments, which no difference row states — and it refuses where every reached case refuses, so `AboveTen(Int.min(0, x))` is an E2010.

A case holds its condition **and** the equality saying the call answered that argument. Without the equality a guard naming the call stands outside every case, and one clause came out established by the cases and refused by the reading that takes the call as an unknown. Where no case is reached the guards cannot all hold, and this says nothing rather than proving the clause vacuously.

## Measured

- souther-examples: **80 E2011 → 70**, none added, no new E2010. The 10: `clamp` at `crm.sou:556,635` and `forecasting.sou:362` and `billing.sou:388`; `Decimal.abs` at `billing.sou:323`; `Decimal.max` at `billing.sou:469,481`; `Decimal.min` at `returns.sou:201`; `Date.addDays` at `quoting.sou:255,676`. `inventory.sou:259` is not among them: its divisor is above zero by `PackSize`'s invariant rather than as a constant this reads, which is what the caller has rather than what the operation answers.
- The issue's probe discharges in full; a construction nothing bounds still warns.
- 7,554 tests green across all modules.

What is left of the issue's list needs what the path knows of the arguments (#750) or rests on `Date.addMonths`, which states nothing here and says so.

## Notes for review

- `Int.abs` is `if n < 0 then 0 - n else n`; the one value with no positive counterpart aborts on overflow, so `r >= 0` holds wherever it answers.
- `Int.floorMod`'s divisor is read as a constant rather than required to be written as one — a name given `100` is that divisor, as everywhere else the check reads a value.
- One clause reads one choosing operation; more than one is left unsettled rather than multiplied out. The spec says so.
- A guard is read as the cases too, and against everything the condition itself established: a condition no case can satisfy is one its branch is never entered under, so the construction beyond it is not reported against. Where that is not visible in the guards, the cases are asked only of a clause the as-written reading left undecided, so one clause cannot come out established and refused at once.
- `NumericDomain` is unchanged. The `Decimal.toInt` tightening on discrete atoms is not here: no program fires it.
- `CompileInvariantOneDenotationTest` used `Int.floorMod` as a call nothing is known about. The rule still holds; the witness is now a product.
- Spec: §invariant-discharge-guarantees, beside §invariant-discharge-terms and §invariant-discharge-preservation.

### DIFF
--- a/souther-compiler/src/main/java/souther/compiler/check/DischargeRules.java
+++ b/souther-compiler/src/main/java/souther/compiler/check/DischargeRules.java
@@ -864,12 +864,19 @@ final class DischargeRules {
         return e instanceof Core.PreservedCall call ? Bound.CHOICES.get(call.operation()) : null;
     }
 
-    /** Those of them the bounding table has, by name, for the test that holds each to a construction
-     * it discharges. */
-    static Set<String> boundedNames() {
-        Set<String> names = new LinkedHashSet<>();
-        boundedOperations().forEach(operation -> names.add(operation.toString()));
-        return names;
+    /**
+     * The bounding table's rows, by the name of the operation each is about — one entry per row, so
+     * an operation bounded at both ends appears twice.
+     *
+     * <p>For the test that holds each row to a construction it discharges. By row and not by
+     * operation: one program needs one of an operation's rows, so a set of names would be satisfied
+     * by a rule that had been written for one end and never fired for the other.
+     */
+    static List<String> boundedRows() {
+        List<String> rows = new ArrayList<>();
+        Bound.BOUNDS.forEach((operation, bounds) ->
+                bounds.forEach(bound -> rows.add(operation.toString())));
+        return rows;
     }
 
     /**
--- a/souther-compiler/src/main/java/souther/compiler/check/DischargeRules.java
+++ b/souther-compiler/src/main/java/souther/compiler/check/DischargeRules.java
@@ -407,6 +407,11 @@ final class DischargeRules {
         return new ResultBound(null, BigDecimal.valueOf(n), rel, new Provided.Always());
     }
 
+    /** The same, where the operation answers that only under a condition on its arguments. */
+    private static ResultBound resultIs(Rel rel, long n, Provided provided) {
+        return new ResultBound(null, BigDecimal.valueOf(n), rel, provided);
+    }
+
     /** {@code result rel argument + offset}. */
     private static ResultBound resultIs(Rel rel, Reads argument, long offset, Provided provided) {
         return new ResultBound(argument, BigDecimal.valueOf(offset), rel, provided);
@@ -421,10 +426,11 @@ final class DischargeRules {
      * past, and it is read the same way — asserted into the domain where a clause is read and where a
      * condition is assumed alike.
      *
-     * <p>{@code Int.floorMod} is here for its lower bound alone and for its upper bound only where
-     * the divisor reads as a constant above zero: the result takes the sign of the divisor, so a
-     * divisor that could be negative bounds it in the other direction, and a divisor the check cannot
-     * read bounds it nowhere. Its {@code 0} is not a case at all — the operation aborts.
+     * <p>{@code Int.floorMod} states both its ends only where the divisor reads as a constant above
+     * zero, and neither of them otherwise. The result takes the sign of the divisor — {@code
+     * floorMod(1, -3)} is {@code -2} — so a divisor that could be negative puts it the other side of
+     * zero, and the lower end is as much the divisor's to decide as the upper one. A divisor the
+     * check cannot read bounds it nowhere. Its {@code 0} is not a case at all: the operation aborts.
      *
      * <p>{@code Decimal.toInt} is within one of what it rounds, whichever mode it is handed. What a
      * single mode does more narrowly — {@code HALF_UP} rounds to within a half — is a second rule and
@@ -434,7 +440,7 @@ final class DischargeRules {
             op("Int", "abs"), List.of(resultIs(Rel.GE, 0)),
             op("Decimal", "abs"), List.of(resultIs(Rel.GE, 0)),
             op("Int", "floorMod"), List.of(
-                    resultIs(Rel.GE, 0),
+                    resultIs(Rel.GE, 0, new Provided.ConstantAboveZero(at(1))),
                     resultIs(Rel.LT, at(1), 0, new Provided.ConstantAboveZero(at(1)))),
             op("Decimal", "toInt"), List.of(
                     resultIs(Rel.GT, at(1), -1, new Provided.Always()),
--- a/souther-compiler/src/main/java/souther/compiler/check/DischargeRules.java
+++ b/souther-compiler/src/main/java/souther/compiler/check/DischargeRules.java
@@ -2,9 +2,12 @@ package souther.compiler.check;
 
 import souther.compiler.ast.Hir;
 import souther.compiler.core.Core;
+import souther.compiler.numeric.NumericDomain;
+import souther.compiler.numeric.NumericDomain.Rel;
 import souther.compiler.types.Type;
 import souther.compiler.types.ValueName;
 
+import java.math.BigDecimal;
 import java.util.ArrayList;
 import java.util.LinkedHashSet;
 import java.util.List;
@@ -15,7 +18,8 @@ import java.util.function.Predicate;
 
 /**
  * What the language's own operations do to the properties the invariant-discharge check tracks
- * (spec §invariant-discharge-preservation).
+ * (spec §invariant-discharge-preservation), and what each guarantees of what it answers
+ * (spec §invariant-discharge-guarantees).
  *
  * <p>These are the tables that decide how much of a model the language tracks rather than leaves to a
  * run-time check, and each is stated per operation because that is the level an author writes at. It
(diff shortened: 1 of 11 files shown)


kawasima:
Both blocking findings are fixed in 725167c0, and the third (the spec row) with them.

**`Int.floorMod`'s lower bound was unconditional.** It is now gated on the same `ConstantAboveZero(at(1))` as the upper one. Verified both directions: `NonNeg(Int.floorMod(x, 0 - 3))` and `NonNeg(Int.floorMod(x, k))` warn again, and the reverse — a construction wanting a negative value, which the unconditional row refused — is no longer refused. Two regression tests, one per direction.

**A guard is read as the cases too.** `assumeCond` now asks whether the condition fails in every reachable case before assuming anything of it, and takes the branch as unreached where it does. That is where the asymmetry was: the cases were read at the construction and not at the guard, so a path they show cannot be taken was walked as though it could.

That covers both reported symptoms: `guard Int.min(a, b) <= 3` under `a >= 10, b >= 10` now reports nothing rather than E2010, and the `kept`-form repro compiles clean rather than dying with `NotOneClause`.

As a second guard against the same class, the cases are consulted only where the clause as written is undecided. Two sound readings can disagree only where the guards contradict, so preferring the as-written one where it decided keeps a clause from being established and refused at once, whatever the domain could not project. What the cases are for — a clause the call-as-unknown reading cannot settle — is untouched: `AboveTen(Int.min(0, x))` is still E2010.

**One site went back to warning.** `inventory.sou:259` is `Eaches(Int.floorMod(eaches.value, pack.value))`, and `pack.value` is above zero by `PackSize`'s invariant rather than as a constant this reads. What a type guarantees of an argument is what the caller has, not what the operation answers, so a bound resting on it would hold where a caller happened to have such a value. Left to #750 and written into the spec so an author can predict it. souther-examples is now **80 E2011 → 70**.

`Decimal.toInt`'s upper row now has a program of its own (`BelowTen` under `guard d <= 9.0m`), as does each `floorMod` end — the rule-to-program test claimed rows and was checking operations. Each of the four rows was removed once to confirm exactly its own test fails.

All modules green.

kawasima:
The blocking finding is fixed in 6e5a06c6, and the three others with it.

**`noCaseSatisfies` ran before the facts the same condition establishes.** Reordered: the sizes and the operation results are taken first — they hold of the value however the condition comes out — and the cases are asked of that domain. Both of your repros are regression tests now, and both compiled to E2010 before the change:

- `guard Int.min(Int.abs(a), Int.abs(b)) < 0` — the bound this PR itself states
- `guard Int.min(List.length(xs), List.length(ys)) < 0` — the same through `sizeFacts`

**The rule-to-program test counted operations where it claimed rows.** `boundedRows()` now returns one entry per row and the fixtures are compared as a list, so a second bound for an operation that already has a program fails until the new end gets one. Each of the four rows was also removed once by hand to confirm exactly its own test fails.

**The same test could not tell a discharge from a fail-open.** It now reads `InvariantChecker.GAVE_UP` and `WATCHING`: nothing given up on, a construction reached, and every construction reached came out `PROVED`. Counting warnings alone passed on a rule that threw on its first call.

**The spec now states the one-choosing-operation limit** — `Int.min(a, b) + Int.min(c, d) >= 0` is left to the run-time check however much is known of the four arguments.

PR description updated to the measured **80 → 70**. All modules green; souther-examples unchanged from the previous run.

Issue #738, opened 2026-08-15
The adequacy report puts an unreached arm at the call that reached it, not where the arm is written


kawasima:
`WrittenAt` says whether a coordinate is where the code it names is written. One reader asks it —
the renderer that builds a diagnostic's body — and every other place that shows or asserts a
location still reads the coordinate as the place.

## The adequacy report puts an unreached arm at the call that reached it

No module path is needed. The standard library is enough, so this is every model that filters or
maps.

```souther
module demo

data Live
data Gone
data Mark = Live | Gone

data Item = { mark: Mark }

data Count = Int
    invariant value >= 0

behavior tally : (items: List<Item>) -> Count
    constructs Count, Live
let tally (items) =
    Count(List.length(List.filter(i -> i.mark == Live, items)))

example tally
    | "every item is live" : ([ Item { mark = Live } ]) -> Count(1)
```

```
branch      1/2
  · no row goes through `else` (15:23)
```

Line 15 column 23 is `List.filter`. There is no `else` there; the arm is in `souther.list`, and the
coordinate is the call the body was copied onto. The coordinate says so:

```
Core.If at 15:23 sourceId=0 writtenAt=OutOfSight[declaration=List.filter]
```

`AdequacyReport.where` decides whether to name a file and never asks what the coordinate stands for.

The machine-readable report drops the answer rather than ignoring it:

```json
"unreached": [{ "label": "else", "kind": "else",
                "at": { "sourceId": "0", "line": 15, "column": 23 } }]
```

so a tool reading this is told the arm is at `m.sou:15:23`.

## The same rule, written down wrongly at the same site

`Adequacy.BranchEvidence.unreached` — which is what picks the occurrence the report shows — says:

> where a copy was stamped with the call site that spliced it — a helper of another module — the
> occurrences are at different places

A helper of another module compiled alongside is not stamped. That equivalence between "another
module" and "no source here" is the one #734 removed, and it is written here in the javadoc of the
method that chooses what the report points at.

## A diagnostic corrects itself rather than saying the right thing once

The diagnostic path does consult the coordinate, and it does it by appending a sentence:

```
Constructing `Pos` here violates `nonneg` on a reachable path: … The code is written in `up.mk`,
which this compile has no source for, so this is where it was reached from and not where it is.
```

The body is true as a whole and its first clause is not. It reads that way because the correction is
the renderer's and the message was written for the other case. Splitting the message records the way
[...]

Issue #734, opened 2026-08-14
An imported helper hides the construction an invariant reaches


kawasima:
A helper that builds a data inside an invariant is refused with `E1105`, and where the reader is
sent depends on which module the helper is written in. Written beside the declaration, the report
points at the construction and labels the clause that reaches it. Imported from another module of
the same project, it points at the call and says nothing about where the construction is — so the
one thing the author has to go and change is the one thing the report does not name.

## The pair

The same helper, twice, differing only in which module it is written in.

```souther
module m
data Yen = Int invariant value >= 0
let atLeastZero (x: Int): Bool = Yen(0).value <= x
data Table = Int
    invariant ok = atLeastZero(value)
```

```
primary   3:34   Yen(0), in the helper
secondary 5:5    invariant ok = atLeastZero(value)
```

```souther
module helpers exposing ( atLeastZero )
data Yen = Int invariant value >= 0
let atLeastZero (x: Int): Bool = Yen(0).value <= x
```
```souther
module m
import helpers ( atLeastZero )
data Table = Int
    invariant ok = atLeastZero(value)
```

```
primary   4:20   atLeastZero(value), the call
secondary none
```

Both are `E1105`, both say the invariant constructs `Yen`, and only the first says where `Yen(0)` is.
The second has no secondary because there is nothing to point at that the first marker does not
already cover: the construction has been given the call's position, so the construction and the
clause are one line and labelling it twice would say nothing.

## Why this is not #248

#248 was the same loss for a prelude combinator, and #256 settled it by deciding the position per
kind of helper: the call site for a prelude helper, whose body is in shipped source the reader does
not have, and the node's own position for a module-own helper, whose body is in the reader's file.

A helper another module of the same project declares is in neither of those. Its body is in a file
the reader has, and it is being treated as the prelude case. So this is not #248 reopened — it is
the rule #256 wrote, holding for one of the two reasons a body's position is dropped and not for the
other.

## What has to be decided

Not obviously "keep the body's position". Where the declaring module is read off the module path,
its declarations come back as text reassembled from what the module carries, and the positions in
that text name a file nobody holds — the same fact `souther.compiler.check.CitableRegion` was added
for in #730. So the answer has at least a fork in it:

- a module compiled from source alongside this one, whose file the reader does have
- a module read off the module path, whose source this compile never saw

and for the second of those, whether the report names the helper some other way, or says that where
the construction is written is not something this compile can show.

## Done when

[...]

PR #754, opened 2026-08-15
Hold what answers a behavior to the module its rows are written for

### LINKED ISSUE #748: Nothing can say the classes an answerer brings are of the module being evaluated
#732 established that a row can be decided by an answerer whose classes are not the ones the row's
values were built in, and the test that holds it says what it deliberately does not hold:

> Written with the same compile's bytecode rather than a second compilation because what is being
> held is the crossing and not the two compiles agreeing. Same bytes, two loaders, and the JVM makes
> them two type universes on its own.

That is the right scope for what it was testing. The question it set aside has no owner: nothing can
say the classes an answerer brings are of the module being evaluated.

## It does not arise today and does arise next

Every answerer today is `Answering.generatedHere()`, applying the `$Impl` of the compile that built
the row's values. One compile, so agreement is not something to check.

An implementation supplied from outside is compiled against a module's classes as some earlier build
emitted them, and the rows it is held to are read from a `.sou` at the time the run happens. The two
are two builds. Nothing makes them the same build, and a project where the model is edited and only
the tests are re-run puts them a revision apart as a matter of course.

## What goes wrong, and how quietly

A derived decoder is generated from a declaration, so the answerer's decoders read what its
declarations said when it was built. Against a model that has moved:

- An invariant narrowed. A row states a value the model now admits; the older `Title` refuses it, and
  what the run sees is a value it built and cannot hand on.
- A case added to a union. A row expects it; the answerer's `FindTodoResult` has no such case, and
  the answer comes back as something the run reads at a case the model does not have there.
- A field renamed or retyped. The neutral form carries the new name; the older decoder finds nothing
  under it.

The first lands on `FIXTURES_VALIDATED`/`INCOMPLETE`/`INFRASTRUCTURE`, which reads as an environment
problem rather than as a stale build. The other two can land on a comparison, and a comparison is
read as a statement about the model. So a run can report a model as one whose rows do not hold, when
what happened is that nobody rebuilt.

This is the same class of mistake `MemoryClassLoader` is child-first to avoid — its javadoc says a
class left from an earlier build answering instead means "an example was compared against an
implementation this compile did not produce, and a correct model was reported as one whose example
does not hold". Here the earlier build is not on a path to be shadowed; it is what the answerer is.

## The material to check with already exists

`ModuleMetadata` writes a module's declarations onto its classes as Souther source text, and
`PublishedModule` reads them back through the same front end. What is checked on the way in is the
ABI:

    if (m.compat() != Backend.BOUNDARY_VERSION || m.header().isBlank()) {
        throw incompatible(m);
    }

`compat` is the boundary revision. Two builds of two different `.sou` files carry the same one, so it
answers a different question than this.

Declaration-level agreement is the right scope, and it is what the classes carry. Everything the
crossing depends on is a declaration: which cases a union has, what a data holds, what an invariant
admits, what a newtype wraps. A helper a fixture applies is run as a method of the *run's* loader,
so a changed helper body is not part of this.

## The options

**Hold the answerer's published module against the one being evaluated.** Read the declarations off
the classes the answerer brings and compare them with the module the rows are written for. Uses what
is already there, says exactly what differs, and costs a parse.

**Read the model from the classes to begin with.** Then they cannot disagree, because there is one
source. It changes what a run is given — the rows would have to travel with the classes, which they
do not today — and it is a larger question than this issue.

**A digest of the source, written beside `compat`.** Cheapest to check and the least informative when
it fails: it says the build is stale and not what moved.

The first is the one to measure first.

## Acceptance

1. A run whose answerer brings its own classes can say whether those classes are of the module it is
   evaluating, before a row is handed to it.
2. A disagreement is reported as a stale build, distinguishable from a row that did not hold and from
   an environment that was not there.
3. What is compared is what the crossing depends on. A difference that cannot change how a value
   crosses is not a disagreement.
4. The check is not `compat`. Two builds of two different sources at one boundary revision are told
   apart.
5. A run whose answerer is the compile's own does not pay for it: there are no two builds to compare.

### LINKED ISSUE #763: The scope a module is resolved against is assembled in two places
Resolving a module needs two things beside the module: what a bare type name denotes here, and what a bare value name reaches. A compilation assembles both in `Names` — `Imports.compute` for the first, `reachableValues` for the second. Reading a module back from published classes assembles them again, in `PublishedUniverse.denotations` / `aliases` / `reachable` (#754).

The two are not the same code, and they answer the same question. That is the shape #755 was: `Front.LibraryNames` answered one thing for a module being compiled and another for one read off the path, and what it cost was a model that compiles in its own project and refuses to be imported anywhere else. Nobody wrote that difference on purpose; it was a `null` check nobody had a reason to revisit.

## What differs today

Written down so the next one is visible rather than found:

| | compiling | reading back |
|---|---|---|
| bare type names | `Names.Imports` | `PublishedUniverse.denotations` |
| aliases | `Names.Imports` | `PublishedUniverse.aliases` |
| own helpers and behaviors | `reachableValues` | `PublishedUniverse.reachable` |
| what an import brings in | `Bodies.publishedNames` | `Bodies.publishedNames` (shared) |
| library names | `Front.LibraryNames` | `Exposing.check(...).exposed()` |
| behaviors of another module | `BehaviorsInScope` | walked from the module |
| what a qualifier stands for | `Front.ModuleNames` | what has been read so far |

Four of those rows are two answers to one question. Each is right today. Each is a place where a rule can be changed once and stay wrong in the other.

The cost is not hypothetical for #748: agreement compares what a build *means*, so a scope that differs from the compiler's is a comparison of something the compiler never read.

## Why it is not a straight extraction

`Imports.compute` decides the scope and reports on it in one pass — a name an import could not bring in is put in scope denoting nothing (`nameless`), so that a use of it takes the error type instead of reporting an unknown type at every use. The reporting and the scope are the same walk, and the reports are `Db`-shaped: they need `Front.Broken`, `Declarations`, `writtenRegistry`.

`reachableValues` is closer to pure, but reads five queries: `Front.Available`, `BehaviorsInScope`, `Front.LibraryNames`, `Front.ModuleNames`, `Front.Behaviors`. Every one of them is a question about a module by name.

## The shape to aim for

Those five are one interface. Something like:

    interface ModuleUniverse {
        Ast.Module module(String name);
        Set<String> moduleNames();
        Set<String> behaviorsOf(String name);
        Map<String, ValueName.Stdlib> libraryNames(String name);
    }

with two implementations: one over the query layer, one over published classes. Then the scope is assembled once, against a universe, and what differs between a compilation and a reading is which universe answers — not what the assembly does with the answers.

The reporting comes apart the same way. The walk answers with what it built *and what it could not do*, as values; the query layer turns those into diagnostics, and a reader that is not compiling ignores them or reads them as "this module cannot be read here". That is the same split `Exposing.check` already has (`Checked(module, exposed, conflicts)`) after #755.

## Acceptance

1. What a bare type name denotes, what a qualifier stands for, and what a bare value name reaches are decided in one place, whatever the module was read from.
2. A rule about imports — a name a module does not expose, a name claimed twice, an alias claimed twice, a module that is not there — is stated once. Changing it changes both readers.
3. Diagnostics are made where a compilation reads the result, not inside the assembly. Reading a published module raises none.
4. The two universes differ only in how a module is fetched. A test that resolves one module both ways gets the same `Hir.Module` for it.
5. `PublishedUniverse` holds no scope-building of its own.

## Not in scope

Which IR the comparison runs over (#754 uses the resolved HIR; whether it should be a later one is its own question), and the reporting of positions for modules read off the path (#760).

kawasima:
Closes #748, #763.

#732 made a run able to be given something other than this compile to apply a behavior, and the test that holds it says what it deliberately does not hold: the two builds agreeing. Nothing could say the classes an answerer brings are of the module being evaluated. It does not arise today — every answerer is `Answering.generatedHere()` — and it arises the day an implementation is supplied from outside a compile (#695), where the implementation was built against some earlier build of the model and the rows are read now.

What goes wrong is quiet. An invariant narrowed since refuses a value the model admits, and that lands on `FIXTURES_VALIDATED`/`INCOMPLETE`/`INFRASTRUCTURE`, which reads as an environment problem. A case added to a union, or a field renamed, lands on a comparison — and a comparison is read as a statement about the model. So a run can report a model as one whose rows do not hold when what happened is that nobody rebuilt.

## What this does

The material was already there: `ModuleMetadata` stamps a module's declarations onto its classes, for the classes an evaluation runs as well as for the ones that ship, and `PublishedModule` reads them back through the same front end. What was missing was a channel to state which declarations an answer reads a row's values by, and an owner for the question.

**An answer states its origin.** Not where its classes were carried in from — that is transport, and a run has no question about transport — but which declarations a crossing into it is read by. Asked per behavior rather than per answerer, since a resolving answerer has an answer of each kind in one evaluation (#732's third review round). Three defences against getting out of the question: the accessor is abstract, so an answer that does not say is a compile error; what says "there is no second build here" is package-private, so an answerer written outside the package can only stand for its own declarations; and an implementation that answers with nothing is refused rather than read as either.

**Declaration agreement, not build identity.** Both sides are read the way this compiler reads any published module — `PublishedModule` puts the declarations back together as source, `Exposing` reads what the import lines bring in, `Scoping` works out what the names can mean and `Resolve` answers them — and the comparison is over what comes back. So it does not decide what anything means: a bare name an import brought in and the same name written in full arrive as the same symbol, a renamed local arrives as the binding it is, and whether two numbers are one number is asked of `ConstEval`, which already says `1.0m` and `1.00m` are.

**Asked of one behavior.** An answer answers one behavior, and what crosses is what that behavior takes, what it answers with, what it constructs and what it is handed injected — and the declarations those reach. So the comparison starts at the behavior and follows one projection, which is the same projection it compares. Which declarations are compared, which modules have to be read to compare them, and which helpers are part of a declaration come out of that one walk rather than out of three rules that can disagree with each other.

What it decides is which settled facts a value crossing depends on: which cases a union has, what a data holds, what an invariant admits, what a newtype wraps, a behavior's signature, which behaviors are injected, and the `let`s a declaration cannot be read without. A `let` body, an example, an import line nothing compared is written with, the order two declarations are written in and how they are spaced are not things a value can be read differently by. A field's name is compared as the word it is, because a decoder reads a value under it.

**Three readers, three vocabularies.** A person gets one diagnostic per behavior (E1927), naming the declaration that moved, or saying that nothing could be told. The row records `FIXTURES_VALIDATED` / `INCOMPLETE` / `ANSWERER_ESTABLISHMENT` / `Applied.Nothing` — never `FAILED`, since a stale build reading as "the model's rows do not hold" is the mistake being removed, and never `FailurePhase.NONE`, since a row that ended undecided while saying nothing went wrong is a state nothing produces. A measure gets `ANSWERER_NOT_ESTABLISHED`, read off the row's phase at the single point where an undecided row becomes a reason, through a switch with no default. `leftNoRowRead` is false: the rows were read, and what is missing is only what the running would have decided.

A run whose answers are the compile's own reads no declarations at all — held by giving four such runs a reader that raises.

## Reading a published module through the one place a scope is assembled

This closes #763 as well, whose first half is already on develop. `Scoping` is now the one place a module's name-resolution environment is derived, against a `ModuleUniverse` that says what it has under a name and nothing about what a name written in a module means. `PublishedUniverse` was the second assembler — its own denotations, its own aliases, its own reachable set — and two assemblers is exactly what the seam exists to remove: the readings drift, and a model compiles in the project that wrote it and refuses to be imported anywhere else.

It now says only which modules to read. A module's declarations reach the modules they name, those are read, and what came back is handed over as a universe; what the names mean is `Scoping`'s and what answers them is `Resolve`'s. The declarations a module is read as declaring are settled once, where the universe is assembled, and resolution reads other modules by what that settled — so a module whose declarations cannot be indexed is an absence to every asker rather than a raise to the second one.

A refusal is this reader taking the module as one it cannot read, which is what `Scoping` leaves to whoever is reading. That makes the two absences worth telling apart: a module these classes do not carry is unknown, so an import of it is refused and the importing module is not read; a module they carry and this compiler cannot read says nothing further, since whatever is wrong with it is reported where it is. Collapsed into one, an import of something absent leaves its names in scope denoting nothing, the importing module reads back with its field at the error type, and a comparison of two builds reads that as a declaration.

## What the comparison is held to

The comparison walks a form's parts when it has nothing else to say about it. That is right for a form of the grammar and is a default for everything else, so a record the compiler puts beside a declaration to say something about itself — which pass rewrote a node, what a coverage point was numbered as — joins the comparison the day it is added, silently, and two builds that mean the same thing start disagreeing over it.

So the walk is required to be a decision: a record a declaration reaches is a form of the grammar, one of the front end's settled answers, one the comparison names, or one it erases. Each is asked of where the type is declared rather than of a list that goes stale beside the thing it lists. Answering it turned up two that had never been decided — what a definition was made as, and what an example row's position contributes to reading what is written at it — and both are erased, being this compile's record of how it built its own tree rather than anything a published declaration carries.

The comparison is also symmetric, which it was not: erasure held a position against a coverage number and called them equal, and the guard on what a collection holds was applied to one side, so a form arriving in the other side's set was compared by an equality that reads what this erases.

## Acceptance

1. A run whose answerer brings its own classes can say whether those classes are of the module it is evaluating, before a row is handed to it — `ARowIsNotHandedToAnAnswerFromAnotherBuildTest`.
2. A disagreement is reported as a stale build, distinguishable from a row that did not hold (the rows end `INCOMPLETE`, and no row-level diagnostic is said) and from an environment that was not there (E1927 against `INFRASTRUCTURE`).
3. What is compared is what the crossing depends on — the reformatting, comment, and moved-body cases all agree, and so does a declaration written beside the behavior that nothing crossing it reaches.
4. The check is not `compat`: two builds of two different sources at one boundary revision are told apart, which is what the whole declaration comparison is for.
5. A run whose answerer is the compile's own does not pay for it — the declarations are behind a `Supplier` that raises if called.
6. `PublishedUniverse` holds no scope-building of its own (#763's acceptance 5): no denotations, no aliases, no reachable set, no `Resolve.Values`.
7. Nothing an answerer hands back stops a compile: an answer that does not say which build it reads by, and a published module this compiler cannot index, are both answers rather than raises.

## Measurements

- `CI=1 mvn -Plint clean verify` BUILD SUCCESS: compiler 4765, fmt 2224, lsp 219, cli 333, bench 6.
- souther-examples: `souther examples --format json` over all 15 projects, 26,653 lines, byte-identical to the same run under develop's jars. Re-run as `mvn -o verify` against this build after each rebase: 12 of 13 modules SUCCESS, and `todomvc` fails to compile `TodomvcTest.java` against its generated types, which it does under develop's jars as well.
- Mutants were shown to kill the test that covers the rule they remove, for each rule this adds: not comparing invariants, cutting the import closure, failing open on unreadable declarations, comparing positions, handing the row over anyway, collapsing the measure's code to `ROW_UNDECIDED`, recording the row as `NONE`, and dropping the module path from what a compile reads declarations of.
- And for the rules review added: reading a module these classes do not carry as one they carry and cannot read leaves a module readable with its field at the error type; giving resolution a registry that indexes on demand lets the raise out of the second ask; and reading a missing origin as this compile's own hands a row to the answerer. Each kills its test.

## What the schema version says about itself

`answerer_not_established` is a word added to the `code` enum, and the version stays at 1. The schema states the rule in its own description: a change that removes or renames anything raises the version and something added does not, since a word added to an enumerated field is one no earlier document carried. It also says what an older reader should do — a reader that has to accept newer documents wants the newer schema, not a higher number — so the one-directional incompatibility is the stated design and not an oversight here.

## Two departures from the plan

The specification and the schema could not be a commit of their own. Introducing E1927 turns three registry tests red at once — every code must name a rule anchor the specification carries, must resolve to a section `souther doc` can answer from, and every `Incompleteness.Code` must be a word the schema knows — so the rule, the code's section and the schema word are in the commit that adds the code.

The design said `Disagree` would carry both sides' stamped declaration text, for a reader to diff. It carries the module and the declaration instead: which declaration moved is what names the difference, and what to do about it is the same whatever the two texts say. Left out deliberately rather than forgotten.


kawasima:
Every finding was reproduced before it was fixed, and each fix is held by a test that fails without it. Six were real; the reflection one was the reason most of the others could exist.

**`ours` did not include the module path** (high). Reproduced: a model importing a compiled module reports `Unreadable` for it, of two builds that agree about everything. `ours` now reads from where the evaluation loader draws its classes — this compilation's first, then the path — behind one named function so the two cannot come apart. Held by `WhatACompileReadsDeclarationsOfIsWhatItCanReachTest`; removing the path fallback turns both its tests red.

**`Unreadable` did not say which side** (medium). Combined with the above, the common failure named the answerer's jar. `Agreement.Unreadable` carries a `Side` now, and the hint is chosen from it: what the answer brought is that build's to fix, what this compile reads is its own path's.

**Every published `let` was compared** (high). Reproduced: an exposed helper no declaration is read through, body changed, reported as a stale build. `ModuleMetadata` publishes invariant-reached helpers *and* the exposing closure; only the first is part of what a value crossing meets. The comparison now takes the helpers a declaration cannot be read without, reached through `ModuleMetadata.reach` itself rather than through a second reading of the same rule. The sibling test holds the other direction: a helper an invariant *does* call still disagrees when it moves.

**Import names were order-dependent** (high). Reproduced directly. They are a set now, and compared by canonical name rather than by spelling.

**One diagnostic per block, not per behavior** (medium). Reproduced with a behavior whose rows are written in two `example` blocks: two E1927s. Now said once per behavior.

**`PipeBehavior.declaredOut` is nullable** (low). Guarded like the sibling `FnDef.declaredReturn` next to it.

**The reflection comparator** (P2 — and the one that mattered most). The finding is right, and the fix is not a projector per `Expr` arm: what was missing is that nobody was being made to decide. The comparison now names what it erases and what it compares as a value, and **stops** on a form it has not been told about. `EveryFormADeclarationIsMadeOfIsClassifiedTest` walks every type reachable from a published declaration — statically, so a form in a corner of the grammar no fixture writes is covered too — and fails with the question rather than with "add it to a list".

It immediately found two forms the walk had been comparing by whatever equality they happened to have: `BigDecimal` (a number a declaration states) and `TypeSymbol` (which declaration a name means). Both are now classified deliberately. Then the fail-fast caught my own `Set` in `bindings()`, which is what it is for.

**The `Disagree` text/diff** stays out, as a decision rather than an omission: the module and the declaration name what moved, and what to do about it is the same whatever the two texts are.

Measurements: `CI=1 mvn -Plint clean verify` BUILD SUCCESS (compiler 4600, fmt 2224, lsp 204, cli 333, bench 6); souther-examples across all 15 projects still byte-identical to develop's jars (26,653 lines).

One thing I checked and left alone: a published module whose invariant calls a bare imported name (`import String ( length )` … `invariant length(value)`) fails to read back with E1025. It reproduces on develop's CLI, so it is not from this change; the tests here write `String.length` and I have not opened anything for it.

kawasima:
Both P1s reproduced before they were fixed, and the P2 turned out to be hiding three live cases. `cb2836ef`.

**Helper-only import** — reproduced exactly as described: an exposed helper's body changed and its now-unneeded `import String ( trim )` dropped gives `Disagree("import")`. The diagnosis is right too — helper projection and import projection were at different depths. The bindings compared are now the ones the compared declarations are *written with*, computed by walking the same forms the comparison walks. So the two projections cannot drift: whatever set of declarations is compared decides which imports are.

**BigDecimal** — you are right, and my comment claiming otherwise was asserting something the language denies. `ConstEval.equal` uses `compareTo` and says why: "`1.0m` and `1.00m` are one number, and a comparison folding the other way would decide at compile time what the run time denies". The comparison does the same now, with a test for the scale difference and a sibling holding that a bound that really moved still disagrees.

**Reflection / `type.isRecord()` as implicit approval** — the finding is right; I disagree with one half of the conclusion and acted on the other.

The half I disagree with: for a form *of the grammar*, comparing by default is the correct default, and it is not symmetric with the leaf case. A grammar form compared when it need not have been reports a build that has not moved — a diagnostic someone reads and answers by erasing it. A grammar form left out reports agreement that was never established, and nobody ever finds out. Given #748 exists to stop a quiet wrong answer, the noisy failure is the one to default to. So `Grouped` would produce a false stale, get read, and be erased — which is the loop working.

The half you are right about: **a record that is not a form of the grammar** is not covered by that argument at all. It arrived with a declaration rather than being part of one, and there `isRecord()` really was approving something nobody decided. Those are named explicitly now (`ValueName`, `BindingOwner`, `MapKeyRepresentation` are meaning; everything else stops the comparison), and `EveryFormADeclarationIsMadeOfIsClassifiedTest` holds the whole reachable set to being classified rather than to being a record.

That immediately caught three that were being compared by accident, all of which do reach a published declaration: `ConstructionOrigin` (where a construction was carried in from), `CoverageOrigin` (what a coverage point was numbered as), and `BindingId` (a binding's owner and its ordinal — a number two builds get by walking). None of them is anything a value crossing is read by; the last one would have gone stale on any reordering. So the mechanism found real cases on its first run, which is the argument for it.

Also from the second review, all reproduced or verified:

- `reading` caught only `CompileException`; a class file this JVM will not parse aborted the compile. It answers `NOT_READABLE_HERE` for every way a class file can refuse to be read now.
- The hint for the run's own side said `{module}` is "which the rows are written for" — it is whichever module in the closure could not be read. Reworded in both catalogs.
- Two consecutive javadoc blocks on `declarationsRead`, with `evaluationLoader` left undocumented. Moved back.
- `PipeBehavior.declaredOut` — already guarded in the previous round.

**Once per behavior, across sources**: I fixed the specification rather than the code. `said` is per verifier and a verifier is per source, so a behavior with rows in the module file and in an attached `examples for` file gets one diagnostic in each. That is what a diagnostic being said where it can be quoted means (#734, #738): dedup across sources leaves a reader of the second file looking at rows that stopped with nothing saying why. The spec sentence now says "once for the behavior in each source its rows are written in", with that reason.

Measurements: `CI=1 mvn -Plint clean verify` BUILD SUCCESS (compiler 4603, fmt 2224, lsp 204, cli 333, bench 6); souther-examples still byte-identical to develop's jars across all 15 projects. PR body updated to the current numbers, and the `Disagree`-without-diff decision is recorded there as a departure rather than left implicit.

kawasima:
All four reproduced, and you are right that patching them one at a time would keep producing the same shape. `d7dcc82b` puts the projection in.

**The qualified-reference false negative is the one that mattered.** Confirmed with the CLI first — `data Todo = { title: example.shared.Title }` with no import line compiles — and then in the comparison: narrowing `example.shared.Title`'s invariant left it `Agree`. That is exactly the failure #748 exists to stop, arriving by a spelling the walk did not look at.

The other three reproduced as described: `S.Title` vs `example.shared.Title` disagreed; an invariant clause named `trim` made `import String ( trim )` look needed (I used a clause name rather than a literal — `namesWritten` collected any `String`, so both land the same way); and the closure followed import lines the bindings had already been narrowed away from.

**One projection now, and both questions asked of it.** A name is read as the declaration it reaches — the import lines say what a bare name and an alias stand for, a qualified name says its module in itself — and that single reading is what the comparison compares and what the walk follows. Consequences, all of them tested:

- a module reached by a qualified name with no import is followed
- an import line nothing compared is written with is not followed, and is not compared
- alias and full spelling agree
- `bindings()` and `reachedUnder()` are gone: an import is now *how a name is read*, not a thing compared, so a binding difference shows up as the name difference it is

This is short of full resolution and deliberately so: what is used is the part of resolution a published declaration carries with it, which is enough to say which declaration a name reaches. If that turns out not to be enough, the next thing it needs is the front end's own resolution over the closure, and the seam is in one place to hand it to.

**`PipeBehavior`** — measured rather than assumed. I made the arm throw and ran a model whose behavior is `price >-> settle` through the comparison: it never arrives, because `ModuleMetadata.signatureOf` publishes the computed signature and it reads back as a declared behavior. So the arm is a refusal now, with a test that goes through it to say so — a day when a composition does arrive fails loudly rather than comparing stages by a rule nobody could read the truth of.

Two mutants confirm the new rules bite: comparing `canonical()` again kills the alias test, and following import lines again kills the qualified-reference test.

Measurements: `CI=1 mvn -Plint clean verify` BUILD SUCCESS (compiler 4607, fmt 2224, lsp 204, cli 333, bench 6); souther-examples still byte-identical to develop's jars across all 15 projects. Spec paragraph and PR body updated to the projection as it now is.

kawasima:
Both reproduced, and the diagnosis is right: the projection was treating every `WrittenName` as a reference to a declaration. `ab25e6f1`.

- renaming `longEnough`'s parameter, with its use renamed too, gave `Disagree` — an invariant that admits exactly what it admitted
- a field called `note` in a module that imports `note` was read as that import, so dropping the import reported a stale build

**A name is compared as what it is doing where it is written.** Three things, and the form that writes it says which:

- **reaches a declaration** (`Var`, `Name`, `TypeRef`) — compared as the declaration it reaches, as before
- **binds a name** (`Binder`, a behavior's `Param`) — held to whatever the other build binds in its place, both directions, so two names on one side cannot both stand for one of theirs. That is the α-equivalence you asked for, and it falls out of the walk: the two sides are traversed together, so binder occurrences pair up positionally and uses are compared through the pairing. No scope tracking, and shadowing can only make it stricter, never looser.
- **is a name** (`Field`, `FieldInit`, `FieldAccess`, each declaration's own) — the word it is

`namesWritten` collects only the first kind now, so the closure follows what the declarations reach and nothing else.

**The classification is enforced, not documented.** A form that writes a name and has not been classified stops the comparison, and `EveryFormADeclarationIsMadeOfIsClassifiedTest` holds every form reachable from a declaration to having been said about. The default that matters is the one I did *not* take: an unclassified form falling through to "is a name" would compare `Title` against `Title` and call two builds agreed about a declaration neither reads the same way. So it throws instead.

On the larger point — running `Resolve` over the closure: I did not, and I think the line is in the right place for what this needs. What the comparison uses is the part of resolution a published declaration carries with it (its import lines, its qualified spellings) plus the pairing the two-sided walk gives for free. That answers "which declaration does this name reach" and "is this the same binding", which is what a crossing depends on. If something turns up that those cannot answer, the front end's resolution goes in at one seam — `Names`, built per module in one place.

Also fixed the two stale comments you spotted: the class javadoc no longer says imports are the edge, and `Origin.Published`'s `@param` says the modules the declarations reach rather than the ones imported.

Measurements: `CI=1 mvn -Plint clean verify` BUILD SUCCESS (compiler 4610, fmt 2224, lsp 204, cli 333, bench 6); souther-examples byte-identical to develop's jars across all 15 projects. A mutant comparing binder spellings again kills the rename test.

One test-harness change worth flagging: `declarationsOf` now rejects only errors, not warnings. An unused import is a warning, and a build that carries one is exactly what one of these tests measures — the old assertion made that model unbuildable for the wrong reason.

kawasima:
One fixed, one I am pushing back on. `ddd8390b`.

**Swallowing every `RuntimeException` — right, fixed.** A fault inside the reader reached the user as "what `{module}` published cannot be read by this compiler", which is a message they cannot act on about something that is not theirs. Two things are caught now, and they are the two a build can cause: `CompileException`, and the bytes not being a class file this JVM reads. The second is `IllegalArgumentException` — measured against `ClassFile.of().parse` on truncated and empty bytes rather than assumed.

**The `IllegalStateException`s from `agreed` — I think raising is correct, and I have written down why.**

The reasoning in the finding holds for anything a *build* can carry, which is exactly what `reading` catches. But an unclassified form cannot come from a build:

- both sides go through this compiler's own front end (`PublishedModule.read` → `CstFrontend.parse`), so the forms that arrive are the forms this compiler parses
- a module published at another boundary revision is refused before any of this (`Backend.BOUNDARY_VERSION`), and one published at this revision is parsed by this grammar

So a form reaching that throw is a hole in what `DeclarationAgreement` has been told — this compiler's gap, not a fact about the two builds. Answering it with `Unreadable` would report a stale build over that gap, and quietly: the user rebuilds, nothing changes, and the hole stays. It is the same reason `sameShape` refuses an unclassified type at all rather than falling back to `equals` — I would be putting the fallback back one level up.

Same for the `PipeBehavior` arm: a composition is published as the signature its stages compute, so it does not arrive; if one does, the premise this comparison is built on has broken.

I have written that distinction into `of`'s javadoc — what a build carries is answered, what this compiler did not decide is raised — with the boundary-revision argument, since without it the two really do look like the same case.

Measurements unchanged: `CI=1 mvn -Plint clean verify` BUILD SUCCESS (compiler 4610, fmt 2224, lsp 204, cli 333, bench 6); souther-examples byte-identical to develop's jars.

kawasima:
Marking this draft. The last round's finding — the `Bound` map is not a lexical scope, and a `let` whose initializer names an import is read as naming the binding — is a false *agreement*, not a false stale, and it is the fourth kind of language semantics this comparator has had to grow. Name resolution, decimal equality, type identity, and now scope and α-equivalence. Six rounds, eleven findings, one root: it is reconstructing what the front end already decided, from raw AST.

So the mechanism is being replaced rather than patched again. The rule it will be built to is:

> `DeclarationAgreement` must not answer any question the Souther front end answers. Its input is already-resolved declaration semantics; it decides only which of those a value crossing between builds depends on.

What that means concretely, measured rather than assumed:

- **The stage is after resolution.** `Hir` carries `TypeSymbol` (which declaration a name is), `ValueName` (what kind of thing it denotes), and `BindingId` (which binding) — every ambiguity a crossing depends on is settled there, and no codegen information has arrived yet. Type checking, totality and coverage buy nothing for this question.
- **The projection is erasure only.** Drop positions, provenance, coverage origins, examples, the `let`s no declaration is read through. Whatever semantic identity survives is used as it stands. No `BigDecimal.compareTo` rule of its own, no name normalisation of its own — a `CrossingView` that computes meaning is the same mistake under a new name.
- **α-equivalence stops being a scope problem.** Two resolved graphs' `BindingId`s are paired positionally as the walk goes; the scopes were decided by the front end. Shadowed names are already distinct ids, so the flat pairing that is wrong today is right there.

Nothing outside `DeclarationAgreement` changes: the `Origin` seam, `TheCompilesOwn`, the three reporting channels, the row lifecycle (`FIXTURES_VALIDATED` / `INCOMPLETE` / `ANSWERER_ESTABLISHMENT` / `Applied.Nothing`), `ANSWERER_NOT_ESTABLISHED`, and their tests all stay.

**Blocked on #755.** Reading the answer's side through the front end needs published modules to mean what they meant where they were written, which is exactly what #755 is. It is being fixed in the shared pipeline — `Exposing.check` returning the module and its table together, and `FromPath` carrying one record per published module — and both of those are where the answer's side has to be read from. Building an entrance of my own first would mean building it twice.

One correction to my own last comment: the negative control I claimed for #755 did not control anything. `Compilation.diagnostics()` drops reports about path modules — they have no source id to be filed under — so the model that fails on the CLI comes back clean there, and I read that as "does not reproduce". It reproduces; the harness was silent. A test on that path has to assert `classes()` too, not only `diagnostics()`.

kawasima:
The mechanism is replaced. `c1130829`, on top of #755.

**What made it possible.** #755 fixed the pipeline so that a published module means what it meant where it was written, and `Exposing.check` now returns a module together with what its import lines bring in. That is exactly what reading the answer's side needs, so no second `Compilation` and no entrance of my own: `PublishedModule.read` → `Exposing.check` → `SyntaxSymbols` → `Resolve`, the compiler's own path, run over a second set of classes. Measured before building on it — a published module's bare `length` comes back as `ValueName.Stdlib("String","length")`, and `UNRESOLVED: 0`.

**What the comparison sees now.** `TypeSymbol` where a type was named, `ValueName` where a value was, `BindingId` where a binding was. So the four kinds of language semantics that kept arriving in review are gone from here:

- alias / bare / fully-qualified — one symbol, nothing to normalise
- lexical scope and α-equivalence — the front end settled which binding each use is of; `BindingId` is already `probe.bind.longEnough.0` on both sides, so a renamed parameter is the same declaration with no scope tracking at all
- decimal equality — `ConstEval.equal`, which is where the rule lives
- which occurrences are names of something — the symbols say it

The false agreement from the last round cannot arise: the initializer's `check` resolves to `ValueName.Helper(a, check)` and the body's to `ValueName.Local`, and those are not equal.

**What went.** `Names`, the `Bound` over spellings, `collectNames`/`namesWritten`/`modulesReached` over written names, `WRITES_A_NAME`/`Says`, `isAFormOfTheGrammar`, the `BigDecimal` rule. About 400 lines, all of it the language's rules written a second time. `EveryFormADeclarationIsMadeOfIsClassifiedTest` went with them — it was holding a classification that no longer exists.

**What stayed.** The crossing policy, which is this issue's to state: `crossingParts`, `readThrough`, the closure, `Agree | Disagree | Unreadable`. And everything outside this class — `Origin`, `TheCompilesOwn`, the three reporting channels, the row lifecycle, `ANSWERER_NOT_ESTABLISHED`.

The rule is written at the top of the class, so the next round has one answer to give: **it must not answer any question the front end answers.**

Measurements: `CI=1 mvn -Plint clean verify` BUILD SUCCESS (compiler 4710, fmt 2224, lsp 204, cli 333, bench 6); souther-examples byte-identical to develop's jars across all 15 projects, 26,653 lines. Three mutants — comparing a use by its spelling, a binding by its name, a number by how it was written — each kill the test that covers the rule they remove.

One test-harness note: `NoPublicWayToTurnASpellingIntoATypeIdentityTest` caught the new reader building an identity from a module and a name. It asks `Registry.identify` now, so the identity is the declaration world's answer rather than one made here — which is what that invariant exists to hold.

kawasima:
Four of five were real and are fixed; the fifth does not reproduce. `a546cee0`.

**1 and 2 were the same gap seen from both ends, and both reproduced.** A module's declarations reach another module's, and neither the reading nor the comparing followed that.

Reading: `import example.money ( Amount, atLeast )` with `invariant atLeast(amount)` came back `null` — the value namespace I handed resolution was `Values.of(module)`, which is documented as what a module reaches *with nothing else in sight*. Two matching builds were reported as an answer nothing could be told about. It is assembled the way `Names.reachableValues` assembles it now, through the same `Bodies.publishedNames`, with imported behaviors and a real `Elsewhere`.

Comparing: with that fixed, tightening `atLeast` from `>= 1.0m` to `>= 5.0m` gave `Agree`. Exactly as diagnosed — `readThrough` was per module and reached only from that module's own data invariants, so a rule written where the type is declared and called by a reader's invariant belonged to neither side of the walk. The closure is now read whole before anything is compared, and what a declaration is read through is gathered across all of it, keyed by the module the front end says each helper is of. Mutants confirm both: restricting `helpersNamedIn` to the walked module kills the new test, and dropping `publishedNames` kills the reading one.

**3 — right, and it was mine.** `rg` returns no caller. I widened `ModuleMetadata.reach` intending to share it, then wrote a second answer over `Hir`, and left a javadoc asserting the sharing. That is the restatement the comment itself warns about. Reverted to private, claim deleted.

**4 — right.** `declares` answered from the class file existing; `PublishedModule.read` answers from the declarations being on it. A class carrying no annotation was reported as `NOT_READABLE_HERE`, sending a reader after a boundary revision that is not the problem. Both ask the same question now.

**5 — does not reproduce, and I would rather not add a guard.** A 400-deep nested invariant is refused by the parser first (`E2104`, "This nests deeper than I read"), so it never reaches the comparison; `PublishedModule.read` raises `CompileException`, `reading` catches it, and the answer is `Unreadable` — the documented contract. A declaration that parses is one this compiler's own depth limit already bounded, and the recursion here is over the tree that limit bounds. Adding a second bound would risk the opposite error: refusing a declaration that is within the limit. If a case turns up where the parser admits something the walk cannot, that is a real finding and I would take it.

Measurements: `CI=1 mvn -Plint clean verify` BUILD SUCCESS (compiler 4712, fmt 2224, lsp 204, cli 333, bench 6); souther-examples byte-identical to develop's jars across all 15 projects.

kawasima:
All four, fixed. `21e5a65f`.

**Set / map falling back to `equals`** — right about the mechanism, and I measured the reach: a probe over a module with `List<Tag>` and `Map<String, Tag>` fields finds no Java `Set` or `Map` anywhere in the resolved declarations (`Souther`'s `Map<K,V>` is a `Hir.TypeRef`), so nothing hits it today. That is not a reason to leave it — the branch is a prepared wrong answer, left from the version where `bindings()` built sets of names. A form of a declaration held in a set or used as a map key stops the comparison now, with the reason: a collection compares what it holds by its own equality, which reads the positions and bindings this erases, so answering by it is a decision nobody made.

**`Bound.bind` writing before refusing** — right. Checked before written now. Nothing reads it after a refusal today, which is the sort of thing that stops being true without anyone noticing.

**`readThrough` from one side only** — right, and I would not call it latent. A helper only their declarations read through is *their* answer about what their values are; choosing what to compare from our side alone compares the two builds by what one of them says a declaration is made of. Both closures are read before either is compared, and the union is used.

**The stray javadoc line** — mine, from widening `reach` and reverting it. `ModuleMetadata` is byte-identical to develop again.

Measurements: `CI=1 mvn -Plint clean verify` BUILD SUCCESS (compiler 4712, fmt 2224, lsp 204, cli 333, bench 6); souther-examples byte-identical to develop's jars across all 15 projects.

kawasima:
All three, and the fourth needs a decision from you rather than a patch from me. `5930bb75`.

**The `FieldAccess` false agreement is real and reproduced.** `r.min >= 0` against `r.max >= 0` gave `Agree`. Erasing every `WrittenName` went one step too far.

I did not move the IR boundary, and I want to give the reasoning rather than just the result. The finding names the right defect but I think it locates it one level out. What was wrong was not "HIR is too early" — it was that I erased spellings *unconditionally*, when the front end only answers *some* of them. So the rule is one rule, not a list of fields to remember:

> A form carrying a spelling and, beside it, what the front end settled it to be is compared by the answer. Everything else is compared as written.

A use by its `ValueName`, a type by its `TypeSymbol`, a binding by its `BindingId` — and a field name by the word it is, because nothing beside it says what that word means and a decoder reads a value under exactly that word. Reading it is not answering a question the front end answered; it is reading the answer, which happens to be the spelling.

Which way round to default is the part that matters. A spelling read where an answer was already there reports a build as moved over how a name was written — noisy, visible, answered by naming the settled form. A spelling passed over where nothing else says what it means is silent, and silence is what #748 exists to remove. So the default is to read.

`AFormThatCarriesItsAnswerIsComparedByItTest` reads the forms rather than a list kept in the test, and holds every one carrying both — six of them — to having an arm. A form that grows an answer and keeps being compared by its spelling fails there. `Hir.Field` and `Hir.FieldAccess` are held to *not* being among them, so the control is real.

If you still want the boundary at Core, that is a bigger change than this and I would rather do it as its own piece of work: `Elaborator.elaborate` needs a `Scope` and a `CheckContext`, so reading the answer's side through it means running type checking over the closure of a second universe. Say the word and I will scope it.

**`injectedBehaviors` — a real regression, mine.** `PublishedUniverse.resolved` returned `Hir.Module` and dropped what travels beside it. Two builds where one implements a behavior and the other leaves it open agreed about the signature and said nothing about the thing that decides whether an implementation may be handed in at all. It travels as far as the comparison now (`PublishedUniverse.Read`), and there is a test where only the injected status differs.

**Reading closure vs agreement closure — right, and the distinction is now in the code.** `PublishedUniverse` still reads widely, because resolution answers every name or none. What is compared is what the compared declarations reach: declarations, signatures, and the helpers those are read through. The test uses a user module, not stdlib — and I had to fix it once: my first version put nothing comparable in the unread module, so it passed under the mutant too. It now declares a type whose invariant differs between builds, so comparing it would be visible.

**The second resolution harness — I agree it is the remaining structural debt, and I would like your call on scope.** `denotations`/`aliases`/`reachable` do assemble what `Names.Imports` and `Names.reachableValues` assemble. The difference is that those are `Db`-shaped: they read `Front.Available`, `Front.Broken`, `Declarations`, `BehaviorsInScope`, and report diagnostics as they go. Sharing them means either giving the query layer a second universe to answer about, or lifting the pure part out from under the reports. Both are real changes to the front end's shape, and neither belongs in a commit whose subject is agreement. I would take it as the next piece of work, with #755's shape as the precedent: fix it where the pipeline is, so every consumer gets it.

Measurements: `CI=1 mvn -Plint clean verify` BUILD SUCCESS (compiler 4717, fmt 2224, lsp 204, cli 333, bench 6); souther-examples byte-identical to develop's jars across all 15 projects. Three mutants — erasing every name, dropping the injected comparison, comparing everything read — each kill the test that covers the rule they remove.

PR #756, opened 2026-08-15
Carry what a module's import lines brought in wherever the module goes

### LINKED ISSUE #755: An invariant that calls a bare imported name cannot be read back from a jar
A module may write `import String ( length )` and then call `length` bare, and that is what the import is for (`exposing`, spec §modules). Where such a module is published and read back from its classes, the importing project cannot compile: the bare name resolves to nothing, and what it is told is to qualify a name it never wrote.

## Reproducing it

Two files, and nothing in the second one mentions `length`.

    // lib.sou
    module lib.text exposing ( Title )
    import String ( length )

    data Title = String
        invariant length(value) > 0

    // app.sou
    module app.uses
    import lib.text ( Title )

    data Note = { title: Title }

Compiled separately, as two projects are:

    $ souther compile -d out lib.sou
    $ souther compile -cp out app.sou
    -- NAMING ERROR  E1025 -------------------------app.sou:5:15

    5|
       ^

    `length` is a standard-library name. Write it qualified (`String.length`, `List.length`)
    or import the name you mean.

Writing the invariant as `String.length(value) > 0` in `lib.sou` compiles both. So the two spellings the language says mean the same thing do not survive publication alike.

## What it costs

An invariant is part of what a type is, so it has to be readable wherever the type is imported — that is why `ModuleMetadata` publishes it as source and `PublishedModule` reads it back through the same front end. A module that imports names bare is readable in its own project and unreadable in every other one, and nothing says so at the point it is published. The author of `lib.text` finds out from a downstream project's build, about a file they cannot see.

What the reader is told is also wrong in a way that costs time: `app.uses` is told to qualify `length`, and there is no `length` in it to qualify.

## Where to look

`Front.FromPath` reads a published module and then drops its library imports:

    Ast.Module m = Exposing.withoutLibraryImports(published.module(),
            Exposing.read(published.module()).kept());

`withoutLibraryImports` is documented as dropping `import List ( ... )` lines "having been checked" — checked, that is, against the bindings of the module being compiled. The published module's own bindings are what its invariant was written under, and this is a second reading of it in another module's context. Whether the fix is to keep the line, to resolve the published module's names before dropping it, or to publish the invariant with its names already qualified, is the question this issue is about; the third would make what a jar carries differ from what the author wrote, which is the thing publishing-as-source exists to avoid.

`PublishedModule.withNeededImports` keeps an import whose names appear in the published text, so the line is on the class — `length` is a word of the invariant. It is dropped after being read back, not before being written.

## Observed alongside, and not the same defect

The diagnostic is reported at `app.sou:5:15`, and `app.sou` has four lines. The coordinate is where `length` is written in `lib.sou`; the file is the one being compiled. So a position that arrived with a declaration from another build is being quoted against the file this compile happens to be reading, rather than being said as a place this compile cannot quote (#734, #738). That would be worth a check of its own even once the resolution above is fixed — a wrong coordinate in a file the reader is looking at is worse than no coordinate.

## Acceptance

1. A project can compile against a published module whose invariant calls a name it imported bare.
2. `invariant length(value) > 0` under `import String ( length )` and `invariant String.length(value) > 0` are read back as the same declaration, since they are the same declaration.
3. Whatever the resolution is, it holds for every published `let` an invariant reaches, not only for the invariant's own expression — a published helper is read back the same way.
4. A diagnostic raised while reading a module back does not carry a coordinate into a file that does not have one.

kawasima:
Towards #755 — acceptance 1, 2, 3, and 4 as far as a primary coordinate goes.

Whether it closes #755 is a decision about the issue's wording, not about this branch. Acceptance 4 says "a diagnostic raised while reading a module back does not carry a coordinate into a file that does not have one", and does not say primary. A second region built over a body spliced in from out of sight is the same sentence about a different region — #760 — and cannot be answered where reports are read: a label naming no source of its own already means "read in the diagnostic's file", so nothing there tells one apart from a hand-made label. It wants provenance where the splice happens. Either #755 waits for #760, or its acceptance 4 is narrowed to the primary coordinate on purpose.

## What was wrong

`import String ( length )` and then `length(value) > 0` in an invariant. The import line is published on the class, and `Front.FromPath` dropped it and threw away the table saying what it had brought in. `Front.LibraryNames` was asked of the source layout — which a module off the class path is not in — and answered an empty table, so every bare name in a published invariant denoted nothing in every project but the one that wrote it.

Measured before the change, all four fail with `E1025`:

| | |
|---|---|
| an invariant calling a bare imported name | ✗ |
| a published helper an invariant calls, calling one | ✗ |
| `import String as S ( length )` | ✗ |
| a module reached only through another off the path | ✗ |

## The two failures this issue names are two failures

Acceptance 1–3 are a semantic loss. Acceptance 4 is a separate observable defect, and it is worse than a wrong caret.

The coordinate in a report about a published module is a line and a column of text this compile reassembled out of what the module published. Every surface quotes a line from a file, and a run with one source has no way to tell an untagged report from a report about no file at all:

```
-- NAMING ERROR  E1025 -------------------------app.sou:5:15

5|
   ^
```

`app.sou` has four lines; `5:15` is where `length` sits in `lib.sou`.

A run with several sources drops the report instead — `Compilation.diagnostics()` files by source, and a module off the path has none. Through the query API the editor reads, the same input answered **no classes and no diagnostics**:

```
bare       diagnostics: {0=[]}   classes: []
qualified  diagnostics: {0=[]}   classes: [app.uses.Note, … 6 classes]
```

`Compiler.compileModules` threw, because `failure()` read `db.allReports()` without filtering by source. Two readings of one set.

## What changed

**The table travels with the module.** `Exposing.check` answers the module without its library import lines *and* what those lines brought in, as one value — there is no longer a way to take the second and leave the first. A module off the path carries the table beside its injected behaviors, the other fact about it this compile cannot work out again. `LibraryNames` is asked of every module this compilation has, and reads the same reading `Front.Exposed` does, so a value an attached `examples for` file declares collides with an import of that spelling in one place instead of two.

**A compilation reads its reports once.** `Compilation.reports()` is the one set; `failure`, `errors`, `warnings` and `diagnostics` all project from it, and `failure(List)` is package-private so nothing outside can go around it. A report about a module off the path is said at the nearest place on the way to it that a source of this compilation writes — the `import` line naming it, or the one naming whichever dependency led there — and the coordinate says it stands in for code written in that module. That is a `Citation.OutOfSight`, which every renderer already reads, so the sentence comes for free on the terminal, in the JSON, and in the editor.

Before and after, on a dependency of a dependency missing from the path:

```
-- MODULE ERROR  E1504 -------------------------app2.sou:2:1     ← before

2|
   ^

Unknown module `lib.deep`.
```

```
-- MODULE ERROR  E1504 -------------------------app2.sou:6:1     ← after

6| import lib.held ( Held )
   ^

Unknown module `lib.deep`. The code is written in `lib.held`, which this
compile has no source for, so this is where it was reached from and not
where it is.
```

## Tests

`AnImportedNameSurvivesPublicationTest` — the four cases above, plus: the two spellings denote the same `ValueName.Stdlib` when read back, measured on resolution facts rather than on the tree, since an import line changes how a name is written and nothing else; and the query-API route emits the same classes for either spelling. All six fail on `develop` and pass here.

`AReportAboutAModuleOffThePathIsSaidWhereItWasReachedTest` — where such a report is said, that the coordinate stands in for the module rather than pointing at it, that a module reached only through another inherits the place, and that it reaches the source the caller does hold. On `develop` that source got zero diagnostics for a compile that emitted nothing.

Full reactor green.

## Not in this change

`Exposing` throws on a jar this compiler disagrees with about the standard library, and that throw escapes `FromPath` and takes the compilation with it. Reading a well-formed module back is this change; what to say about one that cannot be read back at all is artifact-boundary robustness — #757.

A second region built over a body spliced in from out of sight is read in the file the body was spliced into — #760. It reaches the standard library's bodies as much as a published module's, so it is about splice provenance rather than about reading a module back, and it cannot be told from a hand-made label by anything a reader of finished reports can see.

### DIFF
--- a/souther-cli/src/main/java/souther/cli/Main.java
+++ b/souther-cli/src/main/java/souther/cli/Main.java
@@ -316,7 +316,7 @@ public final class Main {
                     : Compiler.analyzedModules(texts, path, warnings, measure);
             // Said first, and whatever the command answers with after. What is wrong with the source
             // is the same news whether the rest of this reports, refuses, or succeeds.
-            List<Located> errors = compilation.errors(compilation.db().allReports());
+            List<Located> errors = compilation.errors();
             List<Located> said = new ArrayList<>(errors);
             said.addAll(warnings);
             report(said, sources, render);
--- a/souther-compiler/src/main/java/souther/compiler/Compiler.java
+++ b/souther-compiler/src/main/java/souther/compiler/Compiler.java
@@ -221,19 +221,19 @@ public final class Compiler {
         compilation.measure(measure);
         Db db = compilation.db();
 
-        CompileException structural = compilation.failure(compilation.structuralReports());
+        CompileException structural = compilation.structuralFailure();
         if (structural != null) {
             throw structural;
         }
 
         db.ask(new Output.All());
-        CompileException failed = compilation.failure(db.allReports());
+        CompileException failed = compilation.failure();
         if (failed != null) {
             throw failed;
         }
         for (String module : compilation.modules()) {
             if (!db.ask(new Output.ConstConstructions(module)).present()) {
-                CompileException bad = compilation.failure(db.allReports());
+                CompileException bad = compilation.failure();
                 if (bad != null) {
                     throw bad;
                 }
@@ -265,7 +265,7 @@ public final class Compiler {
         for (String module : compilation.modules()) {
             compilation.answerWarnings(module);
         }
-        warningsOut.addAll(compilation.warnings(db.allReports()));
+        warningsOut.addAll(compilation.warnings());
         return compilation;
     }
 
@@ -309,7 +309,7 @@ public final class Compiler {
         return driven(() -> {
             compilation.measure(measure);
             compilation.answerEverything();
-            warningsOut.addAll(compilation.warnings(compilation.db().allReports()));
+            warningsOut.addAll(compilation.warnings());
             return compilation;
         });
     }
@@ -436,13 +436,13 @@ public final class Compiler {
         compilation.measure(measure);
         Db db = compilation.db();
 
-        CompileException structural = compilation.failure(compilation.structuralReports());
+        CompileException structural = compilation.structuralFailure();
         if (structural != null) {
             throw structural;
         }
 
         db.ask(new Output.All());
-        CompileException failed = compilation.failure(db.allReports());
+        CompileException failed = compilation.failure();
         if (failed != null) {
             throw failed;
         }
@@ -454,7 +454,7 @@ public final class Compiler {
         List<String> exampleSources = new ArrayList<>();
         for (String module : compilation.modules()) {
             if (!db.ask(new Output.ConstConstructions(module)).present()) {
-                CompileException bad = compilation.failure(db.allReports());
+                CompileException bad = compilation.failure();
                 if (bad != null) {
                     throw bad;
                 }
@@ -472,7 +472,7 @@ public final class Compiler {
         for (String module : compilation.modules()) {
             compilation.answerWarnings(module);
         }
-        warningsOut.addAll(compilation.warnings(db.allReports()));
+        warningsOut.addAll(compilation.warnings());
         if (!exampleFailures.isEmpty()) {
             throw CompileException.ofAllInSources(exampleFailures, exampleSources,
                     ExampleVerifier.legacySummary(exampleFailures));
--- a/souther-compiler/src/main/java/souther/compiler/check/Exposing.java
+++ b/souther-compiler/src/main/java/souther/compiler/check/Exposing.java
@@ -7,8 +7,10 @@ import souther.compiler.diag.msg.ImportMessage;
 import souther.compiler.types.ValueName;
 
 import java.util.ArrayList;
+import java.util.Collections;
 import java.util.HashMap;
 import java.util.HashSet;
+import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
@@ -56,7 +58,19 @@ public final class Exposing {
      * {@code let}.
      */
     public record Checked(Ast.Module module, Map<String, ValueName.Stdlib> exposed,
-                          List<Diagnostic> conflicts) {}
+                          List<Diagnostic> conflicts) {
+
+        /**
+         * Copied, because this is an answer a compilation remembers and an answer it remembers is a
+         * value. What decides whether the work that read one has to be done again is whether the new
+         * answer equals the old, so a caller able to reach into a remembered one could change what
+         * every reader of it sees without anything being asked again.
+         */
+        public Checked {
+            exposed = Collections.unmodifiableMap(new LinkedHashMap<>(exposed));
+            conflicts = List.copyOf(conflicts);
+        }
+    }
 
     /**
      * {@code module} read for its library imports: checked, dropped, and what they brought in.
--- a/souther-compiler/src/main/java/souther/compiler/check/Exposing.java
+++ b/souther-compiler/src/main/java/souther/compiler/check/Exposing.java
@@ -18,14 +18,19 @@ import java.util.Set;
  * module reaches the library either qualified ({@code List.map}) or by importing the names it wants
  * — {@code import List ( map, filter )} — after which it may write them bare.
  *
- * <p>What the imports bring in is answered as a table ({@link #read}) and nothing in the
- * module is rewritten. A name written bare stays written bare, and what it means where it is written
- * is settled once, by resolution, with the bindings in force — an import is the last thing consulted
- * and a binding in force wins over it. A declaration of that name does not shadow it: the two are a
- * conflict, refused on the import line, and the declaration standing as the name's meaning after
- * that is what recovery does with a module that will not be emitted. The
- * {@code import List ( ... )} lines are then dropped ({@link #withoutLibraryImports}), having been
- * checked.
+ * <p>What the imports bring in is answered as a table and nothing in the module is rewritten. A name
+ * written bare stays written bare, and what it means where it is written is settled once, by
+ * resolution, with the bindings in force — an import is the last thing consulted and a binding in
+ * force wins over it. A declaration of that name does not shadow it: the two are a conflict, refused
+ * on the import line, and the declaration standing as the name's meaning after that is what recovery
+ * does with a module that will not be emitted. The {@code import List ( ... )} lines are then
+ * dropped, having been checked.
+ *
+ * <p>The table and the module without those lines are one answer ({@link #check}), because a module
+ * that has had them dropped is unreadable without it. There is no way to take the second and leave
+ * the first: a reader of a module off the class path did exactly that, and an invariant that called
+ * a name its module imported bare then resolved against nothing in every project but the one that
+ * wrote it.
  *
  * <p>It mirrors Elm's {@code import List exposing (map)}: the qualified access always works, and the
  * import merely lets a name be written without its qualifier. A name exposed from two libraries at
@@ -36,8 +41,13 @@ public final class Exposing {
     private Exposing() {}
 
     /**
-     * What the imports bring in, the imports that are not the library's, and the import lines that
-     * name something this module already declares.
+     * A module with its {@code import List ( ... )} lines dropped, what those lines brought in, and
+     * the ones that name something the module already declares.
+     *
+     * <p>One value, because the first two are one fact. The module no longer says what its bare
+     * names mean and the table is the only thing that does, so a caller holding the module holds the
+     * table with it — wherever that module travels, and whether it was read from a source or off the
+     * class path.
      *
      * <p>A collision is answered rather than thrown because it is one module's mistake and not a
      * reason to stop reading. Thrown, it escaped the question that asked for this module and took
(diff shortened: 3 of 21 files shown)


kawasima:
All five findings addressed in 0e0fdfc4 — three from review, two from the automated pass.

## FromPath's own report went round the normalisation — fixed

Measured, and it is as described. `AModuleItNeedsIsNotOnThePath` carries `pos=null` and `module=null`, so it fell through `citable()`'s early return and reached no source at all:

```
error=true module=null      code=E1504 pos=null  saidAt=[]     ← only the command line sees it
error=true module=lib.held  code=E1504 pos=2:1   saidAt=[0]
```

`FromPath` holds the places each module it read was reached from, so the report is anchored through the same `Diagnostic.reachedFrom`. Both are `saidAt=[0]` now.

The conservation assertion is in as `noErrorIsLeftWithNowhereToBeSaid`. It measures `publishSourceIdsOf(found).isEmpty()` rather than `classes().isEmpty()` — the point about a compile that legitimately produces no classes is right, and the second reading would say this had been fixed when it had not.

## Nothing can reach the raw reports — held by a tripwire

The contradiction in `answerEverything()`'s javadoc, which still pointed at `Db#allReports()`, is fixed.

I did not make `Db.allReports()` package-private. Every production call site is closed already, but close to twenty test files read the raw set and most of them are about the store itself — memo invalidation, de-duplication — where reading it raw is the right thing. `EverySurfaceReadsTheSameReportsTest` holds the rule instead: anything under `src/main` other than `Db` and `Compilation` calling `.allReports(` fails. Same strength as a re-entry guard, without rewriting twenty tests.

## One `reachedFrom` — made a list

I treated this as a blocker. Measured:

```
two sources importing the same broken dependency
source 0 has 2
source 1 has 0     ← the author of b.sou is told nothing
```

Under `ofDocuments` the winner comes down to map order, so which file gets the marker is effectively unspecified. An author whose import is failing the build looking at a clean file is the defect this change exists to remove, one level down. `reachedFrom` is a `List<SourcePos>` and everything past the first is a `FindingRegion` label, so delivery rides on the machinery `publishSourceIdsOf` already reads through `belongsToFinding()`. Now `source 0 has 2 / source 1 has 2`.

`ModuleMessage.ItIsReachedFromHereToo` is in both catalogues. It is a finding and not a premise: neither import is measured against the other, and either author has the same thing to do.

## Duplicates after relocation — fixed

Reproduced exactly as reported:

```
raw = 2   (E1506 at two coordinates of the published text)
reports() = 1
source 0 has 1
```

`reports()` reads again on `(module, sourceId, identity)`. `Db.allReports()` de-duplicates on an identity that includes the region, and the region is precisely what this reading erases, so the question has to be asked again after it is gone. Regression test: `twoFindingsInOnePublishedModuleAreSaidOnce`.

## The early return skipped the secondary filter — fixed

`Diagnostic.saidOnlyWhereAReaderCanBeSent()` is split out and applied whether or not the caret moves. A spliced body does have a caret a reader can be sent to and can still carry a second region that cannot be quoted.

## `WrittenAt.outOfSight`'s contract

One sentence added. Where what is out of sight is a module rather than something in one, the name it is reached by and the module's name are the same string; the `List.map` / `souther.list` distinction is about value names. The rule itself is unchanged.

## Left alone

The command line prints two E1504 blocks:

```
Module `lib.held` needs `lib.deep`, which is not on the path either.
Unknown module `lib.deep`.
```

`FromPath` and `Names.Imports` state the same fact separately. The overlap predates this change — before it, one had no coordinate and the other had a wrong one; now both have the right one. Which of the two should own it is a question about the messages, and this branch does not touch it.

Full reactor green.

kawasima:
`ef7efa7c`. Findings 1–4 and the tripwire. 1 and 2 are closed by separating the phases rather than by patching, as suggested. On 3 I was wrong, and it is withdrawn.

## 1. Fixpoint — closed

Measured on the diamond as described (`app.a -> short -> meeting -> leaf`, `app.b -> long -> middle -> meeting -> leaf`, with the leaf off the path):

```
0e0fdfc   source 0 has 2   source 1 has 1     ← b.sou is told nothing
ef7efa7c  source 0 has 2   source 1 has 2
```

`FromPath` is three phases now: walk the graph, propagate source origins to a fixpoint, and only then say anything. The worklist condition is whether a module gained a new origin, so a meeting point that gains one late carries it on down. What was there before wrote down the places a module had when it was read — and where the module had none yet, froze an immutable empty list, losing every later route outright.

## 2. Missing belongs to an edge — fixed

`neededBy` is gone. The reports are built from `edges`, one per `(parent, needed)`. `lib.left -> deep` and `lib.right -> deep` are separate findings from the start, and each report's `OutOfSight(parent)` agrees with the origins it is said at.

`twoModulesNeedingTheSameAbsentOneAreTwoFindings` asserts that the declaration a report stands in for and the source it lands on correspond — the `lib.left` report appears on source 0 and nowhere else.

The point that what is absent is the edge and not the node is the design.

## 3. `saidOnlyWhereAReaderCanBeSent` — withdrawn

I had this wrong. `LabeledRegion.sourceIdOr()` is an existing contract — a label naming no source of its own is read in the diagnostic's — and `aLabelWhoseRegionKnowsNoFileIsReadInTheDiagnosticsOwn` pins it. I read that as "cannot be quoted" and dropped such labels from every report. **The existing test builds a `LabeledRegion` directly, so breaking it through the normalisation does not turn anything red.** That is the dangerous shape, and it went out green.

It is back to dropping a label only when the caret moves, and the reason is narrow: moving the caret means the file the label would be read in is no longer the file it was built against. A report whose caret stays put keeps everything it had.

The case the automated pass raised — a citable primary with a secondary read from the published text, which is what a spliced body produces — is **not** fixed. Saying it was in the previous comment was wrong. A null source does not distinguish that from a hand-made label, so it cannot be answered here; it wants provenance where the splice happens. Filed as #760.

Regression test: `anOrdinaryReportKeepsALabelThatNamesNoSourceOfItsOwn`.

## 4. Legacy message — falls out of 3

The diagnostic now changes only when the caret is relocated, so `Report.saidAt` dropping the raised English holds for the reason it was given. There is no longer a path where removing one label discards it.

## Tripwire — held to one call

Whitelisting `Compilation.java` as a file leaves every method in it able to read the raw set. `theOneReadingReadsItOnce` asserts the count of `.allReports(` calls in it is 1.

## The three that were asked for

- `routesOfDifferentLengthsBothReachWhatTheyMeetOver`
- `twoModulesNeedingTheSameAbsentOneAreTwoFindings`
- `anOrdinaryReportKeepsALabelThatNamesNoSourceOfItsOwn`

Full reactor green.

kawasima:
Both findings measured. Neither holds — no change for either, and here is the evidence.

## `reports()` de-duplicating without a position

`Repeat` carries `Diagnostic.Identity`, and `Identity` has `region` as a component:

```java
public record Identity(Severity severity, String code, String titleKey, Region region,
                       List<LabeledRegion> secondary, String literalMessage,
                       TypeComparison diff, List<Note> notes, String suggestion, Message said) {}
```

It is the same key `Db.allReports()` already uses — `Told(module, sourceId, problem)` — asked a second time, which is the whole of what the change does. Two findings of one problem at two lines have two regions, two identities, and are not merged.

Two errors of one message at two lines of one source:

```
raw = 2
reports = 2
  E1023 at 3:15
  E1023 at 4:15
```

The only thing this merges is what relocation made identical, which is the case it was written for.

## `Ast.Var` needing the same guard as `Ast.TypeRef`

`Ast.Var`'s constructor reads the name it is handed, on its first statement:

```java
public Var {
    if (written.region() != null && region == null) {
```

so there is no such thing as a `Var` without one:

```
new Ast.Var(null, null) throws NPE at construction
```

`TypeRef` is different, and that is why it is asked: `TypeRef.written(String name, …)` returns `name == null ? null : WrittenName.synthetic(name, pos)`, so a reference a pass minted before resolution can genuinely name nothing. The two are asked differently because the two answer differently.

It is also unchanged behaviour — what stood here before read `ref.name()`, which is `written.canonical()`, the same dereference.

## What did change

Two readers in a row have stopped at that asymmetry, and the comment asserted the invariant without saying what enforces it. It now names the constructor (617b6ad2). No code change.

kawasima:
`9d241154`. Finding 1 fixed, 3 and 4 done, 2 left as a decision rather than taken unilaterally.

## 1. `reservedNamespace` — fixed

Reachable, and measured through the route suggested:

```
before   code=E1502 module=null pos=1:1 src=null saidAt=[]
after    code=E1502 module=null pos=2:1 src=0    saidAt=[0]
         source 0 has 1  ->  source 0 has 2
```

It is held as `refused` during the walk and said after the fixpoint, exactly as the missing edges are. That was the whole of what it took — the phase separation had the shape, this producer was simply not on it.

One thing that fell out: a refused module is on the path, so nothing else may call it missing. The missing-edge loop skips it, or an author would be told both that a dependency took a reserved name and that it is not there.

Regression test: `aModuleOnThePathThatTookAReservedNameIsRefusedWhereItWasReached`.

## 3. The edge test was weaker than its claim — fixed

Correct: `at.sourceId()` is the primary's, so the shape it was meant to catch — a labelled region claiming the other author's import — passes it. It reads `publishSourceIdsOf(found)` now and asserts the whole list, and it collects the declarations it saw:

```java
assertEquals(List.of("lib.left".equals(stands) ? "0" : "1"),
        compilation.publishSourceIdsOf(found));
...
assertEquals(Set.of("lib.left", "lib.right"), saidAbout);
```

The second is the one that matters: an edge report dropped outright used to leave the loop passing with nothing in it.

## 4. Defensive copies — done

`Exposing.Checked` and `FromPath.OnThePath` copy what they are given. `Ordered.map`/`Ordered.set` for the two that are order-bearing, `List.copyOf` for the places. The reasoning is the one `Db` states: whether the work that read an answer is done again is decided by comparing the new answer with the old, so a caller able to reach into a remembered one changes what every reader sees with nothing being asked again.

## 2. `Fixes #755` — taken out, and the choice left where it belongs

Right, and I would rather not settle it by editing acceptance criteria myself. `Fixes` is off the PR body, which now says which of the two it is waiting on:

> Either #755 waits for #760, or its acceptance 4 is narrowed to the primary coordinate on purpose.

For what it is worth I lean the same way you do. #760 reaches the standard library's spliced bodies as much as a published module's, so it is about splice provenance rather than about reading a module back — the two share a sentence and not a mechanism. But narrowing an acceptance criterion is the issue author's call.

(Merging into `develop` would not auto-close in any case; this repository closes its issues by hand after the merge.)

Full reactor green.

kawasima:
`b92a917b`. Confirmed and fixed.

Measured before the fix, on the reserved-module route:

```
E1502 :: module.the-module-is-in-the-reserved-namespace
E1504 :: module.unknown-module
```

Two reports that cannot both be true, and the reasoning is exactly as described: `refused` was held apart inside the walk so that nothing would call the module missing, and then left out of what came back out, so `ModuleNames` did not have the name and `Names.Imports` fell through to "unknown module".

`FromPath.Of` carries the refused names beside the modules, and `ModuleNames` reads both. Knowing a name and being able to read what it declares are two questions, and only the second is what `Available` answers — which is the line `Names.Imports` was already drawing. After:

```
E1502 :: module.the-module-is-in-the-reserved-namespace
```

The test says `assertEquals(List.of("E1502"), …)` rather than asking whether the refusal is among what the author was told. You are right that the weaker form would have passed on the broken behaviour — it did.

Full reactor green.

PR #758, opened 2026-08-15
Bound a product and a truncating quotient by what the path knows of their parts

### LINKED ISSUE #750: A product of two values the path bounds is read as a value nothing is known of

kawasima:
Closes #750.

A quantity times a unit price is not negative where both of them are, and the invariant-discharge check said nothing about it. `Terms.scale` answers a linear form only where one side is a constant, so a product of two values became an atom of its own and nothing related that atom to the factors the path has bounds for. The same held of `x / k`. Both guards say exactly what the construction needs, and there is no third guard about the product that an author could write instead.

## What this is not

Not a row in the table of what an operation guarantees (#749). Nothing holds of a product wherever it is written: what holds is decided by what the path establishes about the two factors at the point the clause is read. So it is asked of the domain at the construction, and a construction under guards that say nothing about the factors is reported exactly as before.

## The shape

Three parts, each of which can be held to an answer on its own.

**`numeric/Intervals`** — ranges in, a range out, and nothing else. A product is the least and greatest of the four corner products; an end it reaches is one *some* corner reaches, asked of every corner landing there, since `[-1, 1] * [-1, 1]` is at 1 by two of them and a factor being open at one says nothing about the other. Where an end is at zero, the corners are not the whole of it: a product is constant along the axis where a factor is zero, so zero is a value it takes wherever *either* factor holds zero at all — which a factor may do between its ends, or at an end the other factor's ends do not pair with. Zero is not special in what the ends **are**; it is special in whether an end at zero is **reached**. An unbounded end carries which side it is on, because a corner where two of them meet has an answer.

**`Terms.derivations`** — what an atom outside the affine fragment was computed from, recorded where the atom is named, beside how its values are spaced and for the same reason. A recipe and not a bound: what the value lies between depends on what the path assumed, and the walk that reads the operands is not handed a path. Recipe edges are ordinary atom identities, so `a * b / 100` needs no second identity system. One atom recorded two ways is an internal error, as one atom given two spacings is.

**`DerivedBounds.refine`** — given a domain, read each recipe's operands out of it, put the arithmetic through `Intervals`, hand the domain back with what came out taken as holding. Every bound is derived once from the domain it was given and the results are taken in together; nothing derived is read back to derive something further about the same atom. A bound that tightened under its own answers would be a procedure that decides when to stop, which this is not.

Both readings of a construction are refined — under what the guards established, and under what holds of the values whatever the guards did. Read into the first alone, a product its factors' own types refuse comes out rejected "on a reachable path", which sends an author to guard a path that is not what is wrong.

## What is deliberately not read

`/` on `Decimal` rounds half away from zero at a precision the run time sets rather than truncating (`DecimalMath`), and where an end lands under that rounding is not something this states — the compiler models the operator's semantics nowhere else either (`ConstEval` does not fold `DIV` at all). A divisor the guards only bound away from zero is not read: the rule is about the sign and the magnitude of a number that is written, and what a type guarantees of an argument is what the caller has, not what the operation answers. Multiplication is read for both `Int` and `Decimal`: it is exact in either, and an `Int` that overflows aborts rather than wrapping.

A derived bound refuses as readily as it discharges, so a construction that warned may now be an error.

## Measured

- `CI=1 mvn -Plint clean verify` BUILD SUCCESS — compiler 4656, fmt 2224, cli 333, lsp 204. Each of the first three commits is green on its own (4605 / 4624 / 4643).
- souther-examples, 15 projects, against a jar built from `develop`: **71 E2011 to 56**, none added, no new error, and **every project's generated classes byte-identical**. The 15 are the sites #750 lists: `crm.sou:255`, `joboffer.sou:169,170,179`, `inventory.sou:247`, `yearendadjustment.sou:165-167,272-278`. Unchanged by the review fixes below — the zero fix only ever widens attainment, so it can turn a false E2010 into a warning and never the other way, and no corpus site was resting on it.
- Two of the 17 sites the issue names still warn, each for a stated reason. `inventory.sou:260` divides by a `PackSize` — a value its type bounds, not a written number. `yearendadjustment.sou:263` divides `Int.max(0, …)`, and what `Int.max` answers is modelled as cases rather than as a bound on the atom (#749), so the domain has no range on the numerator to read.
- Seven mutants killed by the tests that are meant to: reading only the first corner that lands on an end; not exchanging the ends for a negative divisor; extending the quotient rule to `Decimal`; refining only the guarded reading; not asking the ranges whether a zero end is reached; deciding the fail-open boundary by a list of failures rather than by their kind; and removing the derivations entirely, which is what holds the guard-equates-a-product witness.

## Two defects review found, and what they moved

**Zero attainment.** `Intervals.product([0, 1], (0, 1))` came back `(0, 1)`. A product is constant along the axis where a factor is zero, so zero is a value it takes wherever either factor holds zero at all — which a factor may do between its ends, or at an end the other factor's ends do not pair with. Read off the corners alone it was bounded strictly away from a value it reaches, and that is not a lost bound: `NonPos(a * b)` for `a in [0, 1]` and `b in (0, 1)` was **E2010**, and with `a` pinned to zero both ends came back open at zero, which is a range holding nothing — the domain took in a contradiction and discharged a construction whose invariant the value is nowhere near. The corners give where the ends are; an end at zero asks the ranges themselves whether the product reaches it.

**The fail-open boundary.** `gaveUp` told a limit of the analysis from the analysis contradicting itself by naming the three refusals it knew of, which is a copy of the distinction the types already carry. Neither of the two this branch adds was on it. `TheCheckDisagreesWithItself` is what the boundary asks about now, and a check that gains a new way to contradict itself is on the right side of it by being one.

The spec sentence saying a derived bound is not read into the values-alone reading was wrong in the same direction and is fixed: each reading derives its own, so a bound resting on a guard is in the guarded reading only and one following from the values' own guarantees is in both.

## Note on the first commit

`Terms.affineOf` took a `Known` and never read it — every use passed it down its own recursion and nothing asked it anything. Dropping the parameter is what makes "the recipe does not depend on the path" a thing the compiler checks rather than a thing the comment says. No behavior changes.


kawasima:
Both correctness issues are fixed, in `847379a5` and `6c860275`. Reproduced each before changing anything, and each has a witness that fails on the code as it was.

**1. Zero attainment (blocker).** Confirmed exactly as described. `Intervals.product([0,1], (0,1))` returned `(0, 1)`, and the two consequences both reproduce at the compile level: `NonPos(a * b)` for `a in [0,1]`, `b in (0,1)` was **E2010**, and with `a` pinned to zero the derived range held nothing, the domain went bottom, and a construction whose invariant the value is nowhere near came out **discharged with no diagnostic at all**.

Fixed on your model: the corners give where the ends are, and an end at zero asks the ranges themselves whether the product reaches it — `one.admits(Count.ZERO) && other holds a value`, either way round. Not a zero special case bolted onto the corner rule, but the reachability that follows from the product being constant along the axis where a factor is zero.

The javadoc and the PR body claimed "Zero is not a case of its own". Withdrawn in both, and the distinction you drew is what replaced it: not special in what the ends *are*, special in whether an end at zero is *reached*. The spec section says it too now.

Witnesses: `zeroIsReachedWhereAFactorReachesItEvenWhereTheOtherFactorReachesNoEnd` and `aFactorThatIsOnlyZeroAgainstAnOpenFactorIsZero` at the unit level, `aProductWithAFactorThatCanBeZeroIsNotRefusedForBeingAboveZero` and `aProductPinnedToZeroIsReadAsZeroAndNotAsNoValueAtAll` at the compile level. With the zero rule mutated away, the first fails with the false E2010 and the second with "nothing was thrown".

**2. The fail-open boundary (major).** Right, and it was worse than an omission — `gaveUp` decided by naming the three refusals it knew of, which is a copy of the distinction the types already carry. Took the structural fix: `TheCheckDisagreesWithItself`, which `OneTermTwoKinds`, `NotOnePlace`, `NotOneClause`, `OneTermTwoDerivations` and the cycle check all are, and which the boundary asks about by kind. A check that gains a new way to contradict itself is on the right side of it by being one.

The cycle check was a bare `IllegalStateException` saying in a comment what its type did not; it is `DerivedBounds.AnAtomComputedFromItself` now. `WhatTheCheckWillNotGiveUpOnIsAskedByItsKindTest` holds the boundary directly — with the list restored, the two new ones are silently given up on.

**3. Quadratic `refine` (non-blocking).** Left out, with one correction and one caveat.

`Terms` is per `InvariantChecker`, and `InvariantChecker.analyze` is called per behavior body (`SpecChecker:415`) — so the map accumulates **per behavior**, not per module. Your framing is the accurate one.

The caveat is that seeding from the derived atoms in `owed`'s numeric forms is not behavior-preserving. A guard can relate a construction's term to a product it does not name — `guard total == a * b` with a clause about `total` — and `boundsOf(total)` reaches the product only through `closedDiff`, so a product no clause mentions still has to be derived. The root set that closes it without losing that is the clause atoms **plus** every atom the domain relates at all, which needs `NumericDomain` to say which atoms it mentions. Happy to file that as its own issue with those acceptance criteria.

**Measured after both fixes:** `CI=1 mvn -Plint clean verify` BUILD SUCCESS, compiler 4655. souther-examples unchanged from the first measurement — 56 E2011, the same positions, classes byte-identical against the `develop` jar. The zero fix only ever widens attainment, so it can turn a false E2010 into a warning and never the other way; that no corpus site moved is what it should be, and the case sat in a hole in the test matrix — the zero rule was covered against an unbounded end, where `End.times` answers it, and never against a bounded factor reaching neither of its ends.

kawasima:
Two more from the low pass, both real, both fixed — `1d19ef40` and `12273a39`. Measured each before changing anything.

**A divisor no `Int` holds threw, and the throw was swallowed.** `x / (9223372036854775807 + 1)`: the walk composes a written divisor's arithmetic over `BigDecimal`, so the constant reaches a number no `Int` is, and `longValueExact` threw. An `ArithmeticException` is not `TheCheckDisagreesWithItself`, so `gaveUp` recorded it as a limit and the construction was **reported as nothing at all** — where the same construction over a divisor of `100` is an E2010. Confirmed by running both through the CLI: `100` gives the error, `MAX + 1` gives silence.

Not a crash, which is what makes it worth having a witness: the fail-open path did exactly what it is for, and what was wrong was the rule throwing where it should decline. A number outside what an `Int` holds is now a divisor there is no rule about, so the clause is owed as any value nothing is known of leaves it owed. `aDivisorNoIntCanHoldIsNotOneTheRuleReads` expects E2011 and fails with `[]` on the code as it was. The spec's divisor condition says it too now.

**The `Known` drop was unfinished.** Four callers in `Predicates` — `constantOf`, `shiftFact`, `piecewiseOf`, `resultFacts` — kept taking a `Known` only to pass it to something that had stopped reading it. That is the same dead weight `161b4162` removed, one call away, and leaving it there makes that commit's claim less true than it says. Dropped. `noCaseSatisfies` keeps its own; it reads `k.numbers()`.

On the sub-threshold note about `serialVersionUID`: `OneTermTwoDerivations` matches `OneTermTwoKinds`, which is the sibling in the same file and has never declared one. Adding it to only the new one would make the file inconsistent in the other direction, and `-Plint` is quiet either way, so I left it.

**Measured after both:** `CI=1 mvn -Plint clean verify` BUILD SUCCESS, compiler 4657. souther-examples unchanged again — 56 E2011, the same positions, classes byte-identical against the `develop` jar.

PR #768, opened 2026-08-16
Ask the compile what may be written here, and keep what it said

### LINKED ISSUE #765: A name that arrived from an import is offered only while the file already parses
`Analyzer.completions` (`Analyzer.java:1135`) builds its candidates from the document alone: the
keywords, the top-level definitions of this file, the names its import lines spell out, and the
params and `let` bindings of the definition the cursor sits in. Every other question the server
answers — hover, definition, references, code actions, rename — is asked of the workspace compile
and takes `(uri, position, graph)`. Completion takes a text and a position, so it has no module to
ask about.

Two of those four sources are the compiler's to answer. What a bare name reaches here is settled by
`Names.Imports` (`query/Names.java:360`) and `Exposing.check`, and reading the import lines a second
time in the LSP is that rule written twice, to go out of agreement the first time either moves.
`Names.NameScope(module)` gives `Symbols.visible()`, "the definitions reachable here by a bare name:
this module's own plus the imported ones" (`check/Symbols.java:89`), and each one arrives as a
declaration of something, so an item's kind is read instead of assumed.
`Front.LibraryNames(module)` gives the standard library names the module's import lines exposed, as
`Exposing.check` read them rather than as a second reader of the same lines.

## The import lines are not the scope

An `examples for` file writes no import line at all, and the module's declarations are in scope bare
there. `crm.examples.sou` says so itself: "No import is needed: the same module's data and behaviors
are in scope bare." Completion in that file offers the keywords and whatever the file itself spells,
and nothing of the model its rows are about — which is the place an author writes the model's names
most densely, and the place the document's own text says least about what is in scope.

The rest follows from reading the lines rather than asking what they did. A name an import line
spells that the named module does not expose is offered, though nothing here denotes it.
`import up as U` contributes nothing. Every item's kind is `FUNCTION`, so a data type is offered as
a function.

## The import source is worse than it reads

It goes through `CstFrontend.parse` (`Analyzer.java:1150`), which throws on the first CST error
anywhere in the file (`CstFrontend.java:38`), so it contributes nothing whenever the document does
not parse — which is most of the time completion is asked for. Over the 16 sources of
`souther-examples`'s crm, account and hr, all of which parse as written, inserting a two-character
line leaves 116 of 6660 edits parsing; of the 6544 broken edits, the number whose import lines
`CstFrontend` could still read is 0. (The insertion is made after every line, including places no
expression is ever written, so the 116 understates how often a real edit parses. The 0 does not: it
follows from where the throw is.)

## Answering while the file will not parse

A file that will not parse is held out of the compile and its module named in `Front.Broken`
(`Analyzer.java:283-296`), so both queries go absent for exactly the file being typed in:
`LibraryNames` falls to `Exposing.Checked` and then to the module path (`Front.java:559`), and the
edited module is neither.

The import lines themselves survive — the error-tolerant CST kept the `IMPORT_DECL` nodes identical
to the intact file in 4298 of those 5211 broken edits that had any. Reading them back here is not
the answer, for the reason above: what a line brings in is `Exposing.check`'s to say.

So the names that came from elsewhere are what the last compile that answered said they were, kept
per document and offered while the current one cannot answer. What the document itself declares is
read off its CST every time, so a definition being written now is offered before it compiles.

Per document and not per module, because a module is not one file. Seen from `crm.examples.sou`, the
declarations of `crm.sou` are elsewhere — this document does not declare them and cannot re-read
them — and a partition drawn at the module would drop exactly those while `crm.sou` is the file
being edited.

What goes stale is any change to what reaches the document from elsewhere, made while the compiler
cannot answer for its module. The usual one is an import line added to a document that does not
parse.

## Acceptance

1. Completion is asked of the workspace compile: which bare spelling reaches what is the compiler's
   answer, not a reading of the document's import lines.
2. Candidates are offered while the document does not parse, and the number of them that came from
   elsewhere is not zero — which is what it is now.
3. What the document being completed declares is read from that document rather than from the
   remembered scope, so a definition being typed is offered before it compiles. A sibling document
   of the same module is elsewhere, not own.
4. An item carries enough to choose between two candidates with the same name — at least `detail` —
   and `LspServer` sends it rather than dropping it at `LspServer.java:317`.

kawasima:
Closes #765.

Completion was the one question the language server answered from the
document alone. Everything else — hover, definition, references, code
actions, rename — takes a position and the workspace; this took a text, so
it had no module to ask about and read the import lines itself to work out
what they brought in. That is the scope assembly's rule written a second
time, and it was also read through a parse that throws on the first syntax
error anywhere in the file, so it contributed nothing whenever the
document did not parse.

## What #767 had already settled

Most of the compiler side of this was in place. `Resolve.Values` is
already split into `Reachable` — a value — and `Elsewhere`, the lookup a
pass makes while it reads, and `Names.ModuleScope` already answers a
module's assembled scope with `Scoped.reachable()` on it, whose javadoc
says it is what a reader wanting to know what may be written here asks
for. So this adds no query. The server asks `Names.NameScope` for the type
namespace, because kinds are read off declarations, and
`Names.ModuleScope(...).reachable()` for the value namespace.

Two things were left, and they are the first two commits.

Resolving a bare value name read the three tables in turn, so the order
they are consulted in lived in the reader. An enumeration would have to
state it again, and two statements of one rule agree until either moves —
so it is written once, on the table, and the reader asks it.

The type scope handed over the declarations a bare name reaches and
dropped the spellings each is reached by. The value namespace already
answers the pair; this is the other half of the same question.

## Answering while the document will not parse

The names that came from elsewhere are what the last compile that could
answer said they were, kept and offered while the current one cannot. It
is an availability fallback and not a second source of names: written only
where the compiler answered, read only where it did not, never merged. So
deleting an import from a document that parses takes its names away at
once.

Kept per document rather than per module, because a module is not one
file. Seen from an attached `examples for` file, the declarations of its
module's own source are from elsewhere — that document does not declare
them and cannot read them off its own text — and a partition drawn at the
module would drop exactly those while the other file is the one being
edited. That file is also where reading import lines was least of an
answer: it writes none, and its module's data and behaviors are in scope
bare there.

What the document declares is read off its tree every time, so a
definition being written is offered before it compiles. Those spellings
are kept out of what is remembered rather than filtered on the way out,
which is what stops a name lingering under the one it was renamed from.

An item now carries where it came from and the server sends it, a kind is
read off the declaration rather than assumed, and the bindings in force at
the cursor are offered ahead of an import that spells the same name.

## Measured

Over the sixteen sources of the crm, account and hr examples, inserting a
two-character line after every line gives 7603 edits, none of which parse.
Before, the number of candidates from anywhere but the document itself was
zero on every one of them; now it is non-zero on every one, and constant
per file — what the compiler last said does not move while the document
does not parse. `account.sou`, which writes no import line at all, offers
one: `NoAccount`, an implicit unit data synthesized from
`-> Balance | NoAccount` that no `data` line spells. `org.sou` offers
four, which is exactly `UserId` plus `fold, length, isEmpty`.

`CI=1 mvn -Plint clean verify` is green (syntax 98, fmt 2224, compiler
4713, lsp 215, cli 333), and each of the first two commits is green on its
own. The `examples --format json` output of five example projects is
byte-identical to develop's.

Six mutations were run to check the tests fail for the reasons they state.
Never reading the remembered answer kills five; remembering what the
document declares kills the rename test; never forgetting a dropped
document kills the eviction test; offering bindings last kills the
shadowing test; assuming a kind kills the kind test; leaving the library
names out of the one table kills the two compiler tests. Nothing else
moves in any of them.

## A correction to the issue

The issue argued from a line written `exposing (..)` contributing the
declarations it brings in. There is no such form: an import's name list is
an explicit list of identifiers, and so is a module header's `exposing`
clause. The issue has been rewritten to argue from the attached
`examples for` file instead, which shows the same thing in the language as
it stands — the set of imports written in a document is not the set of
names it reaches.

### DIFF
--- a/souther-compiler/src/main/java/souther/compiler/check/Resolve.java
+++ b/souther-compiler/src/main/java/souther/compiler/check/Resolve.java
@@ -54,6 +54,9 @@ public final class Resolve {
 
     private final SyntaxSymbols symbols;
     private final Reachable reachable;
+    /** {@link Reachable#byName()}, worked out once: a bare name is looked up for every name this
+     * pass reads, and the table does not change while it reads. */
+    private final Map<String, ValueName> reaches;
     private final Elsewhere elsewhere;
     /** Every name this pass answered, in the order it met them. */
     private final List<TypeUse> denotations = new ArrayList<>();
@@ -105,6 +108,7 @@ public final class Resolve {
     private Resolve(SyntaxSymbols symbols, Values values) {
         this.symbols = symbols;
         this.reachable = values.reachable();
+        this.reaches = reachable.byName();
         this.elsewhere = values.elsewhere();
     }
 
@@ -176,6 +180,27 @@ public final class Resolve {
             helpers = Collections.unmodifiableMap(new LinkedHashMap<>(helpers));
             behaviors = Collections.unmodifiableMap(new LinkedHashMap<>(behaviors));
             exposed = Collections.unmodifiableMap(new LinkedHashMap<>(exposed));
+        }
+
+        /**
+         * Every bare spelling this reaches something by, and what writing it would mean.
+         *
+         * <p>The order the three tables are consulted in is a rule, and it is written once — here.
+         * An import brings a library name in, and everything the module already has is what that
+         * spelling means instead. Resolving a bare name reads this, so a reader listing what may be
+         * written here reads the answer resolution reads rather than a second assembly of the same
+         * three tables: two of those would agree until one of them moved, and nothing would say
+         * which had.
+         *
+         * <p>A binding in force and a type written as a value are not here. Both are answered before
+         * this table is consulted — the first from the bindings that hold at the position, the
+         * second from the type scope — so neither is a fact about the module.
+         */
+        public Map<String, ValueName> byName() {
+            Map<String, ValueName> reached = new LinkedHashMap<>(helpers);
+            behaviors.forEach(reached::putIfAbsent);
+            exposed.forEach(reached::putIfAbsent);
+            return Collections.unmodifiableMap(reached);
         }
 
         /**
@@ -1260,20 +1285,10 @@ public final class Resolve {
             return new ValueName.OfType(written, d.type(),
                     applied ? null : ConstructionOrigin.own());
         }
-        // a helper or a behavior, applied or handed over by name — which the inliner expands into a
-        // block that applies it
-        ValueName.Helper helper = reachable.helpers().get(written);
-        if (helper != null) {
-            return helper;
-        }
-        ValueName.Behavior behavior = reachable.behaviors().get(written);
-        if (behavior != null) {
-            return behavior;
-        }
-        // A name an import let this module write without its qualifier. Asked last: an import brings
-        // a name in, and everything the module already has — a binding in force, its own
-        // declarations — is what that name means here instead.
-        return reachable.exposed().get(written);
+        // A helper or a value of this module, a behavior it reaches, or a name an import let it
+        // write without a qualifier — asked of the one table that says which, so that a reader
+        // listing what may be written here reads the same answer this does.
+        return reaches.get(written);
     }
 
     /**
--- a/souther-compiler/src/main/java/souther/compiler/check/Symbols.java
+++ b/souther-compiler/src/main/java/souther/compiler/check/Symbols.java
@@ -3,20 +3,19 @@ package souther.compiler.check;
 import souther.compiler.ast.Hir;
 import souther.compiler.types.Denotation;
 import souther.compiler.types.TypeKey;
-import souther.compiler.types.TypeSymbol;
 
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.HashMap;
-import java.util.List;
+import java.util.LinkedHashMap;
 import java.util.Map;
 
 /**
  * A {@link Scope} and a {@link Declarations} together, for the readers that have both to hand.
  *
  * <p>Not a thing of its own. It answers nothing itself except the one question that genuinely needs
- * both — {@link #visible} takes the identities a bare name reaches here and asks what each is — and
- * hands over its two parts otherwise. The two answer different questions and fail in different ways:
+ * both — {@link #reachable} takes the identities a bare name reaches here and asks what each is —
+ * and hands over its two parts otherwise. The two answer different questions and fail in different ways:
  * a spelling nothing here writes is not a declaration that did not come out, and while one object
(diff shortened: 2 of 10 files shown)


kawasima:
Three findings, all real. Fixed in 9b0c5cc4, plus one the review did not raise.

**Bindings offered where they are not in force.** Correct, and it is my
change that made it matter. `collectLocalBindings` took every param and
`let` bound anywhere in the definition, with no idea where the cursor was.
That was already imprecise; putting the bindings first made it wrong, and
worse than the example suggests — the displaced candidate loses its origin
too, so the author sees a bare `price` with no detail and cannot tell it
from a real local. The collector now walks into a construct that confines
what it binds only when the cursor is inside it, and skips a binding
written below the cursor. A pattern is not such a construct, so an arm's
own bindings still hold over its body.

Held by two tests in both directions, and the position-blind collector
fails the first of them with exactly the reported symptom:

```
expected: <CompletionItem[label=price, kind=3, detail=up]>
 but was: <CompletionItem[label=price, kind=6, detail=null]>
```

**The whole-workspace re-parse.** Measured before deciding. On the crm
example — seven sources, 3898 lines — one completion took 9.3ms, of which
7.3ms (78%) was that loop; on hr, 6.0 of 7.6ms. It grows with the
workspace rather than with the edit, so it is worth fixing rather than
noting. The sort is now kept with the text it was read from: 9.3ms → 1.1ms
on crm, 7.6 → 0.88 on hr. Every other request that arrives with the
workspace pays the same cost today and stops paying it too.

The test for that took two attempts, which is worth recording. The first
held only the recovery direction and stayed green when the reading was
pinned to whatever it first said — pinned clean, a document that stops
parsing is handed to the compile as a clean source, fails there for its own
reasons, and looks identical from here. Pinned broken is the observable
direction, and reaching it needs a document that will not parse before it
does.

**`Symbols.visible()`.** Deleted. Zero callers, and you are right that it
had quietly stopped answering what its name says: keyed on the spelling, a
declaration reachable under two of them would come back twice.

**One more, from a separate reading.** Eviction of the remembered names ran
as a side effect of asking for completion, which is not where a workspace
losing a file is observed. A URI names a document and can be used again, so
a URI removed and written again with different content inherited the old
document's names — reproduced as `[Amount, Colour, price, length]` leaking
into an unrelated document. It is now dropped where this analyzer is told
what the workspace holds, which every request arrives through and which a
file created or deleted on disk reaches as a diagnose.

`CI=1 mvn -Plint clean verify` is green (compiler 4713, lsp 219, cli 333).
Eight mutations now, each killing exactly the tests that state the rule it
breaks and nothing else.

Issue #769, opened 2026-08-16
A diagnostic whose primary region is no place is anchored to the file the report was filed under


kawasima:
Found while doing #762, and left out of it: that issue is about a source identity put beside a
position that already answers for one. This is the other half of what the measurement turned up, and
it is a different question — here the position does not answer, and a place is assembled for it out
of something else.

## What it measures

Every diagnostic built over one compile of the suite, with its primary region classified the way
`DiagnosticPlace.of` classifies a secondary's:

| | |
|---|---|
| `InSource` — the region names a source | 2972 |
| `Unavailable` — code out of sight | 481 |
| no region at all | 22 |
| **refused: the region names no source and says the code is written at it** | **53** |

The 53 come from fourteen codes: E1023 (14), E1004 (8), E1802 (5), E1402 (4), E1506 (3), E1301 (3),
E1014 (3), E1008 (3), E2001 (2), E1803 (2), E1317 (2), E2301 (1), E1818 (1), E1022 (1), E1017 (1).

## Why it matters

A secondary is built through `DiagnosticPlace`, which refuses that shape outright:

> a region naming no source is a place nobody settled, and a report may not settle it from where it
> happens to be shown

A primary is not held to it. `Spot.primary(d, sourceId)` takes the source from where the report was
filed — `Db.Found.claimedSourceId`, which falls back to the key that found it — and pairs it with
the region's line and column. So a reader is sent to a file, and quoted a line, that the numbers are
not of. That is the reading `DiagnosticPlace` exists to stop, reached through the one region that
does not go through it.

## What this is not

Not a constructor to fix. Fourteen codes is a spread across the model, and the 53 are not obviously
one thing: a diagnostic may have no place to point at, or may have lost a provenance it once had, or
may be about a node nothing placed, or may be carrying a region built wrongly. Dropping all of them
into a "no anchor" case would name a category before reading what is in it.

## Acceptance

1. Every diagnostic either has a primary place, explicitly has none, or says why its primary cannot
   be located — and which of the three it is, is in the value rather than worked out downstream.
2. Nothing assembles a place out of a region that names no source and a source that came from
   somewhere else.
3. `Spot` no longer holds a source beside a region, or the component says which question it answers.
   `NoValueHoldsASourceBesideOneThatAnswersForItTest` names `Spot` as the one open case and is where
   that is closed.

kawasima:
Closed by #794 (merged into develop as 5ae50bb9).

Issue #757, opened 2026-08-15
A jar this compiler disagrees with about the standard library takes the compilation with it


kawasima:
Split out of #755, where the two were found together and only the first is fixed.

`Front.FromPath` reads each module off the class path and asks `Exposing.check` what its import lines bring in. That check raises rather than reports for two of the three things an import can get wrong:

- `NameIsNotAStandardLibraryFunction` — the line names something this compiler's library does not publish
- `NameIsPublishedByTwoModules` — one bare name arrives from two modules at once

Neither is reachable through a jar built by a compiler that agrees with this one: the declaring project would have refused the line. Both are reachable when it does not — which is the case `BOUNDARY_VERSION` exists for, and the case where the compiler should be at its clearest.

`Db` does not catch `CompileException`, so a throw here leaves `FromPath.compute` and ends the compilation. What the author sees is a message about an `import` line, at a coordinate in a file they do not have, for a module they did not write. Every other file's diagnostics go with it.

## What this is about

Not "the module has a defect in it" — #755's shape — but "this artifact cannot be read back at all". The answer is the same machinery #755 introduced: a report said where the module was reached from, naming the module the way `TheModuleWasCompiledByAnotherSouther` already does, rather than a throw carrying a coordinate nobody holds.

Worth being narrow about which failures convert. A known artifact-validation failure becomes a report; an internal bug does not, or a compiler defect arrives at the author as a diagnostic about their dependency.

## Acceptance

1. A module on the class path whose import lines this compiler cannot read is reported, not raised, and the rest of the compilation still answers.
2. The report is said somewhere this compilation can quote, and names the module the code is in.
3. What converts is stated: an artifact-validation failure this compiler knows the shape of, and not any `CompileException` reaching that point.

kawasima:
Closed by #777.

Reading a module off the class path could fail seven ways and every one of them raised, out of a walk over the whole compilation, from inside a text this compiler assembles and nobody holds. The two the issue names are among them; `TheModuleWasCompiledByAnotherSouther`, cited here as the shape to follow, was another.

[...]

PR #787, opened 2026-08-16
Split which text a position is in from whose code it carries, and settle where a report may point from it

### LINKED ISSUE #785: An out-of-sight citation says its place is one the reader holds, and nothing makes it so
Found while doing #757, and left out of it because closing it means splitting a type that #762 had just finished settling.

`Citation.OutOfSight` says what a report may claim about code this compile has no file for:

```java
sealed interface OutOfSight extends Citation permits OutOfSightCitation {
    SourceProvenance provenance();

    /** The call the body was spliced into. A place in a file the reader holds. */
    SourcePos reachedFrom();
}
```

The second sentence is not enforced. A coordinate a parser made in a published module is out of sight from the moment it exists — `SourcePos.standingInFor` says so itself:

> Not where a stand-in is first decided. Whether code is out of sight is settled where a text becomes positions, by the caller that knows what the text was, and a coordinate a parser made says so from the start — a text put back together out of what a module published is read by a parser like any other, and line 4 of it is a line of nothing anybody holds.

So `Citation.of` answers `OutOfSight` for that coordinate, and `reachedFrom()` hands back the coordinate itself: line 4 of a text nobody holds, under a contract that says it is a place the reader holds. It becomes true only after `Diagnostic.reachedFrom` has moved the caret to an import line, and nothing distinguishes the two states.

## What it costs

Nothing today, on the routes that were measured. #757 anchors every report about a module off the class path before it is reported, and pins the caret to the importing line. What is left is a type whose accessor promises something its arms do not all satisfy, at the layer every surface reads a place from.

That was enough on its own once. `Citation` exists because "a coordinate is put to two uses and only one of them can be wrong", and the defect it was written for was three renderers out of four publishing a coordinate as the place code is written. This is the same reading, one level in: a surface that trusts `reachedFrom()` to be quotable is right about every report that went through the anchoring and wrong about any that did not.

## What it would take

`isOutOfSight()` is true in two states — written out of sight and this coordinate is inside that text, and written out of sight and this coordinate is where this compile met it. Telling them apart is a change to `WrittenAt` and `Citation`, which #762 has just finished making the single authority on where a report points, so it is worth doing deliberately rather than as part of something else.

## Acceptance

1. A citation that offers a place a reader can be sent to is one whose place is in a source this compilation holds, and a coordinate inside code out of sight is not that citation.
2. Which of the two a coordinate is, is settled where it is made rather than read back out of it.
3. Measured over the test suite the way #762 measured its own: the count of delivered citations claiming a reached-from place that this compilation does not hold is zero, and a check keeps it there.

kawasima:
Closes #785.

`Citation.OutOfSight.reachedFrom()` said its position was "a place in a file the reader holds" and nothing made it so. A position a parser made in a published module is out of sight from the moment it exists, so the same accessor handed back line 4 of a text nobody has — true only after a report had been anchored to an import line, and nothing distinguished the two states.

## What it measured

Over one compile of the test suite, before the change:

| | |
|---|---|
| out-of-sight citations built | 872 |
| whose `reachedFrom()` names no source | 614 |
| that reached one of the four surfaces reading it | 15, all naming a source |

Nothing was wrong in the output, because every one of those surfaces asked `sourceId != null` again before trusting the accessor. That is the defect said the other way round: a question the type was supposed to answer, answered privately by each reader.

Two things the issue does not mention turned up in the same measurement. `sourceId()` is compared with null by nine places, five of them outside `diag`, and each reads the absence as an answer to a question of its own — file this under the module's source, there is no module, open the module's document, this report is not anchored yet, read it in the diagnostic's own file. Five answers to three questions. And `standingInFor` is reached by splices whose call site is not a place the reader holds: one inside a text this compile cannot name, one inside a published module's own text.

## The two questions were not independent

The first shape this took held a source identity and one boolean saying whether the code was written at the position. That boolean was two facts. `isOutOfSight()` is read in production by three places and they were asking different things: the inliner asks which text a declaration came from, while the empty-literal rule and an underline's width ask whether the position was borrowed. Those answers differ for exactly one state — a position inside a module's published text, whose line 4 is where that module's author wrote that code, and which no reader holds a file for.

So the boolean was "is it written here **and** can a reader be shown it", and that is why a splice into a text nobody named had nowhere to land: it rewrote which text the position was in, turning a snippet into a module's published text.

Measured before changing anything: splitting the two and pointing each reader at its own question leaves the whole suite green. The conflation was carrying two questions under one answer that happened to agree everywhere it was asked.

## What a position holds

    which text : a source this compile holds | a text it cannot name | what a module published
    whose code : this position's own | copied here from p

Six states, every one reached, each with a one-sentence meaning. Every pair is legal, which is the condition under which holding a pair is right — the same rule that put the two halves of a position together, applied to components that are actually independent. `TextRead` and `WrittenAt` are gone into it.

A splice keeps the text and replaces the code. It can do that for every state, because there is a state to land in.

Where the code is written follows from both: code copied here is written where it came from, and code of its own is written wherever the text is — a module for a published one, nothing to state for a text a reader is looking at. A published text is identified by the module and not by the name a reading reached it under, because the route is a fact about the reading.

`NoText` is not introduced. Nothing distinguishes a snippet's position from one a pass minted, and inside a compile the second is all there is: two static `NOWHERE` sentinels. Separating them is the question about whether such a position is a place at all, which is #769's — and it is why the unnamed text keeps one value, since a per-parse identity cannot be defined for a sentinel. What that value does not promise is written on it.

### One projection per question

A placement answers nothing outside its own package. Both of its questions are package-private: a predicate publishing either would be the nullable source identity again under a name that reads like an improvement. What is published is one projection per question, each with one authority:

    Citation.of(SourcePos)      what a report may say about where the code is written
    DiagnosticPlace.of(Region)  where a reader may be sent
    SourcePos.quotedFrom()      which text this is, and what to say where it has no name
    SourcePos.wasCopiedHere()   whether the position is this node's own

`DiagnosticPlace.of` is built on `Citation.of` rather than classifying a second time. `QuotedFrom.TextItCannotShow` names the module that published the *text*, and says so — the code at a position in that text can be somebody else's, and that is the citation's answer.

### The accessor whose absence was being read

`SourcePos.sourceId()` is no longer public. Nothing about it was wrong — it is which file a position is in — but it could be compared with null, and that comparison is how a consumer answered "can a reader be sent here" without asking anything that could tell it.

Which of the three a position is stays in one place. What a caller does about it stays with the caller — the store files a report under the source it was reading for, a query about an unnamed text is keyed on no file, the editor shows a placeless answer against the module's document. Those are policy differences, not classifications.

### A citation per place a report can be in

    Written(at)                the file is one the reader holds, and the code is there
    Unplaced(at)               the code is there, and this compilation cannot name the text
    Reached(p, at)             the file is one the reader holds, and the code is in p
    UnplacedElsewhere(p, at)   the code is in p, at a position only the caller can use
    OutOfSight(p)              the code is in p and there is nothing to point at

Five, not six, though a placement has six states: the two states in a text no reader holds differ in whose code is at the position, and a report says that either way — what it cannot do is point, and that is what they have in common. A citation with nowhere to point always says where the code is, because a text no reader holds is one this compile has no file for, and having no file for it is knowing which module published it.

### And a report that can carry it

A primary region could hold a region or nothing, so a finding with nowhere to point arrived as a report with neither — and the reading that moves such a report to an import line had only the module it was filed under to answer from. The overwrite was still there, one boundary earlier than where it was found.

So a primary is a value with room for it: a region a pass built, passed on unread, or nowhere to point and where the code is written. Not held to being a place a reader can be sent to — a primary region is not held to anything yet, 53 of 3545 naming no source while saying the code is written there, and classifying those is #769's. Every surface that only wanted a region still asks for one.

The move then asks one question instead of two, because whether to move a report and what to move it with are the same question: it is moved when this compilation has nowhere to send a reader, with what the report already says. A report in a text this compilation cannot name is left alone — nothing about it says where its code came from, and answering that from the module is the inference again.

Both surfaces read that primary rather than its region. The terminal says where the code is in the words a label with nothing to point at already uses; the document writes it as a value beside the diagnostic, in the fields it already writes for code out of sight. A value nothing reads is a value nobody has, and reading the region alone was the same drop at the last boundary. Neither touches a region a pass built or a report with nothing at all.

Moving a report cannot re-decide what it is about: the provenance is supplied because a report with nothing pointed at has no answer to read, and one that has an answer refuses a different one.

Each reader of the old accessor had to say what it does with nothing to point at. A guard with a declaration to name says it without a marker, and one with nothing to name gets no label. The adequacy document refuses, because the shipped schema says a place has a source, a line and a column, and a position inside a module's published text is a real place the contract has no shape for — that refusal says the decision has not been taken, and is loud on purpose, the places written there being eight over the whole suite.

## Acceptance

1. A citation offering a place offers one whose text this compilation named, and a text it did not name has no place to offer. Both by construction.
2. Which of the states a position is in is settled where it is made: a parse, or a splice that keeps the text.
3. `EveryPlaceDeliveredNamesASourceThisCompilationHoldsTest` reads a compile's reports and its adequacy document the way a terminal and a build read them: places naming somewhere this compilation does not hold, zero. It measures the walk before checking it.

## Verification

Whole suite green after every commit, and after catching up with `develop` twice. souther-examples: 12 of 13 modules green, `todomvc` failing where it already failed on `develop`.

Negative controls, each confirmed to fail for the reason claimed: collapsing two states into one, making a nested splice compose rather than replace, publishing one of the two questions, cutting the anchoring, reintroducing the transition that rewrote which text a position was in, rebuilding a moved report's provenance from its module, and dropping the provenance when the report is built.

Two of those are worth reporting because they failed first.

A report about a module off the path is anchored by two paths — the walk that finds it and the reading that files it — and cutting either alone leaves the suite green. Cutting both produces no false place: it produces no place at all, and a reader told where the code is and sent nowhere. The type made the first defect unconstructible; #757 is what makes the report useful.

And dropping the provenance where the report is built was not caught at all until the test for that half was written. The end-to-end fixtures do not reach it — a module whose invariant reaches a construction in another module's helper is refused when it is built, so the middle module never reaches a reader — so both halves of that boundary are held to the rule directly rather than through a route that does not exist.


kawasima:
Both reviews are answered. The structural point and three of the automated findings turned out to be one defect, one level below where either put it.

## The two questions were not independent

`isOutOfSight()` is read in production by three places, and they were asking two different things. The inliner asks which text a declaration came from. The empty-literal rule and the underline's width ask whether the position was borrowed. Those answers differ for exactly one state — a position inside a module's published text, whose line 4 is where that module's author wrote that code, and which no reader holds a file for.

So `codeIsWrittenHere` was not "is the code written here". It was "is it written here **and** can a reader be shown it", and that is why the splice had nowhere to land for a text nobody named and rewrote which text the position was in.

Measured before changing anything: splitting the two and pointing each reader at its own question leaves all 4800+ tests green. The conflation was carrying two questions under one answer that happened to agree everywhere it was asked.

## What the model is now

    which text : a source this compile holds | a text it cannot name | what a module published
    whose code : this position's own | copied here from p

Six states, every one reached, each with a one-sentence meaning. Every pair is legal, which is the condition under which holding a pair is right — the same rule that put the two halves of a position together in the first place, applied to components that actually are independent.

`quotedFrom` is decided by the text alone. Where the code is written is decided by both: copied code is written where it came from, and code of its own is written wherever the text is — a module for a published one, nothing to state for a text a reader is looking at. That last is what the old model had no room for.

The projections were written out and checked before any code moved. `Citation` does not survive at four arms: `unnamed text × copied from p` has a position, no name for the text, and a provenance, and `DiagnosticRenderer.qualified` reads that provenance today. Collapsing it into `Unplaced` would have silently dropped the sentence saying where the code is written. So there are five, and "nothing to point at and nothing to say instead" is uninhabited by construction — a text no reader holds is one this compile has no file for, and having no file for it is knowing which module published it.

`NoText` is not introduced. Nothing distinguishes a snippet's position from one a pass minted, and inside a compile the second is all there is: two static `NOWHERE` sentinels. Separating them is the question about whether such a position is a place at all, which is #769's. For the same reason the unnamed text stays a single value — a per-parse identity cannot be defined for a sentinel, so that decision belongs with the separation, not before it.

## The findings

- **The `Unplaced` secondary could only throw.** `secondary(Region, …)` goes through `DiagnosticPlace.of`, which refuses that arm. Fixed by deciding what is actually said: a guard with a declaration to name says it without a marker; one with nothing to name gets no label, because there is nothing to say.
- **`sentTo` returning null closed a route to anchoring.** Fixed. A report with nowhere to point is now anchored on the same terms as one that had a position — the provenance comes from the module the walk was about, which is where the other anchoring path already reads it.
- **The adequacy document's refusal.** Re-measured: eight places written over the whole compiler and CLI suite, all of them placeable. Far too few to read as unreachable, and the reviewer is right that the justification was wrong. The refusal stays, and its reason is rewritten: the shipped schema says a place has a source, a line and a column, and a position inside a module's published text is a real place the contract has no shape for. Filling the field from whatever file is to hand, or leaving it out, are both documents the schema forbids. The refusal says the decision has not been taken, and is loud on purpose.
- **The vacuous switch-arm assertion.** The claim it was reaching for spans the arms — a reader is never left with a report that points nowhere and says nothing about why — so that is what it now says, over two compiles, counted before it is checked.
- **Stale `WrittenAt` javadoc** in the inliner and in the anchoring, corrected.

## Verification

All tests green after each commit, and after catching up with `develop` twice (#786 latest).

Negative control on the new model: reintroduce the old lying transition — a splice into an unnamed text rewriting it into a published one — and four of the eight state-space tests fail, each for its own reason, including the fifth citation arm disappearing.

PR #793, opened 2026-08-16
Refuse an invariant whose ends cannot both hold, on every ordered type

### LINKED ISSUE #773: A contradictory Date invariant is not refused where the same Int and Decimal ones are
A `Date` invariant whose two ends cannot both hold is not refused, where the same shape over `Int`
and `Decimal` is. Measured with the released CLI, 0.1.0-rc4, each declaration on its own:

    data X = Int       invariant value > 5 && value < 3                    refused (E1013)
    data X = Decimal   invariant value > 5.0m && value < 3.0m              refused (E1013)

    data X = Date      invariant value > Date("2020-01-01")
                                 && value < Date("2010-01-01")             compiles
    data X = DateTime  invariant value > DateTime("2020-01-01T00:00")
                                 && value < DateTime("2010-01-01T00:00")   compiles
    data X = Time      invariant value > Time("10:00") && value < Time("09:00")   compiles
    data X = Instant   invariant value > Instant("2020-01-01T00:00:00Z")
                                 && value < Instant("2010-01-01T00:00:00Z")  compiles
    data X = Date      invariant value == Date("2020-01-01")
                                 && value == Date("2010-01-01")            compiles

All four date-like carriers, both by ordering and by equality.

## What makes it a gap rather than a limit

These carriers are ordered as far as the rest of the measurement is concerned. A one-sided bound on
each of them produces a line the report counts:

    data T = DateTime
        invariant value >= DateTime("2020-01-01T00:00")

    pick                     injected      rows 1    pending 1
      boundary    1/1

So a `Date` position is not one nothing can be read of — a bound on it is read, and a row at it is
asked for. The count that decides whether a type has any value is not reaching the same reading.

## What to do first

Localize against the sibling that works. `Int` and `Decimal` go through the same declaration and
differ from `Date` in one thing, and the answer is which of the steps between an invariant clause
and `TypeCardinality` stops for this carrier — whether the clause is not turned into a bound, or the
bound has no count to be compared with, or the count is never asked. The symptom above is what is
established; the cause is not, and this issue should not guess it.

Independent of the two issues about equalities admitting no value: this one is about carriers the
ordering already reads.

## Acceptance

1. A `Date` invariant whose ends conflict is refused, and so are the `DateTime`, `Time` and
   `Instant` shapes above.
2. An invariant whose ends do not conflict is not refused. A probe holds the non-conflicting shape
   for each carrier.
3. What the localization found is stated, so that the next carrier added is added to a reading that
   is named rather than to a list that was extended.
4. The refusal names why the count reached zero, as the issue about the refusal's message requires.

### LINKED ISSUE #780: A contradictory ordering over an enumeration is not refused where the same Int one is
An `invariant` whose ends cannot both hold is refused over an `Int` and is not refused where the
position is an enumeration. Measured on the compiler at `e71eb38` plus the reading added for #771,
each declaration on its own:

    data Red
    data Green
    data Blue
    data Colour = Red | Green | Blue

    data P = { c: Colour }   invariant c > Blue                    compiles
    data P = { c: Colour }   invariant c > Blue && c < Red         compiles
    data P = { c: Colour }   invariant c >= Blue && c <= Red       compiles
    data Painted = Colour    invariant value > Blue                compiles

    data Small = Int         invariant value >= 1 && value <= 3
    data P = { n: Small }    invariant n.value > 3                 refused (E1013)

`c > Blue` admits nothing: `Blue` is the last case its enumeration declares, so no value of `Colour`
is above it. `c >= Blue && c <= Red` admits nothing for the same reason read the other way round.

## Why this is a gap rather than a limit

The ordering is already read of these positions everywhere else. An enumeration is a carrier with a
count (`Carrier.Ordinal`), a clause bounding one is recognised as an end that places a line, and the
classes a position holds are narrowed by it — `data StageI = Stage invariant value >= Qualified`
declares three cases and holds two, which the partition reading gets right today
(`ADeclaredCaseIsNotYetAClassThePositionHoldsTest`).

So a bound on such a position is read, and a line is drawn at it. What is not happening is the count
that decides whether the type has any value reaching the same reading.

## Not the two issues about equalities

#771 gives every position an admissible-value set, and it refuses `c == Red && c == Green` and
`c /= Red && c /= Green && c /= Blue`, because those are about which values are named. This one is
about where the values stop, which is the interval algebra's question over the carrier's count. The
two are deliberately separate readings: pushing ordering into the value sets would make one finite-set
evaluator answer for enum equality, enum ordering, numeric ordering and date ordering at once.

This is the same shape as #773 — a carrier whose bounds the boundary reading takes in, and whose
count does not — and the localization may well land in the same place. Whether one change answers
both is for whoever localizes it, and this issue should not guess.

## Acceptance

1. An enumeration whose ends cannot both hold is refused, and so is a one-sided bound that admits no
   case. Both on a record's field and on a name wrapped round the enumeration.
2. An enumeration bound that admits a case is not refused. A probe holds the one-sided and two-sided
   shapes that do admit something.
3. What the localization found is stated, so the next carrier is added to a reading that is named
   rather than to a list that was extended.

kawasima:
Closes #773. Closes #780.

## What was wrong

An invariant whose ends cannot both hold was refused over an `Int` and a `Decimal` and nowhere else.

    data X = Int       invariant value > 5 && value < 3                     refused
    data X = Date      invariant value > Date("2020-01-01")
                                 && value < Date("2010-01-01")              compiled
    data P = { c: Colour }   invariant c > Blue                             compiled

`Int` and `Decimal` are the positions the interval algebra carries: it holds one number per position and only for the numbers a model adds and subtracts. So the count that decides whether a type has any value heard the rule there and heard nothing anywhere else. A `Date` bounded above a date it is bounded below of compiled, emitted classes, and aborted at run time on every construction.

The ends were being read the whole time. A one-sided bound on a `DateTime` draws a line the report counts and asks for a row at, so the reading that turns a clause into an end reached these positions perfectly. What it produced went to the report and to no domain, and whether a value exists is asked of the domains.

## What the change is

**A rule is read against less of a type than writing a value at it needs.** `Carrier` answered both questions and the smaller one was being answered with the larger one's answer, which is why a `Time` went unrefused: settling it asked for a conversion back from the counts, and no reading of it uses one. `OrderScale` is the reading — a literal onto the order, how the values are spaced, where the order stops — and `Carrier` is that with the writing added. `Time` and `Instant` are scales and not carriers, so no line is drawn at either and the spec's list of what a line is drawn on is unchanged. Which types are ordered stays one table: `Carrier.ofValue` narrows `OrderScale.ofValue` rather than deciding again.

**A rule that steps off an order says so.** Sharpening `value > 5` onto 6 asks the order for the count beside the one named, and where there is none — the last case an enumeration declares — the reading answered "no end", which is what it also answers to an equality and to a literal it could not read. The reader's own comment said as much. It is three answers now, and every caller answers for all three.

**`OrderedIntervals` is a component of `ConstraintState` and an arm of `isBottom`.** A range per position over whatever order the position has. `OrderedReading` reads the clauses for it beside the reading that reads them for which values a position may hold — over the same list, at the same moment, each making its own answer of a clause.

It covers every order and not only the ones no other domain holds. Splitting on which positions have a numeric atom would put the affine fragment's shape into the answer, and would leave the same model told two different sentences depending on what carries it. Where two domains both hold a contradiction, `Emptiness.preferred` writes the more particular proof, so `value > 5 && value < 3` and the same shape over a `Date` now carry one proof and one sentence.

**Every position starts at everything its order has.** Not at an unbounded range: what is below the empty string is nothing, and a reading open below took `value < ""` for a rule leaving room underneath.

**One value, two representations, one bridge.** `Terms.asWrittenValue` rendered a preserved call and dropped the call that builds a temporal and the case of an enumeration named by itself. That was a blind spot on one side of the bridge and not on the other: a reader working from the declarations saw `Date("2020-01-01")` and read the bound, and a reader working from the analysis representation was handed nothing for it.

## Beyond the two issues

Two shapes neither issue names fall the same way, and are refused here: `value > "b" && value < "a"` over a `String`, and `value < ""`.

## Spec

`ends-that-cannot-both-hold-admit-no-value` is new under `a-data-can-be-constructed`. `a-line-is-drawn-where-the-values-can-carry-one` now says a rule over a `Time` or an `Instant` is not read *here* — no line is drawn at either, and whether such a rule admits any value is settled on every ordered type.

## Measured

Every module green (compiler 5004, 1 skipped).

The corpus adequacy report is byte-identical to develop over 21582 lines. That proves no regression and nothing else: the corpus holds no temporal invariant bound and no enum ordering invariant, so a probe was written for those two shapes and is identical between the two builds as well.

Two negative controls. Silencing the new domain in `isBottom` and `holdsNothing` fails 21 tests across the five classes about it. Seeding the extents as unbounded fails `value < ""` alone — the stepping orders reach the same answer through `onTheGrid`, so the seeding is load-bearing for `String` and for nothing else.


kawasima:
Review findings addressed in `bb45136b`. I measured each one before acting; three were real and two were not.

## `||` was judged without the order it sits on — real, fixed

Confirmed: `a < "" || b < ""` compiled, while `a < "" || a < ""` was refused. The extent was applied once around the whole reading, so a branch carried what its rules said and not where its order stops — and whether an alternative admits anything is exactly that question, asked of the branch. Whether the whole rule was refused then turned on whether the branches happened to be empty at the same position.

The extent now goes on where a rule becomes a range, which is the one place a position is spoken about; the outer seeding is gone, so there is no second place to keep in step. Two tests: a disjunction over two positions neither of which can hold anything, and the control beside it — a disjunction with one alternative that can be taken must stay admitted, and a position only one alternative speaks about must stay open.

## The sentence named one of three ways of showing it — real, fixed

Three shapes reach the proof and only one writes a pair of bounds. Measured, all three got the same wrong sentence:

    data X = Time    invariant value > Time("23:59:59")            one end, no pair
    data P = { c: Colour }  invariant c > Blue                     one end, no pair
    data X = String  invariant value == "A" && value == "B"        no bounds at all

The message is `NothingIsLeftForThatPositionToHold` now and says what was shown: the rules leave the position nothing to hold. The spec rule is reworded to match (`rules-leaving-a-position-no-value-are-refused`), and it also says that a choice among alternatives none of which admits anything admits nothing — the first finding's rule, written down.

## The size path borrowed a justification that is not true of it — real, fixed

Silence about a rule stating an end past the end of an order rested on the declaration being refused for it. No end of a *size* reaches the reading that refuses declarations — that reading is over the positions a value has, and a size is a number taken of one — so `String.length(value) > 9223372036854775807` was refused by nothing and reported by nothing. The size caller reports again, and the two callers now say which of them they are. Test at the report level, with an ordinary size bound as the control.

## The bridge guard was wider than its rule — real enough, narrowed

Nothing misbehaves today (`ConstEval` folds only preserved calls), but the guard claimed every call to a resolved name. It is held to the calls that answer a temporal, read off the type rather than a list of names.

## The positions map is in declaration order — not a finding

Measured on a record of `Date`, `Int`, `Map<String, Int>`, `String`: `keys = [first, second, third, fourth]`, and no atom is outside the keys. An atom is named from a body key, so a position with an atom has a key and the second pass adds nothing; a size has no key and is not one of these. The claim holds. Written down in the code rather than left to be re-derived.

## `relating` for a rule past the end — left as is

Such a clause compares a position to a constant, so it is not a relation and registering it as one would name a declaration that moved no edge. The comment already says so.

## Measured

Every module green (compiler 5008). Corpus and the temporal probe both byte-identical to develop. Two controls: not meeting the extent where a rule becomes a range fails the two disjunction tests, and telling the size caller it is refused elsewhere fails the size test.

kawasima:
Fixed in `06c79b89`. The finding is right, and measuring it showed the reach is wider than the two cases named.

## What was measured

| clause | before |
|---|---|
| `s < "" \|\| (b == true && b == false)` — ordered ￨ values | compiled |
| `(a == true && a == false) \|\| (b == true && b == false)` — values, two positions | compiled |
| `(a == true && a == false) \|\| (a == true && a == false)` — values, one position | refused |
| `(p < q && q < p) \|\| (b == true && b == false)` — numbers | compiled |
| both branches contradictory in the predicates | compiled |

So the last one refused was refused only because both branches happened to be empty at one position in one language. Which alternative is possible is not a question about where the emptiness sits.

## The root

The connectives were being applied inside each language. `AdmissibleReading` joined value sets and `OrderedReading` joined ranges, each over the same clause tree, neither seeing what the other had found — and the identical connective code sat in both, free to drift.

A choice between two alternatives is a choice between two readings of *the whole value*. So the connectives are written once (`ClauseReading`), the two languages supply leaves, and `StatedByClauses` is the pair whose `either` drops a branch that holds nothing in either of them. Dropping a branch takes its unread rules with it, and rightly: no value satisfies that branch, so what could not be read inside it narrows nothing a value of the type is under.

The first three rows are now refused; the control — a choice with one alternative that can be taken, and a position only one alternative speaks about — stays admitted.

## What is still not shown, and why it is a different change

The last two rows are not fixed, and not by omission. `numbers` and `facts` are not read from the clause list at all: they are what a construction owes (`Predicates`), and an alternative owes a construction nothing a guard could discharge — so neither ever reads a branch of a choice. Giving those two a branch-wise reading is a second reading of the numbers beside the obligations, which is its own change with its own reason to make. That is written down on `StatedByClauses` rather than left for the next reader to rediscover.

I left the spec sentence as it is: it states what the language holds, and `a-data-can-be-constructed` already closes with "the refusal says which of them was shown" rather than promising every way is shown.

## Minor

`Emptiness.EmptyOrderedInterval` still described itself as crossed ends after the sentence over it was generalised. It now says what it is — three shapes, one thing shown — and the stale "four leaves" count in the file header is gone.

## Measured

Every module green (compiler 5009). Corpus byte-identical to develop. Control: dropping neither branch in `either` fails the two new tests and nothing else.

kawasima:
Fixed in `e7d18240`. Confirmed exactly as described, and the root is one sentence:

**Two situations were being answered with one rule.**

An alternative nobody can take leaves the answer to the others, and its evidence goes with it — nothing satisfies it, so what it said narrows nothing a value of the type is under, its unread rules included. Where *every* alternative is impossible that reasoning does not hold: no one of them speaks for the rest, and returning the first one found impossible settles the proof by the order the operands were written in.

Measured before:

| written | shown as |
|---|---|
| `s < "" \|\| (b == true && b == false)` | `ItsRulesCannotAllHold`, no place |
| `(b == true && b == false) \|\| s < ""` | the ordered proof, at `s` |
| `a < "" \|\| b < ""` | at `b` |
| `b < "" \|\| a < ""` | at `a` |

One model, written two ways, refused two ways — which is the same defect this PR exists to remove one level down, where `Int` and `Date` were told different things about one shape.

So where every alternative is impossible the alternatives are kept, by meeting them: the choice admits nothing whichever way round it is read, and every position either of them left empty is one the choice leaves empty. `Emptiness.preferred` and the declaration order then choose, which is where that choice belongs.

This is the rule `Emptiness.AcrossEveryCase` already states for a sum — "what proves it has none is the whole list, and picking one of them to speak for the rest would answer a question about which case is at fault that nothing asked" — reached here for the same reason, and the Javadoc says so rather than restating it as a local trick.

`OrderedIntervals.join` carried the same early return. Its own users cannot reach it with both sides bottom now that the product handles the case, but its algebra should be right standing alone, so it gets the same case.

## Tests, both ways round

- the proof and the place: `AtAField("s", EmptyOrderedInterval)` for the mixed pair, `AtAField("a", …)` for the two positions, each asserted with the operands reversed
- the sentence: `NothingIsLeftForThatPositionToHold` either way round
- the domain's own join: both sides bottom keeps both positions, and `holdingNothing()` is the same set either way

## Measured

Every module green (compiler 5012). Corpus byte-identical to develop. Control: dropping the both-impossible case fails exactly those four tests and nothing else.

kawasima:
Fixed in `f0433037`. The counterexample reproduces exactly as written, and the root is one sentence: **keeping both impossible alternatives by meeting them was the wrong way to keep them.**

A meet is a conjunction, and alternatives were never stated together. So it composed facts across the choice:

| model | named before | named now |
|---|---|---|
| `(a < "" && b == 0) \|\| (a < "" && b == 1)`, `b` declared first | `b` | `a` |
| the same with the operands reversed | `b` | `a` |
| `a < "" \|\| b < ""` | `a` | no position — `ItsRulesCannotAllHold` |
| `b == 0 \|\| b == 1` (control) | admitted | admitted |

## The rule

What a choice leaves a position with is what **every** alternative leaves it with — an intersection, not a conjunction. `join` over two impossible sides keeps the positions both of them left empty.

Where there are none, the choice still admits nothing and no position is at fault. That case needed saying and neither domain could say it: bottom was encoded as "some position is empty", which has no room for "nothing satisfies this, and not because of any one position". Both domains carry it explicitly now.

`leavingNothing()` is how a caller marks a side impossible on evidence from outside it — which is the pair's case, where one language showed the branch impossible and the other found nothing wrong with it. It claims nothing about any position, because what is known is about the whole.

## Finding 2

`AdmissibleValues.join` gains the bottom rules its neighbour already had. You are right that it was reachable only through `StatedByClauses`, which handles bottom first — an algebra that is right only because of who calls it is one the next caller gets wrong. Its own tests now cover: an impossible alternative leaves the choice to the other (either way round), neither alternative possible admits nothing (either way round, same per-position answer), and a side marked impossible from outside names no position.

## Where the proof is written

`ConstraintState.holdsNothing` writes the particular proof only where there is a position to name. Without one it would name whatever place the reader stood at — the declaration's own value — so a choice whose alternatives fail at different positions gets the general sentence, which is exactly what a reading that showed a contradiction and could say no more leaves.

## Tests

- proof and place, both operand orders: `AtAField("a", …)` for the counterexample, `ConflictingRules` for alternatives failing at different positions, `ConflictingRules` for the mixed-language pair
- sentence, both orders, for both shapes
- `OrderedIntervals.join`: a position every side leaves empty is named; a position neither side leaves empty is *not* made empty by the choice (the fabrication case you asked for); a side marked impossible from outside names no position
- `AdmissibleValues.join`: the three above

## Measured

Every module green (compiler 5019). Corpus byte-identical to develop. Control: meeting the two impossible alternatives again fails four tests — the proof either way round, the place, and the sentence — and nothing else.

PR #794, opened 2026-08-16
A diagnostic says what it points at, and a surface says what it is reading

### LINKED ISSUE #769: A diagnostic whose primary region is no place is anchored to the file the report was filed under

kawasima:
Closes #769.

## What it was

A diagnostic's primary arrived downstream as a region and a null, and the source
it was in was told to it from wherever the report had been filed
(`Spot.primary(d, sourceId)`). Neither value said which of four things it was, so
each reader read the null as whichever it had in mind and read the told source as
whichever question it was answering.

## What the measurement said

Instrumented at construction and at both delivery points, over the whole suite:

| where | |
|---|---|
| primaries reaching a renderer | 266 |
| — told source agrees with the region's own | 119 |
| — **told nothing while the region named the file** | **117** |
| — region names no file, told a source anyway | 11 |
| — region names no file, told nothing | 17 |
| — **told source contradicts the region** | **0** |
| reports the compiler anchored | 5346 |
| — region names the file, told agrees | 5334 |
| — no region at all (E1503) | 10 |
| — contradicts (its own fallback's test) | 2 |

So the pair carried no information where the region answered, was the only answer
where it did not, and was never a second, contradicting answer. The issue's 53
turned out to be one state rather than fourteen codes' worth of separate
problems: a parse of a text nobody named (`Citation.Unplaced`), plus test
fixtures building positions by hand.

## What it is now

The region is classified where the report is built, through the citation its
start projects; what a surface knows is a value it hands over.

```java
sealed interface Primary { InSource | InAnUnnamedText | Unavailable | Nowhere }
sealed interface Spot    { InSource | InTextBeingRead }
sealed interface Shown   { ItsSubject | ALabel }
record ReportContext(Optional<SourceId> filedUnder, Optional<TextBeingRead> textBeingRead)
sealed interface SpotResolution { Found | NoPrimary | CodeUnavailable | TextWasNotProvided }
```

A place a reader is sent to carries its own source. A place in a text this
compilation cannot name carries the text the surface says it is reading, which is
not a second answer because `UnnamedRegion` refuses a region that answers for
one. Which file a report is listed under is neither of those, and is on the
context.

Gone rather than renamed:

- `Diagnostic.region()` and `Diagnostic.pos()` — four questions under one
  nullable, and every reader picked one.
- `Compilation.sourceIdOf(Db.Found)` — half of it is answered by the place now,
  and the half that is left is `filedUnderOf`, which the report cannot answer.
- `Spot`'s told source, `Located.primarySourceId`, `Located.inFile`,
  `CompileException`'s parallel `List<Diagnostic>`/`List<SourceId>`, and
  `Main.indexOf`'s "one file handed over answers for any id".

## Two defects this turned up

**A report about a path module could be moved as another module's.**
`Front.needs` and `cannotBeReadBack` said nothing about where their code was, and
`Diagnostic.reachedFrom` takes the caller's word in exactly that state. Saying it
at the producer puts the two under one check. Nothing a reader sees changes.

**The command line quoted the only file it had for a report about another.**
`indexOf` answered index 0 for whatever id it was asked, including for none.

## What is checked

- `AReportAboutAPathModuleSaysWhichModuleItIsAboutTest` — both producers name
  their provenance; a move that agrees goes through; a move to another module's
  code is refused. Negative control: with the old `.nowhere()`, nothing is
  thrown.
- `HowManyFilesWereHandedOverDoesNotDecideWhatIsQuotedTest` — one file and two
  render the same string; a report the command cannot place is not quoted from
  the only file to hand. Negative control: with the shortcut, the header names
  that file.
- `ASecondaryRegionNamesItsFileTest` — a report in a text the caller names is
  quoted from it; one in a text nobody named is not quoted from whatever a
  resolver answers with.
- `AnalyzerTest#aSemanticErrorIsMarkedWhereItIsWritten` — the single-document
  route marks the reported line. This caught a regression made while writing
  this: the marker had fallen to the head of the document.
- `AReportAboutAModuleOffThePathIsSaidWhereItWasReachedTest` — the old
  `twoFindingsInOnePublishedModuleAreSaidOnce` bundled two claims, one of which
  its own denominator carried. Split into an identity test and a walk-completeness
  test with a fixture whose two findings differ in what they say.
- `NoValueHoldsASourceBesideOneThatAnswersForItTest` — `OPEN` is empty, and
  `Diagnostic` joins the types that answer for a source.

## Verification

Full suite green on the tip (4975 tests, from a clean build). Each of the four
commits was checked out and `test-compile`d on its own; only the tip has had the
whole suite run against it.

### DIFF
--- a/souther-bench/src/main/java/souther/bench/Corpus.java
+++ b/souther-bench/src/main/java/souther/bench/Corpus.java
@@ -86,7 +86,7 @@ public record Corpus(String name, List<String> sources, int lines) {
             for (Diagnostic diagnostic : found.getValue()) {
                 if (diagnostic.severity() == Severity.ERROR) {
                     errors.add(name + " source " + found.getKey() + ": " + diagnostic.code()
-                            + " at " + diagnostic.pos());
+                            + " at " + diagnostic.primary());
                 }
             }
         }
--- a/souther-cli/src/main/java/souther/cli/Main.java
+++ b/souther-cli/src/main/java/souther/cli/Main.java
@@ -716,14 +716,14 @@ public final class Main {
     /**
      * Which of the files handed over a source id names, or -1 when it names none of them.
      *
-     * <p>For a diagnostic, which may name no source — a report this compile could pin on none — so
-     * the single file is the answer whatever the id reads. Not for a reason in a report, which
-     * always names one: {@link #namesOf} matches on the id alone.
+     * <p>On the id and nothing else. The one file handed over used to be the answer for whatever it
+     * was asked, including for no id at all, because a report could arrive here without one and the
+     * single file was the only guess to be had. A report says which source it points into now, so
+     * there is nothing left to guess — and the guess was answering for reports it had no business
+     * answering for: handed one file and a report about another, it quoted that file at the
+     * report's numbers, which put a caret past the end of a line the author never wrote.
      */
     private static int indexOf(List<Path> sources, SourceId sourceId) {
-        if (sources.size() == 1) {
-            return 0;
-        }
         for (int i = 0; i < sources.size(); i++) {
             if (Compilation.idOfSourceIndex(i).equals(sourceId)) {
                 return i;
--- a/souther-cli/src/main/java/souther/cli/Main.java
+++ b/souther-cli/src/main/java/souther/cli/Main.java
@@ -662,9 +662,9 @@ public final class Main {
     /**
      * Which of the files handed over a source id names, or null when it names none of them.
      *
-     * <p>One file handed over is the answer however the diagnostic is tagged — which is why one
-     * item is not read as "the source called 0, or nothing". A compile of one source names it like
-     * any other now, so this is for the report that could be pinned on none.
+     * <p>Matched on the id, like everything else this command resolves. A report says which source
+     * it points into, so a caller with one file to hand has nothing to guess at and a caller with
+     * several has nothing to work out.
      */
     private static Path pathOf(List<Path> sources, SourceId sourceId) {
         int at = indexOf(sources, sourceId);
@@ -689,12 +689,10 @@ public final class Main {
      * front of the reader. Both renderings go through {@link #displayNames}, so a run cannot quote a
      * line from {@code a/model.sou} and then say the rows of {@code model.sou} were not read.
      *
-     * <p>Matched on the id and nothing else, which is where this parts from {@link #indexOf}. That
-     * answers for a diagnostic, which may name no source at all — a report this compile could pin on
-     * none — and the one file handed over is then the answer. A reason in a report always names one,
-     * so an id that is none of these files is an id
-     * about a source this command did not hand over, and answering with the only file would be
-     * inventing the very correspondence this is here to stop being guessed at.
+     * <p>Matched on the id and nothing else, as {@link #indexOf} is. An id that is none of these
+     * files is an id about a source this command did not hand over, and answering with the only file
+     * there happens to be would be inventing the very correspondence this is here to stop being
+     * guessed at.
      */
     static SourceNameResolver namesOf(List<Path> sources) {
         List<String> names = displayNames(sources);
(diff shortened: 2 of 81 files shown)


kawasima:
Review findings addressed in 7e1f69e1.

**1 (high) — a label became the anchor for a report with no place of its own.** Confirmed, and it is this PR's own defect one layer over: a report that says it points nowhere was given a line and a column that came from a label. Reproduced with a primary built by `atCodeWrittenOutOfSight` carrying an in-source label, which is what `Adequacy` produces for a boundary finding whose body is out of sight and whose guard is not:

```
before the fix                              after
-- EXAMPLE E1911 ------model.sou:2:3        -- EXAMPLE E1911 ----------model.sou
2| line two here
     ^                                      model.sou:2:3
(the guard's sentence is gone)              2| line two here
                                                 ^
                                            The guard that draws that line.
```

A label anchors only where the report has a place of its own somewhere else — "the two change places" presupposes two. Whether the report points into the file is now asked apart from what the marker goes on, because for such a report they differ: it is still said on the file its label is in, so an author editing that file is told.

`AReportWithNowhereToPointStillSaysWhereItsCodeIsTest#aLabelOfItsOwnDoesNotBecomeThePlaceItPointsAt` pins it. Negative control: with the old selection the anchor comes back as the `ALabel` and the header carries `2:3`.

**2 (low) — two comments on `Main` described the fallback the previous commit removed.** Rewritten; `namesOf`'s stated reason for differing from `indexOf` is gone, because it no longer differs.

**3 (low) — a stacked javadoc.** Removed, and the helper below it now has one of its own.

Chasing 3 turned up two more of the same kind, both from the mechanical rewrite that moved tests off `Diagnostic.pos()`:

- `DiagnosticRenderTest` asserted `region().start() == null ? null : d.code()`, which after the rewrite could not be null, so the line duplicated the one above it and checked nothing. It asserts the primary points into a source now.
- `EveryPlaceDeliveredNamesASourceThisCompilationHoldsTest.citations` cast every primary to `InSource` where the guard it replaced was what kept a report with no place from reaching the cast — in the one fixture that deliberately reaches out of sight.

Full suite green (clean build).

PR #796, opened 2026-08-16
Read a behavior's input once, and hold an `unreachable` against it

### LINKED ISSUE #778: A case declared unreachable is taken out of the measure without the claim being checked
A model states where its combinations cannot arise, and the report takes it at its word: it removes
the class from the denominator and prints the claim back as a finding. Nothing asks whether the case
is reachable.

    module h

    data A
    data B
    data Kind = A | B
    data Out = { n: Int }

    behavior pick : (k: Kind) -> Out
        constructs Out

    let pick (k) = match k with
        | A -> Out { n = 1 }
        | B -> unreachable "B never arrives"

`B` arrives whenever a caller passes it, and the declaration two lines above says so. Measured with
the released CLI, 0.1.0-rc4:

    pick                     implemented   rows 0    pending 0
      partition   axes 1   single-axis 0/0   excluded 1
        · `B` is declared unreachable: B never arrives

It compiles, and the line under `partition` is the model's own claim restated as something the
report found.

## What would refute the claim is what the claim removes

Reaching an `unreachable` is E1911, so a row written at `B` refutes it. That row is the only thing
that ever does — and the exclusion is what stops it from being asked for. The measure drops the one
obligation whose discharge would have shown the model wrong.

## Where it is trusted

`Exclusions.of` follows the body's spine to the first fork, and where that fork is a `match` on an
input position it excludes every case whose arm answers nothing. Whether the case can arrive is not
asked. Its own account of the safe direction is about the exclusions it misses:

> So this misses exclusions that hold. That direction is the safe one: a case left in is a case the
> report asks for, and an author reading a gap they cannot fill is at least told something true
> about their model — an author never told about a gap is not.

Right about a missed exclusion, and silent about a false one, which runs the other way: a case taken
out is a case nothing asks for, and the author is told something untrue about their model.

#426 put the exclusion in for the opposite defect — an arm answering `unreachable` was counted,
which left every correct model permanently one arm short. That reading is not in question here. What
is missing is its other half: an exclusion is a claim, and a claim the measure acts on is one the
measure should hold the model to.

## What the shape turns on

Whether an unreachable case is refused, or left in the denominator and reported, or left excluded
with the claim named as unproven, is one decision and belongs in one place. A case reached at the
first fork from a parameter whose type has it is the easy end: the type is what says it arrives. A
case behind another `match` or behind a guard is the other end, and the answer there may be that
nothing is proven — the same answer `GuardReachability` already gives about an arm.

Whichever way it goes, an exclusion must not be both unproven and silent.

## Acceptance

1. A case declared unreachable that the model's own rules say arrives is refused, and the
   declaration that says it arrives is named.
2. A case whose reachability the reading cannot settle is not silently excluded. What could not be
   settled is said.
3. A case the model does rule out keeps what it has today: out of the denominator, its reason
   printed, and no row asked for.
4. The reading of #426 stands — an arm answering `unreachable`, in a model that states its
   impossible combinations, is still not counted as an arm.

kawasima:
Closes #778.

A case an arm declares cannot arrive was taken at its word: it left the
denominator, and the report printed the claim back as something it had found.
Nothing asked whether the case can arrive — and the row that would have refuted
the claim is the one the exclusion stopped anybody from being asked for.

## What was underneath

Three readings answered "what can arrive at this input position", and they
disagreed:

| reading | what it read | `data Active = Flag invariant value /= Off` |
|---|---|---|
| the signature measure | the type's cases | asked for a row at `Off` |
| the partition | the type crossed with the rules | left `Off` out, correctly |
| the arms | the body's shape | asked for the arm behind `Off` |

`Exclusions` papered over the first and the third with the body's claim. A model
that wrote the claim measured correctly and a model that did not was asked for a
row `E1903` refuses — with no `unreachable` written anywhere near it. The claim
was load bearing, which is why it could not be checked.

## What this does

**One reading, made once.** `InputDomain` walks a behavior's positions and
answers, per position, which distinctions it has and which of them the rules
leave. Every measure counts against it: the signature's cases, the partition's
classes, the arms owed a row, and the rows the generator offers. A `Position`
cannot be written down — the interface is sealed over a package-private record —
so nothing can assemble a reading out of what a body claimed.

**A claim is held against that reading, never read into it.** Three answers and
no fourth:

* the rules refuse the case — the claim says what they say, and the report keeps
  the author's own words;
* the rules leave it standing at a position nothing stands between and the
  caller — refused where it is written, **E1326**;
* nothing settles it — the case keeps whatever it was owed, and the report says
  the claim was not proven.

Refusal is unconditional and admission is not. That is why an arm written for a
refused case is owed no row at any depth, while a contradiction is only ever
reported at the first fork — and why both blank cells of the matrix ADR-0083 was
written for are untouched.

**The fact and the policy are apart.** `ReadingResult` keeps what the rules came
to, `Complete` with nothing kept included; `ObligationDomain` applies the one
widening this compiler still makes there (issue #780) in one function, so no
reader repeats it and fixing #780 changes that function alone.

## Measured

With the built CLI, over probes in the issue's shape:

```
the issue's model      E1326, and `partition 1/2 · no row is in B`   (was: excluded 1)
`Active` with a claim  excluded in both measures, by the rules       (was: by the body)
`Active` with no claim excluded 1                                    (was: 1/3, `Off` owed)
`invariant f /= g`     counted, and said to be unproven              (was: excluded 1)
a nested two-sum table unchanged
```

`souther compile` refuses the first with exit 1. A claim written in a `let` a
behavior calls is judged against the input the call gave it: the diagnostic
points at the helper's `unreachable` and names the calling behavior's parameter.

souther-examples: no measure moves in any of the 14 modules, and nothing there
is refused. The only JSON difference is an `unprovenClaims` array that is empty
everywhere.

Tests: 7792, no failures.

## Notes for review

* Commits are one step each and each is green on its own. The first two change
  no behavior — the corpus is byte-identical under `examples --format json` —
  and the third is where the output moves.
* The corpus does not exercise the value-set crossing: with `admits` mutated to
  refuse nothing, all 14 modules stay byte-identical. What kills that mutation is
  `AClassIsOneThePositionCanHoldAndNothingTakesItAwayTest`, 6 of its 13.
* `NO_CLASS_OPEN_AT_POSITION` is gone. The state it named was made by exclusions
  and cannot arise now that every class of an axis is one a row can be written
  at.
* An unproven claim leaves a gap `--strict` refuses over. That is the cost of
  subtracting only what is proven; the report says which of the two kinds of gap
  it is.
* Still open: a confirmed exclusion names the case and not the rule that refuses
  it. `AdmissibleValues` carries values without the declarations that wrote them,
  and threading that through its connectives is its own change.


kawasima:
Review findings addressed in 1a758a90.

**The one that mattered (#1).** Confirmed as a live defect, and worse than reported: the body reading matched a parameter *by name*, so

* a local shadowing the parameter — `let f = defaulted(f)` — had every read below it judged against the parameter's rules, and a valid model was refused with E1326; and
* the mirror case, a claim inside an expanded helper, was caught only where the helper's parameter happened to share the behavior's spelling. Renaming it from `f` to `g` made the refusal disappear, so what the previous commit claimed about helpers held by coincidence.

A binding is what a name is, so that is what is asked now. Keeping it working took carrying the bindings a reader has passed (`InputReads`): a helper expanded into a body binds the call's argument to its own parameter, and does it *inside* the condition of an `if` rather than above it. Guards, arms and claims all resolve the same way now. Both probes are tests, and mutating the resolution back to spelling fails the first one.

The corpus moves, in the direction of reading more: `hr` gains an axis with two classes (both already covered) and 64 boundary leaves; `crm`'s `closeDate` moves from "the model divides it no way" to "a rule was written that this could not read". One goes the other way — `ordering`'s `calendar.aSunday`, whose rule sits in a lambda handed to a combinator, where the chain to the parameter is not a run of bindings this follows. That is a note about one position and moves no measure; what it had before was the same name coincidence.

**Also fixed:** the queries that no longer read the claims stop asking for them (#2), `Claims.confirmedReasonsFor` and `UnreachableReasons.of` are gone as unread (#3), `ProducedCases` stops naming the deleted class (#4), and the optional's class throws rather than passing a null it did not expect (#7).

**Left as they are, with reasons:**

* #5 — `Inputs` answers absent for a module holding a type nothing could be worked out for, and readers map that to `NONE`. It is reachable only for a model that already failed to compile, where the report is a partial one about a module with a hole in it. Worth a follow-up if a report of such a module is something we want to keep deriving axes for.
* #6 — a claim at an axis past the `MAX_AXES` cap is not printed. The position itself is dropped from the report there, so nothing about it is asked for either; the "never both unproven and silent" rule holds for every position the report speaks about.

Tests: 7779, no failures.

kawasima:
Both findings and the test gap are addressed in 05c3f0ee; develop is merged in (ee9be04d).

**Claim scope (P2-2).** Taken the second way: nested claims are collected and judged, so the specification and ADR keep their promise rather than being narrowed to fit the reader. Every `match` on an input position is read wherever it is written, and a claim carries what stands above it. That is the only thing it decides — a case the rules refuse cannot arrive however deep the fork is, while a case they leave arrives at the *position* and says nothing about whether this fork is reached. First fork: contradicted claims are still refused. Behind another arm: unproven, with its own word.

What that is worth shows up in the corpus. `example.employmentinsurance`'s matrix — the model ADR-0083 was written for — now reports both blank cells:

```
`TwentyYearsOrMore` is declared unreachable: twenty insured years cannot have accrued
before the age of thirty, and nothing here proves it: this arm is inside another, and
what reaches it is not read here
```

That is the report #426 was measured on, and the thing its author could not get from the output: which gaps sit behind an `unreachable` and which are theirs to close. Nothing else in the 14 modules moves and nothing is refused.

The spec and ADR now also state the limit that remains: a `match` on something that is not an input position is not read at all, because it is a claim about a value no rule of the model speaks about.

**Claims off the axes (P2-1).** `PartitionEvidence` carries `claimsOffAxis`, keyed by path, for claims about positions no axis speaks for; the human report and the JSON both print them, and the schema has the field. Regression tests for both ways to have no axis — past the twelve-axis limit, and deeper than the walk reads — and they fail when the projection is stubbed out.

**Test gap.** `AnArmNothingReachesIsNotOwedARowTest` gains the missing case: `data Active = Flag invariant value /= Off` with an `Off` arm that answers a value goes from `branch 1/3` to `1/2`, with the reachable gap still named. Beside it, a control with the invariant removed, where every arm is owed a row. Both fail when the match-arm half of `ArmReachability` is removed.

**Merge.** develop's `whyUnread` change (`unreadFormOf`, telling a bound past the end of an order from one nothing could read) landed in a file this branch had moved; the rule is ported to where the reading now lives, and the conflict is resolved that way rather than by keeping either side whole.

Tests: 7859, no failures.

kawasima:
All four addressed in 57ff585b.

**1 — two producers of the judged claims.** Real, and the one that matters: this branch is built on that invariant. The suggested fix is not reachable directly — `Adequacy.Claimed` asks `Bodies.Checked`, so the check cannot ask it back — so the judging moved into what the body check answers (`Elaborated.claims()`), and the query is a projection of that. One judging, and the two answers that must never differ (a build refusing a claim, a report calling it unproven) now cannot.

**2 — `None` droppable by a finite value set.** Confirmed, and it predates this PR: the old classifier tested `ObservedValue.Absent` against values converted from rules, which never produce one, so `None` held nothing of any finite set either. What changed is the cost — a refusal now takes an arm out of what the rows are owed — so the distinction says what it is: a rule names values and absence is not one of them, so neither answer is settled by a set of values.

**3 — an arm with no `unreachable` becoming a claim.** Not reachable today: `NormalReturn.of` is false only where an `unreachable` is on every path, so the reasons are never empty. Fixed as an invariant rather than by skipping such claims — `Claim` refuses to be built without a place, which is loud if the two readings ever disagree, where dropping them silently would not be. `ClaimDiagnostics` loses the null check that was covering for it.

**4 — the duplicated `@param read`.** Fixed.

Also checked, and agreed with your non-issue: `assign(...)` cannot return null now that `ordered()` filters on `Axis::derivable`.

Tests 7859, no failures; souther-examples byte-identical to the previous commit over all 14 modules.

kawasima:
All four in 2eb7e993, and the first was a hole in the check itself.

**1 — a `match` as a `let`'s value.** Reproduced. The reading walked the spine and called what it ended at the first fork, but what a `let` binds runs *before* the body it binds it for, so a fork every caller reaches was read as one standing behind something and escaped E1326. Fixed at the root: the first fork is taken from the order things run in, and that order lives in one place (`Evaluated`) — the reasons walk was already reading it for its own question, so both now ask the same account of what runs first, including a construction's declared-field order.

`CompileUnreachableTest` had that shape twice and is refused by the fix, which is the right answer for those models. Both tests are about what the emitter does when an abort is reached, so both keep their subject and move to a premise nothing settles — which is also what leaves the value writable at the boundary and able to reach the abort.

**2 — the human report.** The off-axis claims were printed inside the arm that prints numbers, so a behavior whose positions were all dropped or never read had its verdicts reached and not shown. A claim is not a number: printed either way now, and the early return counts a claim as something to say. Your `o.middle.inner.flag` probe reads:

```
  · `Off` at `o.middle.inner.flag` is declared unreachable: the probe never passes Off, and nothing here proves it: nothing was read about this case
  · not read: o.middle.inner (the walk stopped before reaching what is under it)
```

**3 — the order.** Right, and the repository says so itself. `Ordered.map` now, with a test over three behaviors named so that hash order and declaration order differ; it fails with `Map.copyOf` (`[km, ka, kz]` for `[kz, ka, km]`).

**4 — the guard.** Taken as a class boundary, as you suggest: `Coverages` counts what a row is owed with no claim in scope, `ClaimReport` puts what was claimed beside those numbers afterwards, and `Adequacy` names neither — which let the now-redundant `Claimed` query go, since the judging travels with the body check. The tripwire watches the two files that count a denominator inside the package that also reports, which is the gap you found.

Tests 7860, no failures; souther-examples byte-identical over all 14 modules.

kawasima:
All four in 01544ff8. The first two turned out to be one root.

**Duplicate `excluded`, and a decoration that could rebuild the measure.** `excluded` was built per claim *occurrence*, so two arms declaring one case counted it twice; and `ClaimReport` took a measured `PartitionEvidence` and handed back another, so filtering `axis.classes()` by a claim compiled, ran and passed the tripwire. Both are the same mistake: what a measure counts and what a body said about it were one value.

They are a product now, as you suggest. `PartitionEvidence` is the measure and cannot carry a claim at all — the three claim-shaped records are gone from it — and `ClaimAnnotations` is what the body declared, one entry per case of a position with every arm's words kept. The report holds both and joins them where it is written. Nothing is handed a measure to return one, so there is nothing to rebuild.

Your probe now reads `excluded 1` with both arms' reasons folded (`declared unreachable on every path`), and it is a test.

The claims reach a report through one projection (`Bodies.Claimed` turns the judging into words) and never as verdicts a measure could act on. The tripwire holds that `souther.compiler.claims` is named by the check that judges and by that projection, and by nothing else.

**Boundary tests.** Both added: a `match` written as a `let`'s value is refused with E1326, and a claim about a position no axis carries appears in `AdequacyReport.human()`. The claim assertions that were reading evidence fields now read the report, which is where a reader meets them.

**Stale javadoc.** All three, plus a `Generator` reason that had been left with two doc comments.

Tests 7862, no failures; souther-examples byte-identical over all 14 modules.

kawasima:
All three in 8739cf61.

**The count.** Right, and the fix is the framing you gave: `excluded` is a number taken out of the denominator printed beside it, so it is counted over the positions that line counts. A claim about a position past the axis limit is in neither `classes` nor `covered` and is no longer in `excluded`; it is still said further down under its own name.

**The test.** Also right — `contains("excluded 1")` passed on the signature's own line. It reads the partition line whole now, with a control beside it: a confirmed claim at a position past the limit, where the line must carry no `excluded`.

Writing that control was worth more than the assertion. My first fixture used the `one /= two` model, whose claim is *unproven* and so never reaches this count either way: the control was green against the bug. It proves something only with a model whose rules refuse the case (`Active` at a position past the limit), which is what it uses now — and the same mutation that produced your output fails it.

**Documentation.** All three: the two javadocs on `BehaviorReport` with the first detached from the record, the description that outlived the reason it was about, and `Generator`'s account of the four things it says now that one of them is gone.

Tests 7863, no failures; souther-examples byte-identical over all 14 modules.

PR #806, opened 2026-08-17
Answer what a report found about a position in one place

### LINKED ISSUE #791: A rule left unread at a position the report measures is said by one surface and not the other
A report can measure a position and, in the same breath, say that a rule about it went unread — and
only the human surface says the second part.

    module g

    data Prospecting
    data Qualified
    data Won
    data Stage = Prospecting | Qualified | Won

    data StageI = Stage
        invariant past = value > Won

    data Accepted = { at: String }

    behavior classify : (s: StageI) -> Accepted

Human:

    classify                 injected      rows 0    pending 0
      partition   axes 1   single-axis 0/0   (1 not measured: no row names this behavior)
        · not read: s (a rule about it is one this compiler did not read)

Machine, the same run:

    notDerivable: []   notRead: []
    axis classify/s    classes ["Prospecting", "Qualified", "Won"]

Measured with the CLI built at `a0172ea0`.

## Where the two come from

`Adequacy` builds the human line out of two lists: `partition.unread()`, which is every rule this
compiler could not turn into a line wherever it was written, and the `CannotDerive` arm of
`partition.notDerivable()`. `AdequacyReport` writes only the second, split into `notDerivable` and
`notRead`.

So a rule recorded against a position the axes went on to measure reaches one surface and stops
there. Nothing is wrong with either list; what is wrong is that one question — what did this report
find about this position — is answered by reading two of them in one place and one of them in the
other.

The human list also runs two things together under one bullet. A position with no axis and a
position with an axis whose rules were not all read are different things to act on, and they are
printed alike.

## The half nothing writes at all

The same question has an answer neither surface carries.

    data Code = String
        invariant enumerated = value == "a" || value == "b"
        invariant shape = String.matches("[a-z]", value)

    behavior classify : (c: Code) -> Accepted

    partition   axes 1   single-axis 0/0
    axis classify/c    classes ["\"a\"", "\"b\""]

Two classes, and nothing said. The values reading took the first clause in and not the second, so
the classes are the values the model singled out and the second rule may yet refuse one of them —
a row is offered at `"b"` on the strength of a reading that did not run to the end. No `unread`
entry is recorded here, because the clause states no end and the reader that would have recorded
one is about ends.

The material is already there. Since #788 an axis carries how much of what its position's rules say
was read (`Axis.read`, `AdmissibleSet.Completeness`), deliberately without a consumer — the note in
that PR says so. This is the reader it was kept for.

## Why it matters

A consumer keyed on the document cannot see what a reader of the human report sees, and the two
disagree about the same run. Whichever is right, they cannot both be: `souther examples --strict`
and a person reading the same output are being told different things about the same position.

And the account a person does get is the coarser of the two. "A rule about it was not read" of a
position with three classes reads as though the position were unmeasured, which the numbers beside
it deny.

## Acceptance

1. One reading answers what a report found about a position, and both surfaces write from it. A
   position named by one and unnamed by the other is what this is about.
2. A position the axes measure whose rules were not read in full is said as that, and told apart
   from a position with no axis at all.
3. The machine document carries it, so a consumer sees what a reader of the human report sees.
4. What an axis already carries is what answers it, rather than a second derivation of how much was
   read.
5. Nothing is invented: a position whose rules were read in full gains no line, in either surface.

kawasima:
Closes #791.

## What it measures

A report can measure a position and, in the same breath, say that a rule about
it went unread — and only the human surface says the second part. Over the
example corpus, of the positions a rule was left unread at:

| | |
| --- | --- |
| named by both surfaces | 55 |
| **named by the human line and absent from the document** | **64** |

The other half nothing wrote at all. Of the 244 axes a report shows, **85 rest
on a reading that did not run to the end** — 62 stopped where the walk does not
enter what the position holds, 19 at a rule form nothing read, 4 at a rule
relating two positions — and neither surface said so.

```souther
data Assignee = String

data Issue = { assignee: Assignee? }
behavior classify : (i: Issue) -> Accepted
```

`i.assignee` is offered as `None` and `Some`. What `Assignee` says about the
value inside is behind the option and was not read. Nothing said that.

## Why

`Adequacy` built the human line out of two lists — every rule this compiler
could not turn into a line, and the `CannotDerive` arm of the positions it
divided no way — and `AdequacyReport` wrote only the second. One question,
answered by reading two of them in one place and one of them in the other.

And what an axis carries about how far its rules were read has had no consumer
since it was measured. That is the half nothing wrote.

## The shape

- `PartitionEvidence.notRead()` is the one reading, and both surfaces write from
  it. The merge and its deduplication live there rather than in each surface.
- `AxisCoverage.Reading` is `InFull | InPart(reason)`, taken from what the axis
  already carries. Not a second account worked out beside the lists: those
  answer about rules, and this answers about a position.
- `PARTITION_READ_IN_PART` is its own finding kind, so a measured position is
  said apart from one with no axis. The sealed switches over the kinds made
  every reader say what it does with it.
- `BlockReason.of(UnreadReason)` is the one projection between the two
  vocabularies, so a reader holding a reading's completeness and one holding a
  block reason cannot come to different words for one stop.

In the document, `status` and `reason` say whether a number was arrived at and
why there is none. How far a reading got is a different question about a
position that *has* numbers, so it is `extent` and `stoppedBy` — a test that
holds every `status` node to the measure vocabulary is what caught the overload.
The words for what stopped a reading are written once (`$defs/notReadReason`)
and referred to twice.

## What was left alone

Whether the unread remainder could refuse one of the classes offered is not
judged. `Completeness` says how far a reading got; whether what it missed bears
on the answer is a different question at a different altitude, and folding it in
would turn a `complete | partial` observation into a semantic verdict. If it is
ever wanted it is an orthogonal axis, not a wider enum.

The reasons are published as they are. `NOT_REACHED` was measured before this
was written, to check it is not a reader's control flow reported as a fact about
the model: of the stops that fire in the corpus, all are positions whose
contents sit behind a shape the walk does not enter, and every one is guarded by
"is there any rule under what is being left". No depth limit and no recursion
guard fires. It is an honest reading.

## Measured

- `CI=1 mvn -Plint verify` green: syntax 114, formatter 2233, compiler 5115,
  build driver 10, LSP 255, CLI 340, benchmarks 7 and 13.
- The conformance snapshot moved by the `read` field alone (+20/−3 in one
  document).
- `crm` gains 40 `read in part` lines beside its 232 existing `not read` lines;
  the corpus total is the 85 above.

### DIFF
--- a/souther-compiler/src/main/java/souther/compiler/inputs/BlockReason.java
+++ b/souther-compiler/src/main/java/souther/compiler/inputs/BlockReason.java
@@ -22,6 +22,24 @@ package souther.compiler.inputs;
 public sealed interface BlockReason {
 
     /**
+     * What a value reading's account of what it left unread comes to here.
+     *
+     * <p>The one place the two vocabularies meet, so a reader holding a reading's completeness and
+     * one holding a block reason cannot come to different words for one stop. A relation between
+     * two positions is what {@link ComparisonBetweenPositions} already says, whichever rule wrote
+     * it: a {@code guard} comparing two inputs and an {@code invariant} relating two fields leave a
+     * reader the same thing to know. The other two are their own, because what would lift each is
+     * different work — one wants a reader for a form, and one wants the gathering to reach further.
+     */
+    static BlockReason of(souther.compiler.values.UnreadReason why) {
+        return switch (why) {
+            case RELATES_TWO_POSITIONS -> new ComparisonBetweenPositions();
+            case FORM_NOT_READ, ALTERNATIVE_NOT_READ -> new UnreadValueRule();
+            case NOT_REACHED -> new ValueRulesNotReached();
+        };
+    }
+
+    /**
      * The type at the position could not be interpreted: a name denoting no declaration, or a
      * declaration reachable from itself. Such a model compiles, so this is a position a report is
      * asked about and cannot be answered for.
--- a/souther-compiler/src/main/java/souther/compiler/inputs/Crossing.java
+++ b/souther-compiler/src/main/java/souther/compiler/inputs/Crossing.java
@@ -67,11 +67,7 @@ final class Crossing {
      * and one wants the gathering to reach further.
      */
     static BlockReason stopped(souther.compiler.values.UnreadReason why) {
-        return switch (why) {
-            case RELATES_TWO_POSITIONS -> new BlockReason.ComparisonBetweenPositions();
-            case FORM_NOT_READ, ALTERNATIVE_NOT_READ -> new BlockReason.UnreadValueRule();
-            case NOT_REACHED -> new BlockReason.ValueRulesNotReached();
-        };
+        return BlockReason.of(why);
     }
 
     /**
--- a/souther-compiler/src/main/java/souther/compiler/partition/ReportedReason.java
+++ b/souther-compiler/src/main/java/souther/compiler/partition/ReportedReason.java
@@ -34,11 +34,14 @@ public final class ReportedReason {
                     UndividedPosition.Reason.UNSUPPORTED_TRAVERSAL;
             case BlockReason.UnreadComparisonForm _ ->
                     UndividedPosition.Reason.UNSUPPORTED_SYNTAX;
-            // Both of these are a rule this could not read, which is the one thing a reader of a
-            // document is being told. Which reader of the clause gave up, and whether it gave up on
-            // the form or never arrived at the clause, are this compiler's own business.
-            case BlockReason.UnreadValueRule _, BlockReason.ValueRulesNotReached _ ->
-                    UndividedPosition.Reason.UNSUPPORTED_SYNTAX;
+            case BlockReason.UnreadValueRule _ -> UndividedPosition.Reason.UNSUPPORTED_SYNTAX;
+            // Its own word, and not the one above. Both are rules this reading did not turn into a
+            // line, and a reader acting on them is doing different work: one wants a reader for a
+            // form that was seen, and one wants the gathering to reach the rules at all. Collapsed
+            // together, a position whose rules nothing had looked at was reported as an expression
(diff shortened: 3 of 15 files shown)


kawasima:
Pushed `e4ef2246` for the review.

**Blocking — the wrong cause.** `NOT_REACHED` now has its own published word, `rules_not_read_at_all`, rather than borrowing `unsupported_syntax`. The split is at `ReportedReason`: `UnreadValueRule` keeps `unsupported_syntax` (a rule that was read and could not be used) and `ValueRulesNotReached` gets the new one. I did not map it to `unsupported_traversal` — which limit stopped the walk is not recorded and is not promised, so naming the shape would be reading this month's corpus back as a rule.

`UNSUPPORTED_SYNTAX`'s own javadoc claimed the third case ("a rule this never reached at all is a third"); that sentence is gone, so the enum and the schema now say the same thing.

Measured: the option case answers `rules_not_read_at_all`, and the guard-over-helpers case in the catalog corpus still answers `unsupported_syntax` — the snapshot's existing word did not move.

**The test held presence, not value.** It asserts `"rules_not_read_at_all"` now, and the human line against its sentence. You are right that this is how the mis-projection passed.

**`notRead`'s contract.** "Nothing is established about these either way" was true of the list it used to be written from and is false of this one. The schema now says what separates it from `notDerivable` is that the model does state something here, and that it is not a statement the position was unmeasured — whether the reading of a position ran to the end is what that axis's `read` says.

**The javadoc's quantifier.** `notRead()` said "every position this reading could not finish reading the rules about", which is the misuse the change exists to stop. It now says what it is — the positions this reading carries an unread *finding* about — and names `AxisCoverage.read()` as the account of a position.

**Formatting churn.** Reverted. The schema is develop's file with the change applied by hand: +17/−6 rather than +484/−111.

PR #813, opened 2026-08-17
Ask a departure of the declaration it stands in, and ask once

### LINKED ISSUE #810: A departure in one line is asked of the whole module it was found in, and asked three times
`souther-fmt` takes 66.3 seconds over 2224 tests, and 58 of those seconds are in four classes:

    30.32 s   498   EveryDeviationStatesADifferenceTest
    17.99 s   495   EveryDepartureFromTheCanonicalFormIsSomeRulesTest
     6.15 s    20   RepairingWhatTheRulesSayWritesTheCanonicalFormTest
     3.67 s     3   AProbeSurvivesAtEveryTokenBoundaryTest

It is not the formatter. The whole corpus — 25 sources, 1023 lines, 52.6KB — parses and writes its
canonical form in 48ms, and writing one is flat in the size of the source, 0.78 ms/KB at 14KB and
0.90 at 116KB. What the sweep spends is spent asking about text it did not need to ask about.

## A candidate is a whole module, and the question is one line

`EveryDepartureFromTheCanonicalFormIsSomeRulesTest#sweep` finds 493 departures. Each is a canonical
form with one line of it written differently, and each is carried as the whole module that line was
found in — 2,445,105 characters of text for 493 questions about single lines. 129 of them are 14KB
or larger, and `Deviations.of` costs 75ms on one of those, so those 129 alone are 9.7 of the 12.1
seconds one pass over the sweep takes. Answering about the line means writing the canonical form of
the four hundred lines around it and holding every family against all of them, twice — the loop
reaches its fixpoint in exactly two rounds for 492 of the 493, which is the design and is not the
cost.

The smaller text already exists. `Declarations` cuts a canonical form into the declarations it is
made of and records, for each, whether that declaration alone writes the same text it stands as in
the whole; `written(i)` then builds the candidate inside its declaration. All of it is used to
refuse a candidate and none of it is kept. Carrying it instead:

    493 departures, 354 cut to their declaration (71%)
    text:  2,445,105 -> 342,912 characters   (14.0%)
    time:      11,814 ms -> 967 ms           (12.2x)

What decides whether something is a departure should stay where it is. The sweep asks the whole
file, dedupes by `kind of shape` read from the whole file's layout, and writes what it found to
`every-departure-there-is.txt`. If only the text a property is asked of gets smaller, those 493
recorded pairs do not move, and that they do not move is what says the cut dropped nothing. The
declaration a candidate is cut to is one whose canonical form was checked, so it is a source in its
own right and the property is asked of it as of any other.

The 29% that cannot be cut are the ones the cut cannot answer for — a line in the header, a
declaration that does not stand on its own, and the two ways of writing a line that reach the
declaration next door, which `reachesTheNextOne` already names. They stay whole.

## The same report is written three times

`Deviations.of` is a function of its text, and the sweep's 493 reports are written once for
`aDepartureIsNamedByARule`, once for `aDeviationQuotesTwoDifferentAnswers`, and a third time for
`andTheSweepReachesBothOfTheGroupRules`, which walks every departure again only to count that two
rules were reached. Three passes at 12.1 seconds each is what those two classes are.

The sweep is already held once in a static, for the same reason. The reports should be too: what
the sweep hands out is a departure and what the rules said about it, so a check reads an answer
rather than asking for it again. That also removes a pass the sweep pays before any check runs — it
parses and writes the canonical form of every candidate to decide it is one, and `Deviations.of`
then parses and writes it again.

## What this is worth

The sweep itself is 3.4 seconds and stays: what admits a candidate is the whole file, and that is
the sentence the oracle rests on. Past it, one pass over the departures goes from 12.1 seconds to
about 1, and there is one pass instead of three.

`Deviations.of` is also superlinear in the size of its text, which is why the big candidates cost
what they do. That is a defect in the rules rather than in this sweep and is filed on its own as #809; it
makes the text that is left cheaper, and this issue does not wait on it.

kawasima:
#810. The sweep's 500 departures were each carried as the whole module the line was found in, and
the reports over them were written three times.

## A candidate is asked of the declaration it stands in

`Declarations` already cut a canonical form into the declarations it is made of, checked that each
writes the same text alone that it stands as in the whole, and built the candidate inside its own
declaration — all of it to refuse a candidate, and none of it kept.

`admits` returned a boolean where there are three answers. It is `says` now:

- `Refused` — the declaration is not written that way either, so the file is not. The candidate goes.
- `Answered(text, canonical)` — the same question asked of a seventh of the text, and it is what the
  departure carries.
- `NotAsked` — the header, a declaration that does not stand on its own, and the two ways of writing
  a line that reach the declaration next door. The whole file answers both.

Folded back into a yes and a no, `NotAsked` would either drop a candidate the whole file admits or
carry a text nothing here checked the canonical form of.

What admits a candidate has not moved: the sweep asks the whole file, dedupes by the `kind of shape`
read from the whole file's layout, and writes what it found to `every-departure-there-is.txt`. Those
500 pairs are the same 500 after this, which is what says the cut dropped nothing.

    500 departures, 357 cut to their declaration (71%)
    text: 2,445,251 -> 345,778 characters (14%)

## One report instead of three

`Deviations.of` is a function of its text, and the sweep's reports were written once for
`aDepartureIsNamedByARule`, once for `aDeviationQuotesTwoDifferentAnswers`, and a third time for
`andTheSweepReachesBothOfTheGroupRules`, which walked every departure again only to count that two
rules were reached. The sweep is already held once in a static; the reports are held with it, and a
check reads an answer rather than asking for it again.

`Deviations.of` also wrote the canonical form of its own source twice — once to hold the report
against at the end, once as the first round's. It writes it once, and each round after a repair
writes the one for the text that repair produced.

## Checked

    mvn -pl souther-fmt test

    EveryDepartureFromTheCanonicalFormIsSomeRules   18.55 s -> 5.25 s   502 tests
    EveryDeviationStatesADifference                 31.33 s -> 5.24 s   505 tests
    RepairingWhatTheRulesSayWritesTheCanonicalForm   5.43 s -> 5.39 s    23 tests
    the module                                      34.5 s  -> 7.9 s   2247 tests

`Repairing` does not move. Its cost is not the second write of a source it already had, so the one
removal that could have reached it does not.

`CI=1 mvn -Plint clean verify` — BUILD SUCCESS, 1:31.

`AProbeSurvivesAtEveryTokenBoundaryTest` (3.67 s over 3 tests) is in the issue's table and not in
what it asks for; it is untouched.

Closes #810

### DIFF
--- a/souther-fmt/src/main/java/souther/compiler/fmt/Deviations.java
+++ b/souther-fmt/src/main/java/souther/compiler/fmt/Deviations.java
@@ -1,7 +1,6 @@
 package souther.compiler.fmt;
 
 import souther.compiler.cst.CstParser;
-import souther.compiler.cst.SyntaxNode;
 
 import java.util.ArrayList;
 import java.util.Comparator;
@@ -56,14 +55,16 @@ public final class Deviations {
      * that one leaves nine more.
      */
     public static Report of(String source) {
-        SyntaxNode root = CstParser.parse(source).root();
-        String canonical = Formatter.canonicalize(root).text();
+        // The canonical form the report is held against at the end is the first round's own, and
+        // each round after a repair writes the one for the text that repair produced. Asked for
+        // separately, the first two are the same text written twice from the same source.
+        Formatter.CanonicalForm form = Formatter.canonicalize(CstParser.parse(source).root());
+        String canonical = form.text();
         List<Deviation> out = new ArrayList<>();
         Set<String> seen = new LinkedHashSet<>();
         List<List<Repair.Edit>> repaired = new ArrayList<>();
         String text = source;
         for (int round = 0; round < 8; round++) {
-            Formatter.CanonicalForm form = Formatter.canonicalize(CstParser.parse(text).root());
             Witnesses.Pairing pairing = new Witnesses.Pairing(text, form);
             List<Witness> witnesses = all(text, form, pairing);
             for (Witness w : witnesses) {
@@ -96,6 +97,7 @@ public final class Deviations {
             }
             repaired.add(edits);
             text = next;
+            form = Formatter.canonicalize(CstParser.parse(text).root());
         }
         return new Report(sorted(out), text.equals(canonical));
     }
--- a/souther-fmt/src/main/java/souther/compiler/fmt/Deviations.java
+++ b/souther-fmt/src/main/java/souther/compiler/fmt/Deviations.java
@@ -28,6 +28,10 @@ public final class Deviations {
     /** One decision a source did not take, and where in it that shows. */
     public record Deviation(int line, int column, String rule, String canonical, String source) {}
 
+    /** How many times the rules are asked before a report gives up on reaching the canonical
+     * form. */
+    private static final int ROUNDS = 8;
+
     /**
      * What a source has against it, and whether that is all of it.
      *
@@ -64,7 +68,7 @@ public final class Deviations {
         Set<String> seen = new LinkedHashSet<>();
         List<List<Repair.Edit>> repaired = new ArrayList<>();
         String text = source;
-        for (int round = 0; round < 8; round++) {
+        for (int round = 0; round < ROUNDS; round++) {
             Witnesses.Pairing pairing = new Witnesses.Pairing(text, form);
             List<Witness> witnesses = all(text, form, pairing);
             for (Witness w : witnesses) {
@@ -97,6 +101,12 @@ public final class Deviations {
             }
             repaired.add(edits);
             text = next;
+            if (round + 1 == ROUNDS) {
+                // No round is left to ask about this text. Writing its canonical form here would
+                // be a pass nothing reads, and this is where the source can be one the formatter
+                // refuses — so it is not written at all.
+                break;
+            }
             form = Formatter.canonicalize(CstParser.parse(text).root());
         }
         return new Report(sorted(out), text.equals(canonical));
--- a/souther-fmt/src/test/java/souther/compiler/fmt/EveryDepartureFromTheCanonicalFormIsSomeRulesTest.java
+++ b/souther-fmt/src/test/java/souther/compiler/fmt/EveryDepartureFromTheCanonicalFormIsSomeRulesTest.java
@@ -47,8 +47,8 @@ import static org.junit.jupiter.api.Assertions.assertTrue;
  *
  * <p>What admits a candidate is the whole file it was found in, and what it is then asked about is
  * the declaration that line stands in, where that declaration can answer for it. Both are a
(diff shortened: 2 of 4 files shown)


kawasima:
Follow-up in e069a863: both review findings, and three more measured cuts.

## Where the time still was

Instrumented, over one sweep (ms): whole-file `format` 2816, reports 860, the declaration check 599,
whole-file `parse` 150, everything else 207. The 918 whole-file formats split into 457 where a
declaration had answered and the whole file decides, **310 where nothing could answer**, and 151 in
files that are one declaration.

## The header, and the sources that have none

A header line had no answerer, so every candidate in one went to the whole file. The first
declaration now carries the header — a header line is asked in front of the same text it has in
front of it in the whole file, and the first declaration is the only one whose source would
otherwise write it twice.

And an empty header still wrote the newline after it, so every declaration of a source that opens
with a declaration was reconstructed with a blank line in front. Thirteen declarations were recorded
as ones the formatter writes differently — which is true of a text beginning with a blank line — and
every candidate in them went to the whole file.

    918 candidates at the whole file -> 790
    cut to their declaration: 357/500 -> 423/500
    carried: 345,778 -> 248,118 characters

The 500 recorded pairs are unchanged, which is what says neither cut dropped a candidate.

## The same report, twice, in the third class

`RepairingWhatTheRulesSayWritesTheCanonicalForm` wrote a report for every departing source twice —
once to hold against the formatter, once to count what it said. It holds them now, as the sweep
does. What decides that a source departs is still `Formatter.format`, not the report: a source the
rules say nothing about is what that check exists to catch, so it is not derived from what they said.

## The review's two

`of` wrote the canonical form of the text the eighth round produced, which no round reads: a wasted
pass on every source that does not converge, and a throw site — a report that had run out of rounds
threw where it used to return what it had. It stops before writing it.

Taking the reports in the sweep does mean a candidate the rules cannot answer at all ends the sweep
rather than one parameterized test. The row is named where the report is taken, so the failure says
which candidate it was.

## Measured

    mvn -pl souther-fmt test                         at #813    now

    EveryDepartureFromTheCanonicalFormIsSomeRules      5.25 s   4.59 s
    EveryDeviationStatesADifference                    5.24 s   5.28 s
    RepairingWhatTheRulesSayWritesTheCanonicalForm     5.39 s   3.51 s
    the module                                         7.9 s    7.0 s    2247 tests

`CI=1 mvn -Plint clean verify` — BUILD SUCCESS, 1:31.

What is left is the oracle's own: 790 whole-file formats, because what admits a candidate is the
whole file. Formatting is ~0.9 ms/KB and ~20x the parse of the same text, so a faster floor than
this is the formatter's own work, not the sweep's.

PR #815, opened 2026-08-17
Ask a unit's question of that unit, not of the file

### LINKED ISSUE #814: A group's opportunities are found by reading every opportunity the layout has, and found twice
A group's opportunities are found by reading every opportunity the layout has, and found twice.

`Repair.opportunitiesOf` is asked which of a group's opportunities the source settled differently. It answers by reading the whole layout:

    for (Opportunity o : canonical.layout().opportunities()) {
        if (o.settledBy() != group
                || Witnesses.brokeInSource(source, pairing, o.at()) == o.broke()) {
            continue;
        }

A layout has an opportunity at every place any group could break, so this is the whole file for one group's question. Both `Witness.Conditional` and `Witness.RunTogether` are projected through it, once per witness, which makes a report `Θ(G·O)` in the groups a source departs at and the opportunities the file has.

And each of those is asked twice. `Repair.where` is how `Deviations.of` finds where a witness lands, and it evaluates the whole projection to keep one number:

    static int where(String source, Formatter.CanonicalForm canonical, Witnesses.Pairing pairing,
            Witness w) {
        List<Edit> edits = of(source, canonical, pairing, w);
        return edits.isEmpty() ? -1 : edits.get(0).from();
    }

`Deviations.of` calls it for every witness of a round and then calls `Repair.edits`, which calls `of` for every witness again. So a group's opportunities are swept twice per round, and what the first sweep built is thrown away.

`Witnesses.lines` is the same shape a level down: `Repair.indentation` asks for the canonical form's lines once per indentation witness, and that walk is over every break the layout has.

## Where this stands against the two issues beside it

Not #809, which was about the token count inside each of these — `brokeInSource` used to walk the run for each opportunity, and #811 took that factor out. What is left is the group-by-opportunity product, which #811 does not touch and does not claim to.

Not #810 either. That one is about the sweep in `EveryDepartureFromTheCanonicalFormIsSomeRulesTest` evaluating whole-module candidates repeatedly, which is a test's question about which text to ask about. This is the projection itself.

## What is known about how much it costs

Little, which is part of why it is filed rather than fixed. Over a source with one departure per declaration — five to forty deviations against a few thousand opportunities — a report is flat: 4.21, 3.45, 3.74 and 3.68 ms/KB at 14K, 29K, 58K and 116K characters. `G` is small there and the product does not show.

What would show it is a source departing at a large fraction of its groups, and no such input was built from the corpus: collapsing whole declarations onto one line does not parse. So the product is read off the code rather than measured, and a first step here is an input that exercises it — without one, a fix cannot be shown to be worth anything.

## What the answer probably is

The opportunities grouped by the group that settles them, once per layout rather than once per witness, which is the same move #811 made for the token runs: the layout already knows, and every witness is asking it again.

Whether `where` should evaluate the projection at all is a second question. What it wants is where a witness lands, and building every edit to read the first one's offset is the projection doing work for an answer that is not the projection's.

kawasima:
Closes #814.

Two of the projections a report runs read the whole file to answer about one unit,
so a source that departs at a large fraction of its units cost the units multiplied
by the file. The issue reads that product off the code and says an input that
exercises it is the first step, since the corpus has none — real code departs at a
handful of its units however long it is.

## The input

It does not need a source collapsed onto one line, which is what the issue found
does not parse. A source departs from a group's decision as much by breaking where
the canonical form writes one line as by the other way about:

```
let f0 (a: Int): Int = g(
    a,
    a
)
```

which the canonical form writes as `let f0 (a: Int): Int = g(a, a)`. For a level, a
record type written down the page exactly where the canonical form writes it and
indented two columns instead of four departs at its level and at nothing else. Both
hold the canonical form still and move only how many units departed.

`souther-bench` grows a `report` measurement from that: two shapes at four sizes,
the departure density held at every unit, and the same shape with one departure
beside each as the control. The distance between the pair is the figure, in ms/KB —
a report that is slow because the source is long is slow in both lines.

With the sweep still in place, the dense line doubles its ms/KB with the size and
the control is flat:

```
REPORT n=200  level, every    24299 chars     78.6 ms ( 3.314 ms/KB)
REPORT n=200  level, one      25891 chars     16.9 ms ( 0.670 ms/KB)
REPORT n=400  level, every    48699 chars    304.5 ms ( 6.403 ms/KB)
REPORT n=400  level, one      51891 chars     29.2 ms ( 0.576 ms/KB)
REPORT n=800  level, every    97499 chars   1361.0 ms (14.294 ms/KB)
REPORT n=800  level, one     103891 chars     59.5 ms ( 0.586 ms/KB)
REPORT n=1600 level, every   195699 chars   6403.2 ms (33.505 ms/KB)
REPORT n=1600 level, one     208491 chars    126.0 ms ( 0.619 ms/KB)
```

## The fix

`Repair.Round` gathers each of the two under the unit that is asked about, once for
a round of witnesses. The opportunities go under the group that settles them. The
lines go under every level they are written under and not only the innermost,
because a level that moves takes what is nested inside it along — that is where the
fix was first left half-done: hoisting `Witnesses.lines` out of the loop stopped the
walk being repeated and left the sweep, and the shape stayed quadratic.

`Repair.where` goes with it. It evaluated the whole projection to keep one number,
and `Repair.edits` then evaluated it again to write the text. `edits` is now
`composed(each(...))`; `each` hands back one list per witness and the report reads
the offset off the list of the witness it is about.

| | before | after |
|---|---|---|
| `level, every` n=1600 | 6403.2 ms (33.505 ms/KB) | 220.6 ms (1.154) |
| `level, one` n=1600 | 126.0 ms (0.619) | 126.9 ms (0.623) |
| `group, every` n=3200 | 926.8 ms (6.640) | 554.4 ms (3.971) |
| `group, one` n=3200 | 399.7 ms (3.689) | 419.8 ms (3.874) |

## What holds it

`AWitnessIsProjectedThroughItsOwnUnitAndNotTheWholeFileTest` builds a second `Round`
by sweeping and hands both to the same projection, over the corpus and over three
sources written to depart at every group, at every level, and at a level with levels
under it. Every way of building the gatherings wrongly that it catches — filing a
line under its innermost level alone, gathering every opportunity under one group,
reversing either order — leaves the existing suite green, which is the corpus not
having the shape.

`EveryShapeThatIsTimedStillDepartsTest` holds the benchmark's shapes to being those
shapes: that the dense one departs at every declaration, that the control departs at
one, that neither departs the other's way, that the pair has one canonical form, and
that the report over each is whole.

Reports over the dense shapes are byte-identical before and after — line, column,
rule and both sides.

`souther-fmt` moves from `test` to compile scope in `souther-bench`, which is an
executable measurement tool rather than a test suite.

### DIFF
--- a/souther-bench/pom.xml
+++ b/souther-bench/pom.xml
@@ -37,6 +37,12 @@
             <groupId>net.unit8.raoh</groupId>
             <artifactId>raoh</artifactId>
         </dependency>
+        <!-- Called from here: the report measurement asks the formatter what a source has against
+             its canonical form, which is a cost this module times like any other. -->
+        <dependency>
+            <groupId>org.souther-lang</groupId>
+            <artifactId>souther-fmt</artifactId>
+        </dependency>
         <dependency>
             <groupId>org.junit.jupiter</groupId>
             <artifactId>junit-jupiter</artifactId>
@@ -49,11 +55,6 @@
              so every one of them is named here — saying so as dependencies is what makes `-am`
              build them, and what makes the checks cover what they claim to whichever way the tree
              was built. -->
-        <dependency>
-            <groupId>org.souther-lang</groupId>
-            <artifactId>souther-fmt</artifactId>
-            <scope>test</scope>
-        </dependency>
         <dependency>
             <groupId>org.souther-lang</groupId>
             <artifactId>souther-build-driver</artifactId>
--- a/souther-bench/src/main/java/souther/bench/Bench.java
+++ b/souther-bench/src/main/java/souther/bench/Bench.java
@@ -18,6 +18,7 @@ import java.util.List;
  *   run     what the generated code costs to run, per element
  *   scale   how a whole-workspace compile grows with the number of modules
  *   values  how one module's compile grows with the number of values it declares
+ *   report  how what a source has against its canonical form grows with the source
  * </pre>
  *
  * <p>Every corpus is compiled once and checked before anything is timed. A language change that
@@ -31,7 +32,7 @@ public final class Bench {
     public static void main(String[] args) {
         Report report = new Report(System.out);
         List<String> wanted = args.length == 0
-                ? List.of("cold", "warm", "phase", "edit", "run", "scale", "values")
+                ? List.of("cold", "warm", "phase", "edit", "run", "scale", "values", "report")
                 : new ArrayList<>(List.of(args));
 
         List<Corpus> corpora = Corpus.all();
@@ -78,6 +79,10 @@ public final class Bench {
         }
         if (wanted.contains("values")) {
             Values.measure(report);
+            report.blank();
+        }
+        if (wanted.contains("report")) {
+            Reporting.measure(report);
         }
     }
 }
--- a/souther-bench/src/main/java/souther/bench/Reporting.java
+++ b/souther-bench/src/main/java/souther/bench/Reporting.java
@@ -1,0 +1,131 @@
+package souther.bench;
+
+import souther.compiler.fmt.Deviations;
+
+/**
+ * How a report of what a source has against its canonical form grows with the source.
+ *
+ * <p>A report is asked round by round, and each round projects every rule's answer onto the text.
+ * Two of those projections used to read the whole file to answer about one unit — which of a group's
+ * places the source settled differently was found by sweeping every opportunity the layout has, and
+ * which lines a level writes by walking every line the canonical form has. That costs the units a
+ * source departs at multiplied by the file, and what says whether it is still there is a source that
+ * departs at all of them.
+ *
+ * <p>The corpus does not have one. Real code departs from the canonical form at a handful of its
+ * units however long it is, so the product stays inside the parse and a report over it is flat in
+ * ms/KB whether the projection reads a unit or a file. So the sources here are generated with the
+ * departure density held at every unit, and it is the size that moves — which is the axis the
+ * product is in: a file twice as long has twice the units departing and twice the file to read them
+ * from, and four times the cost if a unit's question is asked of the file.
+ *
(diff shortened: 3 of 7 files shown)


kawasima:
On the eager floor `Round` adds, since the review could not tell it from noise: it
is measurable on its own, and it is about one percent.

Building a `Round` against a source with one departure, beside the whole report over
the same source:

| shape | chars | `Round` | of which `lines` | report | floor |
|---|---|---|---|---|---|
| `group, one` n=400 | 13546 | 1.56 ms | 0.59 | 92.0 ms | 1.7% |
| `group, one` n=1600 | 54946 | 1.57 ms | 1.19 | 175.9 ms | 0.9% |
| `group, one` n=3200 | 110946 | 3.01 ms | 2.17 | 374.1 ms | 0.8% |
| `level, one` n=800 | 103891 | 0.92 ms | 0.81 | 61.5 ms | 1.5% |
| `level, one` n=1600 | 208491 | 2.37 ms | 2.20 | 123.3 ms | 1.9% |

Linear in the file, and most of it is the `Witnesses.lines` walk rather than either
gathering.

So the 399.7 → 419.8 ms on `group, one` at n=3200 is not the floor: the floor is six
times smaller than that difference. At n=1600 the two measured pairs both go the
other way — 195.1 → 180.1 and 188.6 → 181.6 — and the same code measured under two
harnesses gave 374.1 and 419.8 for the same source, which is the spread the +5% sits
inside.

That also bounds the other direction the review mentions. Handing the projection the
`CanonicalLine` list `Witnesses.indentation` already built would save the `lines`
column above and nothing else, so it is worth at most the same one percent — for a
coupling between which witnesses a round found and what its projection reads. Left
alone on those grounds rather than on the grounds that it is small.

Issue #803, opened 2026-08-17
A relation between a behavior's input and its answer can only be written by changing the output's type


kawasima:
ADR-0003 opens by counting three ways the model constrains a domain — a `data`'s
own invariant, a `behavior`'s precondition, and its postcondition — and then
settles two of them. An invariant violation's destination is decided. A
precondition is not a contract: one that can really fail is an ordinary business
case written with `guard ... else`, and one that cannot is reified onto the type
of whatever is then constructed. The postcondition is named in that opening
sentence and never returned to. This is the third one.

What has no home is a relation between what a behavior is given and what it
answers.

    behavior findTodo : (id: TodoId) -> Todo | NotFound

That the `Todo` this answers with is the one that was asked for is the whole
meaning of a lookup, and there is nowhere to write it. Not on `Todo`: a `Todo`
carries its own id and knows nothing of a request, so it cannot say that its id
is the one someone asked for. Not on `NotFound` either, for the same reason. The
relation names a parameter and an answer, and reifying it onto a type needs one
value carrying both sides.

The same gap with a shape instead of an id:

    data Area           = { prefecture: String, district: String }
    data SearchCriteria = { area: Area }
    data Building       = { name: String, area: Area }

    behavior findBuildings : (cond: SearchCriteria) -> List<Building>

Every `Building` this answers with is in `cond.area`, and nothing in the model
says so. The types do not say it: `SearchCriteria.area` and `Building.area` are
both `Area`, but a search over neighbouring districts or along a rail line
answers with buildings whose area is deliberately not the one asked for. Sharing
a nominal type says the two are comparable, not that they are equal. So the
relation cannot be inferred from the declaration and has to be written.

## Where the language sends it today

The check does not cross behaviors, and `invariant-reify` says why: inferring a
relation across a call would be an implicit cross-unit contract, and `depends on`
and an exposed composition's output are always declared rather than inferred.
What it declines is the inference. The consistent way to carry a relation, it
says, is to reify it as a type invariant, which the seeding then assumes
downstream for free.

That answer works when the related values fit in one data. `WithdrawRequest {
balance, amount } invariant amount <= balance` is two fields of one input, and
the behavior that takes it gets the relation without a guard.

It reaches further than it first appears, and where it reaches it should be
used. `businesstrip`'s `事前承認する` guards `承認者ID == 申請.申請者.上長ID` and
answers `事前承認済み`, which carries the applicant through `...出張申請共通項目`
[...]

PR #816, opened 2026-08-17
Say how much of a rule the check reads, and show it where the rule is written

kawasima:
Commit 4 of #803: the classification and the editor become consumers of the contract IR.

## What it does

A data's clause is classified — derivable / exact match / runtime only — and shown to whoever is writing it. A behavior's rule was not, and it is the same question asked of the same expression fragment. Now it is asked once and answered per rule.

- `InvariantChecker.capabilityOf` splits into the data entry it had and a reading over a typed statement plus the names that stand for themselves. Which names those are is what tells the two kinds of clause apart, and it is settled before the classification rather than inside it.
- `check/ContractDischarge` holds the answers under the rule ids they belong to, plus which of the answer's cases no rule names. The unit is the rule and, under a rule, the conjunct: `ensures Found | Missing -> value.id == id` is classified once for `Found` and once for `Missing`, because what `value` is differs between them.
- `query/Bodies.ContractCapabilities` is the key. It reports nothing — whether a clause is well formed was decided by `Contracts`, which owns the contracts and what reading them found.
- The LSP hover on an `ensures` clause lists what the check reads of each of its cases, under the case's own name; the hover on a behavior's name says which of its cases nothing is stated about.
- `HelperInvariants` becomes `ClauseHelpers`: it has expanded both kinds of clause for two commits.

The hover says what is **read**, not what a caller is given. Nothing assumes a rule at a call yet — that is commit 5 — and a hover claiming otherwise would promise a guarantee that is not there.

## The representation, and why the contract is read twice

A rule is classified in the tree the analysis has rules about (`InliningPolicy.DISCHARGE`), as a data's clause is. The contract the emitter runs on is expanded the other way and cannot be read back into this one, so the declaration is read again into the analysis tree — by the same reading, which `BehaviorChecker` now splits from the holding. The rules that come out carry the same `RuleId`s, because an id is a coordinate in the declaration and both readings walk the same declaration. Nothing matches the two lists up afterwards.

That is the one deviation from the plan, which had a `contract.forDischarge()` view. A view over the executable contract cannot recover an operation that its expansion already turned into a fold.

`RuleId`'s comment claimed its ids were not counted off a walk, and `arm` is. It now says what the coordinates are and what moves them; `(clause, selector)` cannot be the identity, because two arms of one clause may name the same case.

## Verification

Full reactor green under `CI=1 mvn -Plint` — syntax 114, fmt 2251, compiler 5184, build driver 10, lsp 264, cli 340, runtime 113, bench 13, 0 failures.

New: 7 tests on the query (one answer per case, per conjunct, the clause's name on each, the cases nothing is said about, a behavior that states nothing) and 6 on the hover (both cases of one arrow shown, arms read to different depths shown apart, the unstated case on the behavior, and hovering elsewhere unchanged).

### DIFF
--- a/souther-compiler/src/main/java/souther/compiler/check/BehaviorChecker.java
+++ b/souther-compiler/src/main/java/souther/compiler/check/BehaviorChecker.java
@@ -47,8 +47,11 @@ public final class BehaviorChecker {
      */
     public static BehaviorContract contractOf(Hir.SpecBehavior behavior, String module, Sig sig,
                                        Symbols symbols, Map<String, Type> helpers) {
-        List<Diagnostic> found = new ArrayList<>();
-        BehaviorContract contract = read(behavior, module, sig, symbols, found);
+        Reading reading = read(behavior, module, sig, symbols);
+        BehaviorContract contract = reading.contract();
+        // The rules it did read, held to what a rule has to be. Two mistakes in one declaration are
+        // two things for an author to fix, and this is the reading that reports them.
+        List<Diagnostic> found = new ArrayList<>(reading.unread());
         for (Rule rule : contract.rules()) {
             collect(found, () -> checkRule(behavior, contract, rule, helpers, symbols));
         }
@@ -83,17 +86,38 @@ public final class BehaviorChecker {
      */
     public static BehaviorContract contractAsRead(Hir.SpecBehavior behavior, String module, Sig sig,
                                                   Symbols symbols) {
-        List<Diagnostic> unread = new ArrayList<>();
-        BehaviorContract contract = read(behavior, module, sig, symbols, unread);
-        if (!unread.isEmpty()) {
-            throw CompileException.of(unread.get(0));
-        }
-        return contract;
+        return read(behavior, module, sig, symbols).whole();
     }
 
-    /** The declaration as rules, with what could not be read of it collected into {@code found}. */
-    private static BehaviorContract read(Hir.SpecBehavior behavior, String module, Sig sig,
-                                         Symbols symbols, List<Diagnostic> found) {
+    /**
+     * What reading a declaration came to: the rules it made, and what it could not read.
+     *
+     * <p>The two are one value because the contract alone does not say it is short of anything. A
+     * declaration whose arm could not be read makes a contract that states nothing about the cases
+     * that arm named, which is indistinguishable from a declaration that said nothing about them —
+     * and a reader deriving what is left unstated would report the opposite of what the author wrote.
+     * So what was lost arrives with what was kept, and a caller says which it is asking for.
+     */
+    private record Reading(BehaviorContract contract, List<Diagnostic> unread) {
+
+        /**
+         * The contract where the whole declaration was read.
+         *
+         * @throws CompileException where any of it was not, which is reported by the reading that
+         *     holds the declaration to its rules rather than said twice
+         */
+        BehaviorContract whole() {
+            if (!unread.isEmpty()) {
+                throw CompileException.of(unread.get(0));
+            }
+            return contract;
+        }
+    }
+
+    /** The declaration as rules, beside what could not be read of it. */
+    private static Reading read(Hir.SpecBehavior behavior, String module, Sig sig,
+                                Symbols symbols) {
+        List<Diagnostic> found = new ArrayList<>();
         ValueName.Behavior name = new ValueName.Behavior(module, behavior.name());
         if (sig == null) {
             // The signature could not be made, which is reported where that failed. A rule is
@@ -138,7 +162,7 @@ public final class BehaviorChecker {
             }
             clauses.add(new Clause(written.name(), rules, written.pos(), written.region()));
         }
-        return new BehaviorContract(name, params, sig.outputType(), clauses);
+        return new Reading(new BehaviorContract(name, params, sig.outputType(), clauses), found);
     }
 
     /**
--- a/souther-compiler/src/main/java/souther/compiler/check/BehaviorChecker.java
+++ b/souther-compiler/src/main/java/souther/compiler/check/BehaviorChecker.java
@@ -47,6 +47,43 @@ public final class BehaviorChecker {
      */
     public static BehaviorContract contractOf(Hir.SpecBehavior behavior, String module, Sig sig,
                                        Symbols symbols, Map<String, Type> helpers) {
+        List<Diagnostic> found = new ArrayList<>();
+        BehaviorContract contract = read(behavior, module, sig, symbols, found);
+        for (Rule rule : contract.rules()) {
+            collect(found, () -> checkRule(behavior, contract, rule, helpers, symbols));
+        }
+        if (found.size() == 1) {
+            throw CompileException.of(found.get(0));
+        }
+        if (!found.isEmpty()) {
+            throw CompileException.ofAll(found, DiagnosticRenderer.legacyBody(found.get(0)));
+        }
+        return contract;
+    }
+
+    /**
+     * The same declaration read again, in whatever representation {@code behavior} is written in, and
+     * held to nothing.
+     *
+     * <p>What the rules are — which case each applies to, what {@code value} is there, which rule is
+     * which — is a property of the declaration and not of the representation, so this reads the same
+     * rules under the same {@link RuleId}s as {@link #contractOf} does of the tree that runs. What is
+     * not done again is the holding: a clause that is not well formed is one mistake, and the reading
+     * that reports it is the one the author's build is made of. An arm this cannot read contributes
+     * no rule, as it does there.
+     *
+     * @throws Unanswerable where there is no signature to read the rules against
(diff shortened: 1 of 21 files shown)


kawasima:
Fixed all three findings, plus the `RuleId` wording. Pushed as e671992c.

**1. Positions and conjuncts came from the expanded tree.** Confirmed before fixing: `ensures positive(value.rank) && id.value > 0` on line 11 was answered at `3:31` — inside the `let`. The hover kept nothing, since it keeps the rules whose position falls in the clause the cursor is in, so a rule written through a helper showed no classification at all. Splitting after expanding was the same defect: a helper whose body is `a && b` came back as two answers to a rule written as one.

Fixed the way the invariant path already does it — read what was written, split and place against that, expand each conjunct as it is classified. `ClauseHelpers.ensuresForDischarge` now hands over the written declarations beside the expansion to read them under (`WrittenEnsures`), since neither is usable alone. Now answered at `11:13` and `11:37`, two conjuncts where the author wrote two.

**2. `Nothing is stated about ...` was attached by spelling.** Right — `declaringDef` matches characters, so a parameter named after a behavior found the behavior. It is now asked only where the cursor is on the token the behavior is declared under, and by the canonical name (`nameOf`) rather than the raw token text, so a decomposed spelling is the same name the compiler filed the answer under.

**3. `casesNothingIsSaidAbout` claimed cases whose arm was dropped.** Confirmed: `ensures Found | Other -> ...` where `Other` is not a case of the answer produced `rules=[], casesNothingIsSaidAbout=[Found, Missing]` — the hover told the author nothing was stated about `Found` while `Found` was on their screen. `contractAsRead` now reads all of the declaration or none of it; the refusal is reported where the declaration is held to its rules.

**`RuleId` wording.** `arm` counts the arms of the whole behavior, not of its clause — that is what the binding `value` is is numbered by, so one number for both is one thing to keep true. The comment says so now.

Four new tests, each verified to fail without its change (reverted in place and re-run): the position and the dropped arm on the query, the spelling and the canonical name on the hover. The helper hover has one on the LSP side too.

One note on process, since it nearly cost me a wrong conclusion: `mvn -pl souther-lsp test` without `-am` linked the tests against the installed compiler jar, and the LSP hover test failed against the *old* behavior while the compiler-side probe showed the fixed position. Same source, two answers. All numbers below are from a full reactor run.

Full reactor green under `CI=1 mvn -Plint` — syntax 114, fmt 2251, compiler 5186, build driver 10, lsp 268, cli 340, runtime 113, bench 13, 0 failures.

kawasima:
分類規則を直しました（013673c9）。指摘の構造は正しく、ただし今日の再現形は違ったので、実測を先に置きます。

## 実測

| 書き方 | 修正前の答え |
| --- | --- |
| `value.rank > id.n`（直書き） | DERIVABLE |
| helper が `rank > floor` だけ | EXACT_MATCH |
| helper が `rank > floor && String.startsWith(...)` | EXACT_MATCH（obligation 1個） |

つまり **helper 呼び出しは1つの term として読まれていて、body の中まで分解されていない**。`Predicates.obligations` が `&&` を複数 obligation に割るのは written の `&&` に対してで、展開後の body には届いていません。なので「DERIVABLE と表示される」形は今日は作れませんでした（numeric 側が展開で消えるため、答えはむしろ弱い側に倒れる）。

`owed.unreadable()` を見ていない点はご指摘のとおりで、`Owed.and` が `clauses` 非空かつ `unreadable=true` を作れる以上、コードとしては誤答を書いています。

## 直したもの

**規則を「見つかった読み方をすべて答える」に変更**しました。`ClauseDischarge.readings(asABound, asATerm, unread, at, why)`。1つの written conjunct に対し、bound として読めた／term として読めた／読めなかった、をそれぞれ答えます。位置は1つ（部品は1つとして書かれたので自分の位置を持たない）。ご提案の composite capability そのものです。worst-kind への縮約は採りませんでした——`DERIVABLE && EXACT_MATCH` を EXACT_MATCH と呼ぶと「nothing weaker states it」が嘘になる、というご指摘の通りだからです。

`everyRuleRead`（`FieldDomains` が bounds を信じてよいかを決める経路）も全読み方を見るようにしました。ここは表示でなく判断なので、`numeric && unreadable` を DERIVABLE と答えていた形は実害side でした。

**テストは規則を直接押さえています。** 今日のソースでは heterogeneous な conjunct が作れない以上、ソース経由のテストは「作れないから緑」になります。`EveryReadingOfAClauseIsAnsweredTest` が `readings` の8通りを直接固定します（bound+term → 2件、bound+unread → RUNTIME_ONLY を含む、reason は必要な時だけ評価、など）。ご提案の2本はこの中の2ケースに相当します。

## 前回指摘への「禍根を残さない」対策

1. **3手順を1関数に**。`ClauseHelpers.conjunctsToRead`（written を割る→位置→各断片を展開）。invariant 側と ensures 側で二重に手書きしていて、片方だけ順序が逆だったのが原因なので、順序を書く場所を1つにしました。
2. **性質テスト**。`AClassificationIsPlacedInsideTheClauseItIsAboutTest` が、両種の clause について「答えの位置がその clause の region の中にある」ことを検査します。helper を上と下の両方に書いたコーパスで、順序を戻すと両方赤になることを確認済みです。
3. **部分的な読みを型で持てなくした**。`BehaviorChecker.read` は読めなかったものを contract と一緒に返し、`whole()` を通らないと contract が取り出せません。
4. **綴りを compiler の問いに渡せなくした**。`contractOf` は `SyntaxToken` を取り、正準化は内側でやります。

Full reactor green（`CI=1 mvn -Plint`）— syntax 114, fmt 2251, compiler 5193, build driver 10, lsp 268, cli 340, runtime 113, bench 13, 0 failures.

なお、実測で分かった**別件**を1つ残します: 同じ規則を helper 経由で書くと直書きより弱く分類されます（DERIVABLE → EXACT_MATCH）。展開後の body を読み通していないためで、過大評価ではなく過小評価なので今回は直していません。コミット5 で caller assumption に使うときに効いてくるので、そこで扱います。

kawasima:
Turned the point fixes into an arrangement that does not admit them (94bd4057).

## What kept recurring

Reading a clause is four facts: what the author wrote, where they wrote it, which representation the analysis reads it in, and whether all of it was read. They are not independent — the position belongs to the written form, the analysis wants the expanded one, and an expansion carries the positions of the body it copied in. Every reader derived them separately, so a reader could hold a matched set or a mismatched one with nothing to tell them apart. That is one defect per reader, forever: the data reader placed its answers on what was written, the rule reader placed them on what it had expanded, and neither said which was right.

## The arrangement

`ClausesForDischarge` owns the expansion and never hands it out. What it gives back is a `ClauseReading` — a written conjunct that knows where it is and what it comes to — and the classification takes no position at all, because a position that can be passed can be passed from the wrong tree.

Counting the degrees of freedom before and after:

| | before | after |
| --- | --- | --- |
| places that build a DISCHARGE expansion for classifying | 3 | 1 |
| classification entry points taking a `SourcePos` | 3 | 0 |
| places writing "split, place, expand" | 2 (one backwards) | 1 |

Both old mistakes now fail to compile, from inside `souther.compiler.check` as well:

```java
declaring.expansion().inline(written, owner);              // シンボルを見つけられません
InvariantChecker.capabilitiesOf(reading, wherever, ...);   // SourcePos を Typing に変換できません
```

`everyRuleRead` shows nothing to anybody, so it now asks what the check made of a clause and not where to say it (`ClauseDischarge.kindsRead`). It had been inventing an anchor to get an answer that threw the anchor away.

Behaviour is unchanged — same 5193 compiler tests, and the placement property still fails when the order is reverted. Full reactor green under `CI=1 mvn -Plint`: syntax 114, fmt 2251, compiler 5193, build driver 10, lsp 268, cli 340, runtime 113, bench 13, 0 failures.

What is still a convention rather than a type: `ClauseHelpers.invariantsForDischarge` builds its own DISCHARGE expansion for the check's own reading of whole clauses. That answers a different question — what a construction owes, read against values rather than placed in front of an author — so it is not folded in here; it is named in `ClausesForDischarge`'s doc as the thing that is deliberately not this.

kawasima:
All three confirmed and fixed in f55fd984.

- **Typing failure was silent.** The reading two lines below records it, and says why: a clause this compiler could not type and an analysis that fell over typing it both come out `runtimeOnly`, and only one of them is something the model says. Both readers now type through one step that records it.
- **`readOnlyAsBounds` typed outside any guard** while its sibling typed inside one, so the same clause fail-opened on one route and would have thrown out of `everyRuleRead` on the other. It goes through that step too.
- **Javadoc.** Correct, and worse than it looked: the two blocks standing over `everyRuleRead` were already two before this branch — a claim carried by nobody, twice — and my insertion made a third, leaving the method with none of them. `readOnlyAsBounds` now sits after the method, and the two blocks are one saying both things.

Full reactor green under `CI=1 mvn -Plint` — compiler 5193, lsp 268, 0 failures.

PR #817, opened 2026-08-17
Let a caller assume what the answer was declared to be

kawasima:
Commit 5 of #803: what a behavior declares about its answer is now what a caller that reached the case may assume.

## What it does

Until here an `ensures` was a run-time check and nothing else. The relation it declares fits on neither type — the answer does not carry the input — so a caller had to re-establish it with a guard it had no way to derive.

```souther
behavior findIt : (id: Id) -> Found | Missing
    ensures Found -> value.rank > id.value

let use (id) = {
    guard id.value >= 0
        else Missing { asked = id }
    match findIt(id) with
        | Found as found -> Ranked(found.rank)   -- `Ranked`'s `value > 0` is settled here
        | Missing as missing -> missing
}
```

Take the clause away and that construction is the unproven one (E2011) it always was. Every test asserting silence carries that control, since silence is also what a check that never ran looks like.

## How

- **One reading.** `StatedContract` reads a behavior's rules into the representation the analysis has rules about and types them there, once. The editor's classification (commit 4) is now derived from that same reading, so what an author is shown and what a caller is given cannot drift apart.
- **Contracts in sight.** `Bodies.ContractsInSight` puts that answer under the name each behavior is called by — this module's own and the ones it borrows, each read from the module that declares it.
- **In `PathEngine`, not in the check.** PR#804 removed the second reader of path semantics; putting this in the checker would have grown it back. `InvariantChecker` holds no policy about it.
- **Which rules a caller reaches is read off the arm's pattern.** One case named → that case; several → only that it is one of them, so a rule about one of them is a rule about a value the arm may not have. No new abstract domain: the pattern is the proof the checker already holds.
- **The call is followed through the names it was given**, so `let r = findIt(id)` then `match r` works. Derived from what the walk recorded (`Denotations` already holds what each binding was given) rather than remembered beside it.
- **Conjunct by conjunct**, as everything else about a clause: a rule one half of which names something unreadable still says the other half.

## What does not reach a caller yet

A behavior whose answer has no cases is never matched, so its answer is named only by a `let` — and a name for a call is not a location, so there is no term to read a rule against. I tried making it one: it takes away the record that lets the match path follow the call, because a binding is either an alias for its initializer or a location and `Denotations` holds one of the two. That is a change to what a name for an answer means, and it belongs in its own piece of work. The specification says so rather than leaving it to be discovered.

## Verification

Full reactor green under `CI=1 mvn -Plint` — compiler 5197, lsp 267, cli 340, syntax 114, fmt 2251, build driver 10, runtime 113, bench 13, 0 failures.

Five new tests: through a match, through a name, an or-pattern assuming nothing about one of its cases, a rule half of which is unreadable, and a relation declared in another module reaching its caller here.

### DIFF
--- a/souther-compiler/src/main/java/souther/compiler/check/ContractDischarge.java
+++ b/souther-compiler/src/main/java/souther/compiler/check/ContractDischarge.java
@@ -48,33 +48,21 @@ public record ContractDischarge(List<RuleDischarge> rules,
     public record RuleDischarge(BehaviorContract.RuleId rule, ClauseDischarge capability) {}
 
     /**
-     * What the check can read of {@code contract}, whose rules are as their author wrote them. Each
-     * is expanded into the representation the check reads ({@link InliningPolicy#DISCHARGE}) as it is
-     * classified, so what comes back is placed and split where it is written.
+     * What the check can read of what {@code stated} states, rule by rule.
      *
-     * <p>Handed a contract rather than a declaration. What a rule is about, what {@code value} is
-     * there and which rule it is are the reading's answers ({@link BehaviorChecker}), and this asks
-     * one question of them. The contract handed here is the same declaration that reading reads for
-     * the tree that runs, so its {@link BehaviorContract.RuleId}s are the ids of the rules that run —
-     * a reader holding both is not left matching them up.
+     * <p>Handed the rules as the analysis holds them rather than the declaration they came from.
+     * Which case a rule is about, what {@code value} is there, where each conjunct was written and
+     * what it types to are all settled by the one reading ({@link StatedContract}), and this asks one
+     * question of it — the same reading the check at a call takes what it may assume from, so what an
+     * author is shown and what a caller is given cannot drift apart.
      *
-     * @param expansion what turns a piece of what was written into the tree the check reads. Applied
-     *                  here, one conjunct at a time, rather than to the declaration before it was
-     *                  read: an expansion carries the positions of the body it copies in, so a rule
-     *                  expanded first would be placed inside the helper it names and split where that
-     *                  helper's author split it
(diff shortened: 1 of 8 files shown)


kawasima:
Both points taken. 880a5e12 adds the boundary test; the granularity is #818.

**1. The published-class boundary.** You are right that the cross-module test was compiling both modules in one invocation, which is not the path `ContractsInSight` claims. The new test compiles the library to classes, drops its source, and compiles the consumer on its own against them — so the declaration goes out as metadata and comes back through the front end before anything is assumed from it. The construction after the match is settled; with the library's clause taken away it is unproven again, which is the control every other one of these carries.

It passed as written, as you expected — the metadata writer already follows the helpers an `ensures` reaches. What the test adds is that it goes on passing, since nothing said that had to keep being true.

**2. Query granularity.** Agreed, and filed as #818 rather than fixed here. One thing the issue records that is worth saying: the obstacle is not just scope. The callees are not known when the table is built — the check is handed its inputs up front and has no way to ask a question mid-walk — so a per-call-edge dependency needs the set of behaviors a body calls to be an answer of its own. That is a different change from this one.

---

The background reviewer also raised two LSP points; both are about #816's code and neither survives a look at the current tree.

- `contractOf` reading `.value()` without `present()`: `Answer.present()` is defined as `value != null`, and the call site null-checks the value — the same check, spelled on the value.
- `contractOf(uri, graph, ident.text())`: that was the raw-spelling defect from the first review of #816 and it was fixed there. `contractOf` takes a `SyntaxToken` now and canonicalizes inside, so a spelling cannot be passed at all.

Full reactor green under `CI=1 mvn -Plint` — compiler 5198, lsp 267, cli 340, 0 failures.

PR #820, opened 2026-08-18
Hold a written value to what the behavior states

### LINKED ISSUE #803: A relation between a behavior's input and its answer can only be written by changing the output's type

kawasima:
Commit 6 of #803, the last one: an `example` row and a `fake` row are held to the `ensures` of the behavior they are written for.

## What was wrong

A row and a fake row each write an input and the answer that goes with it — the pair a clause relates — and neither was held to one. A model could record that `findTodo` answers a `Todo` carrying another id than the one asked for and compile clean, because the clause was checked only where the behavior answers and a recorded row answers nowhere.

## What holds them

The declaration's own check. A row's operands are compiled and applied as the module's own code, so what a row writes exists as a live value while the module is compiled, and `$Ensures.check` is invoked over it (`examples/EnsuresChecks`, the one place a check is loaded, called and unwrapped). Nothing reads a clause a second time.

Refusals are **E1928** at an `example` row and **E1929** at a row of a `fake` table, and a row that stops there is `FIXTURES_VALIDATED` / `FAILED` / the new `FailurePhase.ENSURES`.

The check runs before anything is applied, so a row waiting for an implementation is held too — its values are written whether or not a `let` exists yet.

## What a row has to have written

What is held is decided by what the row wrote, and an arm is part of that: the specification reads an arm as a reference to the answer, and the fixture side reads it the same way. A bare case name naming a **unit case** states the whole answer — a unit case has one value, so naming it is writing it — and such a row is held like any other, so `ensures NotFound -> id.value > 0` is decided by a row stating `(Id(0)) -> NotFound` with no construction written.

Two forms write only one side and are held where the behavior answers: a `_` row of a fake states no input, and a `with` answers whatever it is asked.

A bare case name **carrying fields** is not held today. The arm does name the answer, so a rule reading only the inputs is decidable from such a row — what stands in the way is that the check is asked about an answer rather than about an arm, which is a second entry point against ADR-0104's ABI. Filed as #821 and written down in the specification as the limit it is, rather than reworded into a property of the row.

## A refused table answers nothing

A table with an E1929 row is not stood in with, as a table that will not build is not: a row reaching it stops at `FAKE_RESOLUTION` and says nothing of its own, since running it would put the rest of the behavior in a state the model rules out.

It is not compared with the rows recorded for the behavior either. E1919 says two descriptions differ and neither is the right one; E1929 says the declaration decides and the fake is the side that is wrong. Said about one pair those contradict each other — measured as `[E1929, E1919]` before the fix.

## Changed from the plan

Whether a behavior has a check to run is asked of `Bodies.Contracts`, not of whether `$Ensures` loads. A class that will not load is a compile that did not emit what it said it would; reading the loader for this would take that for a behavior with nothing to say.

`EnsuresChecks` is asked by declaration rather than by word — it takes a `ValueName.Behavior` and raises when the module is not its own, instead of answering "declares nothing" for a behavior it holds no contracts for. Measured: a `fake` for an imported injected behavior does not reach this at all today, because `Bodies.Signatures` of the consumer does not carry the imported behavior and `resolveFake` refuses the row with E1908. That refusal is pinned by a test, so the day it is lifted this fails loudly rather than passing everything.

## Evidence

- The three shapes the plan named as negative controls — a body breaking its own clause, a row contradicting one, a fake table contradicting one — were measured compiling with zero diagnostics before any of this was written. Four of the nine tests written first were red for exactly that reason.
- `assertedExpected` answers the value and the assertion from one evaluation. Measured with no constant in it: two rows name one helper once each, one in the input and one in the expectation, and cost the same. Broken on purpose to check the instrument — 9 steps against 12.
- Conformance answers regenerated with `-Dsouther.conformance.update=true`: no diff.
- Full reactor green under `CI=1 mvn -Plint` — compiler 5215, fmt 2257, cli 340, lsp 267, 0 failures.

## Where it is written down

`specification.adoc` gains `[#example-ensures]` and the `[#e1928]` / `[#e1929]` catalog entries; `[#ensures]` points at it. ADR-0104 records that this is the policy ADR-0093 said the language did not have — and that it does not take that decision back: two descriptions of one behavior disagreeing still has no right side (E1919 stays a warning naming both), while a written value against a declaration is a different pair and the declaration decides it.

Closes #803.


kawasima:
Review addressed in `7d2729de`, with the remaining half of the first point filed as #821.

**Bare cases.** Accepted. A unit case is now built and held — a unit case has one value, so naming it is writing it, and the example in the review (`ensures NotFound -> id.value > 0` against `(Id(0)) -> NotFound`) is refused. A bare case name *carrying fields* is still not held: the arm names the answer, so a rule reading only the inputs is decidable there, but the check is asked about an answer rather than about an arm and closing it is a second entry point against ADR-0104's ABI. That is #821 rather than a rewording, and the spec now states the limit in those terms.

**Imported dependencies.** Measured before changing anything, and the escape does not reproduce: `Bodies.Signatures` of the consumer does not carry an imported behavior, so no table is built for it, and `resolveFake` refuses the row outright with E1908 — a `fake` names an injection target of its own module. The identity criticism is right all the same, so `EnsuresChecks` now takes a `ValueName.Behavior` and raises when asked about another module's behavior instead of answering "declares nothing". The E1908 refusal is pinned by a test, so the day it is lifted the fixture side fails loudly rather than passing everything. Moving to `ContractsInSight` belongs with the design that admits such a fake, not ahead of it.

**Invalid fake reaching the parent example.** Taken. I had read it as harmless because the crossing check catches it — measured, the row is `FAILED`/`INVOCATION`, not `HELD` — but one cause reported twice, and a row whose outcome is decided by a state the model rules out, are both wrong. A table with a refused row is no longer stood in with: the row stops at `FAKE_RESOLUTION` as it does for a table that will not build, and the refusal is said once where the table is written.

Also folded in from the automated pass: `EnsuresChecks.found` is a `ConcurrentHashMap` (a row is a thread, and one that overruns its budget is abandoned rather than stopped), and a check that cannot be reached raises `IllegalStateException` rather than `FixtureException`, which callers catch to report a written fixture at fault.

Full reactor green under `CI=1 mvn -Plint` — compiler 5214, fmt 2257, cli 340, lsp 267. Conformance answers unchanged.

kawasima:
Fixed in `76ae9688`, and the PR body is updated.

Reproduced the contradiction first, on exactly the pair from the review: the codes said were `[E1929, E1919]`. So a refused table now answers nothing in the disagreement reading either — where a table that will not build already stood — and the refusal stays where it was, said once where the fake is written.

The gate is inside the reading rather than after it, since holding a table runs the module's code and that is what a reading is what it costs of.

The point about the two diagnostics is the part worth keeping in the record: E1919 is written for two descriptions with nothing to choose between them and says neither is the right one; E1929 is a description held to a declaration and names one. Together about one pair they are not a duplicate but a contradiction. That reasoning is now in the specification under `[#e1929]` and in ADR-0104, so the next reader meets it where the codes are defined rather than having to reconstruct it.

Full reactor green under `CI=1 mvn -Plint` — compiler 5215, fmt 2257, cli 340, lsp 267. Conformance answers unchanged.

PR #827, opened 2026-08-18
Seed a call once where it stands, not once for every node above it

### LINKED ISSUE #826: The seeding walk re-reads the subtree at every node of the walk
`InvariantChecker.walk` calls `PathEngine.answering(e, ...)` at every node, and `answering` reads all
of `e`'s descendants down to the next branch or block. `walk` then recurses into those same
descendants, each of which reads its own subtree again. Every call is seeded once per ancestor:
`seedAt`, `predicates.obligations`, `assuming`, and a fresh set copy in `Known.speaking` on each
pass. Measured over the conformance corpus, the seeding ran four times for one distinct answer.

The redundancy is not wrong, only paid for. Facts are threaded down, so a call seeded at an ancestor
is already in `Known` by the time the descendant's own reading happens, and seeding it again lands
on the same subjects. What it costs is roughly quadratic in the depth of a body, in the most
expensive part of the check.

## Why it is written that way

A construction is judged when the walk reaches it, and the answers it is built from stand underneath
it. Seeded at each call's own step instead, the construction would be judged before the values it
was handed had said anything. So the reading has to run ahead of the walk — the question is how far,
and at present the answer is "the whole subtree, again, from every node".

It already stops in the places where reading ahead would be wrong rather than merely repeated: at a
branch, because a call in the arm beside this one is one this path never evaluates; at a block; and
at a binding's body, because what a name denotes is not settled until the binding is entered.

## What would fix it

Seeding each call once, at the node that owns it, and letting the threading carry it. That is a
change to how the walk is structured rather than to what it decides, so it should be able to hold
the verdict of every construction site fixed — which is the gate it should be held to.

Beside it, `Terms.subjectOf` visits a field access's target twice per level (once through `atomOf`,
once through the identity walk), so a long field chain costs more than the chain is long. Same fix,
same gate.

## Not a correctness issue

Filed apart from the semantic model deliberately: identity, equivalence, knowledge and reporting
policy are settled, and this is the traversal that carries them. Found while reviewing #825.

kawasima:
Closes #826.

`InvariantChecker.walk` asked `PathEngine.answering` at every node, and `answering` reads all of a
node's descendants down to the next branch or block. The walk then recursed into those same
descendants, each of which read its own subtree again — quadratic in the depth of a body, in the
most expensive part of the check.

## The walk

The reading has to run ahead of the walk: a construction is judged when the walk reaches it, and the
answers it is built from stand underneath it. How far ahead is `answering`'s question, and it
already stops where reading ahead would be wrong rather than merely early — at a branch, at a block,
at a binding's body. So those are the places that need to read, and nothing else does: what is read
is threaded down, and seeding the same call again from a descendant lands on the subjects it already
holds.

`entering` is that door. `walk` is a step inside a region, over what the `entering` that owns it
read; every descent that is not a region of its own hands `k` on unchanged, so a step that reaches
nothing stands in a region that reached nothing.

One place the two did not line up: `answering` stops at a block while the walk descended into one
through `forEachChild`, so a closure reached as an ordinary value — a call `Combinators` does not
recognise — got a case of its own.

## Beside it

Two readings of one expression, asked apart.

`InvariantChecker.name` asked `Terms.atomOf` and `Terms.subjectOf` of the same value and each walked
it, though where a value is a number the domain carries its atom *is* its subject, recorded with how
that type's values are spaced. `positionOf` answers both from one reading.

`Terms.pathKey` asked `locationOf` again at every step of a chain, though `Location.of` answers a
chain by answering its root — a chain that is nowhere is one whose every step is nowhere.

## The gate

Held to the verdict of every construction site. Over the conformance corpus and the three bench
corpora, all 233 sites answer with the same verdict and the same clauses as before.

What it cost, over a body of nested calls in one region:

| depth | seedings | nodes read | check |
|---|---|---|---|
| 1 | 4 → 1 | 31 → 11 | 10.9ms → 6.3ms |
| 4 | 22 → 4 | 52 → 14 | 5.0ms → 1.6ms |
| 8 | 60 → 8 | 94 → 18 | 8.6ms → 2.2ms |
| 16 | 184 → 16 | 226 → 26 | 23.8ms → 3.4ms |

Over the corpora themselves, whose bodies are shallow: 35 seedings to 23, 8567 nodes read to 4344.

The other two are small there, and honestly so. `name` stops at `FIELDS_SEEDED` levels of fields
plus the newtype chain, and the corpora write chains of length 0 or 1, so the identity walks go from
30677 to 30248. What depth of chain the quadratic actually cost is not something this repository has
an input for.

`AnAnswerIsSeededOnceWhereItStandsTest` reads `PathEngine.SEEDED`, a seam beside
`InvariantChecker.WATCHING` and `DerivedBounds.WATCHING` and for their reason: two readings that
differ only in how often they seed answer alike about every program, so the property has nowhere
else to be read. It fails on the old shape at depth 1 — `expected: <1> but was: <4>`.

Full reactor green under `CI=1 mvn -Plint` — runtime 113, syntax 114, formatter 2257, compiler 5238,
build driver 10, lsp 267, cli 340, bench 20, 0 failures. Conformance snapshots unchanged.


kawasima:
Review is right that the invariant had a source-reachable counterexample, and `ff726a1f` closes it.

Two corrections to the finding, both from running it rather than reading it.

**It is a factor of two, not a factor of N.** `without` rebuilds the outer expression once per binding entered, and what that costs is one pass over what stands beside it — not one per level of expansion. Measured with the `SEEDED` seam over `Pair { l = bump…(a), r = step…(a) }`, at 4 sibling answers:

| helper nesting | seedings / distinct |
|---|---|
| 1 | 8 / 4 |
| 2 | 8 / 4 |
| 3 | 8 / 4 |

**It is a residual, not a regression.** The same input on `develop` gives 18 / 4. The commit under review took it to 8 / 4; this one takes it to 4 / 4.

**The conditional site had the same shape**, which the review did not name. A conditional given to a value is walked once with each branch put where it stood, so everything beside it was read three times — one outer reading plus one per branch. Measured 12 / 4, now 4 / 4. The two branch readings are the design (`answering`'s javadoc: "each branch's own reading seeds what stands in it"); the outer one was the same redundancy.

The fix is the direction the review proposed: `entering` now takes what is walked and what of it is unread, so the rebuild the judgment needs is separated from the reading. Those two sites are the only ones that walk a tree they did not read — every other `entering` is handed a subtree of its own.

`AnAnswerIsSeededOnceWhereItStandsTest` holds all three shapes, and each is red without the part of the walk that answers it (nested chain 4 for 1, helper binding 2 for 1, conditional 3 for 1).

Verdicts unchanged over all 233 construction sites. Full reactor green under `CI=1 mvn -Plint`.

On `Terms.positionOf`: agreed there is no defect, and I have made `atomOf` ask `subjectOf` too, since the two spellings of one expression are what made it worth a second look.

kawasima:
P2 confirmed and fixed in `8385a7fd`, and chasing it turned up a second one that the previous commit had put there.

| shape | develop | 914d06d4 | ff726a1f | 8385a7fd |
|---|---|---|---|---|
| answer in the condition | 2/1 | 2/1 | 2/1 | 1/1 |
| answer in a branch an arm stands over | 1/1 | 1/1 | **2/1** | 1/1 |
| answer beside a conditional an arm stands over | 3/1 | 3/1 | 1/1 | 1/1 |

Row two is mine. Handing `value.then()` over as `unread` unconditionally is too much where the conditional sits past somewhere the reading stops — the region that owns the branch reads it when the walk gets there — and the whole expression was too much where the reading had reached it. One argument, wrong in both directions depending on the shape. Rows two and three are the same defect seen from either side.

You are right that swapping `entering` for `walk` is not enough. What decides it is whether the reading reached where the conditional stands, and the search that found it is what knows: `conditionalIn` marks a site `pastTheReading()` wherever it descends past a place `answering` stops — a binding's body, an arm, an attempt's success, a departure. Then a site the reading reached has its condition in `within` already, so the condition is walked and the branch is what is unread; a site past the reading has its condition entered and nothing unread at the top.

Narrower than holding a read frontier through the walk, though. `bindingIn` stops exactly where `answering` stops, so a binding standing in a value is always one the reading reached — `conditionalIn` is the only search that goes further, and it is the only one asked.

The test holds five shapes now. At `914d06d4` four fail, at `ff726a1f` two.

Verdicts unchanged over all 233 construction sites. Full reactor green under `CI=1 mvn -Plint`, compiler 5242.

kawasima:
Right that nothing was on the remote — that was on me, it is pushed now as `b7958215`.

But it does not make your probe 1/1, and it is not meant to. The main-source diff is **zero non-comment lines**. What changed is that the rule is written down and the table is asserted.

**Your example is not a counterexample.** Two controls:

| where the answer stands | seedings per answer |
|---|---|
| in an arm standing under a conditional written in that arm (your shape) | 2 |
| in an arm standing under a conditional written **outside every arm** | 2 |
| in an arm, with **no conditional anywhere** | 1 |

Where the conditional stands relative to the answer does not move the number. Taking the conditional away does. So the two is the two readings — opening a conditional a value is handed reads the expression once with each branch standing where it stood. The same two judge every construction inside it and merge what they found (`Judgment.of`); `Capture`/`say` exist for exactly that. An answer read once per reading is this walk being read twice, not an answer being read again.

It could not be one, either. An arm's contents are read under the denotations `enteringArm` introduces, which do not exist before the branches are split — there is nowhere earlier to read them that lands on the same subjects. And the walk and the judging are doubled there whatever the seeding does, so it would not move the order.

**On the shape of the last three rounds.** Each was one probe, one patch. That is how a rule ends up recorded as the list of cases someone tried, so it is now stated once: `walk` is owed that `k` holds what `answering` derives over `e` as far as that reading goes from `e`; `entering` meets it; `unread` is the part not yet met. The three call shapes are derivations, including the null one — where the part was put in past somewhere the reading stops, the requirement never reached there. A consequence, not an exception.

The test held six shapes, one per round. It holds one table now — ten places × three sizes, plus nesting depth as its own slope:

    standing on its own                                        n
    beside a binding an expansion introduced                   n
    in the condition a conditional is decided by               n
    beside a conditional, in the region it is opened in        n
    in a branch of that conditional                            n
    in the condition of a conditional an arm stands over       n
    in a branch of a conditional an arm stands over            n
    in an arm, with no conditional anywhere                    n
    in an arm standing under a conditional                    2n
    in an arm standing under a conditional written in that arm 2n

Your shape is the last row, asserted at 2 with the two controls beside it. At `914d06d4` the table fails on the binding row, at `ff726a1f` on the condition row.

If you want the 2 gone, that is a change to what a reading is rather than to where the seeding runs, and it belongs in its own issue — not in #826.

PR #832, opened 2026-08-18
Draw a line where an `ensures` compares an input

### LINKED ISSUE #823: A comparison in an ensures draws no boundary a row is offered at
`souther examples --generate --boundaries` offers a row at each line the model draws, and the lines come from two rules. `OriginRef` names them: `InvariantOrigin`, a clause of a `data`'s invariant, and `GuardOrigin`, a comparison in a behavior's body. A comparison written in an `ensures` draws a line too, and is not among them.

```souther
    ensures asked = NotFound -> id.value > 0
```

`id.value = 0` is a value the stated relation changes at: below it and at it the behavior may not answer `NotFound`, above it it may. `--boundaries` offers no row there, and a row at that value is one the model has a reason to want.

## Why it is not already there

The clause is not read by any of it. `BehaviorContract` appears nowhere in `GeneratedRows`, in `Generator`, or in the partition machinery, so no cut is derived from a clause and there is no obligation for a row to meet.

## What has to be decided first

Not the emission — the kind of obligation it is. `BoundaryObligation` says the two existing kinds are met differently:

> An obligation from an invariant is met by a row whose value is the boundary. One from a guard is not: the comparison has to have produced a value as well, because a value can reach the input of a behavior without reaching the comparison that cares about it.

A clause is not a guard in that sense. It is not reached conditionally: every rule runs when the behavior answers, so a row that hands the behavior the threshold has reached the comparison by construction. That puts it on the invariant side of the distinction — met by writing the value — which is the answer to settle explicitly rather than to inherit from whichever existing arm the code is written next to.

`OriginRef` is sealed, so an `EnsuresOrigin` arm makes every reader of an origin answer for it: the report's wording, what the generated row says it stands for, and how `BoundaryAssessment.Coverage` measures it.

Two more to settle while there:

- Which side of the relation the cut belongs to. `id.value > 0` under a `NotFound` arm is a cut on an *input*, which is what the generator offers rows at. A comparison reading `value` is a cut on the answer, and a row cannot be written at it — the row does not choose what the behavior answers. Only the first is an obligation a row can meet.
- Whether a cut a clause and a guard both draw is one line or two. `OriginRef` already answers this for an invariant and a guard that repeat each other: they merge into one partition and stay separate obligations, because reaching one says nothing about the other. A clause is checked whenever the behavior answers, so what reaching it means may differ again.

Related: #822, which is about the clause being visible next to the rows an author has to answer — the same information, on the other side of the same command.

kawasima:
Closes #823.

`souther examples --generate --boundaries` offers a row at each line the model draws, and the lines came from two rules. A comparison written in an `ensures` draws one too, and no row was offered there:

```souther
    ensures asked = NotFound -> id.value > 0
```

## What the issue asked to settle first

**The kind of obligation, and it is neither of the two that were here.** The existing arms bundle two questions that turn out to be independent:

| | met by | the line's other side |
|---|---|---|
| `InvariantOrigin` | writing the value | none — outside it nothing exists |
| `GuardOrigin` | writing the value **and** getting the comparison to answer | values a row can write |
| `EnsuresOrigin` | writing the value | values a row can write |

A clause is not reached conditionally — every rule of a declaration runs when the behavior answers — so a row handing the behavior the threshold has reached the comparison by construction. But both sides hold values a row can write, so the value **and its neighbour** are owed a row. Inherited from whichever arm the code sat next to, one of those two would have been wrong. `OriginRef` answers the two questions (`comparisonSite()`, `besideTheCut()`) instead of being matched on, and `Partitions`/`Coverages` ask them.

**Which side of the relation.** A comparison reading `value` is a line on the answer and no row can be written at one. Nothing turns it away: a term over the answer names no position of the input, so it draws nothing. Only what a rule states outright, too — the walk descends through `&&` and stops everywhere else, since a disjunct is not stated on its own.

**One line or two.** Two, the way an invariant and a `guard` at one value already are. A row on the line settles the clause's obligation and leaves the guard's owed where it never reached the comparison:

```
    boundary    1/4
      ! no row is at charge/a.n = 101 (ensures charge (small))
      ! no row is at charge/a.n = 100 (guard@15:9)
      ! no row is at charge/a.n = 101 (guard@15:9)
```

## Also here

A clause comparing one input against another draws a line between the two, as a `guard` doing so does, met the same way — by a row putting one count in both positions. `assessBetween` was written as though only a guard could draw that shape.

What a comparison draws is read in one place per shape (`ComparedLine`, `ComparedTerms`); each reader finds its own comparisons and adds what meeting the line takes. `InputReads` carries which tree it is walking, since a clause's rules are written against the declaration's bindings and a body's against the implementation's.

A position a clause compares that nothing could read is named as unread — otherwise the report says the model draws no line through it while the clause two tokens away says otherwise. Not for a rule relating the answer to an input; the conformance corpus caught that one.

## A pre-existing defect this uncovered

The first commit stands alone and reproduces on `develop`. What is generated for a gap looked the assessment up **by the value alone**, so where two rules drew a line at one value and only one was a gap, it could take the met one and stop the run:

```
internal compiler error: IllegalStateException: the assessment at charge/a.n = 100 says
A_ROW_IS_ALREADY_THERE, which is not a gap
```

Nothing about the report was wrong — both lines were counted and printed. The finding already carries the rule it is about, so the lookup asks for it.

## Verification

Full reactor green under `CI=1 mvn -Plint` — compiler 5251, syntax 2257, lsp 267, cli 344, 0 failures. Conformance snapshots unchanged. The first commit was built and its test run on its own worktree.

`specification.adoc` updated at §boundary-coordinates and §a-comparison-of-two-positions-draws-a-line-between-them.


kawasima:
Rebased onto `develop` (now `3206f172`, PR #828), and pushed a third commit addressing the review.

## Blocking: a measure in an `ensures` drew no line

Confirmed, and it was worse than a miss — the position came back as one this compiler could not read a rule about:

```
· not read: name (a rule about it is one this compiler did not read)
```

`StatedContract` elaborates with `forDischarge()`, so `String.length(name.value)` arrives as `Core.PreservedCall` while `GuardThresholds.termOf` recognised only the `Core.Call` a body's condition holds. Fixed where the review suggested: `NumericMeasures` already owns *which* operations count, so it now owns reading one out of either shape (`measureIn` → `(ValueName.Stdlib, Core)`), and `termOf` asks it. No branch in `ComparedLine`, so the comparison reader stays one.

```
boundary    1/2
  ! no row is at find/String.length(name) = 3 (ensures find (NotFound))
//     | "name=0 <= x <= 3" : (Name("")) -> <?>
//     | (Name("xxx")) -> <?>
```

Regression tests over `String.length` and `List.length` in `WhatAClauseDrawsALineOnTest`.

## Same cause, second form: a rule stated through a helper

A call is expanded where the rules are read, so `ensures small = Refused -> big(a)` arrives as the helper's body under a binding. The walk stopped at anything that was not a comparison, and with nothing in either answer the report said `· not derivable: a.n` — the model draws no line through the position, which the clause two tokens away contradicts. That is exactly the sentence this PR claims to prevent.

A `let` is not a choice, so the walk goes through it the way the body reader does. It now draws the line rather than merely naming the position:

```
  ! no row is at charge/a.n = 100 (ensures charge (small))
  ! no row is at charge/a.n = 101 (ensures charge (small))
```

And where the walk does stop on a form it has no rule for (`Bool.not(...)`), the position is named as unread — the silence was stronger than a guard's.

## Non-blocking: the justification mixed semantics with the run

Taken. "Every rule runs when the behavior answers" is not what makes this true — under short-circuit the right side of an `&&` need not be evaluated. Restated in `EnsuresThresholds`, `OriginRef.EnsuresOrigin`, `comparisonSite()`, `Coverages` and the spec: a guard's line is about control flow, and a clause states a relation whose line is covered by the input the relation changes at. That is also why a disjunct draws nothing — it is not part of what the rule requires.

## Non-blocking: `atEdge`'s narrowed match

`edges` is `Boundaries(name)`'s answer and every finding was made from one of its assessments, so the two cannot diverge without the readings having come apart — which is what the fall-through refuses. Said in the javadoc rather than left for the next reader to reconstruct.

Full reactor green under `CI=1 mvn -Plint` — compiler 5262, syntax 2257, lsp 267, cli 344, 0 failures. Conformance snapshots unchanged.

PR #834, opened 2026-08-18
Put a behavior's `ensures` over the rows that ask what it answers

### LINKED ISSUE #822: A generated row does not carry what the behavior already states about its answer
`souther examples --generate` prints rows for what nothing covers, with the inputs filled and the answer left as `<?>` — a placeholder the grammar does not accept, so a row cannot be uncommented without being answered. The row is a question: these inputs, and what does the old system answer?

Where the behavior carries an `ensures`, the model has already answered part of it, and the block does not say so.

```souther
    ensures found = Todo -> value.id.value == id.value
    ensures asked = NotFound -> id.value > 0
```

A generated row at `TodoId(0)` is one whose answer cannot be a `NotFound`, and a generated row at `TodoId(7)` is one whose answer, if it is a `Todo`, has `id = TodoId(7)`. Whoever fills in `<?>` — a person reading the block, or an agent writing an implementation against it — is deriving that from the source a second time, or not deriving it at all.

## Why it is not already there

`GeneratedRows` reads the adequacy report and the generator's outcome. It holds no contract: `BehaviorContract` appears nowhere in it, and neither does anything else that would let it say what a behavior states.

Nothing is wrong today as a result. A generated row states no answer, so it writes only one side of the relation and is not held to the clause (spec `[#example-ensures]`); the row cannot contradict the declaration because it makes no claim about the answer at all. What is missing is not a check but a fact next to the question.

## What this would be

The clause as it is written, printed with the rows for that behavior — the same words the author wrote, not a reformulation. Where the row sits at a boundary that a clause draws (#TBD, filed separately), saying which clause drew it is the same note the boundary rows already carry for a guard or an invariant.

Worth deciding while writing it: whether a clause naming a case rules that case out for a given row's inputs — `ensures NotFound -> id.value > 0` says a row at `TodoId(0)` is not answered by `NotFound` — and whether the block should say that much, or only quote the clause and leave the reading to whoever fills the row. The first is a small inference and is what a reader actually wants; the second cannot be wrong.

Not urgent, and it is worth the most where the answers are being harvested rather than invented: an injected behavior whose rows are the record of what it owes (#695).

kawasima:
Closes #822.

`souther examples --generate` prints rows for what nothing covers, with the inputs filled and the answer left as `<?>`. Where the behavior carries an `ensures`, part of that answer is already written down, and the block did not say so.

Now the clauses are quoted over the rows of the behavior they are written on:

```
// generated by `souther examples --generate`: 2 rows to fill what nothing covers.
// `ensures` written for `findTodo`:
//     ensures found = Todo -> value.id.value == id.value
//     ensures asked = NotFound -> id.value > 0
//
// example findTodo
//     | "id=0 <= x <= 0" : (TodoId(0)) -> <?>
//     | "id=0 < x" : (TodoId(1)) -> <?>
```

Once per behavior, and only where a row is being offered. A clause is written on the behavior, so it says the same thing about every row of it.

## The source, not a rendering of the rule

The quote is cut out of the file by the clause's own region. `BehaviorContract` holds the same words resolved, typed and specialized per case; rendering that would hand back the compiler's equivalent for the author's, and the quote would move the day a normalization changed without anybody having edited the model.

## Read off the settled module, not off the contracts

`Bodies.Contracts` answers a different question — which clauses the checker could read — and drops the behaviors whose reading it refused. `souther examples` says what is wrong with the source and then reports what it observed, so a refused behavior's rows are offered anyway: one whose arm names a data that is no case of its answer is refused (E1619) and still gets a row per case of a sum it takes. Taken from the contracts, the block would print nothing over those rows, which reads as a behavior that states nothing rather than one whose statement was refused.

That case is reachable and is the third test.

The heading is ``` `ensures` written for ``` and not "what the behavior states", for the same reason. What is claimed is that these words are in the declaration.

## What is not in this

No evaluation of a clause, no case ruled out for a row's inputs, no specialization per row. `ensures asked = NotFound -> id.value > 0` says a row at `TodoId(0)` cannot be answered by `NotFound`, and deriving that is contract × input → answer space — a feature, not a rendering, and worth its own issue.

`Generator.GeneratedRow`, `OriginRef` and `BehaviorContract` are untouched. Which clause drew a boundary is a row's provenance and is already in the report (#823); what a behavior states is context for every row of it, not provenance for one.

## Notes

The old four-argument `GeneratedRows.of` is removed rather than kept beside the new one — left there it would be a way of calling this with the clauses dropped, which is the omission being fixed. The call sites are tests.

`commands.md` gains a paragraph under `examples`. No change to `specification.adoc` (this is the presentation of `--generate`, not the semantics of `ensures`) and none to the JSON report (the block is a separate channel already).

Verified: full build green, `CI=1 mvn -o -Plint verify` green, and the output above is from the built jar rather than a test.


kawasima:
P2 を直しました (`c44a48f7`)。指摘の通りで、`Bodies.Contracts` を避けた理由を一段下で自分が破っていました。

## 直したもの

`quoted()` の `null` が `Map` の absence に落ちる形をやめました。absence が運ぶ事実を1つにしています。

- **書かれていない** — pass が書いた節。著者のものは何も無く、見出しが主張する内容もそれ。これは absence のまま。節ごとではなく、切り出す前に一度だけ聞きます。
- **書かれているが切り出せなかった** — text が無い / region が text の外 / 節の一部だけが書かれている。全部 `IllegalStateException`。

どれもモデルからは到達しません。`Db.forget` は source の text とそのモジュールについての答えを同時に落とすので、text だけ消えた節がここへ届くのは store が半分だけ落ちた状態です。region の数字は同じ text を読んで作られたものですし、節は宣言から読まれて、それを書き換える pass は位置をそのまま持ち越します。

## テスト2本

**helper 展開** — これは指摘の通り、この実装の最も重要な性質が無検査でした。入れる前に、その読み違いが空想でないことを確認しています。

```
ensures asked = NotFound -> positive(id)
```

が becomes する rule は `Expansion[callee=example.todo.positive, ..., body=Binary[op=GT, left=FieldAccess[... pos=6:32 ...]]]` で、helper の body を helper 自身の位置で持っています。式から作った renderer なら、著者が10行目に書いた `positive(id)` の場所に6行目の `id.value > 0` を出します。引用は `positive(id)` のまま出ることを固定しました。

**tab indent** — `cut()` が `Region.column` を `String` の index として扱っている件。1段目の下の行から落とすのは code unit の数で、column がまさにそれです。tab が進む桁数として読むと、tab インデントの節は全行で text の外から切り始めます。

## 検証

全モジュールのテスト緑、`CI=1 mvn -o -Plint verify` 緑。

Issue #835, opened 2026-08-18
A checked body reads what every name in its module means


kawasima:
Declaring a type re-checks every body of the module. Nothing calls it and nothing writes its
spelling.

Measured on the module #829 was measured on — `shop.orders` with `caller`, `called` and an
`Amount` — by adding `data Discount = Int`, over the inputs of `Bodies.CheckedBehavior`:

| input | value differs |
|---|---|
| `Names.Meanings(module)` | yes |
| everything else it reads | no |

`IncrementalCompilationTest.declaringADataRechecksTheBodiesBesideIt` pins it.

## It is the rule `Db` states, one level up

A collection gathered per module is an index, and a question asked per definition depends on the
entries it reaches rather than on the index. `Names.Meanings` is an index — what each type spelling
written in a module denotes, and what each alias reaches — and `Bodies.CheckedBehavior` takes it
whole through `Names.derivedSymbols`. `caller` does not write `Discount`, so what that spelling
means is no part of what `caller` is checked against.

#829 cut two edges of this shape and left this one. `Bodies.CalleeSigsForBody` is the module's
callable signatures cut down to what one body names. `Names.Meanings` is itself the cut that stops
a behavior being declared, which adds a value name and no type name.

The reach is the module and not past it. `Meanings` holds a module's own declarations plus what its
import lines named, so declaring a type in an imported module that the importer does not name does
not arrive — measured, and `declaringABehaviorInAnImportedModuleDoesNotRecheckTheBodiesThatBorrowFromIt`
is the same shape for behaviors. What a body reads of another module's declarations is already
asked one declaration at a time, through `Shapes.DerivedDef`.

## The narrowing that closed the behavior half does not carry over

`Bodies.CalleeSigsForBody` projects over `Bodies.BehaviorsReached`, a frontier that already existed
and that a body cannot get past: a helper cannot reach a behavior at all (E1818), so expanding one
into a body brings no name that frontier has not got.

Nothing draws that line for type names, and drawing it is deciding what a body's scope is.

What a body's check does fetch is recorded already — each declaration it reads is a
`Shapes.DerivedDef` dependency of `Bodies.CheckedBehavior`. Over the conformance corpora:

| body | spellings its module's meanings hold | declarations the check fetched |
|---|---|---|
| `conformance.catalog.discount` | 7 | 7 |
| `conformance.shipping.出荷する` | 11 | 7 |

Two bodies is not a sample, and the fetched set is a lower bound rather than the frontier: a
[...]

PR #838, opened 2026-08-18
Answer with what a module's names mean, and check a body against what it names

### LINKED ISSUE #829: A checked body reads its module's scope and callable signatures whole
Adding a behavior to a module re-checks every body in it. No `ensures` and no body is involved.

Measured on a module with two behaviors, adding a third that nothing calls, over the inputs of `Bodies.CheckedBehavior`:

| input | value differs |
|---|---|
| `Shapes.Scope(module)` | yes |
| `Bodies.CalleeSigs(module)` | yes |
| everything else it reads | no |

`CheckedBehavior(module, behavior)` is deliberately about one body. #818 narrowed the contracts it reads to the behaviors the body reaches, so editing an `ensures` no longer touches the bodies beside it. These two are the same shape and are not narrowed: a module-wide table read whole, which gives every body of the module one identity.

`CalleeSigs` is the smaller of the two. What a body wants of it is the signature of what it calls, which is the frontier #818 already built — `Bodies.CallTargets(module, behavior)` answers it, so this is a projection over an answer that exists.

`Shapes.Scope` is not. It carries types, constructors, helpers and behavior resolution together and a body reads it for all of those, so slicing it per body is a question about what a body's scope is rather than a mechanical narrowing. Doing it alongside `CalleeSigs` would settle that question by accident.

`IncrementalCompilationTest.declaringABehaviorRechecksTheBodiesBesideIt` pins what happens today, so narrowing either table turns it red.

No soundness question, and nothing observable until a module is large enough for the re-check to be felt.

kawasima:
Closes #829.

Adding a behavior to a module re-checked every body in it. Two inputs of
`Bodies.CheckedBehavior` changed, and they were not the same fault.

`Bodies.CalleeSigs` changed because the map really did gain an entry — the module's index of
everything callable in it, read whole by a question asked per body.

`Shapes.Scope` changed while saying the same thing. Its answer, `Symbols`, reads declarations by
asking the store, so two of them are the same when the store is and never when one replaces the
other. Measured over that edit: `denotedNames`, `namesInScope` and `reachable` all came out equal,
and of the six parts of the assembly the scope is built from, the one that moved was the value
namespace — which nothing building a scope reads.

## What the commits do

**Build a scope where it is read.** `Scoping.Meanings` — the module name, what each type spelling
denotes, what each alias reaches — is what `Names.Meanings` answers, cut out of the whole assembly.
A scope is built inside whichever compute reads it, through `Names.derivedSymbols`. That is also
what makes the declarations it fetches dependencies of the asker, one declaration at a time.
`Shapes.Scope`, `Names.NameScope` and `Compilation.symbols` are gone; the LSP wanted what a
spelling reaches, which `Names.Reachable` answers as a map of declarations.

**Check a body against the signatures it names.** `Bodies.CalleeSigsForBody` projects the module's
index over `BehaviorsReached`. The two trees involved name the same behaviors because a helper
cannot reach a behavior at all (E1818).

**Say the rule on `Key`.** An answer is a value; an object that reads the store is not an answer.
A coarse producer does not have to be split — what a consumer reads has to stop where its meaning
stops, and a key between the two is where that happens.

**Ask two stores the same input.** `EquivalentDatabasesAnswerTheSameTest` compiles each conformance
corpus into two fresh stores and compares what they kept. It separates an answer that is the same
thing said twice and compares unequal anyway from one that says something different; the second is
required to be empty. Finding which takes a walk of the pair, because a record holding a way of
reading a store declares an `equals` and still says no.

Known and not fixed, by cause, with what each has to become:

| cause | |
|---|---|
| `byte[]` | a value — reached from five questions, one thing to fix |
| the store, through `Scoping.Scoped`'s way of asking the modules around a module | a capability, out of the answer |
| `souther.compiler.inputs.InputDomain` | unclassified, so read it first |
| `souther.compiler.query.Bodies$Elaborated` | unclassified |

Exactly those: fixing one without striking it off fails, and so does adding one.

## What is left

Declaring a type still re-checks every body of the module, because it really does change what the
names of the module mean. That is #835, and `declaringADataRechecksTheBodiesBesideIt` pins it.

## Checked

`CI=1 mvn -o -Plint verify` green across every module. Each new check was made to fail on purpose
first: making `Scoping.Meanings` compare by identity turns the incremental tests red with two
identical-looking values, and it turns up in the register test named with its path; putting a value
that does not reproduce into a scope turns up under the other assertion instead.

### DIFF
--- a/souther-compiler/src/main/java/souther/compiler/check/Scoping.java
+++ b/souther-compiler/src/main/java/souther/compiler/check/Scoping.java
@@ -90,7 +90,7 @@ public final class Scoping {
         Resolve.Values values = new Resolve.Values(
                 reachable(universe, subject, imports), new OfTheUniverse(universe));
         refused.addAll(oneSpellingTwice(subject, ownData));
-        return new Scoped(m.name(), denotations, aliases, values, imports, refused);
+        return new Scoped(new Meanings(m.name(), denotations, aliases), values, imports, refused);
     }
 
     /**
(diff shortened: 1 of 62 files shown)


kawasima:
Four fixes in `64f0ee88`, measured rather than taken on faith.

**1. The walk's loop guard never fired.** Confirmed: the set was keyed by address and a `Pair` is made where it is asked about, so `add` always succeeded. Instrumented over the conformance corpora — 0 stops as written, **24** on a set that compares its members, so the guard has work to do. The budget was not being reached today (0 exhaustions), so nothing was being missed; the protection simply was not there. Now a `HashSet<Pair>`, which is what `Pair`'s own equality was written for.

**2. The register was cause-wide.** Confirmed by experiment: adding `byte[]` to a brand-new answer left the test green. It now registers each cause with the set of questions it is reached from, and the same experiment fails, naming `Meanings` and `ModuleScope` as new places `byte[]` reached. Paths stay out of the comparison and in the failure message, so restructuring what an answer holds inside itself still does not churn.

**3. `derivedSymbols` / `resolvedSymbols` are package-private again.** Every production caller was already inside `souther.compiler.query`; only tests wanted them from outside. A test-scope `souther.compiler.query.Scopes` reaches over the boundary from inside it, which is a thing only a test source can do.

**4. Added `bothStoresAreAskedTheSameQuestions`.** Its own sentence rather than a line of the other, since "asked the same" and "answered the same" are two claims. Ran it five times over the corpora — stable — and checked it fails by asking one store an extra question.

`CI=1 mvn -o -Plint verify` green across every module.

PR #844, opened 2026-08-18
Run an injected behavior's recorded rows against a bound implementation

### LINKED ISSUE #695: An injected behavior's recorded rows and stand-ins are run against a bound implementation

kawasima:
Closes #695.

A row recorded against an injected behavior was held to everything it could be held to without being
run, and there it stopped. ADR-0088's title said why — *until it can be run*. This is the other half:
an `Answering` whose answerer brings its own classes, and a face that lets a test reach one recorded
row at a time.

```java
BoundExamples examples = SoutherExamples.of(Path.of("src/main/souther/todo.sou"))
        .bind(new FindTodoImpl(db));

for (RecordedRow row : examples.rows()) {
    db.truncate(table(name("todos"))).execute();
    if (stored.is(row)) {
        db.insertInto(table(name("todos"))).values(1L, "write the SQL", false).execute();
    }
    assertHeld(examples.evaluate(row));
}
```

## What is here

**A binding names its behavior through the ABI and its origin through its own classes.** `bind` takes
the instance and nothing else. Which behavior it implements is the binary name
`SoutherJvmAbi.nameOf(GeneratedClass.BehaviorInterface(module, behavior))` gives that behavior's base,
looked for in the instance's supertypes — taking `FindTodoImpl` apart would restate the emitter's
naming rule and go on answering after it moved. Its origin is `Origin.Published` over
`ClassFileDeclarations` reading its own loader's class resources. Neither is asked of the caller,
which is what makes it impossible to state either wrongly. It answers for what it implements and
delegates the rest to `generatedHere`, the resolution being per behavior.

**The API is enumeration and single evaluation.** `rows()` and `evaluate(row)`, and no loop. What
changes between rows is the world the implementation answers out of, and that is the caller's. No
JUnit, no `DynamicTest`, no hooks. `RowKey` holds a `RowIdentity.Named` so a key for an unnameable
row cannot be written, and `evaluate` takes the enumerated row so an unnamed one still runs.

**`Applied` gets `Bound`**, carrying nothing. No measure branches on it: `hits` answers what this
compile's instrumentation saw, not who applied the behavior, and a test holds a bound row's `hits`
empty rather than a branch being added.

**`FailurePhase.INFRASTRUCTURE` becomes `VALUE_CROSSING`**, which is what its one producer says of
itself. Nothing outside Java named the old constant.

**A `fake`'s explicit entries are observed against the implementation** —
`AsStated | OtherThanStated | Unobserved(reason)`, with no severity. ADR-0093 already states a
disagreement without choosing a side, and a fake written to reach a composite's error path disagrees
because that is what it is for. `alsoBy` is the static half of an apportionment; the dynamic half
needs world identity, which only the caller holds.

**What an injected implementation answered is held to its `ensures` (E1930).** ADR-0104 put that check
where an injected answer enters generated code; a behavior only the application's Java calls has no
such crossing in the compilation, so a row applying a bound implementation is the only place its
clause is ever run. Nothing in the generated surface carries the contract either — an injected
behavior's base class is the same class with a clause and without one.

## Three things the issue text had wrong, found by measuring

**The check was silent for every implementation until the answer was brought back.**
`CaseGen.jumpUnlessMatches` guards each rule with an `instanceof` against the class this compile
emitted, so an answer of another loader's classes matches no guard and every rule is skipped. The
issue said reading an answer back is loader-free already — true for the comparison, which reads by
name and by the accessor every data has, and not true for handing a value to generated bytecode. The
answer now goes out through the neutral form and back in through this module's own decoder, read at
the case it turned out to be, which is the line the Decoder draws for a value arriving from outside.

**A second `fake` table for a target that already has one is not refused.** The model in
`aTableShadowedByAnEarlierOneStatesNothingToObserve` compiles with no diagnostic at all, and dispatch
takes the first table — so the second states nothing the implementation could contradict. It is
excluded from the enumeration; the test goes red when the rule is removed, the dead table becoming
the one that answers. Whether the compiler should refuse such a table is its own question.

**A bound row's `Counting` is not `Read(0, ∅)`.** `Title`'s invariant is this compile's code and the
row's fixture goes through it, so the counting covers the fixtures and stops at the behavior. What
the tests hold is that it is `Read` and not `Unread`, and that `hits` is empty.

## Not here

No JUnit adapter — it is a consumer of this face and its own issue, deliberately not built alongside
it. No second example semantics, no second staleness rule, no new test-result model, no persisted
evidence, and no world state in the language.

## Verification

All modules: 8483 tests, 0 failures. `CI=1 mvn -Plint verify` BUILD SUCCESS.


kawasima:
All six addressed. The first two were correctness, as the review said.

## 1 — source set and module path (P1)

`of(Path)` read one file and handed the compile `ModulePath.EMPTY`, so a module writing its rows in
an `examples for` file or importing another user module found them unreachable rather than failing.
The entrance is now `of(List<Path> sources, ModulePath dependencies)`, with `of(List<Path>)` and
`of(Path)` over it and `ofSources(List<String>, ModulePath)` for text.

Which module the rows are of is no longer `modules().get(0)`. `bind` asks the implementation which
module's behaviors it implements, and refuses an instance that implements behaviors of two — a source
set is not ordered by what it means, so binding by file order was a guess.

The deferred-compile alternative was considered and not taken: `of()` compiling eagerly is what makes
a model that does not compile fail before any binding, and the dependency path is a fact about the
project rather than about whichever instance is bound first.

`TheFaceReadsWhatAModelIsWrittenAsTest` runs a module that imports another and keeps its rows in a
companion file.

## 2 — one gate, asked as one (P1)

The diagnosis was right and the fix is structural rather than two `if`s. Whether anything applies the
behavior and whether what does was built against this module are now one sealed `Handing`
(`MayApply | NothingApplies | NotEstablished`), which a caller with values to hand over cannot
consider half of. `checkRow` and `observe` both switch over it.

`observe` also runs under the deadline `evaluate` runs under — the review is right that an
implementation that hangs would otherwise hang the caller's loop. `StandinObservation.Reason` gains
`TheImplementationIsOfAnotherBuild` and `TheObservationRanOut`.

`SoutherExamples.withBudget(Duration)` came out of writing the test: what a bound implementation waits
for is a database or a socket, and how long that may take is not a compile's to know.

`AnEntryIsHandedOverOnTheTermsARowIsTest` holds the disagreement case, a same-declarations control,
the non-answering implementation, and the overload case.

## 3 — `RowKey` (P2)

Correct, and it defeated the type's own purpose. It is a final class with a package-private
constructor, made only by `row(behavior, name)`. It also refuses another enumeration's rows rather
than answering `false` about them, which was the same failure one call along. A test asserts
`RowKey.class.getConstructors().length == 0`.

## 4 — `applyOf` (P2)

Also correct, including the `IllegalArgumentException` escape. The base is asked for the `apply` it
declares and that signature is resolved on the instance; where the base declares none — a single-input
behavior implements `Behavior<In,Out>` — the interface's erased one is what a row enters through.
`IllegalArgumentException` is now caught as `ImplementationNotReached`. The test binds an
implementation carrying `apply(String)` and `apply(Integer)` beside the behavior's.

## 5 — `CompileException` (P2)

`SoutherExamples` throws `CompileException.ofAllReported(...)`, keeping positions, codes and
sentences. A test asserts the diagnostics survive and are still located.

## 6 — `StandinEntry.table()` (P3)

Returns `SourcePos` now, plus `at()` for the entry itself.

## Verification

8501 tests across all modules, 0 failures. `CI=1 mvn -Plint verify` BUILD SUCCESS.

Two things I did not do. Module identity in `RowKey` is not added: the key is refused by another
enumeration, which closes the same hole without giving an address a namespace it has no reader for
yet. And a public source-set abstraction is not introduced — `List<Path>` plus `ModulePath` is what
the compiler already models, and a third representation of a source set is the kind of thing this API
is meant to stand in front of rather than add to.

kawasima:
The three blocking findings and the two follow-ups are in `b1f1879`.

## 1 — `evaluate` kept only the outcome (P1)

Right, and the reasoning is right too: `RowOutcome` is what an adequacy measure reads and was
designed for, not a test result. A row that failed a comparison carries `FAILED`, `COMPARISON` and
both arms, which is the same thing a right implementation produces for a right row.

`evaluate` now answers `RowEvaluation(RowOutcome outcome, List<Diagnostic> diagnostics)`.
`ANSWERER_ESTABLISHMENT` is covered too: `one()` asks the same `Handing` and, where it says
`NotEstablished`, files the behavior-level diagnostic at the row's position — a bulk run says that
once for the behavior and every row is in one report, while a row handed over alone is the only
place its reader looks.

`shown(Locale)` is there for a consumer building an assertion message. Writing it turned up that
`Messages.render(said(), locale)` gives "This example does not hold." — the two values are on
`diagnostic.diff()`, so `shown` prints those and the notes. The test asserts both values appear.

The language is handed in, not picked. Two tripwires in `souther.compiler.diag` caught the first
attempt at `Locale.getDefault()` and then at `Messages.resolveLocale`, and both are right: what
answers a reader has to say which reader, and that is the consumer.

## 2 — header-less single file (P1)

A regression I introduced with the source-set fix, exactly as described. Single-source keeps its own
route now:

```
of(Path)                    -> ofDocuments({path: text}) + Front.DefaultName = stem
of(Path, ModulePath)          same
ofSource(String)            -> Compilation.ofSource(text, "Main", EMPTY)
ofSource(String, ModulePath)  same
of(List<Path>, ...)         -> ofDocuments, and delegates to the above when the list holds one
```

`ofDocuments` rather than `ofSource` for the Path form so the source is named by its path — a refusal
then quotes the file rather than the index. `aRefusalNamesTheFileItIsAbout` holds that.

## 3 — `observe` swallowing unexpected throwables (P2)

Right. It makes the cut a row makes now: overspending and stack exhaustion are `TheObservationRanOut`,
cancellation is rethrown as cancellation, and every other `RuntimeException` goes out. What the
implementation itself threw never reaches there — it is `InvocationFailure` and was already answered.

## 4 — source identity (P2)

`RecordedRow.at()` added, with the unnamed-ordinal-per-source reason in its javadoc. The Path form
names its source by the path, so `CompileException` quotes the file. The multi-source form uses
`ofDocuments` keyed by path for the same reason.

## 5 — thread safety (P2)

Made safe rather than documented away, since the API does claim parallelism is the caller's. The one
thing an evaluation mutates is the agreement cache, which is now a `ConcurrentHashMap` of
`ConcurrentHashMap`. Everything else a bound row touches is per-row (`FixtureReader`), already
concurrent (`EnsuresChecks.found`), or stateless (`Crossing`). A bound behavior takes no stand-ins,
so nothing defines a class during evaluation.

## 6 — finishing touches (P3)

`BoundExamples.moduleName()` added. Overloads are symmetric now — `of` and `ofSource`/`ofSources` each
take an optional `ModulePath`. `withBudget` stays on `SoutherExamples`: the budget is read when
`bind` builds the run, so moving it onto `BoundExamples` would mean rebuilding one, and its javadoc
says it must be set before `bind`.

`anImportedModuleIsReadFromTheDependencyPath` is the `ModulePath.ofClassPath` test you asked for —
`example.ids` compiled to a directory, `example.app` and its companion read from files against it.

## Verification

8503 tests across all modules, 0 failures. `CI=1 mvn -Plint verify` BUILD SUCCESS.

kawasima:
`57525ba`. Five of the seven were real and are fixed; one I am rejecting with the measurement; the
thread finding changed the design and is written down as such.

## The thread (3, 4) — adopted

Confirmed: `Deadline.ofMillis` spawns a `souther-reading` thread for every body. Narrower than
stated in one respect — a `DSLContext` holding a `Connection` works across threads, so the PR's own
example was not broken — but `ThreadLocal`-bound context is, and that is common enough in Java
integration testing that documenting it would not do.

`Deadline.onTheCallersThread()` is what the face binds with, so `evaluate` and `observe` both apply
where they were called from. The clock goes with it: counted limits are counted in the code and
thrown from it, so they hold either way, and a wall clock guards code this compile generated — of
which there is none past the crossing. `withBudget` is removed rather than left meaning nothing.

The round-2 guarantee that a non-returning implementation still returns is dropped, deliberately.

`aBoundImplementationRunsWhereItWasCalledFrom` holds both operations against
`Thread.currentThread()`.

This is the finding that taught the ADR something: world ownership includes execution context, not
just database setup. That sentence is now in ADR-0105, with why the clock was given up.

## 1 — E1930 never ran for a scalar answer — confirmed

`typeOf` → `valueTypeCandidate` → `identify` answers only for a generated value class, so a `-> Int`
answer is a `Long` with no case and the gate returned "kept" for every implementation. `Crossing`
walks `BoundaryOutput` now and an answer with no case crosses at the output's shape.
`aScalarAnswerIsHeldToWhatTheBehaviorDeclares` is `FailurePhase.ENSURES` where it was `COMPARISON`.

## 2 — gated on `injected` — confirmed

The gate reads `Applied.Bound`. The reason for the check is that nothing else ran it, which is a fact
about what applied the row, not about how the behavior is written.

## 5, 6 — confirmed

`withBudget`'s javadoc is moot now, the method being gone. `RecordedRow` has `equals`/`hashCode`,
identity on the enumeration and the written row.

## 7 — not a finding

The three arms answer differently because the three representations are different. `CodecGen` builds
a decoded map as a `java.util.LinkedHashMap` (`CodecGen.java:435`, `new_(CD_LinkedHashMap)`) and a
list as an `ArrayList`; only `Set` is a `PersistentHashSet`, which is why `Sets.fromList` appears in
that arm alone and in `JsonBoundary`'s. `Crossing` mirrors what the runtime holds. Making the three
uniform would introduce the second representation the comment warns about.

## Verification

8504 tests, 0 failures. `CI=1 mvn -Plint verify` BUILD SUCCESS.

kawasima:
`89143ca`. Both blocking points, and the collection test.

## 1 — `bind` had become a behavior-override API

Confirmed and closed as injected-only. The spec, the ADR and `rows()` all said a binding makes
runnable the rows that had nothing to run them; the code answered for any behavior whose ABI base
the instance matched. `bind` now refuses an instance for a behavior written with a body, and says
what is wrong with the instance rather than letting a row fail later.

I did not push the rule down into `Answering`. That seam is what applies a behavior for a run, and a
run may legitimately be given an answerer that applies anything — `AnAnswererInAnotherLoaderRunsARowTest`
applies *bodied* behaviors from a second loader, which is #732's whole point. Putting an
injected-only check there would break the seam to enforce a face's rule. So the behaviors a bound
answerer answers for are handed to it as a set: the admitting side's statement of what the instance
was supplied for, made once instead of derived twice where the two could differ. `Answering.bound`'s
javadoc now says which layer owns which question.

`aBehaviorThatImplementsItselfIsNotBound` holds the refusal.

## 2 — the ADR

Three sentences the code had left behind, all fixed: the observation running under a deadline because
"an implementation that does not come back would otherwise hang the caller's loop", `withBudget`, and
`The result is RowOutcome`. The stale comment in `ExampleVerifier.observe` went with them. The
injected/body wording needed no change once (1) closed injected-only — it was already right.

## 3 — the collection test

`AnAnswerIsReadIntoThisBuildAtEveryDepthTest` holds the `BoundaryOutput` walk at each depth a
declared type stands: a collection's elements are this build's classes, a nominal is, and a scalar
crosses as itself.

Held on the crossing rather than through a clause, and that was a decision rather than convenience. I
tried the `ensures List.all(...)` form first and spent four attempts on E1804/E1506 without learning
what an `ensures` admits — which is a separate question from what the crossing must do, and guessing
at it would have made the test about the wrong thing. Asserting the element's `Class` against the
run loader's is the direct statement: an element left as the other build's class is exactly what
makes an emitted guard match nothing, which is the defect the walk was added for.

## Verification

8508 tests, 0 failures. `CI=1 mvn -Plint verify` BUILD SUCCESS.

kawasima:
`6456a09`. The execution-policy finding is fixed by moving the boundary rather than by choosing a
side, and the Javadocs are closed.

## The seam

You are right that `withDeadline(onTheCallersThread())` was giving up more than the clock, and that
the fix is not to go back to the worker. What it was giving up is the reason
`EvaluationPolicy.workerStackBytes` is 64 MiB said outright: a counted depth limit of 50,000 is set
to be reached *before* the stack runs out, and on a JUnit thread it is not. One row could be decided
in a build and `INCOMPLETE` here.

I did not split `RowWork` into phases, and the reason is a fact the code makes plain:
`EvaluationContext.begin`/`Probe.begin` are per-thread and span the row, so two phases on two workers
would take the counting apart. The boundary that keeps both properties is the one you drew — what
each side owns — realised as a hand-off rather than a split:

```
caller thread                     souther-reading worker (64 MiB)
  evaluate(row) ──────────────────► fixtures, expectation, ensures-on-stated
                                    │
  runs it  ◄── Handoff.handOver ────┤  implementation.apply(...)
           ──── the answer ────────►│
                                    │ crossing, ensures-on-answer, comparison
  RowEvaluation ◄───────────────────┘
```

One hand-off per application, the worker alive on either side of it, so nothing the row was counting
is interrupted — and the implementation runs in the world the caller arranged. `Deadline` gained
`crossingBackToTheCaller(stackBytes)`; `Handoff.onThisThread()` is null for every run a compile
makes, so the compile-time path is untouched.

`TheRowStaysOnItsWorkerAndTheApplicationDoesNotTest` holds all four halves of it: the row's own work
is not on the calling thread, the handed-over work is, the worker reaches a depth a default stack
would not, and what the application threw comes back as it was thrown.

The ADR now says what is actually given up, and why running the whole row on the caller's thread
gives up something silently.

## Javadocs

`bind` said "whatever behavior it implements" — which is the wording the implementation went on to
take literally, so it is worth being exact about: it says "each behavior written without a body that
it implements" now, with an `@throws` for the three refusals. `one` said it answers "the row's
outcome and nothing beside it". The `RowEvaluation` in the class Javadoc was the old internal class,
now `RowWork`.

## Verification

8512 tests, 0 failures. `CI=1 mvn -Plint verify` BUILD SUCCESS.

PR #845, opened 2026-08-18
Name a fork by the construct written, not by the one it lowered to

### LINKED ISSUE #841: Coverage provenance preserves a construct's identity but loses what source construct it was
Three source constructs draw a boundary line, and a report calls all three `guard`. Measured on `develop` at `38026cc7` with `souther examples`:

| source | boundary origin | arm the branch measure names |
| --- | --- | --- |
| `if temp >= 240 then 1 else 2` | `guard@5:5` | `then` / `else` |
| `guard temp >= 240 else Low` | `guard@9:5` | `then` / `else` |
| `[ temp \| temp >= 240 ]` | `guard@5:5` | `then` / `else` |

Only the middle row's origin and only the first row's arms are words the author wrote. A reader following `guard@5:5` back to an `if` is looking for a form that is not there, and one following `then` back to a comprehension is looking for a word the language has no place for in that construct.

## This is one defect, not two

The origin word and the arm words come from the same place. `CoverageSites` walks the tree that runs and reads the source's meaning off its shape:

```java
case Core.If iff -> {
    armOf(Site.Kind.THEN, "then", iff, iff.origin(), 0, iff.then(), inside);
    armOf(Site.Kind.ELSE, "else", iff, iff.origin(), 1, iff.els(), inside);
    guards.add(new GuardRef(behavior, iff.origin(), ...));
}
```

and says so in its own class comment:

> A `guard … else` is an `if` by the time it gets here, so it needs no case of its own.

That is correct about Core and wrong about what is being measured. `Core` is the lowered form (ADR-0021), `guard` is sugar for `if` (ADR-0020), and a comprehension is rewritten by `Lower` — so by the time coverage is numbered, one node stands for three source constructs. Naming the construct, and naming what each of its outcomes means, are questions about the source, and they are being answered from the execution shape.

`OriginRef.describe` spelling `guard@` is the same answer one layer further on. Nothing on the path from the source to the report carries what the author wrote.

## What is missing is the kind, not the identity

`CoverageOrigin` already exists for exactly this crossing. It survives helper expansion and lowering so that copies of one fork are one obligation, and `lowered(part)` derives a comprehension's guards from the comprehension. It preserves **which** source construct this is.

What it does not carry is **what** that construct was. That is the whole of the gap.

## Design

**One owner for the source kind.** `CoverageOrigin` gains it. `AstBuilder.construct()` is the only place one is made and its five callers are the five source constructs a row can be owed for — a comparison, a list comprehension, an `if`, a `match`, a `guard` statement. The counter increments per call, so two origins cannot share an ordinal and adding the kind does not move identity. Every rewrite from Ast through Core already copies `origin()`, so expansion preserves it without a new rule to keep.

No Core node gains a surface-form field. `Core` keeps knowing only what runs.

**Outcomes become data.** `Site` carries a `String label` frozen at analysis time — `"then"`, `"else"`, `"case UnderThirty"`, `">="` — and a `Site.Kind` that puts a source outcome, a lowering mechanism and an instrumentation site in one enum. Both are replaced by a value that says what an outcome means:

```java
sealed interface SourceOutcome {
    record Held(HeldBy by) implements SourceOutcome {}
    record Failed(FailedBy by) implements SourceOutcome {}
    record Matched(List<TypeSymbol> cases) implements SourceOutcome {}
    record Compared(Hir.BinOp op) implements SourceOutcome {}
}

sealed interface HeldBy { record Condition() ...; record Construction() ...; }
sealed interface FailedBy { record Condition() ...; record Construction(Optional<String> clause) ...; }
```

It does not carry the construct. `Site` already reaches the origin through its `Obligation`, and a second copy would let a site say `GUARD` in one component and `IF` in the other. The meaning of an outcome is the pair, and the pair is `Site`.

`isArm()` moves onto the outcome: a `Compared` is not an arm and the other three are. The special case that `Site.Kind.COMPARISON` is today.

**The pair is checked where it is made.** Not every combination exists — a comprehension has no construction to succeed at, a `match` has neither. `Site` holds an exhaustive switch over the admissible set, so adding a construct or an outcome stops there and is decided about rather than falling through:

```
IF            × Held/Failed(Condition | Construction)
GUARD         × Held/Failed(Condition | Construction)
COMPREHENSION × Held/Failed(Condition)
MATCH         × Matched
COMPARISON    × Compared
```

**Words are chosen where a reader is.** No reader-facing string stays in `CoverageSites`. A phrase inserted into a sentence is localized at render time through `Localizable`, which is what `tok.*` and `kind.*` already do and what ADR-0101 leaves room for beside its rule that a value selecting a whole wording makes two records. So `if` / `guard` / the comprehension's own word, and `then` / the body past a guard / the element a comprehension drops, are catalog entries in both shipped locales rather than Java strings.

## Acceptance

| source | origin kind | outcomes |
| --- | --- | --- |
| `if c then a else b` | `IF` | `Held(Condition)` / `Failed(Condition)` |
| `guard c else x` | `GUARD` | `Held(Condition)` / `Failed(Condition)` |
| `[e \| c]` | `COMPREHENSION` | `Held(Condition)` / `Failed(Condition)` |
| `if T(v) as x then a else …` | `IF` | `Held(Construction)` / `Failed(Construction(clause))` |
| `guard T(v) as x else …` | `GUARD` | `Held(Construction)` / `Failed(Construction(clause))` |
| after helper expansion | unchanged | unchanged |
| after comprehension lowering | `COMPREHENSION`, derived per guard | per guard |

Every surface a reader meets says the same thing about one construct: the boundary origin in the report and in the JSON, the E1916 sentence and the label beside it, the E1918 sentence, and the branch measure's arms.

The last two rows are held by machinery rather than by a rule someone must remember, so they are checked by breaking the machinery: `lowered()` dropping the kind, and `HelperInliner` minting an origin instead of copying one, each have to turn a test red.

## Constraints

- `CoverageOrigin` is the only owner of the source kind.
- No `Core` node carries a surface form.
- `SourceOutcome` does not duplicate the construct kind.
- No reader-facing string is decided in `CoverageSites`.
- `adequacy-schema-1.json` grows by addition only: `then`, `else`, `case`, `constructed` and `departure` stay, and what no author wrote gets a word of its own. The schema's own preamble is the rule — a removal or a rename raises the version, an addition does not.

## Out of scope

Whether the measure should still be called `branch` and its unit an *arm*. ADR-0089 names it "arms of the body", and a `guard`'s successful side and a comprehension's filter are not arms the author wrote — but changing that is a decision about what Souther measures, not a repair of provenance, and it can be taken on its own once the outcomes are data.

kawasima:
An `if`, a `guard` and a guard comprehension are one `Core.If` by the time coverage is numbered (ADR-0020, ADR-0021), and `CoverageSites` read the source's own words off that node — saying so in its own class comment: *"A `guard … else` is an `if` by the time it gets here, so it needs no case of its own."* That is right about Core and wrong about what is being measured.

Measured on `develop` at `38026cc7` with `souther examples`, and again here:

| source | boundary origin | ways through it |
| --- | --- | --- |
| `if temp >= 240 then 1 else 2` | `guard@5:5` → `if@5:5` | `then` / `else` — unchanged |
| `guard temp >= 240 else Low` | `guard@9:5` — unchanged | `then` → `continued`, `else` unchanged |
| `[ temp \| temp >= 240 ]` | `guard@5:5` → `comprehension@5:5` | `then` / `else` → `kept` / `dropped` |

Only the middle row's origin and only the first row's arms were words anyone wrote.

## What was missing

`CoverageOrigin` already crosses lowering and expansion so that copies of one fork stay one obligation, and `lowered(part)` already derives a comprehension's guards from the comprehension. It preserved **which** construct this is. What it did not carry is **what** that construct was, and that is the whole of the gap — the origin word and the arm words are two symptoms of it, which is why they are one change.

## What this does

**One owner for the source kind.** `CoverageOrigin` gains a `CoverageConstruct`. `AstBuilder.construct(kind)` is the only place one is made and its five callers are the five constructs a row can be owed for; the counter increments per call, so no two origins share an ordinal and adding this does not move identity. Every rewrite from Ast through Core already copies `origin()`, so an expansion preserves it with no new rule to keep. No `Core` node learns a surface form.

**Outcomes become data.** `Site` carried a `Kind` that put a source outcome, a lowering mechanism and an instrumentation site in one enum, and a `String label` frozen in one language at analysis time. Both become a `SourceOutcome` — `Held(HeldBy)`, `Failed(FailedBy)`, `Matched`, `Compared`.

It does not carry the construct. `Site` reaches it through its `Obligation`, and a second copy would let one site say `GUARD` in one component and `IF` in the next — in a change whose point is a single truth for provenance. The meaning is the pair, so the pair is what `OutcomeName.of` is total over, and `Site`'s compact constructor refuses a combination no construct of the language has (a comprehension attempting a construction, a `match` settling a condition). `isArm()` moves onto the outcome, which is the special case `Site.Kind.COMPARISON` used to be.

`HeldBy` and `FailedBy` are separate because only the failing side of an attempt names a clause, and because `Held(boolean)` would be a boolean whose meaning the type does not state.

**Words move to where a reader is.** No reader-facing string is decided in `CoverageSites`. A phrase dropped into a sentence is localized at render time through `Localizable`, which is what `tok.*` and `kind.*` already do and what ADR-0101 leaves beside its rule that a value selecting a whole wording makes two records. `E1916`'s guard case becomes `NoRowIsAtTheLineAConstructDrew`, its label becomes `TheConstructThatDrawsTheLine`, and `E1918` takes its arm as a phrase:

```
No row is at price/temp = 240, which is where the `if` draws its line.
No row goes through the empty list the comprehension yields in `pick`.
`pick` の内包表記が空リストを返す側を通る行が、ありません。
```

**The schema grows by addition only.** `then`, `else`, `case`, `constructed` and `departure` all stay and are all still written; `continued`, `kept` and `dropped` are words no earlier document carried, and `construct` is an optional key saying which construct an `else` belongs to. That is the rule the schema's own preamble states, so `schemaVersion` stays 1.

## How the acceptance matrix is held

`AnOutcomeIsNamedByWhatWasWrittenTest` walks the five source shapes, the admissible pairs, and a comprehension with two guards inside a helper called twice — one construct, lowered into two forks, each copied twice.

The last two rows of the matrix are held by machinery rather than by a rule someone has to remember, so they were checked by breaking the machinery and watching the test go red:

| break | red |
| --- | --- |
| `lowered()` drops the kind | 4 |
| `HelperInliner` mints an origin instead of copying one | 2 |
| `Lower` stops deriving a fork per guard | 1 |

`CI=1 mvn -Plint verify` is green.

**One thing the green does not say.** The conformance corpus does not exercise any of this: its only boundary origins are `guard@1:73:5`, which is a real `guard`, and it has no unreached arm at all. The corpus says this broke nothing; what says the routes are fixed is the test above and the five models run by hand.

## Out of scope

Whether the measure should still be called `branch` and its unit an *arm*. ADR-0089 names it "arms of the body", and a `guard`'s successful side and a comprehension's filter are not arms anyone wrote — but that is a decision about what Souther measures rather than a repair of provenance, and it can be taken on its own now that the outcomes are data.

Closes #841

### DIFF
--- a/souther-cli/src/test/java/souther/cli/AComparisonInAnEnsuresIsOfferedARowTest.java
+++ b/souther-cli/src/test/java/souther/cli/AComparisonInAnEnsuresIsOfferedARowTest.java
@@ -60,20 +60,24 @@ class AComparisonInAnEnsuresIsOfferedARowTest {
     }
 
     /**
-     * A clause's line is met by writing the value; a guard's at the same value is not.
+     * A clause's line is met by writing the value; a fork's at the same value is not.
      *
      * <p>The distinction the whole thing turns on, in one report. The row hands the behavior a
      * hundred and takes the branch above the comparison, so it never reaches
-     * {@code a.n.value > 100} — the guard's line at a hundred is unmet, and the clause's is met,
+     * {@code a.n.value > 100} — the fork's line at a hundred is unmet, and the clause's is met,
      * because every rule of a declaration runs whenever the behavior answers.
+     *
+     * <p>The fork is named {@code if} because that is what this body writes. Matched on the word so
+     * that the two rules are told apart by what drew each of them; a report naming every fork
+     * {@code guard} would have this pass while sending a reader after a construct that is not there.
      */
     @Test
-    void aClausesLineIsMetByWritingTheValueAndAGuardsIsNot() throws Exception {
+    void aClausesLineIsMetByWritingTheValueAndAForksIsNot() throws Exception {
         List<String> gaps = boundaryGaps(reportOf(REACHED_BY_ONE_RULE_ONLY));
 
         assertTrue(gaps.stream().anyMatch(line -> line.contains("charge/a.n = 100")
-                        && line.contains("guard")),
-                () -> "no row got the guard's comparison to answer at a hundred: " + gaps);
+                        && line.contains("if@")),
+                () -> "no row got the fork's comparison to answer at a hundred: " + gaps);
         assertFalse(gaps.stream().anyMatch(line -> line.contains("charge/a.n = 100")
                         && line.contains("ensures")),
                 () -> "the row writes a hundred, which is the whole of what the clause wants: "
@@ -107,7 +111,7 @@ class AComparisonInAnEnsuresIsOfferedARowTest {
      * <p>A clause comparing one input against another draws a line where the two hold one count. It
      * is on neither of them, so it has no axis and no class either side — and what meeting it takes
      * is the same answer the clause's other lines get: the row that puts one count in both
-     * positions has met it. The guard drawing the same line is not met, because the row takes the
+     * positions has met it. The fork drawing the same line is not met, because the row takes the
      * branch above the comparison.
      */
     @Test
@@ -115,8 +119,8 @@ class AComparisonInAnEnsuresIsOfferedARowTest {
         List<String> gaps = boundaryGaps(reportOf(A_LINE_BETWEEN_TWO_POSITIONS));
 
         assertTrue(gaps.stream().anyMatch(line -> line.contains("book/from = to")
-                        && line.contains("guard")),
-                () -> "no row got the guard's comparison to answer on the line: " + gaps);
+                        && line.contains("if@")),
+                () -> "no row got the fork's comparison to answer on the line: " + gaps);
         assertFalse(gaps.stream().anyMatch(line -> line.contains("book/from = to")
                         && line.contains("ensures")),
                 () -> "the row puts one count in both positions, which is what the clause wants: "
--- a/souther-cli/src/test/java/souther/cli/AnArmNothingReachesIsNotOwedARowTest.java
+++ b/souther-cli/src/test/java/souther/cli/AnArmNothingReachesIsNotOwedARowTest.java
@@ -305,9 +305,13 @@ class AnArmNothingReachesIsNotOwedARowTest {
     // --- what happens if the proof is wrong -------------------------------------------------------
 
     private static CoverageSites.Site arm(int index) {
-        return new CoverageSites.Site("classify", CoverageSites.Site.Kind.THEN, "then", null,
-                index, index,
-                new CoverageSites.Obligation("classify", CoverageOrigin.written("t", index), 0));
+        return new CoverageSites.Site("classify",
+                new souther.compiler.coverage.SourceOutcome.Held(
+                        new souther.compiler.coverage.SourceOutcome.HeldBy.Condition()),
+                null, index, index,
+                new CoverageSites.Obligation("classify",
+                        CoverageOrigin.written("t", index,
+                                souther.compiler.types.CoverageConstruct.IF), 0));
     }
 
     /**
--- a/souther-compiler/src/main/java/souther/compiler/coverage/CoverageSites.java
+++ b/souther-compiler/src/main/java/souther/compiler/coverage/CoverageSites.java
@@ -4,6 +4,7 @@ import souther.compiler.ast.Hir;
 import souther.compiler.core.Core;
 import souther.compiler.diag.Citation;
 import souther.compiler.diag.SourcePos;
+import souther.compiler.types.CoverageConstruct;
 import souther.compiler.types.CoverageOrigin;
 import souther.compiler.types.TypeSymbol;
 
@@ -32,8 +33,15 @@ import java.util.Map;
  * invariant clause gets none either: it is a property of a type, not a fork in this body, and whether
  * the rows reach its edges is a different measure asked another way.
  *
- * <p>A {@code guard … else} is an {@code if} by the time it gets here, so it needs no case of its own.
- * Helper bodies are not walked: a non-recursive helper is already inlined into the body that uses it,
+ * <p>A {@code guard … else} and a comprehension's condition are both an {@code if} by the time they
+ * get here, so the walk has one case for the three of them — and what they are is not read off that
+ * case. Which construct the author wrote is carried from where the source was read
+ * ({@link souther.compiler.types.CoverageConstruct}), and what one way through it means is a
+ * {@link SourceOutcome} beside it. Deciding either from the shape of the lowered node is answering a
+ * question about the source out of the tree that runs, which is how a comprehension came to be
+ * reported as a {@code guard} with a {@code then} arm.
+ *
+ * <p>Helper bodies are not walked: a non-recursive helper is already inlined into the body that uses it,
  * and a recursive one is a shared method rather than a fork in any one behavior.
  */
 public final class CoverageSites {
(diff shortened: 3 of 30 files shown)


kawasima:
All three are fixed in cc1f5d0f. Taking them in the order you raised them.

## 1 — `COMPARISON` on every binary expression

Confirmed, and worse than the report said. `AstBuilder.binary()` mints one origin for whatever `binOp()` accepts, which is `EQ NE LT LE GT GE AND OR ADD SUB MUL DIV CONCAT` — so `xs ++ ys` carried `kind = COMPARISON`, and so did the `&&` in `a >= 10 && b > 0`, which `CoverageSites.comparisons()` walks into rather than numbering. The value was untrue of a valid program on the day it was introduced.

`BINARY` now. What the source wrote is a binary expression; being read as a comparison a row can reach is what happened to some of them, and `SourceOutcome.Compared` was already saying that. The admissible pair is `BINARY × Compared → COMPARISON`, and neither axis names the other any more.

Held by `everyBinaryTheSourceWroteIsABinaryAndOnlySomeAreSites`, over `if a + b >= 10 && a > 0 then a * b else a - b`: six binary expressions, all `BINARY`, two of them sites.

## 2 — `isAGuard()`

`wasDrawnInABodyFork()`. Your reading of the risk is the one that decided the name: the predicate is now the thing `Adequacy.warning()` reads to choose the construct-aware message, so a reader repairing the old name into `kind == GUARD` would have broken `if` and comprehension and left `guard` passing — silently, since the set of findings would not move. The javadoc says that outright, so the next person to notice the mismatch finds the reason rather than the mismatch.

Six call sites, four of them tests.

## 3 — reader-facing words in `CoverageSites`

You are right that the declaration and the implementation had come apart, and moving is the better of the two ways to close it — `Site` had two consumers with different needs and would have grown a method per consumer.

`label` is now `ArmVocabulary.label(Site)` in `souther.compiler.report`, which is where both readers of it live: the table's line, and the document's `branch.unreached[].label` and `findings[].subject`, which join on that word and would join to nothing if either were spelled a second time. `GeneratedRows` reads the same function.

The catalog phrase is now `Adequacy.phraseFor(Site)`, beside the sentence that says it.

Both switch on `OutcomeName`, not on the product — so the pair is still admitted in one place, and a construct added to the language arrives at each encoder as a case with no word rather than one quietly answered with a neighbour's.

## On the `equals` note

Agreed, and for the reason you give. `AstBuilder` takes a fresh ordinal per construct, so no two origins can share `(module, ordinal)` and the kind cannot change what any comparison answers; a broken one being a different value is the direction that gets noticed. `CoverageOrigin`'s compact constructor also holds `ordinal < 0` and `NOT_WRITTEN` to each other, so `isWritten()` and the kind cannot disagree.

`CI=1 mvn -Plint verify` is green on the rebased tree, and the five models still read as they did in the description.

PR #862, opened 2026-08-19
Read a quotient's divisor as a range, as a product's factors are

### LINKED ISSUE #858: A quotient by a value the path bounds away from zero is read as no fact, though a product of two such values is bounded
A quotient's divisor is read only where it is a written number. Where the path bounds it away from zero — a day count declared between one and thirty-one, a guard above zero — the quotient is a value nothing is known of, and the construction under it is reported as a clause the model does not establish. A product of two values in the same position is bounded.

## Measured

`Money` is `Int` with `invariant value >= 0`; `Days` is `Int` with `invariant value >= 1 && value <= 31`.

| written | reported |
|---|---|
| `Money(a.value + b.value)` | — |
| `Money(a.value * b.value)` | — |
| `Money(a.value / 2)` | — |
| `let k = 2` then `Money(a.value / k)` | — |
| `Money((a.value + b.value) / 2)` | — |
| `Money(a.value / b.value)` | E2011 |
| `guard b >= 1` then `Money(a / b)` | E2011 |
| `guard b > 0` then `Money(a / b)` | E2011 |
| `Money(a.value * b.value / c.value)` | E2011 |
| `AtMost1000(a.value / b.value)` | E2011 |

The bound that goes unread is not one end or the other: an invariant bounding above alone is owed the same way. Binding the divisor to a written constant first is read, so the line is not how the divisor is spelled — it is whether the form has any coefficients at all.

## Where

`Terms.quotient` answers null unless the divisor's form is a bare constant, and `Derivation.Quotient` holds that constant as a `long`. `Intervals.truncatingQuotient` takes the same `long` and puts the dividend's two ends through it. A divisor with a coefficient never reaches any of the three, so nothing is recorded and the atom the whole quotient is named as stays unrelated to what it was computed from.

`Terms.quotient`'s own doc states this as the rule: "A divisor the path only bounds away from zero is not read either — it is the sign and the magnitude of a written number that the rule is about." So this is a decision that was taken, not an omission, and `AProductIsBoundedByWhatThePathBoundsItsFactorsToTest.aQuotientByAValueTheGuardsOnlyBoundIsStillOwed` asserts it.

## Why the decision does not hold up

The sign and the magnitude are what the rule needs, and a range held away from zero has both. `*` is already read this way: `Intervals.product` takes two ranges and reads the four corners, and a product of two values the path bounds is discharged. A quotient by a divisor of one sign is monotone in each operand exactly as a product is, so the same four corners bound it — dividing by the near end of the divisor is one end of the answer and by the far end is the other. Truncation toward zero is monotone, so putting the corners through it after is sound.

What a divisor a range admits zero at leaves is still nothing, and for a reason that has not changed: the run time aborts there (spec §stdlib-int), and a range straddling zero sends the quotient past every value on both sides. That case stays as it is.

A guard above zero over the whole numbers arrives here as an end at one — `NumericDomain` sharpens a strict bound on a discrete atom onto the adjacent count — so the guard an author actually writes is one this can read.

## What it costs an author today

The quotient can be bound to a name and guarded:

```
let q = a.value / b.value
guard q >= 0
    else Bad
Money(q)
```

That discharges. What it costs is a refusal in the behavior's answer type for a condition that cannot happen — `Bad` is unreachable, and every caller has to handle it. Apportioning by a day count is where this is written, so it is not a corner of the language.

## Not to lose

- `aDivisorNoIntCanHoldIsNotOneTheRuleReads`: a written number no `Int` holds is not a divisor the operator has, and the clause stays owed. Reading the divisor as a range must not make that number readable.
- `aDecimalQuotientIsNotRead`: `/` on `Decimal` rounds to a precision the run time sets, and where that leaves an end is not read. The rule stays over whole numbers.
- The ends of a quotient are widened rather than sharpened; an operand end that is not reached must not tighten the answer.

Related: the answer of a `List.fold` is bounded no way either, for a different reason — #859.

kawasima:
Closes #858.

## What was wrong

`Derivation.Quotient` held its divisor as a `long`. The representation can only name a written number, so `Terms.quotient` had to refuse every divisor with a coefficient in it and `Intervals.truncatingQuotient` could only take one. The rule as stated — that it is the sign and the magnitude of a *written* number that the rule is about — described that field rather than the arithmetic. Division by a range held off zero moves one way along each operand with the other held, so its extremes sit at the corners of the box the two ranges make, exactly as a product's do.

What it cost an author was a refusal in the behavior's answer type for a condition that cannot happen: the quotient had to be named and guarded, and every caller then had to handle a case that is unreachable.

## What it now does

The divisor is a form beside the numerator, and three things are settled with it.

**Applicability is decided where the recipe is read, not where it is recorded.** One expression is read under more than one set of assumptions, and the same divisor is held off zero by one and not by another. `Derivation` records what was computed; `DerivedBounds` decides what follows. Choices about the *operator* — `/` on `Decimal` rounds at a precision the run time sets and still has no rule — stay at the recording site, because they are not path-dependent.

**`Intervals.truncatingQuotient` bounds the values the operation produced.** `/` does not answer everywhere: a zero divisor aborts, and so does the one pair whose quotient no `Int` holds. A pair it aborts on leaves no value to bound, so a range coming back — top included — does not say the operation is defined for every pair the operands admit. That contract is written into the API doc, and it is why `Long.MIN_VALUE / -1` needs no case of its own. A corner where both ends run past every value is no number and is not a candidate; nothing is lost, because a divisor held off zero has at most one such end.

**The operator's divisor is a value of its own type.** `Derivation.Quotient` carries the divisor's extent, and the reading meets what the path proves with it. Where the two share nothing, this operator has no divisor here and no fact is derived — *not* a range holding no value, which taken into the domain would be a contradiction, and a contradictory domain proves every clause there is. A rule with no operands to fire on is not a proof that the path has no execution. This is what keeps `x / (9223372036854775807 + 1)` owed, and it now states that once for every spelling rather than only for a written constant.

Read only on the divisor, deliberately. The dividend's extent would only sharpen a bound that is already sound; the divisor's decides whether the rule applies at all.

## Three things told apart

| | means | where |
|---|---|---|
| domain bottom | there is no execution here | `NumericDomain` |
| no overlap | this rule has no operand | `DerivedBounds`, before `Intervals` is called |
| quotient top | there is an execution, and no bound to give | `Intervals.truncatingQuotient` |

## Spec

`§invariant-discharge-arithmetic` is rewritten from *A quotient by a written divisor* to *A quotient*. It leads with the corner hull; monotonicity appears as the reason the corners are the extremes, qualified with the direction turning on the other operand's sign. The condition is written as what the facts at the construction establish, not as a shape of today's `Bounds` — so a `NumericDomain` that later carries a disjunctive range needs no spec change. The rule about what `Int.floorMod` guarantees (`§invariant-discharge-guarantees`) is a different question and is untouched: that one holds wherever the call is written.

## Also

`NumericDomain.Bounds.isEmpty` is renamed `saysNothing`. Beside the `holdsAValue` this change needed, the two names said opposite things about one range: a range with neither end holds every value, and one whose ends have crossed has both ends and holds none.

## Checked

`CI=1 mvn -o -Plint verify` — BUILD SUCCESS across every module.

The new witnesses were held to failing without the change: restoring the old gate turns the five discharge and refutation cases red, and dropping the extent meet turns those five plus `aDivisorNoIntCanHoldIsNotOneTheRuleReads` red. That last one now also reads `InvariantChecker.GAVE_UP` and the verdict, since this check is fail-open and a warning count alone would pass on an analysis that stopped at the divisor.

A defect found while writing it: `signOf` read a divisor whose lower end is negative and strict — `(-5, -1]` — as lying above zero. `aDivisorBelowZeroIsBelowItWhicheverOfItsEndsAreReached` covers it. No discrete carrier reaches that shape today, since a strict bound on one is sharpened onto the adjacent count.

No conformance corpus answer moved, which means no corpus program divides by a range. Worth adding one, separately.


kawasima:
P1 を直しました（`c2cd875d`）。指摘のとおりです。

`[0, 100] / [0, 5]` は `Int` なら正常終了する除数が `{1..5}` で、結果は `[0, 100]` に収まる。「零にいくらでも近い除数」は dense にしか存在しないので、私が書いた理由は `Int` について偽でした。しかも `Intervals` は range しか受け取らず、range は values の間隔を語らないので、そもそもその判断ができる場所ではない。「ここにあるのは算術だけ」と宣言しているクラスで、算術でないことを算術の顔で答えていました。

零の除外は discharge rule の premise なので、`DerivedBounds` の no-overlap と同じ場所に上げました。

```java
Bounds divisor = boundsOf(q.divisor(), ...).meet(q.divisorExtent());
if (!divisor.holdsAValue() || divisor.admits(Count.ZERO)) {
    return new Bounds(null, null);
}
return Intervals.truncatingQuotient(boundsOf(q.numerator(), ...), divisor);
```

`truncatingQuotient` の precondition は「値を一つ以上持ち、零を含まない」の二つになり、答えは corner hull だけになりました。`@param` に、零を含む range を答えられない理由（間隔次第で答えが変わる／whole numbers では有界、dense では非有界）を書いています。

`DerivedBounds.quotient` の javadoc には、これが商についての主張ではないことを明記しました。

> Neither answer is a claim about the quotients. In particular the second is not a statement that they run past every value: over the whole numbers a divisor between zero and five divides by one at the nearest, and the successful divides are bounded. What is said is that this rule does not establish where they are.

指摘のあった非対称も消えています。`MIN / -1` は premise と無関係な successful-result arithmetic の partiality なので transfer の中に残り、零は premise なので外に出た。両方が partiality だから同じ扱い、という混線がなくなりました。

仕様も premise として書き直しました。

> Nothing follows from this rule unless what is established at the construction keeps zero out of every divisor it admits. That is the condition for stating anything and not a claim about the divides that do answer — over the whole numbers a divisor between zero and five divides by one at the nearest, and where those quotients lie is simply not what this rule says.

`Intervals` 側のテストは `aDivisorThatAdmitsZeroBoundsTheQuotientNoWay` を `aRangeAdmittingZeroIsRefusedRatherThanAnswered` に置き換え。除数が `(0, 5]` で strict に零を外している場合は依然として本当に非有界なので、そちらのテストは残しています（ここは range の端だけで決まる）。

この premise が効いていることも確認しました。`|| divisor.admits(Count.ZERO)` を外すと `aQuotientByAValueNothingHoldsAwayFromZeroIsStillOwed` が赤になります（`Intervals` の precondition 違反を fail-open が飲み込んで、E2011 が出なくなる）。

`CI=1 mvn -o -Plint verify` は全モジュール BUILD SUCCESS。プログラムから見える挙動は変わっていません。

kawasima:
バックグラウンドのレビューから4件。3件を `7b414d3a` で直し、1件は指摘が誤りです。

**直した3件**

`OrderedInterval` の javadoc が `NumericDomain.Bounds` の `isEmpty` を名指していました。改名で消えたメソッドです。しかもこのコメントは二つの型の「空」が逆の意味だと説明するために存在するので、まさにその疑問を持った読み手が最初に着く場所でした。

`sameDerivation` が二つの extent を record の `equals` で比べていました。`BigDecimal.equals` に落ちるので `0.00` と `0` が別の場所になる。`Count` の javadoc が「代数の比較は全部 `compareTo` か `Place.sameAs` を通る」と書いている通りに直しました。今日は両方が同じ `Carrier.extent()` 由来なので発火しませんが、発火した場合に出るのは `OneTermTwoDerivations` — 握り潰さない内部エラーなので、一致している二つの読みが scale の違いで hard error になっていた。

`extentOf` が `Carrier.ofValue` の戻りを無条件に参照していました。null を返しうる（`BOOL, RAW -> null`）。`/` がどの operand を取るかは型付けが決めることでここの保証ではないので、carrier が無ければ規則を降ります。null 参照は fail-open の catch に飲まれて behavior 全体の解析を黙って落とすだけなので、読めないものは降りるという既存の扱いに揃えました。

**直していない1件**

仕様の `<<primitive-arm>>` 参照が誤りで `<<stdlib-int>>` にすべき、という指摘。これは違います。`§primitive-arm` の末尾（`specification.adoc:3407`）に

> Where the model has already ruled the failure out — a `+require+` (<<attempted-construction>>) or the divisor type's own invariant keeps the divisor away from zero — the union need not arise: the operator `+/+` divides, and its abort on zero (<<stdlib-int>>) cannot be reached.

とあり、私の文が指している「abort を排除するのは除数の型か `require`」はここに書かれています。`§stdlib-int` は abort する事実を述べる節で、排除の仕方は述べていない。参照先はこのままにします。

`CI=1 mvn -o -Plint verify` は全モジュール BUILD SUCCESS。

Issue #867, opened 2026-08-19
An expansion binding a value the arithmetic cannot read is one atom, so a step loses the places under it


kawasima:
`Terms.affine` reads through a binding an expansion introduced — that is what makes a helper's
arithmetic the caller's arithmetic — and it gives up on the whole binding where the value bound to
it is not a number. A helper taking a record binds a record, so a reduction's step that calls one is
a single atom, and the walk is owed (E2011) whatever the step does.

```souther
data Yen = Int
    invariant nonNeg = value >= 0

data Item = { n: Yen }

let dbl   (n: Int) : Int = n * 2
let fee   (i: Item): Yen = i.n
let seven (i: Item): Int = 7
```

| step of `Yen(List.fold(<step>, 0, xs))` | reported |
|---|---|
| `(acc, i) -> acc + i.n.value` | — |
| `(acc, i) -> { let v = i.n.value ... acc + v }` | — |
| `(acc, i) -> acc + dbl(i.n.value)` | — |
| `(acc, i) -> acc + fee(i).value` | E2011 |
| `(acc, i) -> seven(i)` | E2011 |

Rows three and four differ in the type of the helper's parameter and in nothing else. Row five says
it is not about what the helper answers: the body is the literal `7`, and a step answering a written
number is one the rule discharges when it is written at the call
(`AWalkFromASeedIsBoundedByCheckingOneStepTest.aStepThatIgnoresItsAccumulatorIsTheSeedJoinedWithWhatItAnswers`).

Row four's step, after expansion, is

```
Binary[ADD, Read(acc),
       FieldAccess[target = LetIn[binder = i,
                                  value  = Read(i: Item),
                                  body   = FieldAccess[Read(i), n]],
                   field  = value]]
```

and the form recorded for it is

```
LinearForm[constant=0, coefs={ handed(foldFrom(...), 0)                    = 1,
                               let(handed(foldFrom(...), 1), #0.0.n)       = 1 }]
```

The element's field comes out named `let(<element>, #0.0.n)`. What the element's type guarantees is
recorded at `<element>.n` (`StepInputFacts.java:119`, `terms.under(root, path)`). Two names for one
place: the fact never reaches the atom, and `namedByTheStep` drops it besides.

## Where it is

`Terms.affine`'s `case Core.LetIn` (`Terms.java:257`) reads the bound value first and yields null
where that has no form, and `affineOf`'s leaf answers null for a value that is not a number. So a
[...]

PR #872, opened 2026-08-19
Read a `match` handed as a value as the arms it answers one of

### LINKED ISSUE #866: A `match` handed as a value is not opened, so what its arms and its type say is lost
An `if` standing where a value is handed is opened — read once with each branch there, under what
the condition settles — and a `match` in the same position is not. What a `match` answers is one
atom nothing is recorded against, so a construction downstream of it is owed (E2011) even where
every arm is written out, and even where the declared type of the value it answers states the very
clause the construction needs.

```souther
data Yen = Int
    invariant nonNeg = value >= 0

data Fast
data Slow
data Plan = Fast | Slow

let twice (a: Yen): Yen = Yen(a.value * 2)
```

| written | reported |
|---|---|
| `twice(if n > 0 then a else a)` | — |
| `twice(match p with \| Fast -> a \| Slow -> a)` | E2011 |
| `twice(match p with \| Fast -> Yen(4500) \| Slow -> Yen(3500))` | E2011 |
| `Yen((match p with \| Fast -> a \| Slow -> a).value * 2)` | E2011 |

Row three is what says how much is lost: both arms are numbers written out, above the clause's own
end, and neither reaches the construction. Rows two and four say the rest of it — `a` is declared
`Yen`, so `a.value >= 0` holds of whichever arm ran, and there is no way for the reading to get at
it.

Row one is the control. The same shape written with `if` discharges, so what the rows above report
is the form and not the branching.

No helper and no reduction appears in any of these. A helper is how the case is met in practice: a
`let` whose body is a `match` is expanded at the call, so what stands at the parameter is the
`match`, and the parameter's declared type stops bounding it. That is what an author sees as "the
invariant of a helper's answer is not assumed", and it is neither about helpers nor about answers.

## Where it is

`ConditionalSite` (`InvariantChecker.java:2032`) holds a `Core.If`. `conditionalValueIn` and
`conditionalIn` (`InvariantChecker.java:2134`) walk *into* a `Match` looking for an `If` underneath
it, and never offer the `Match` itself. What they find is opened at `InvariantChecker.java:1305` —
once per branch, in the scope the conditional stands in.

## The rule

A `match` standing where a value is handed is one of its arms, and which one is decided by its
scrutinee. So it is read once with each arm standing there, under what that arm's pattern settles of
the scrutinee, and what the readings find is said once — which is what the walk already does for a
`match` in tail position (`InvariantChecker.java:1382`) and for an `if` in a value position.

An arm's binding is the part an `if` has no counterpart to. An arm that binds a name has it in scope
for its body alone, so the site carries that binder the way `ConditionalSite.of` carries a `let`'s
today.

## Not this issue

A `match` written inside a reduction's step is read by `Terms` and not by this walk, and `Terms` has
no arm for one either. Fixing this leaves that case reported. The step's own reader is #867.

kawasima:
Closes #866.

An `if` standing where a value is handed was opened and a `match` in the same position was not, so
what a `match` answered was an atom nothing was recorded against — a construction downstream of it
was owed whatever the arms were written as, and whatever the type of the value they answered
declared.

```souther
let asBig (a: Yen): Big = Big(a.value)   // Yen: value >= 0, Big: value >= 1000

let charge (p) = asBig(match p with | Fast -> Yen(4500) | Slow -> Yen(3500))   // E2011
```

A helper is how it is met in practice: a `let` whose body is a `match` is expanded at the call, so
what stands at the parameter is the `match`. To an author that reads as the invariant of a helper's
answer not being assumed, and it is neither about helpers nor about answers.

## What the change is

The concept was *a conditional a value is handed*, and its type was `Core.If`. It is *a case split*
now — a node answering one of several arms, where which one is decided by something a reading can
assume. `SplitSite` holds the node, `armsOf` enumerates the arms with what choosing each one
settles, and the walk never asks which form it was handed. An arm is entered with its scrutinee, so
what it binds stands for the value already there.

Three things moved with it, and each is where adding a `Match` case to the walk would have left
something behind.

**`say` and `Judgment.of` were written for two readings.** A `match` has an arm per case. `say`
folds now; what that rests on is `Judgment.of` being a join written so that the order branches are
combined in decides nothing, and that is said where the fold is.

**The arm-lifting machinery was the workaround for the missing rule.** `PathEngine.enteringLiftedArm`
and `SplitSite.of(Core.Case)` reached an `if` written inside an arm by replaying the arm's binder
with no scrutinee, and the javadoc called it a limit — nothing there said which value the arm
opened. Opening the `match` where it stands removes the limit, so they go, and `opens`'
`scrutinee == null` branch goes with them rather than staying a shape no producer makes.

**The opening budget was counted in splits.** Three nested `if`s are eight readings of a body, which
is what `BRANCHES_OPENED = 3` meant; three nested `match`es over a sum of twelve are 1728 — the same
bound naming a cost two orders of magnitude apart. `READINGS_A_PATH_MAY_COST = 8` opens a split
while the path has not already multiplied past it, so `if`s open exactly as far as they did, a wide
`match` is opened once however wide it is, and nesting them stops.

| a body handing N distinct matches over a C-case sum to one sum | splits opened | readings | time |
|---|---|---|---|
| C=12, N=4, before | 157 | 1884 | 667 ms |
| C=12, N=4, after | 1 | 12 | 24 ms |
| C=8, N=4, before | 73 | 584 | 178 ms |
| C=8, N=4, after | 1 | 8 | 16 ms |
| C=2 (an `if`'s width), N=4, either | 7 | 14 | 15 ms |

## What moved that was already written down

Two checked-in answers, both in `AnAnswerIsSeededOnceWhereItStandsTest`. An answer standing in an
arm was seeded twice because the arm was entered again under each branch of a conditional standing
over it. The arm is a reading of its own now and a split beside it is opened inside that reading, so
the answer is seeded where it is put in and the readings below start from where that seeding stands.
That leaves every place in the list seeded once and the `readings` column saying nothing, so the
test states *an answer is seeded once wherever it stands* instead — keeping the slope over a growing
number of answers that #826 put there.

## Tests

`AMatchHandedAsAValueIsReadWithEachArmStandingThereTest`, ten of them. Every case asserting silence
has a neighbour differing in one thing that is not silent: an arm below the clause, a value of a
type declaring nothing, a field beside the one the arm answers, an unproven construction written in
the scrutinee. The five asserting silence were watched failing with the production change taken
away; the five neighbours pass either way, which is what makes them controls.

`CI=1 mvn -o -Plint verify` green across every module.

## Not in this change

A `match` written inside a reduction's step is read by `Terms`, which has no arm for one either.
That is #867, a separate reader with a separate gap, and it stays reported.


kawasima:
Pushed `ea27275c` for the review.

**The budget was reading the wrong side of the multiply.** `cost < LIMIT` gates on what the path has cost already and then spawns `cost * width` readings, so a path fifteen readings long admitted a split of any width there is. `opens(cost, width)` asks what opening it would bring the path to. The first split on a path is still opened however wide it is, and the reason is written where the bound is: a sum's cases are the model's, and a rule that read fewer of them the more cases the model named would be a rule about the declaration rather than about the expression.

**The finding about arms was real, and its window was narrower than stated.** A conditional in a *value* position inside an arm — a tail-position one costs no budget and was never affected. Measured over sums of three to nine cases, with the conditional in an arm:

| cases | develop | branch at 8 | branch at 16 |
|---|---|---|---|
| 3, 4 | discharges | discharges | discharges |
| 5, 6, 7 | discharges | **owed** | discharges |
| 8 | owed | owed | **discharges** |
| 9+ | owed | owed | owed |

So eight or more was already owed on develop; the regression was five to seven. The bound is sixteen now, which moves no answer anywhere and makes a sum of eight cases better than develop. It also opens a fourth conditional where three were opened before, which is the same bound spent on the shape it was written for. Both ends are pinned in `HowFarSplitsAreOpenedIsBoundedByWhatARereadingCostsTest`.

**Three switches over one set of forms** is right and is now one. `splitOf` answers what a split asks and what its arms are, and what choosing an arm settles is deferred behind `Choosing` — so the width is read without entering anything, which is also what the budget needed.

**Three arms.** `AMatchHandedAsAValueIsReadWithEachArmStandingThere` gains a three-case sum: an arm below the clause said wherever it stands (the failing arm is the last one), and two arms leaving two different clauses unsettled leaving both (`unsettled=lo, hi`). Mutating `say` to fold only the first two readings makes both fail — the second by turning an E2011 into an E2010, since dropping the third reading makes a refuting arm decisive.

`CI=1 mvn -o -Plint verify` green.

PR #873, opened 2026-08-19
Say a rule went unread where the reading of ends gave up

### LINKED ISSUE #868: An invariant clause that places no end is dropped without a word, though the guard reader names the same rule as unread
ADR-0090 says the two producers of a line say the same thing when they cannot read a rule:

> A rule this could not read is named, whichever rule it was. A `guard`'s comparison already said
> so; an invariant's bound did not, and a bound it dropped left the position looking like one no
> rule bounds — which at a position whose only rule was that bound made the report state the
> opposite of the declaration. The two producers now say it in the same words, because they answer
> the same question: this position was written about and this could not draw the line.

They do not, for a clause that compares an arithmetic form. A `guard` names it:

```
  · not read: order.straw (a rule about it is one this compiler did not read)
  · not read: order.straw (the comparison relates it to another position rather than dividing it)
```

An `invariant` of the same shape says nothing at all. For

```
data Cm = Int
    invariant value >= 0

data Parcel = { length: Cm, width: Cm }
    invariant Int.add(length.value, width.value) <= 150
```

`souther examples` reports

```
boundary    0/2
  ! no row is at quote/parcel.length = 0 (invariant Cm #1)
  ! no row is at quote/parcel.width = 0 (invariant Cm #1)
```

The `Parcel` clause is nowhere: not as a line, not as a rule that went unread, not as a position
this could not divide. A reader is left with a model whose only stated rule about `length` looks
like `value >= 0`.

The rule is not being ignored elsewhere. A row written past it is refused at construction, so the
discharge check reads it and holds a value to it:

```
Input #1 of `quote` could not be built: the value the row writes did not produce a value:
invariant violated on domain.sum.Parcel
```

## Where it goes

`DeclaredBounds.of` walks the conjuncts of each declared invariant and steps past anything that is
not an end:

```java
if (!((measure == null ? InvariantBound.of(each, carrier)
        : InvariantBound.ofSize(each, measure))
        instanceof InvariantBound.Read.AnEnd placed)) {
    continue;
}
```

`InvariantBound.Read.NoEnd` carries nothing, so there is no reason to forward. Its own words name
three ways a clause reaches it:

> An equality states both ends at once and a disequality states neither; a literal the order does
> not read is not an end either.

A comparison of an arithmetic form is a fourth, and it is not in that list. Whether the three named
ones should be reported is the same question — an `invariant a /= 0` is a rule about the position
that this reading turns into no line, and a report saying so is what tells a gap an author can
close from one they cannot.

`UnreadRule` is keyed by the path it is about. A clause relating two coordinates has no single one,
and `GuardThresholds` already answers that by emitting one per side — which is why the `guard` above
names `order.straw` and `order.choco` separately. The same route is open here.

## What this is worth on its own

A report that says nothing about a rule is worse than one that says it could not read it, and this
is the direction ADR-0090 chose everywhere else. It is also a precondition for reading these rules
at all: until "this compiler did not read your clause" is sayable, there is no honest state to
start from.

kawasima:
Closes #868.

## What was wrong

The issue reports that an `invariant` clause comparing an arithmetic form is dropped without a word, while a `guard` of the same shape names both the positions it compares. The mechanism it names — `InvariantBound.Read.NoEnd` carrying nothing, an arithmetic comparison being a fourth case — is half of it. The reading already names both positions:

```
data Plain = { length: Int, width: Int }
    invariant Int.add(length, width) <= 150
```

reports `not read: p.length` and `not read: p.width`. Adding one clause to the fields' own type silences both:

```
data Cm = Int
    invariant value >= 0
data Parcel = { length: Cm, width: Cm }
    invariant Int.add(length.value, width.value) <= 150
```

The difference is not the clause. The invariant side kept "this could not be read" as what a position is left with if nothing divided it (`Position.valuesUnread` → `PendingPosition`), and `Cm`'s bound gives each field a cut — so the position is no longer pending and the account travels no further. A line read from one rule answered for the rule beside it. The `guard` side already fixed this shape; `GuardThresholds.walk` says so in its own comment.

The vehicle for a per-rule account exists (`Position.unreadRules()`), but its only producer walked the position's own newtype chain and so could never see a clause written on the value the position sits in.

## What this changes

The reading that draws lines keeps its negative answer. `InvariantChecker.direct` is the one place an invariant clause is read for an end — the record's own clauses and the field types' alike, rebased onto positions — and it discarded `NoEnd`. It now emits a `FieldDomains.Unread` at every position an ordering comparison names, with the three reasons a `guard`'s comparison is read with, and `Partitions` collects them whether or not the position was divided.

- `InputDomain.unreadBoundsAt` is deleted, and `InvariantBound.statesAnEnd` with it — it was that walk's only reader.
- A coordinate with no carrier is now a coordinate. Dropped from `byName`, a rule about a position no line can be drawn on named nothing; `data Holder = { stage: Qualified } invariant stage >= Prospecting` now answers in the words a `guard` at the same field already used.
- The end reading outranks the value reading for what a pending position is left with, so one clause gets one sentence.

## Two things found on the way

- The deleted walk was **unobserved**. Disabling it left all 5502 compiler tests green — everything it could find was already said by the value reading's fallback.
- Where it did fire, one clause got **two lines naming different limits**: `(it is compared against values no line can be drawn on here)` beside `(a rule about it is one this compiler did not read)`. That is fixed by the ranking above.

## Left out

A size rule stepping past the last whole number (`List.length(value) > 9223372036854775807`) was named by the deleted walk as a form this compiler could not read, which it reads perfectly. It is silent now rather than wrong. What such a rule states is that no size stands there, and no reading says that yet.

## Verification

`CI=1 mvn -o -Plint verify` green — 5515 compiler tests, all modules, checked-in conformance answers unchanged.

14 new tests, each checked against a broken build: disabling the emitter fails five of the eight readings and three of the six report assertions; removing the ordering guard fails on an equality becoming a line nobody wrote; restoring the old ranking fails on one clause coming back with two causes.

### DIFF
--- a/docs/adr/0090-a-types-partition-is-the-equivalence-class-and-a-threshold-gives-the-boundary.md
+++ b/docs/adr/0090-a-types-partition-is-the-equivalence-class-and-a-threshold-gives-the-boundary.md
@@ -7,7 +7,8 @@ line is drawn on is a numeric *term*, not a position — the content of a locati
 one. Revised again for #622: which values a term's line can be drawn on is one table, and a rule
 written over values it does not hold is reported as unread rather than dropped. Revised again for
 #649: what decides whether a clause draws a line is which terms the clause reaches, not which
-declaration it is written on.
+declaration it is written on. Revised again for #868: a rule left unread is a rule about where the
+values stop, and is answered per rule rather than per position.
 
 ## Context
 
@@ -244,6 +245,26 @@ an invariant's bound did not, and a bound it dropped left the position looking l
 bounds — which at a position whose only rule was that bound made the report state the opposite of
 the declaration. The two producers now say it in the same words, because they answer the same
 question: this position was written about and this could not draw the line.
+
+That question is what the sentence covers, and it is narrower than every rule about the position. An
+ordering compared against something no end came out of is one of these. An equality names a value
+rather than an end, a denial takes one away, and a format, a membership or a quantifier says which
+values exist: none of them is a line, so none of them is a line that went unread. A report has
+nowhere to put one, and naming it would send an author after a boundary nobody wrote. What such a
+rule leaves open is which values stand at the position, which is another reading's question and is
+answered where that reading gives up.
+
+Which limit stopped a comparison is one answer for both producers and not a classification each of
+them makes. A relation asks for a class about two positions, a carrier nothing orders asks for that
+order, and what is left is a form this does not take apart — and whether a comparison is a relation
+is asked of what its sides *name*, however deeply. `x < y + 1` relates two positions as surely as
+`x < y` does. Read off whether a side *is* a position, the two producers gave one shape two
+different answers the moment they classified it apart.
+
+And it is asked per rule. A position carries more than one statement, so a line read at it says
+nothing about the rule beside it. Held as what a position is left with when nothing divides it, a
+bound on a field's own type answered for the record's clause about that same field, and two
+declarations differing by one bound said opposite things about the clause above them.
 
 An invariant's bound gives a boundary and not a partition: everything outside it is refused at
 construction, so there is no class on the far side to cover. A `guard`'s line has values on both
--- a/docs/adr/0090-a-types-partition-is-the-equivalence-class-and-a-threshold-gives-the-boundary.md
+++ b/docs/adr/0090-a-types-partition-is-the-equivalence-class-and-a-threshold-gives-the-boundary.md
@@ -261,6 +261,13 @@ is asked of what its sides *name*, however deeply. `x < y + 1` relates two posit
 `x < y` does. Read off whether a side *is* a position, the two producers gave one shape two
 different answers the moment they classified it apart.
 
+Two positions and not a position on each side. `x < x + 1` puts one on either side of the
+comparison and names one position, so there is no second one for a class to be about: what a reader
+would have to be given is a reading of the form. Asked as "does each side name something", both
+producers sent an author looking for a position the model never wrote — which is the distinction
+`Relates` already draws for the readings that turn a clause into values, held here too because it
+is the same question about the same shape.
+
 And it is asked per rule. A position carries more than one statement, so a line read at it says
 nothing about the rule beside it. Held as what a position is left with when nothing divides it, a
 bound on a field's own type answered for the record's clause about that same field, and two
--- a/souther-compiler/src/main/java/souther/compiler/check/FieldDomains.java
+++ b/souther-compiler/src/main/java/souther/compiler/check/FieldDomains.java
@@ -61,7 +61,7 @@ public final class FieldDomains {
      * those is a reading that found no rules.
      */
     public static final FieldDomains NONE =
-            new FieldDomains(Map.of(), Map.of(), Map.of(), Map.of(), List.of(), Map.of(),
+            new FieldDomains(Map.of(), Map.of(), Map.of(), Map.of(), List.of(), List.of(), Map.of(),
                     new ReadingEvidence(), Map.of(), true, Set.of(THE_VALUE), NO_POSITIONS,
                     ConstraintState.top(), null, null, null, Map.of(), () -> true);
 
@@ -69,6 +69,9 @@ public final class FieldDomains {
     /** The ends the record's own clauses place, which is a different question from the range they
      * leave — see {@link #placedAt}. */
     private final List<InvariantChecker.Direct> directs;
+    /** The rules saying where a coordinate's values stop that no end came out of — see
+     * {@link #unreadAt}. */
+    private final List<Unread> unread;
     /** What each clause reaching this value raises, keyed on the rule it is. */
     private final Map<RuleRef, Required> raised;
     /** Which readings took each clause in, as each of them said so. */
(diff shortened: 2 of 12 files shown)


kawasima:
[P1] `x < y + 1` の分類差を直しました。`f150f399`。

指摘どおり再現しました。修正前:

```
byRule  (invariant x < y + 1)    · not read: p.x (a rule about it is one this compiler did not read)
byGuard (if b.x < b.y + 1)       · not read: b.x (the comparison relates it to another position ...)
```

最小案（side ごとに `coordinatesIn` を取って `!left.isEmpty() && !right.isEmpty()`）ではなく、提案されていた「分類自体を共通概念にする」ほうを採りました。3つの理由だけ共有して**選び方**を共有しなかったことが、両方を持った最初のコミットで既にずれた原因なので。

`check/UnreadComparison` が規則を持ち、位置の引き方は各 reader に残ります（`Relates` と同じ分け方）。side は3択:

- `NamesNothing` — ここには読んでいる位置が無い
- `HoldsOne` — 位置が式の中にある（`y + 1`、`Int.add(length, width)`）
- `IsOne(ordered)` — side が位置そのもの。carrier があるかを答える

`GuardThresholds.why` と `InvariantChecker.unreadEnds` はどちらも side を作って `UnreadComparison.why` を呼ぶだけになりました。guard 側の外向きの挙動は変わりません（既存 5515 件が緑のまま）。

修正後は両方 `ComparisonBetweenPositions` です。

テストは指摘の形で2本:

- `ALineReadAtAPositionSaysNothingAboutTheRuleBesideItTest#andSoDoesOneWhereTheSecondPositionIsInsideAnExpression` — reading 直下で `x` / `y` 両方が `ComparisonBetweenPositions`
- `AnInvariantSaysARuleWentUnreadTheWayAGuardDoesTest#andForAComparisonWithOnePositionOnEachSide` — 同型の `invariant x < y + 1` と `if b.x < b.y + 1` を report で**直接比較**

効いているかは狭い問い方に戻して確認しました。2本とも赤になり、`expected: <ComparisonBetweenPositions[]> but was: <UnreadComparisonForm[]>` が出ます。

---

**#868 との仕様差** も指摘に沿って ADR-0090 に書きました（同コミット）。ヘッダに `Revised again for #868` を足し、本文に境界を追記しています。

> That question is what the sentence covers, and it is narrower than every rule about the position. An ordering compared against something no end came out of is one of these. An equality names a value rather than an end, a denial takes one away, and a format, a membership or a quantifier says which values exist: none of them is a line, so none of them is a line that went unread.

あわせて「分類は両 producer で1つの答え」「所見は規則ごと」も同じ節に書きました。`/=` を unread に戻す変更が #868 本文を根拠に合理化されないように、です。

`CI=1 mvn -o -Plint verify` 緑（compiler 5517件、全モジュール BUILD SUCCESS）。

kawasima:
[P1] same-position identity を直しました。`c754e594`。

指摘どおり、両方の producer で再現しました。

```
byRule  (invariant x < x + 1)   · not read: r.x (the comparison relates it to another position ...)
byGuard (if b.x < b.x + 1)      · not read: b.x (the comparison relates it to another position ...)
byShifted (if b.x + 1 < 10)     · not read: b.x (a rule about it is one this compiler did not read)
```

guard 側は以前からこうだったので、前回の共通化はそれを invariant 側へ一般化していたという指摘も実測どおりです。

`Side` が「名指したか」でなく「どの position を名指したか」を持つようにしました。

```java
NamesNothing<K>()
NamesInside<K>(Set<K> positions)     // 旧 HoldsOne。javadoc 自身が Int.add(length, width) を例に
                                     // 挙げていたので、名前も実態に合わせた
IsOne<K>(K position, boolean ordered)
```

relation の判定は指摘の3条件です。

```java
!left.positions().isEmpty() && !right.positions().isEmpty() && named.size() > 1
```

identity は reader ごと（guard は `TermPath`、invariant は place）。`UnreadComparison` は等値比較しかしないので `K` は生成型パラメータのままにしてあります。invariant 側は **place** で持ちます（`FactSubject` ではなく）。1つの place が複数の名前に答えるので、名前で持つと同じ place の2つの名前が両辺に来たときにまた「別の position」になるためです。

結果は指摘の表どおりになりました。

| | |
|---|---|
| `x < y` | relation |
| `x < y + 1` | relation |
| `x < x + 1` | form |
| `x + 1 < 10` | form |

テストは指摘どおり両側に:

- `ALineReadAtAPositionSaysNothingAboutTheRuleBesideItTest#butOnePositionOnBothSidesIsNotTwoPositions` — reading 直下で `UnreadComparisonForm`
- `AnInvariantSaysARuleWentUnreadTheWayAGuardDoesTest#andNeitherNamesASecondPositionWhereThereIsOnlyOne` — guard / invariant 双方で `"another position"` と報告されないこと、かつ両者の文言が一致すること

`named.size() > 1` を外すと2本とも赤（`expected: <UnreadComparisonForm[]> but was: <ComparisonBetweenPositions[]>` と `expected: <false> but was: <true>` ×2）。

ADR-0090 にも1段落足しました。

> Two positions and not a position on each side. `x < x + 1` puts one on either side of the comparison and names one position, so there is no second one for a class to be about ... which is the distinction `Relates` already draws for the readings that turn a clause into values, held here too because it is the same question about the same shape.

`CI=1 mvn -o -Plint verify` 緑（compiler 5519件、全モジュール BUILD SUCCESS）。lint 警告は増えていません。

Issue #863, opened 2026-08-19
BOUNDARY has no reachable Unaccounted, so a completeness counted over it is counted over one arm


kawasima:
`CoverageObligation` says of itself that an arm nothing raises is a question every model answers by
nobody having asked it, and that a completeness counted over such a set says more than it knows.
`BOUNDARY` is the other half of that sentence, and it is in the enum today.

`FieldDomains.answered` reads:

```java
case BOUNDARY -> new RuleAccounting.Outcome.Accounted(RuleAccounting.Reader.THE_END_READING);
```

with the reason written beside it: raised only where an end was read off the rule, so the reading
that raised it is the one that answered it. So

```
BOUNDARY was raised   ≡   the end reading took the rule in   ≡   BOUNDARY is answered
```

and there is no reachable `Unaccounted`. Every question the accounting can leave standing today is
an `ADMITTED_VALUES`. A measure counted over both arms is counted over one.

## The question

Whether `BOUNDARY` is a question the model raises, answered by whichever reading takes it in — which
is what `CoverageObligation` is for — or a fact derived from one reading having succeeded, in which
case it is a reading's account of itself and belongs where `ADMITTED_VALUES` is not.

The two are told apart by a reader that could raise it and not answer it. If a rule can place a line
that the end reading does not take in, `BOUNDARY` is the first kind and the arm is simply not
reachable yet. If no such rule can exist, the arm is the second kind and says nothing.

Asked from the other side, this is the same question: whether "nothing composed a value at this
edge" belongs under `BOUNDARY`. It does not — whether a row can be written is existential over the
whole value and no rule answers for it, which is what #854 settles — but the reason it does not is
the reason `BOUNDARY` may not be a model question at all.

#851 adds `PARTITION`, raised by a comparison in a body and answered somewhere else entirely. That is
the first arm whose producer and whose reader are not the same step, and it should make the shape of
`BOUNDARY` legible by contrast. Worth holding until then.

Follows #842. Split from #854.

kawasima:
Answered by #883 (`a80dff34`), which #851 became.

This asked whether `BOUNDARY` is a question the model raises or a reading's account of itself, and proposed to tell them apart by finding a rule that places a line the end reading does not take in.

[...]

PR #874, opened 2026-08-19
Say what the projection states, and stop asking a second walk what the rules could have become

### LINKED ISSUE #854: Exactness is evidence about the projection that was made, not a question the rules raise
`FieldDomains.allRulesRead()` is what an edge stands on, and #842 asked whether that is a question
of the same kind as the ones `CoverageObligation` holds — something a rule raises and a reading
answers — or a property of a reading. It is the second, and the reason settles the shape of the fix.

## The decision

Exactness does not go into `CoverageObligation`.

An obligation is a question the model raises, and it is written in the model's words: which values
may stand at a position, where a line falls. Whether the range is exact is not in that vocabulary.
It asks whether the interval the bounds are read from holds everything the rules said, which is a
question about the representation this compiler chose. `value /= 0` is a hole, and a range is all
`NumericDomain` holds; give it a domain that holds holes and the same declaration is exact. The
model did not change and no reader was added or lost. Put exactness in the accounting and the count
of questions the model raises starts moving with the abstract domain.

Coverage and precision are independent in both directions, and there is a witness each way:

- `invariant nonzero = value /= 0` is taken in whole by the reading that turns clauses into sets of
  values. Every `ADMITTED_VALUES` question is answered and the edge at 0 is a row nobody can write.
- `value * 2 >= 4` is read whole, moves the floor, and `allRulesRead()` is false all the same
  (`AQuestionIsAnsweredByWhicheverReadingTookTheRuleIn.aClauseCanBeAccountedForWithoutBecomingABound`).

## What that leaves

`allRulesRead()` is a conjunction of three answers from three sources:

1. `seeded` — which holds `Seeded.everyClauseRead()`, the completeness of the seeding's own reading
   of the root declaration's clauses. The name says something else. Nested clauses do not reach it:
   `PathEngine.seedAt` returns a `Known` and never touches `read`.
2. `constraints.numbers().projectionIsLossless()` — `losses.isEmpty()` on the domain that was built.
3. `InvariantChecker.everyRuleRead()` — a second walk of the declarations that never consults the
   projection. The record arm builds a fresh `InvariantChecker`, types the clause again and requires
   every `ClauseDischarge.Kind` to be `DERIVABLE`; the newtype arm derives the `Carrier` again, walks
   the newtype chain again, splits the conjuncts again and asks `InvariantBound` / `InvariantConstraints`
   whether each could have been a bound.

The third is the one to remove. It classifies a rule by its shape and asks what could have become a
bound, rather than reading what did. It is also why the sentence above about swapping the abstract
domain is not true today: a domain that held holes would leave `everyRuleBecameABound` returning
false for `value /= 0` regardless, because a disequality places no end
(`InvariantBound.Read.NoEnd`).

## What is owed

Exactness is evidence about the projection that was actually produced, not a classification of the
rules that might have produced it. The projection reading returns it beside the state:

```
ProjectionEvidence
    Exact
    Approximate(causes)
```

The causes do not share a key, and forcing them into one would put a rule's name on something that
has none:

| cause           | key                         | where it comes from                                  |
| --------------- | --------------------------- | ---------------------------------------------------- |
| `Unavailable`   | path                        | `Gathering.missed(path)` / `notGathered`. A clause that could not be stated is one whose position is exactly what is unknown about it, so there is no `Clause.Ref` to name |
| `Unrepresented` | `Clause.Ref` × position     | the projection's own record of what it adopted        |
| `Lossy`         | `Clause.Ref` × atom × `Loss`| `NumericDomain`'s loss, attributed at the site that has both |

`Unrepresented` is not `ReadingEvidence.left`. That one says no reading adopted the conjunct, and it
is the right answer to `ADMITTED_VALUES`; a conjunct read as a fact rather than as a numeric form is
adopted, so `Predicates.subjectsIn` names its subject, so `settle` records nothing standing — and no
bound came of it. The two questions have to stay apart or #842's evidence answers #854's question
with the wrong reading.

Where it is produced: `Predicates.assume(owed, k, held)` already walks `owed.clauses()` and sends
`known()` and `numeric()` to `Known.taking(...)` while `fact()` goes elsewhere. That loop is where
"went into the numeric domain" and "went in as something else" are already told apart, and it
returns only a `Known` today. Leaving the evidence there adds no reader.

`NumericDomain` keeps answering "this assume lost this kind of information about this atom" and gains
no dependency on rule identity; the compiler layer attributes the rule.

Then these go:

- `InvariantChecker.everyRuleRead` and the two re-readings under it
- `Seeded.everyClauseRead`, or what is left of it under a name that says what it holds
- the boolean consumer of `NumericDomain.projectionIsLossless()`
- `FieldDomains.allRulesRead()` and `Position.everyRuleOfTheValueWasRead()` — the name states
  "every rule was read", and the proposition wanted is that the projection is precise enough to
  conclude a concrete value from an edge
- `CutEvidence.Present.uncertain`, which carries the evidence instead. A reader that only wants the
  boolean derives it. Left as a boolean, the reason is lost again at the partition boundary, with
  `FieldDomains` holding it one call earlier.

`settle` should stop calling `Predicates.obligations` a second time. Replacing its `Known.top()` with
the `Known` actually in force is not the fix — an assumption's reading does not require prior
knowledge — the fix is that the first read leaves its evidence.

## How it is verified

`everyRuleRead()` is not a gold standard: `value * 2 >= 4` is a case it gets wrong. Comparing the new
evidence against it and calling every disagreement a regression would keep the defect.

Conformance is a semantic-delta detector and not an evidence-completeness detector — a model with no
boundary reports the same thing whether the evidence is right or wrong — so the characterization comes
first, and it fixes the answer rather than the agreement:

```
value * 2 >= 4    Exact
value /= 0        Approximate(Lossy DROPPED_DISEQUALITY)
a rule read as a fact and not as a numeric form
                  Approximate(Unrepresented)
a clause that cannot be stated
                  Approximate(Unavailable)
```

Beside them: a nested clause that cannot be stated; a top-level and a nested `Owed.unreadable`; a
conjunction with one unprojectable half; a cycle stop on a recursive type, which must not make a rule
already read approximate; a rule under an optional or a collection.

Follows #842. The `BOUNDARY` question is #863.

kawasima:
Closes #854.

Whether a row can be written at an edge stood on `FieldDomains.allRulesRead`, and #854 asked whether
that is a question of the same kind as the ones `CoverageObligation` holds. It is not: exactness is
evidence about the projection that was made, not a question the rules raise. Put it in the accounting
and the count of questions the model raises starts moving with the abstract domain — `value /= 0` is
a hole because a range is all `NumericDomain` holds, and a domain that holds holes makes the same
declaration exact with no rule changed and no reader added.

## What replaces it

`ProjectionEvidence` — `Exact`, or `Approximate` with the causes. The causes do not share a key:

| cause | key |
| --- | --- |
| `Unavailable` | a path. A clause that could not be stated is one whose position is exactly what is unknown about it |
| `Unrepresented` | a rule and a position |
| `Lossy` | a rule, the atom the algebra dropped it about, and what it dropped |

Produced where the bounds are made. `Predicates.narrowableIn` is the numeric-only sibling of
`subjectsIn` — `subjectsIn` counts the key a guard settles by name, which is why
`value == 3 || value == 5` names a subject there and narrows nothing here. `NumericDomain` gains no
dependency on rule identity: it goes on answering what it lost about which atom, and the compiler
layer attributes the rule.

## What went

`allRulesRead`, `InvariantChecker.everyRuleRead` with `readOnlyAsBounds`, `everyRuleBecameABound` and
`readAsABound` under it, `Position.everyRuleOfTheValueWasRead`, and the boolean in
`CutEvidence.Present`, which carries the evidence and derives `uncertain()` from it.
`Seeded.everyClauseRead` stays — `StepInputFacts` asks it a different question.

## Four things this turned up

**A stop answers two questions.** What a position admits is short wherever a rule about it went
unread, however the value it is written under is reached — a rule inside an optional narrows that
value when there is one. Whether an edge may be promised is about what a construction has to satisfy,
and a rule about a value the construction need not make refuses nothing at an edge of a field every
value has. `PathEngine.declining` reported all of them through one `missed(path)`; the stop now says
which of the two it was, and the conditions are asked apart so that each can say it.

**A line is where the rules leave the position, not where a clause wrote it.** #854's own model —
`within = value >= 0 && value <= 10` beside `nonzero = value /= 0` — has an exact projection: the
first puts the value on one side of zero, and the algebra sharpens the denial to an end rather than
dropping it. The range is `[1, 10]`, and the line was still at the 0 `within` wrote.
`LocalInspection.axisBounds` already clamps a written end to the projection — "the value moves and
the names do not" — and was being handed what the value a position sits in projects onto it, which a
newtype's own value has nobody to be projected onto it by. `leftAt` answers the other question, and
the line moves to 1 while `within` keeps it. `nonzero` did not draw a new one.

**What counts as the projection is not only the interval algebra.** A rule that placed an end is in
the bounds too — that is how an ordered value which is not a number gets its range, and read through
the algebra alone a bounded `Date` field is a rule the bounds do not hold, taking every boundary
beside it down with it. A rule that raises no coverage question at all — a pattern, a membership — is
invisible if the question is asked per obligation, and a pattern three records down is still a way a
`Root` can be refused. So `Unrepresented` is asked of every rule the reading was handed.

**`Rules` needed no new arm.** It already keeps "a caller that stopped" and "a kind that carries no
clause" apart for reach and for values; the projection is the third question with the same two
answers, and `FieldDomains.NONE` answers as the first for both.

## And the second readings

`settle` read every conjunct twice more — once through `Predicates.obligations` and once through
`StatedByClauses.adoptedIn` — because the seeding reports per clause and it wants per part. Both
readings now say what they made of each part as they read it, keyed by the part as the tree holds it.
Absent rather than empty where a part is missing, and it throws: a conjunct nothing saw is not a
conjunct that constrained nothing. No lookup missed anywhere in the corpus.

`Known.top()` went with it rather than being replaced by the `Known` in force. It was never the
defect — an assumption's reading requires nothing to have been said beforehand — the defect was that
there was a second call.

## Measured

Against every model the tests compile, comparing the old answer with the new one: 348 where the old
oracle was needlessly approximate, and 3 where it said exact and the range was not — one model,
`invariant odd = n * n >= 4`, which the newtype arm of the same walk got right. Two defects of my own
turned up the same way and are worth writing down: `Owed.Subject` carries `measured` and I collapsed
it to the path, which checked a rule about a string's values against the range of its length; and
`Unavailable` read every stop rather than the ones a construction cannot get out of.

`WhatMakesARangeExactIsNotWhatAnswersACoverageQuestionTest` holds the grid. Every row was measured
before it was written down.

5509 tests green, `CI=1 mvn -Plint verify` clean.

The `BOUNDARY` half of #854 is split out as #863.


kawasima:
All three are real and all three are fixed in `1542e536`. Each was reproduced first, and the
characterizations asked for are checked in.

**A rule is represented where every part of it is.** `narrowedBy` was unioned over the whole clause,
so one conjunct answered for the conjunct beside it. Measured before:

| written | was | now |
| --- | --- | --- |
| `invariant a = value >= 1` / `invariant b = value * value >= 4` | Unrepresented | Unrepresented |
| `invariant c = value >= 1 && value * value >= 4` | **exact** | Unrepresented |
| `invariant mixed = value >= 1 && (value == 3 \|\| value == 5)` | **exact** | Unrepresented |

Kept per part now, and `Direct` carries the part it was read from so the end a part placed excuses
that part rather than the rule. One thing that had to come with it: a conjunction is itself recorded
as a part, and its own set is the union of its conjuncts' — asked as well, a rule whose halves are
each held in a language of its own (a `Date` bounded at both ends, read by the comparison rather than
by the algebra) answers for neither half and fails on the node above them.
`ACarrierNothingReadsUnmeasuresItsSiblings` caught exactly that. A conjunction whose conjuncts are
recorded beside it is skipped: it says what they say.

**A count's line is clamped by the range taken of the count.** `leftAt` answered with the position's
own values whatever axis the caller was on, so `data L = List<Int>` came back `leftAt = null` and the
end as written stood. It takes the axis now rather than falling back, because a `String` is measured
both ways. `aLineOnACountIsClampedByTheRangeTakenOfTheCount` pins the written end at none, the range
at one, and no range on the values.

**A loss is a fact about the step it happened at.** Both orders are exact now, and a hole nothing
excludes is still a hole:

| written | was | now |
| --- | --- | --- |
| `floor = value >= 1` then `nonzero = value /= 0` | exact | exact |
| `nonzero = value /= 0` then `floor = value >= 1` | **Lossy** | exact |
| `nonzero = value /= 0` alone | Lossy | Lossy |

What the algebra was handed is kept beside what it dropped, and `projection()` asks the state the
reading ended in whether it holds it — `NumericDomain.entails` reads `NE` as "proves it on one side
of zero", which is the same test the assume step makes and could not make yet. Not a second reading
of the clause: the reading's own result checked against its own final state. The provenance gap you
noted goes with it — a second rule dropping the same loss at the same atom is no longer invisible to
a before/after diff, because there is no diff to be invisible to.

`known` also stopped counting. `Clause.known` is filled by `sizeFacts`/`resultFacts` with relations
true of every value — a size is never negative — so any rule mentioning a count looked like a rule
that narrowed the count. `Set.size(value) == 3 || Set.size(value) == 5` was exact and is now
`Unrepresented`.

5514 green, `-Plint verify` clean.

kawasima:
Both reproduced, both fixed in `1a0b7a0c`. They took the same shape out of the evidence, so the
answer is one change rather than two.

| model | was | now |
| --- | --- | --- |
| `nonzero = value /= 0` / `nonfive = value /= 5` / `floor = value >= 1` | **exact** | `Lossy(nonfive)` |
| `data R = { a: Int, b: Int }` / `sum = a + b >= 10` | **exact** | `Lossy(sum, KEPT_UNPROJECTABLE)` |
| `floor` then `nonzero` | exact | exact |
| `nonzero` alone | Lossy | Lossy |

**What was given, not what changed.** The diff was trying to recover an event from a state summary,
and the state does not hold one — `Map<A, Set<Loss>>` says which kinds an atom has and not how many
times each happened. Rather than have `assume` return the losses it produced, the simpler answer is
that the events are not needed: every form the algebra was handed is kept, and the projection is
asked at the end whether it holds each of them. A form it holds needs no loss to have been recorded,
and a form it does not hold is a cause whether the loss was the first at that atom or the second.
`Gathering.projected` and the before-and-after domains are gone with it, and the one per-part channel
now carries what the reading made of each part: the positions it spoke of, the atoms it was handed a
form for, and the forms.

**The projection, and nothing beside it.** You are right that this was the more important of the two,
and `proveBaseLe`'s own javadoc had already said it — "The kept relations are not read here — this is
what a residual is proven against, and reading them would let one stand on another".
`NumericDomain.projectionEntails` is `entails` proving against that base, so a form kept as written
is not a premise for its own proof. The distinction is written down on both: a caller deciding
whether a construction discharges its invariant wants everything the state holds, and a caller
deciding how much of the rules the bounds state wants what is projected out of it.

**Characterizations.** `aHoleShutByABoundIsAnsweredAndTheOneBesideItIsNot` runs all six orders of the
three clauses and asks that the cause name `nonfive` in every one of them — so the shut hole is
answered and the open one is named, whoever wrote the clauses in whatever order.
`aFormKeptAsWrittenIsNotProvenByItsOwnBeingKept` holds the second.

5552 green, `-Plint verify` clean.

kawasima:
All three in `7454439f`.

**The position now comes from the part that is short of the bounds.** You are right that this was not
a display nit — the type says a rule and a position, and the position was being read off the rule.
Measured before and after on `data Length = { a: Int, b: Int }` with `invariant both = a >= 1 && b * b
>= 4`: the causes named `a` and `b`, and now name `b`. What each part raises is kept beside what the
rule raises, at the point the part is settled, so the failing conjunct answers with its own
questions. Written apart as two invariants the answer is the same, which is the point: where the
author put the `&&` is not a fact about which position is short of its rules.

`theConjunctTheBoundsDoNotHoldNamesItsOwnPosition` holds it, and it asserts the paths rather than
`exact()` — as you say, the cause set is the thing under test now.

**`Lossy` cannot be made exact the same way, so it says so.** The domain records which kinds of loss
an atom has and not which rule left it with each. Two rules leaving one atom short in different ways
would be named with both kinds, and there is nothing in the reading to separate them without putting
rule identity into `NumericDomain` — which is the boundary this change exists to keep. So the
javadoc now states the precision it has: the rule is exact, the atom is exact, and what was lost is
the atom's account of itself. If that turns out to be worth more, the producer to change is the
algebra's own record and not this reader.

**CI.** Reproduced and fixed. The failing test is in `souther-cli`, which is why every local run I
reported was green — `-pl souther-compiler` never builds it. It is
`ABoundaryIsAValueTheRecordCanHoldTest.aNewtypeTakenStraightIsHeldToItsOwnRules`, the same model this
branch already moved at the boundary, one module further out: the report was checked in against the
reading that drew the line at the 0 `within` wrote, and `nonzero` takes the 0 away, so `N` is `[1,
10]` and the line is at the 1. Both edges are rows an author is owed now rather than edges nothing
promises — the denial became an end, so what is missing is the row and not the reading.

`CI=1 mvn -Plint verify` over the whole reactor is green, which is the check I should have been
running rather than the one module.

Issue #876, opened 2026-08-19
E1916 reports the same test-design fact as the adequacy report and names it in the other vocabulary


kawasima:
After #869 the adequacy report names the technique it is doing: a border is closed or open, and each one owes an `ON` point and an `OFF` point (ISTQB CTAL-TA v4.0 §3.1.1, spec `[#example-report-vocabulary]`). The diagnostic built from the same finding was left alone, so one missing row is now told twice in two vocabularies:

```
souther examples
  ! no row is at the ON point rate/n = 100 (if@10:5)

E1916
  no row is at that boundary …
```

Both come from one `Adequacy.Finding` of kind `BOUNDARY_UNMET`, which already carries the role — `AdequacyReport` reads it and `ExampleMessage.NoRowIsAtThatBoundary` / `NoRowIsAtTheLineAConstructDrew` do not. A reader who has just been handed a vocabulary is asked to translate again at the diagnostic, which is the cost #869 set out to remove.

The question is not "add ON/OFF to E1916". It is whether a diagnostic reporting the same test-design fact should share that vocabulary, and it is a different question from #869 because the answer has to hold three things #869 did not touch:

- **Localization.** The report is written in one language; a diagnostic is a message catalog in every shipped one. `ON` and `OFF` are terms of art rather than English words, so whether they are translated, transliterated, or left as they are is a decision per catalog.
- **The code names a rule.** E1916 is `every-guard-boundary-has-a-row`. If the message says `border` where the anchor says `boundary`, the diagnostic and the rule it points at stop agreeing — so either the anchor moves, which is a stable-identifier change, or the sentence says why they differ.
- **What the sentence is for.** The report answers "how far do the rows reach"; a diagnostic answers "what do I fix". They may want different words for one fact, and that is a decision to take rather than assume either way.

Split out of #869, which deliberately stopped at the report and the JSON.

kawasima:
Merged into `develop` as `d60810ab` (PR #887).

E1916 now names which point of the border it is about, in the same word the adequacy report uses, and the anchor is `every-border-has-a-row-against-its-line` — one requirement over every border a rule draws, rather than one written about guards while the code reported invariant and `ensures` borders under it.

Two things the old wording said of both points were fixed along the way: `which is where the rule draws its line` is true only of the point carrying the value the rule was written with, and the hint was keyed on position against the line where the sentence names a role. Which of the two carries the line's value turns on the border being closed or open, so both readings were inverted on every open border.

PR #879, opened 2026-08-19
Say which point of a border a row is owed at, in the words the technique already has

### LINKED ISSUE #869: The adequacy report invents words for a technique that already has them, so what it counts is unreadable to whoever practises it
`souther examples` derives equivalence partitions from a model's declarations, draws a line where a
rule places one, and asks for a row at the line and at the value beside it. That is a test design
technique with a name, a syllabus and a body of practice — domain testing (ISTQB CTAL-TA v4.0
§3.1.1, generalising equivalence partitioning and boundary value analysis) — and the report names
none of it. It says `partition axes`, `single-axis`, `boundary`, `obligation`, `pairs`.

The values already agree with the definitions. For a closed border and an open one:

```
guard n <= 100    →  ! no row is at closed/n = 100     ! no row is at closed/n = 101
guard n <  100    →  ! no row is at opened/n = 100     ! no row is at opened/n = 99
```

against the syllabus:

> For a closed border, an ON point lies on this border. For an open border, an ON point lies inside
> the equivalence partition and is closest to the border according to the given precision.
> For a closed border, an OFF point lies outside the equivalence partition and is closest to the
> border according to the given precision. For an open border, an OFF point lies on this border.

100 and 101 for the closed border, 99 and 100 for the open one. Souther produces exactly these and
does not say which is the ON and which is the OFF.

## The correspondence

| CTAL-TA v4.0 §3.1.1 | the report today |
| --- | --- |
| equivalence partition | `class` |
| border | the line |
| closed / open border | `ComparedTerms.holdsAtTheLine`, not surfaced |
| ON point, OFF point | the value, the value beside it |
| IN point | `single-axis` |
| OUT point | — |
| coverage item | `obligation` |
| atomic condition | unnamed |
| combinatorial testing (§3.1.2) | `pairs` |

`pairs` counts combinations across two positions, which is the neighbouring technique rather than
this one, and reporting it beside the boundary counts invites a reader to add them up.

## What this is not

Not a coverage criterion to claim. The syllabus names two — simplified domain coverage (Jeng et
al., 1994) and reliable domain coverage (Forgács et al., 2024) — and naming either as satisfied
would be a single word standing for a whole report, which ADR-0089 refuses:

> Nothing is reduced to a single score, at any level.

So the report says which coverage items a border owes and which of them a row is at, and a reader
who works to one of those criteria decides for themselves whether it adds up. That reading is also
the honest one while some coverage items cannot be produced at all: a border on a `Decimal` has no
OFF point, because no smallest step is named for a decimal, and the syllabus asks for one "according
to the given precision" — a precision this language does not carry.

Spec `[#example-report-vocabulary]` fixes the words the report uses and is where this is decided.
The vocabulary section exists because a line of the report carries no diagnostic code and so has
nowhere to be looked up; borrowing a published vocabulary answers that better than inventing one.

kawasima:
Closes #869.

`souther examples` was doing domain testing (ISTQB CTAL-TA v4.0 §3.1.1) and naming none of it. It said `single-axis`, `boundary`, `pairs`; the technique says equivalence partition, border, ON point, OFF point, coverage item. A reader who practises it had to translate, and the invented words could be looked up nowhere — a line of this report carries no diagnostic code, which is why `[#example-report-vocabulary]` exists.

## What the report says now

```
    partition   axes 3   equivalence partitions 6/6
    combination pairs 6 reached / 6 known reachable, 6 untried
    border      0/2
      ! no row is at the ON point rate/n = 100 (if@10:5)
      ! no row is at the OFF point rate/n = 101 (if@10:5)
```

`combination` moved to a line of its own, under the same condition it was written under before — these counts used to be the tail of the partition line. Counting combinations across two positions is the neighbouring technique (§3.1.2); printed beside the border counts a reader could add the two into a total neither is part of.

Each border line says which role the value it names serves in: the ON point is inside the partition the border bounds and closest to it, the OFF point outside and closest. Both are values a row can be owed at one border and they differ by one step, so `no row is at n = 100` and `no row is at n = 101` were two sentences a reader could not tell apart. Which role a value has turns on whether the border is closed or open — the same cut is the ON point of `n <= 100` and the OFF point of `n < 100` — so it comes off neither the value nor the side. It is derived once, from the side and the rule, rather than worked out again by the human report, the JSON, and the finding.

No coverage criterion is claimed. The syllabus names two, and either printed as a verdict would be a single word standing for a whole report, which ADR-0089 refuses.

## What is *not* renamed, and why

`single-axis` did **not** become `IN points`. The count is `AxisCoverage.covered()` — how many equivalence partitions some row is in — and the syllabus's IN point is stricter: a value inside a partition and *away from its borders*. A row written on the line is that partition's ON point and is still counted here, so `guard n <= 100` with a single `n = 100` row has every partition covered by this measure and no IN point anywhere. Putting the published name on it would have been the exact failure #869 is about, one level down. It is `equivalence partitions`, which is what it counts.

Deriving IN points properly is a measurement change rather than a vocabulary one, and it has to reach the generator: `Generator` classifies rows with its own `whereIn(row, axes)`, so a report saying a partition has no IN point while `--generate` offers nothing for it would be the two disagreeing about what is owed. Filed as #880.

## The defect underneath

Deriving the role needs one bit — whether the cut value itself satisfies the rule. `GuardOrigin` and `EnsuresOrigin` carried it. `InvariantOrigin` did not: `LocalInspection` was handed a `DeclaredBounds.End` whose `Endpoint` says exactly this and dropped it, and `BoundaryLine` wrote `false` into the drawing for want of anything to read.

Assuming every bound is closed would have been correct for everything a report shows today, and wrong in principle. Measured at the construction site, both kinds are built:

```
Int     value > 5      →  inclusive 6   →  ON
Decimal value > 5.0m   →  exclusive 5   →  (dropped further down)
```

So the closed end is a fact about how far the derivation gets, not one about bounds.

## What a border owes, measured

Worth stating because #869's body has it differently — a correction is [posted there](https://github.com/souther-lang/souther/issues/869#issuecomment-5339220137).

A strict bound on a `Decimal` derives **no boundary obligation at all**. It is not an ON obligation whose OFF neighbour could not be constructed. The denominator is unaffected either way, but the sentence to tell a reader differs: the second does not say Souther owes a coverage item here.

More generally "a decimal has no OFF point" does not hold. Where the carrier names no neighbouring value, which role goes missing depends on where the cut falls — a closed border keeps its ON point, an open one its OFF point. And `besideTheCut` answers nothing for an invariant, so a bound owes its edge and nothing beside it: the spelling moves where that point is (`>= 5` → 5, `> 5` → 6) and both ends that arrive are inside. ON and OFF are roles, not a count of two per border.

## JSON

`point` added beside `side`; nothing renamed, so `schemaVersion` stays at 2, which is what the schema's own contract says. The two are not one fact: `side` says where the value sits around the cut, which is what says how to read `value`; `point` says what a row written there is for. Neither key produces the other.

## Tests

`ABorderSaysWhichOfItsTwoPointsALineIsTest`, 10 cases, written as pairs differing in one operator — an absolute expectation on one side alone holds against an implementation that answers `ON` everywhere, which is what the answer was before it was derived. Inverting the derivation kills 5; inverting the bound's bit kills 2 on the invariant path.

`pointRole` is package-private and refuses a side no rule places a row on. It was total over the side and answered from the arithmetic alone, so `<= 100` with `BELOW` came back `OFF` — of 99, a value well inside the partition. `aSideNoRowIsOwedOnIsRefusedRatherThanGivenTheNearerWord` pins the domain, including that a bound refuses every side but the cut.

Existing expectations were updated on both sides. Negative assertions matter here: `assertFalse(contains("no row is at X"))` would have stopped matching the new sentence and passed unconditionally, so each was either re-pointed at the full new text or, where no positive on the same subject collides, widened to cover both roles.

`CI=1 mvn -o -Plint verify` green across the reactor on every commit.

## Not in this PR

- Counting IN points properly — #880.
- E1916 still reports the same finding in the old vocabulary — #876, which brings localization, the coupling between a diagnostic code and its spec anchor, and whether a diagnostic should share a report's words at all.


kawasima:
Pushed `818a11a7`. All three review items addressed.

**[Blocker] `IN points` misnamed the measure** — right, and the fix is the rename rather than a stricter count. `covered()` is built in `Coverages` by classifying each row into a partition class with nothing excluded for sitting on a border, so `guard n <= 100` with a single `n = 100` row covers the partition while having no IN point in it. Now `equivalence partitions`.

Not a holding name. "Did any row reach this partition" and "did a row reach it away from its borders" are two questions, so the stricter one arrives as its own measure rather than by taking this name back — the name will not move twice. Filed as #880, which also carries the part that makes it more than a report change: `Generator` classifies rows with its own `whereIn(row, axes)`, so a report saying a partition has no IN point while `--generate` offers nothing for it would be the two disagreeing about what is owed.

The spec sentence went with it, and the OUT-point paragraph is gone. You are right that "it is the neighbour's IN point, so counting both asks for one row twice" states a general rule about the technique that the syllabus does not hold — reliable domain coverage wants both. It now says IN and OUT are not derived, which is a limit of this compiler.

**[Blocker] The spec kept the pre-correction Decimal reading** — fixed in both places, and your point that "no OFF" does not generalise is the reason the sentence was wrong rather than merely stale. Which of the two goes missing depends on where the cut falls: a closed border on a decimal keeps its ON point and loses the neighbour, an open one keeps its OFF point and loses the ON. A strict bound on a decimal is a third thing again — no border at all.

**[P2] `pointRole` was total over a side it is not defined on** — this was the best catch. `pointRole(<= 100 origin, BELOW)` returned `OFF` of 99, a value well inside the partition. It is package-private now, `BoundaryObligation.pointRole()` is the way to ask, and a side outside `AT` or `besideTheCut(origin)` throws rather than getting the nearer of two words. `aSideNoRowIsOwedOnIsRefusedRatherThanGivenTheNearerWord` pins it, including that a bound refuses every side but the cut.

`CI=1 mvn -o -Plint verify` green across the reactor.

On `mergeable: false` — that was a stale computation while GitHub had not finished comparing; `git merge-base --is-ancestor origin/develop HEAD` holds. Worth re-checking on the new head.

kawasima:
Correction to my previous comment: `mergeable: false` was not stale. `develop` had moved — #874 and #875 merged — and there was a real conflict. I asserted otherwise without running the check; the command I quoted fails.

Merged `origin/develop` in (`c0c48600`), which GitHub now reports `MERGEABLE`.

One conflict, in `ABoundaryIsAValueTheRecordCanHoldTest`, and it was a semantic one rather than textual. #874 changed what the projection states, so under `invariant nonzero = value /= 0` the value 0 is no longer a line of `N` at all and the line moved to `f/n = 1`. My side had only reworded that assertion to carry `the ON point`. Took develop's side whole: the assertion it replaces no longer exists, so there is nothing there to name a point of.

`CI=1 mvn -o -Plint verify` green on the merged tree. GitHub build is running.

Issue #870, opened 2026-08-19
IN and OUT are counted per equivalence partition, and a domain coverage criterion counts them per border


kawasima:
The coverage items of domain testing are keyed on the border (ISTQB CTAL-TA v4.0 §3.1.1):

> One ON, one OFF, one IN, and one OUT point for each border defined by one of the operators
> `<`, `≤`, `>`, or `≥`.

and the same point can be of different types for different borders. Souther keys two of the four on
the equivalence partition instead. `single-axis` counts partitions some row is in, which answers
"is there an IN point somewhere" and not "does this border have one"; the far side of a `guard`'s
line is a partition of its own, so its representative is an OUT point that nothing reports as one.

What a border carries today depends on which rule drew it. For one position under an invariant, a
`guard` and an `ensures` all at once:

```
! no row is at judge/req.cost = 0        (invariant Cost #1)        ← ON
! no row is at judge/req.cost = 0        (ensures judge (asked))    ← ON
! no row is at judge/req.cost = 1        (ensures judge (asked))    ← OFF
! no row is at judge/req.cost = 100000   (guard@21:5)               ← ON
! no row is at judge/req.cost = 100001   (guard@21:5)               ← OFF
```

| drawn by | ON | OFF | IN | OUT |
| --- | --- | --- | --- | --- |
| `guard` | yes | yes | per partition | per partition |
| `ensures` | yes | yes | per partition | per partition |
| `invariant` | yes | no | per partition | no |

Re-keying IN and OUT onto the border is most of this. The `guard` and `ensures` rows above already
have the values; what is missing is that a border is what owes them.

## The invariant border

An invariant border has no OFF and no OUT, and this is not an omission to close. ADR-0090:

> An invariant's bound is a boundary and not a partition. Everything outside it is refused at
> construction, so there is no class on the far side to cover — what remains is an edge worth a row.

The decision stands: nothing outside the bound can be constructed, so no row can be written there,
and the type is what answers for those two points. What has to change is that the report says so.
Left blank, a border shows two of four coverage items and a reader counting them finds Souther short
without being told why; said, it reads as the model having discharged them, which is what happened.

The vocabulary for this exists. Spec `[#example-adequacy]` already writes *excluded* for a case the
model's own rules refuse — counted out of what the rows are held to, reported, and not counted
against the model — and extends it to classes:

> A class the rules refuse is *excluded*, as a case of the signature is.

A point in a class the rules refuse is the same reading one step further in.

## What it does not settle

[...]

PR #887, opened 2026-08-19
Say which point of the border a warning is about, in the report's words

### LINKED ISSUE #876: E1916 reports the same test-design fact as the adequacy report and names it in the other vocabulary

kawasima:
Closes #876.

After #869 and #870 the adequacy report names the technique it is doing: a border owes a row at four
points, and the two against the line are called `ON` and `OFF` (ISTQB CTAL-TA v4.0 §3.1.1). The
diagnostic built from the same finding was left alone, so one missing row was told twice:

```
souther examples
  ! no row is at the ON point submit/cost = 100 (guard@21:5)

E1916
  No row is at submit/cost = 100, which is where the `guard` draws its line.
```

Now:

```
No row is at the ON point submit/cost = 100 of the border the `guard` draws.
Hint: A row on the line is what tells a rule written `<=` from one written `<`.
```

## What the sentence gained

`point` is a component of both E1916 messages, read off the `PointRole` the finding already carries.
`Adequacy.warning()` reads it; nothing on the finding side moved.

`ON` and `OFF` are the same word in every catalog — the syllabus's terms rather than English words
for them, so they are handed over as a name is and what a catalog holds is the sentence around them.
This is a decision rather than a consequence of `PointRole` living in a module `souther-syntax`
cannot see. `everyMessageSaysEachValueItCarries` is what makes it hold in every language.

Which point it is crosses how the rule is named. The two messages are still told apart by whether the
rule has a place or a name; both gained the point.

## Two things the old wording said of both points

`which is where {rule} draws its line` is true only of the point carrying the value the rule was
written with — `ON` for a closed border, `OFF` for an open one. The other is one step over, so the
sentence told it the line is where it is not. The border is named by the rule that drew it instead.

The hint was the same shape. A row on the line tells `<=` from `<`, which is the `ON` point's answer;
the `OFF` point now has its own.

## The anchor

`every-guard-boundary-has-a-row` asked for a row at each line a `guard` draws, and the paragraph
under it hands an invariant's edges to another measure. But nothing filters `BOUNDARY_UNMET` by
origin — the checked-in conformance report has borders whose origin is `invariant Sku #1` owing an
`on` point — so E1916 fires for invariant and `ensures` borders too, and the spec was already citing
the guard's anchor for a clause's line.

It is now `every-border-has-a-row-against-its-line`, one requirement over every border a rule draws,
stated where the rules are rather than beside the arms. What is specific to a guard stays in
§example-branch-coverage and says so: the arms do not stand in for the line.

## Not touched

`DOMAIN_POINT_UNCOVERED`, the report, the JSON schema, the conformance corpus. `IN` and `OUT` are
reported and warned about by nothing, which is #870's decision; the `PointRole` switch in
`warning()` is exhaustive, so a finding built for one of them stops there rather than borrowing a
neighbour's hint.

## Verification

`CI=1 mvn -o -Plint verify` green across all modules, 17611 tests.

`ADiagnosticNamesThePointInTheReportsWordsTest` takes the guard's `ON` point and the `ensures`'s
`OFF` point, so both roles and both origins are covered by two cases. Confirmed red before green:
passing a constant `"ON"` fails the `OFF` case, and giving both points the `ON` hint fails the hint
case.


kawasima:
Fixed in `4e790397`. The finding was right and the defect was mine.

`ON` and `OFF` are not "on the line" and "one step past it". Which point carries the value the rule
was written with turns on the border being closed or open, which the spec's own table says:

```
guard n <= 100    ON  n = 100    OFF n = 101
guard n <  100    ON  n =  99    OFF n = 100
```

So the split was correct on every closed border and inverted on every open one — and the test model
had one of each. The `ensures ... id.value > 0` border is open, so its `OFF` point at zero is where
the clause put its line, and the hint was telling the author about a row one step past it.

## What changed

Option A. The two hints are now said in the role's own words:

```
ON   A row at the nearest value the border keeps in is what shows the border does not fall a step further in.
OFF  A row at the nearest value the border keeps out is what shows it does not fall a step further out.
```

Settled by `PointRole` alone, so the switch is now a correct design rather than a coincidence.

Not option B. Bringing `OriginRef.holdsAtTheValue` into the diagnostic would put the hint on the
closed/open axis while the sentence above it names the role — a second reading of one finding, which
is the cost this PR exists to stop charging.

`<=` versus `<` leaves the hint with it. That is the border being closed or open, a property of the
border and not of either of its points (`a-border-owes-four-points`), and the report already says it
there.

## Tests

The hint test now covers all four points — the closed border's `ON` and `OFF` and the open border's
`OFF` and `ON` — so a hint keyed on the line cannot pass it. Beside it, a second test asserts the
inversion itself: `ON, OFF, OFF, ON` across the four values, which is the thing that makes the first
test a control rather than a restatement.

Confirmed discriminating: swapping the two hint arms fails the loop.

`CI=1 mvn -o -Plint verify` green, 17613 tests.

PR #888, opened 2026-08-19
A finding carries what it is about as the value the measure established

### LINKED ISSUE #884: A finding carries what it is about as a list of objects, so every renderer reads it by position
`Adequacy.Finding` carries what a finding is about as `List<Object> args`, and what each element is
follows from the kind and from nothing written down. Thirty-two reads across `AdequacyReport` and
`GeneratedRows` index into it by position, and every one of them is a claim about a kind's arity and
about the type at each place that only the producer knows.

```java
public record Finding(Kind kind, String behavior, MeasurementStatus status,
                      Citation at, List<Object> args) {
```

The type says a finding carries some objects. What holds it together is that `partitionFindings`
writes them in an order and each renderer reads that order back.

## What it costs

Two defects in #883 came out of it, and neither was caught by anything but a reader.

A `RuleCitation` was put in the list so that a rule with no name could be printed as where it is
written. One renderer knew and one did not, so a JSON `subject` came out as
`WrittenAt[at=WrittenCitation[at=54:5]]` — a record's `toString`, published. Nothing typed the
element, so nothing said the string branch was wrong.

Then the obligation was added at index 3 so a renderer could file a question under the measure that
answers it, and read back with a cast. Adding an element ahead of it would have compiled and broken
the cast at run time, and the arity is checked nowhere.

Both are the same shape as what #851 was about, one layer over: a value crosses a seam narrower than
what the far side asks of it, and the far side fills the gap with whatever is nearest — an index and
a cast.

## The shape a fix would have

A finding's arguments are per kind, and the kinds already exist. Something like a payload per kind,
sealed over the same set the `Kind` enum names, so that what a renderer reads is what the producer
wrote and a kind gaining a field is a compile error at every reader rather than a silent index.

`Kind` is the natural place to hang it: it already carries the diagnostic code and whether a build
refuses over the finding, so which arguments a kind has is the same kind of fact.

Not a rewrite of the measures. What the findings say and which ones exist do not change; what changes
is that the list of things a finding is about stops being a list.

## Scope

`Finding` is shared by every measure — signature, partition, border, branch, pairs — so this is
wider than the one kind that exposed it. Which is also why it is worth doing once rather than each
time a kind grows an argument: `RULE_UNACCOUNTED` is at five elements today, and it got there in
three separate changes.

Found while doing #851.

kawasima:
Closes #884.

`Adequacy.Finding` carried what it was about as `List<Object> args`, in the order some
message key took them. What each element was followed from the kind and from nothing
written down, and every renderer indexed into it and cast.

The root of it is a layer under the untyped list. The javadoc said `args` were "the
message's arguments", and only 4 of the 12 kinds ever become a message — the rest fall
through to `default -> throw`. So the payload shape of every kind was set by what a third
of them needed for one of three readers, and each later reader that wanted something the
message did not need appended to the same list. It ended up answering two questions with
different lifetimes at once: the words of a sentence, and the identity of what the
sentence is about.

Three things in the code said so before this change:

* `behavior` was both a field of `Finding` and an element of `args`, because a message
  key took it.
* Index 1 meant a bare value for `boundary_unmet` and `operator + " " + value` for
  `domain_point_uncovered`, and `atEdge` read index 1 as a bare value.
* `text(said, at)` returned `""` past the end, so arity was checked on neither side.

And the four arguments of the border kinds turned out to be a copy of
`BorderAssessment.Point`, recomputed where the finding was built —
`p.against()`, `p.asked()`, `p.border().axis()`, `p.border().origin(...)`. The report
printed the same line twice through two paths, fifteen lines apart.

## What changed

`About` is a sealed type, one shape per kind, 11 shapes for 12 kinds (the two border
kinds come off one point and differ by the role). `Finding` takes an `About` and no
`Kind`; `kind()` is a switch over it, so a kind and what it is about cannot disagree, and
`About` does not know its own kind.

Where a measure already held the item, it is carried whole rather than copied out field
by field: `BorderAssessment.Point`, `PartitionEvidence.Unanswered` (whose own javadoc
says it is handed on whole rather than taken apart, and which was taken apart into six
elements one seam later), `CoverageSites.Site`, `UnreadPosition`, `UndividedPosition`,
`OmittedAxis`, `AxisCoverage`. Ten of the twelve needed nothing new.

Two had no value to be, so the measures grew one:

* `InputCaseEvidence` says which of the inputs it is. A type that says it is the evidence
  of one input and cannot answer which one only means something beside the list it came
  from. Zero-based; `#1` is the report's word.
* `AxisCoverage.uncovered()` answers `AxisClass`, a class that knows the position it is a
  class of.

Words moved to whoever writes a sentence. The two switches turning a reason and an
obligation into English are the report's now, and so is the choice of which name a
question is shown under — both surfaces were already overruling the one the finding
arrived with, for a subject that is a comparison.

`atEdge` no longer matches a copy of the axis, the value, the rule and the role back
against the assessments to find the item the finding came from. It reads the point, and
the "these two readings disagree" failure goes with it.

## What the types hold

Three states a producer used to be trusted for, from review:

* An input's evidence and the list it sits in both say which input it is — the second is
  what the document publishes as the order of `signature.inputs`. `SignatureEvidence`
  holds them to each other where both exist, the way `BorderAssessment` holds its items
  to what the border says it owes. An input at no position is refused.
* Every `About` shape holds what it names. The list this replaced refused a null element
  as a side effect of being copied, so the shapes had to be told to keep what the copy
  was doing by accident.
* `AxisClass` refuses a class that names no position, which is the whole of what it was
  added for.

## Behaviour

One intentional output change, in its own commit: `axis_class_uncovered` names the
position. `no row is in \`No\`` becomes `no row is in \`No\` at q.flag`, and the
document's `subject` becomes `C (at grade)`. The schema already argued for this of the
inputs — a case carries `(in #1)` because two parameters of one type give two findings a
class name cannot tell apart — and did not say it of the axes, which is the same fact one
position over.

`AClassNoRowIsInNamesThePositionItIsAClassOfTest` builds the collision: one behavior, two
positions of one type, one case reached at neither, so the two axes differ in nothing but
which position they are of. With the change backed out the document publishes

```json
{"kind":"axis_class_uncovered","disposition":"reported","subject":"No"},
{"kind":"axis_class_uncovered","disposition":"reported","subject":"No"}
```

two lines under `"No (in #1)"` and `"No (in #2)"`.

Everything else is byte-identical. Worth being exact about what checked that: the
conformance corpus's checked-in answers did not move, but that corpus produces only three
finding kinds (`domain_point_uncovered`, `partition_not_derivable`,
`partition_not_read`). It does cover the border path, which is the largest rewrite here.
The other kinds rest on the unit tests.

8803 tests across the reactor, no failures.

## Merged with #887

#887 landed on `develop` six minutes after #884 was filed, in this renderer: a boundary
warning now names which point of the border it is about, and its hint is keyed on the
role. Both are kept — the sentences and the hints are #887's, read off this branch's
point rather than off four fields copied out of one. `point(said)` goes with the list it
indexed into. `ADiagnosticNamesThePointInTheReportsWordsTest` passes unchanged, which is
what says the composition is the one #887 asked for.

## Notes for review

* Renderers read `About` with record deconstruction patterns, not accessors. That is a
  convention and not something the type enforces: a type pattern (`case Leaf l ->`) still
  compiles when a record gains a component, while a deconstruction pattern does not.
  `sealed` catches a shape being added; deconstruction catches a shape changing.
* A review rule that falls out of this: `toString()`, `describe()`, `said()`, string
  concatenation and `i + 1` inside `Findings` are the symptom. There were 13 of them and
  there are now 0 — the remaining `behavior.name()` calls fill the finding's own field.


kawasima:
Both blockers are real. Fixed in a20c0d05, along with the third point.

**1. `SignatureEvidence.inputs` order vs `InputCaseEvidence.at`.** Right, and it is the
shape this PR is against. `SignatureEvidence` now holds them to each other in its
compact constructor, which is where both exist and the only place they can be said to
agree — the same thing `BorderAssessment` does with its items and what the border says
it owes. `InputCaseEvidence` refuses a negative position. The wire is unchanged.

I considered deriving `at` from the list index instead of checking it. Rejected: a
constructor that silently rewrites what a caller wrote takes away a reading, and the
caller who put an evidence at the wrong index would get no answer at all.

**2. `About` leaves admitting null.** Also right, and the observation that this made the
new type *weaker* than the old one is the part worth acting on — `List.copyOf` was
refusing null elements as a side effect, so the replacement lost something the original
did by accident. All 11 leaves now say it themselves.

On the fixture: the point about the order being backwards is well taken, and it goes
further than a test smell. `TypeSymbol`'s constructor is package-private on purpose —
"a caller that could build one from two strings is a caller that could arrive at an
identity without having been handed one" — so a fixture minting subjects is fighting the
same design in the same direction. The synthetic `finding(kind, status)` is gone. Both
dispositions a finished compile shows are read off a real one, and `undecided` moved to
`CompilePartialAdequacyTest`, beside the assertion that no code is warned for it, which
is the same fact from the other side. `everyKindABuildMayRefuseHasADiagnosticCode` in
`WhatStrictRefusesIsWhatTheRowsDoNotCoverTest` already holds the kind/code agreement, so
nothing is lost by not looping every constant here.

**3. The collision test.** Added `AClassNoRowIsInNamesThePositionItIsAClassOfTest`: one
behavior, two positions of one type, one case reached at neither, so the two axes differ
in nothing but which position they are of. Checked it goes red with the change backed
out — the document publishes

```json
{"kind":"axis_class_uncovered","disposition":"reported","subject":"No"},
{"kind":"axis_class_uncovered","disposition":"reported","subject":"No"}
```

two lines under `"No (in #1)"` and `"No (in #2)"`, which is the defect stated better
than I stated it in the PR body.

8803 tests across the reactor, no failures. No further output changed: the delta is
still the one commit.

PR #889, opened 2026-08-19
Keep a unit data out of the set a constructs clause declares

### LINKED ISSUE #882: Unit data should not belong to a behavior's `constructs` set
`constructs` currently includes a unit data whenever its name is written as a value.

That is consistent with the current expression semantics: writing a unit-data name constructs its
single value. It is not, however, a useful member of a behavior's construction set.

For example:

```souther
data Kind = Domestic | Overseas
data Request = { kind: Kind, cost: Int }
data Decision = { accepted: Bool }

behavior b : (r: Request) -> Decision
    constructs Decision

let b (r) =
    if r.kind == Domestic
    then Decision { accepted = true }
    else Decision { accepted = false }
```

Today this is E1002 unless `Domestic` is added to `constructs`.

That makes an ordinary observation of a case change the behavior's declared construction authority:

```souther
r.kind == Domestic
```

requires `constructs Domestic`, while:

```souther
match r.kind with
| Domestic -> ...
```

does not, because the latter occurrence is a pattern rather than a value expression.

The compiler is internally consistent here. The problem is the rule.

## Why unit data are different

`constructs` exists for two related purposes:

1. to state which data a behavior has authority to construct; and
2. to make it visible from the declaration whether the behavior mints a new domain value or merely
   passes an existing one through.

Neither distinction carries useful information for a unit data.

A unit data:

* has no fields;
* cannot declare an invariant;
* has exactly one value;
* already requires no protected construction path in practice.

Constructing `Domestic` cannot choose any information that an existing `Domestic` did not already
contain. There is no observable distinction between minting another value of a unit data and passing
the unit value through.

This is already reflected in the closed-construction design: unit data have nothing to validate and
are not what construction authority protects. Injected behavior bases also already receive unit
output factories independently of the behavior's `constructs` entries — the backend walks the return
type's cases for those, and reads `constructs` for the field-bearing data and newtypes.

The declaration therefore gains noise rather than useful model information from:

```souther
constructs Domestic
```

especially when the unit value merely appears in a comparison, a guard departure, a collection, or
an argument.

## Proposed rule

Writing a unit-data name remains a unit-data construction in expression semantics.

However:

> A unit-data construction does not contribute to a behavior's construction set and requires no
> construction authority.

Consequently, a unit data MUST NOT be named in `constructs`.

`constructs` continues to describe non-unit data construction exactly: when the clause is present,
the declared set and the implementation's inferred set must still match exactly.

For example:

```souther
data Missing

behavior find : (...) -> Member | Missing
```

needs no:

```souther
constructs Missing
```

while:

```souther
data Member = { id: Id }

behavior make : (id: Id) -> Member
    constructs Member
```

is unchanged.

## Consequences

All of these still construct the unit value, but none contributes `Domestic` to the construction set:

```souther
Domestic
f(Domestic)
[Domestic]
r.kind == Domestic
guard allowed else Domestic
```

A helper that mentions or returns a unit data likewise contributes no unit data transitively to its
calling behavior's construction set.

Non-unit construction remains unchanged:

```souther
Member { id = id }
OrderId(raw)
[Member { id = id }]
```

still contributes `Member` / `OrderId`, including through helpers and blocks.

An explicit unit entry such as:

```souther
constructs Domestic
```

should be rejected as meaningless rather than treated as an optional declaration. Allowing it both
with and without the entry would weaken the existing exact-match meaning of `constructs`.

That rejection needs a diagnostic of its own rather than E1006. E1006 says

> declares `constructs {data}` but never builds {data}

and under this rule the body does build the unit value — only not into the construction set. Saying
E1006 there would state something the new rule denies. E1006 keeps its meaning, and a new code
(E1026 is the next free one in the family; E1003 is retired) says that a unit data is not part of a
construction set and the entry is to be removed.

## How this was found

The original report for this issue appeared to be a coverage/partition failure around:

```souther
r.kind == Domestic && r.cost <= 100000
```

That diagnosis was wrong.

The measured model declared:

```souther
constructs Auto, Manual
```

and therefore did not compile: the `Domestic` expression was inferred as another construction and
caused E1002. Behavior elaboration failed, so the later coverage report had no body to measure —
which is what `border not measured`, `not derivable: r.cost`, and `axes 1` were saying, alongside
`rows 0` and `branch not applicable (this behavior has no body)` in the same report.

Removing the explicit `constructs` clause makes the same body elaborate, and the guard/border
readers find the expected comparisons: `axes 2`, `borders 2`, a line at `r.cost = 100000` / `100001`,
and a singleton border at `r.kind = Domestic`.

So the coverage reader was not losing the sum comparison. The measurement exposed that ordinary use
of a unit value currently leaks into `constructs`.

## Specification / implementation work

* Change §constructs so unit-data construction is excluded from the construction set.
* Keep §unit-data's rule that writing the name constructs the unit value.
* Update the closed-construction text to distinguish construction from construction authority.
* Remove unit data from `DataChecker.collectConstructs`.
* Reject unit data explicitly named in `constructs`, under a new code rather than E1006.
* Keep E1002/E1006 exactness over the remaining, non-unit construction set.
* Update injected-behavior examples/tests that currently write unit cases in `constructs`.
* Keep unit output factories available to injected Java implementations independently of
  `constructs`.

kawasima:
Closes #882.

A unit-data construction no longer contributes to a behavior's construction set, and a unit data may
no longer be named in `constructs`. Writing `Domestic` still constructs the `Domestic` value; what it
no longer does is put `Domestic` in the set the clause declares.

## Why the rule, and not an exception for comparisons

`DataChecker.collectConstructs` recorded a bare unit-data name wherever it stood, so the position the
name was written in decided the clause:

```souther
if r.kind == Domestic      // demanded `constructs Domestic`
match r.kind with
| Domestic -> ...          // demanded nothing — a pattern is not an expression
```

Exempting comparison operands would have fixed one position out of several. The distinction the
clause exists to record — whether a behavior mints a new value or passes an existing one through —
does not exist for a unit: it has no fields, carries no invariant, and has one value. Authority is
not what a unit needs either; a case is reachable through the decoder or the `protected` factory
rather than by being named.

Naming it is refused rather than made optional. Were `constructs Member` and `constructs Member,
Missing` both correct for one body, the exact match E1002 and E1006 keep would say nothing.

## E1026 rather than E1006

E1006 says the body never built the data. Here the body may well build it, and the entry is wrong
anyway — so E1006 keeps the meaning it has, its loop skips units, and E1026 answers for the entry.

The check is asked of the clause and of nothing else, so an injected behavior and an implemented one
get the same answer. Every behavior's clause is read before the first entry is raised, which is what
E1002 and E1006 already do within one clause.

## What the spec said and the compiler did not

Two sentences were describing a compiler that had moved on, and both were measured before being
changed:

- `[#java-base-class]` said an injected base carries a factory for each unit data **listed in
  `constructs`**. `Backend.java` builds them from the **output type's cases**. Declaring a unit that
  is not an output case produces no factory today either, so nothing here is lost.
- `[#constructs]` said an injected behavior MUST always write the clause. It is not enforced, and a
  units-only injected behavior compiles with no clause and still receives both factories.

E1305 loses its unit allowance with them: every entry is a non-unit data now.

`[#unit-data]` gains the recommendation that goes with this — where a case list introduces the type,
the separate `data` says one thing twice and is the spelling that goes stale when the case is
renamed.

## How #882 was found, and why its first diagnosis was wrong

The issue reported a coverage failure: `r.kind == Domestic && r.cost <= 100000` came back with `axes
1`, `border not measured`, and `not derivable: r.cost`, and named `GuardThresholds.read` as the
place to start.

The measured model did not compile. It declared `constructs Auto, Manual`, the `Domestic` expression
was inferred as another construction, E1002 fired, and behavior elaboration failed — so the coverage
report had no body to measure. `rows 0` and `branch not applicable (this behavior has no body)` were
in the same report.

With the clause omitted, the same body gives `axes 2`, `borders 2`, a line at `r.cost = 100000` /
`100001`, and a singleton border at `r.kind = Domestic`. The guard and border readers were never
losing the sum comparison.

## Migration

155 fixture and corpus files drop the unit entries they carried. The conformance answers move by the
two lines the clauses took up — no other difference.

The tests that measured the old rule are rewritten rather than deleted. Most used a unit as the entry
that put a clause there at all, which E1002 needs to run; those get a field-bearing case instead.
`CompileBareIdentifierTest` and `CompileUnitValueTest` turn over with the rule and now assert both
halves — the compile emits the class, and naming it is E1026.

`CI=1 mvn -Plint verify` is green across the reactor.


kawasima:
Followed up on the review's one soft observation — that the E1006 loop's skip for unit entries is
unreachable, since `TypeChecker` raises E1026 before that check can see one.

It is reachable, and load-bearing. `TypeChecker.checkModule` and `Bodies.CheckedBehavior` are
separate queries: the second asks about one body and not about a module, so the clause check has not
run when the body's own reader meets the entry. A probe that threw instead of skipping never fired
across the compiler suite — every path the tests take goes through `Compiler.compile` — but fires at
once when the per-behavior query is asked directly.

Without the skip, that query answers absent and the author is told the body never builds a unit it
does build, with a hint to the wrong edit.

`47451b51` adds the case that tells the two apart and points the comment at it. Confirmed red with
the skip removed, green with it back.

kawasima:
All five addressed in `975970ba`, on top of the `develop` merge (`9fc300b3`, no conflicts).

**P1 — E1305.** The hint said `Expose {data}, or make it a unit data.` Taking the second half now
earns E1026 on the same entry, so it says `Expose {data}.` alone, the body sentence drops "neither a
unit data nor", and `DeclarationMessage.ExposeItOrMakeItAUnitData` becomes `ExposeIt`. `[#e1305]`
says why the old way out is gone, and that a unit case of the output needs neither.

**P1 — ADRs.** ADR-0002's Decision now names the unit as the deliberate exception and ties it to the
rationale already at the end of that file: authority guards `[#invariant-on-construct]`, and a unit
has nothing to validate. ADR-0006 and ADR-0017 both said `constructs` is required of an injected
behavior *because it drives factory generation* — corrected to say the unit factories come from the
output type, and that a clause with nothing left to name is not written.

**P2 — `DataChecker.Constructs`.** You are right that the compiler had kept as one synonym what the
specification just split. The javadoc now says `Constructs` holds what a body contributes to a
construction set, that `builds` answers false for a unit the body does build, and that a reader
wanting every construction wants its own walk. `collectConstructs` is described by what it collects
rather than as "the data types `e` constructs". `Declarations` loses "what a compilation may
construct is governed by `constructs`". I kept the type names — the rename to `ConstructionSet`
reaches further than this PR should, and the contract is what was misleading.

**P2 — the test oracle.** Correct, and worse than you put it: `demo.立替` is emitted by the
declaration whether or not the body writes the name, so that assertion held before the change too.
It runs the behavior now and checks the value each arm returns.

**P3 — the factory comment.** Rewritten to say the two factories come from two places, with why
`Member` gets none.

Also took the wording fix: **"A behavior whose construction set is empty writes no clause"** — the
old phrasing described the unit itself.

On the §unit-data recommendation: it does not follow from the `constructs` rule, and the text now
says so rather than implying it. Keeping it here is deliberate — it was asked for alongside this
change — but it reads as a modelling convention adopted at the same time, not as a consequence.

`CI=1 mvn -Plint verify` green across the reactor.

Issue #842, opened 2026-08-18
A position's value-set reading is printed under the partition measure, in the vocabulary of position division


kawasima:
```souther
module example.rooms

data Length = Int
    invariant min = value >= 1
    invariant max = value <= 100

behavior price : (length: Length) -> Int
let price (length) =
    if length.value >= 50 then 1 else 2

example price
    | "long" : (Length(50)) -> 1
    | "short" : (Length(49)) -> 2
```

```
    partition   axes 1   single-axis 2/2
      · read in part: length (a rule about it is one this compiler did not read)
    boundary    2/4
      ! no row is at price/length = 1 (invariant Length (min))
      ! no row is at price/length = 100 (invariant Length (max))
```

The clause the first line says went unread is the clause the third line draws a boundary from, named. Both sentences are true of different questions, and neither says which question it is about.

## The two questions

`read in part` is written off `Axis.read`, which is an `AdmissibleSet.Completeness`. What that field records is stated on `Axis` itself:

> how much of what the rules say about this position's **values** was read

Which values the position may hold — not how far the rules that divide it were read. `AdmissibleReading` turns a clause into a set of values only where the clause is an equality or an inequality; everything else, ordering comparisons included, falls through `unreadable` to `FORM_NOT_READ`. `BlockReason.of` maps that to `UnreadValueRule`, `ReportedReason` maps that to `UndividedPosition.Reason.UNSUPPORTED_SYNTAX`, and `Adequacy` writes it as *a rule about it is one this compiler did not read*.

So `value >= 1` leaves the value set open, which is what the line reports, while the boundary measure reads the same clause and names it. Three controls, same model otherwise:

- drop the invariant — the line goes, and the boundaries go with it
- write `invariant only = value == 7` — the line goes, and the reading is complete
- write any ordering comparison — the line appears

## What the report does with it

The reading is a fact about the value set, and it is printed under the `partition` heading, in the vocabulary of position division. `UndividedPosition.Reason` is the enum for why a position could not be divided; `BlockReason`'s own javadoc keeps the two vocabularies apart on purpose, and `ReportedReason` is where the value-set reading borrows the division words. A reader is handed the borrowed sentence with nothing naming the question it answers, two lines above a measure that read the same clause successfully.

Rewording the sentence alone leaves the arrangement: the next measure to consult `Axis.read` prints the same borrowed words under its own heading. What is owed is that the line say which reading is short — the values this position may hold — and that the reason come from a vocabulary about that.

## Beside it

[...]

PR #893, opened 2026-08-20
Say which rule went unread, and let the reading that knows say it

### LINKED ISSUE #886: A position's unread rule is reported without naming the rule, in the measure beside the one #842 fixed
```
      · not read: r.cost (a rule about it is one this compiler did not read)
```

The position is named and the rule is not. That is the sentence #842 was about — an author told that a
rule about a position went unread, with nothing saying which rule — and it was fixed for the
accounting and not for the measure beside it.

The type says why:

```java
public record UnreadRule(TermPath at, BlockReason why) {}
```

whose own javadoc opens

> Asked of the rule and not of the position. One position carries more than one statement, and a line
> read at it says nothing about the rest — a threshold on `x` does not answer for the bound beside it
> that nothing could read.

It carries no rule. What it is asked of and what it holds are two different things, and the sentence
written from it can only name the position.

## Where it is lost

`InvariantChecker.unreadEnds` has the rule, the conjunct, the coordinate and the reason.
`FieldDomains.Unread` now carries all but the carrier. Then `InputDomain.unreadRulesAt` builds an
`UnreadRule` from each and keeps two of them:

```java
UnreadRule said = new UnreadRule(path, each.why());
```

so the rule and the conjunct stop there, and the dedupe below it is by position and reason — two
rules stopped by one limit at one position become one line, which is what the type says it is not
for.

The guard producer has the same shape: `GuardThresholds.compared` builds one per position mentioned,
keyed on the position.

## What it would take

The rule reaches the accounting already, as a `RuleRef` with a `RuleCitation` beside it — the two
that #883 settled: an identity to tell one rule from another, and a handle for a reader to find it.
An `UnreadRule` wanting to name its rule wants those, not a second way of saying which rule.

The report line then says which rule, as the accounting's line already does:

```
      · not read: invariant Length (even) — a rule about `r.cost` this compiler did not read
```

## Why it is filed rather than fixed with #851

#851 is the accounting: which questions the model raises and what answered them. This is the measure
beside it — why a position could not be divided — and it publishes through `PARTITION_NOT_READ` and
`UndividedPosition.Reason`, which #851 does not touch. Same defect, different measure.

Found while doing #851.

kawasima:
Closes #886.

A position's unread rule was reported without naming the rule:

```
  · not read: r.cost (a rule about it is one this compiler did not read)
```

That is the sentence #842 was about, fixed for the accounting and not for the
measure beside it. `UnreadRule` was asked of the rule and held a position and a
limit, so the sentence could only name the position — and two rules stopped by
one limit at one position came out as one line, deduplicated on a key the rule
was not part of.

It now reads:

```
  · not read: invariant Length (even) — written in a form this compiler does not read, about `r.cost`
```

## What changed

**The rule travels with the finding.** It was in hand at every producer and
dropped at the seam: the reading of ends records which clause it gave up on,
the guard walk has the comparison, the `ensures` walk has the clause.
`UnreadRule` now carries the pair a question already carries — an identity that
tells one rule from another, and a handle a reader looks it up with — and is
keyed on `(rule, position, limit)`.

**Which of those a finding may have is a property of its reason.**
`BlockReason` splits into `AboutARule` and `AboutThePosition`. A finding built
on the first owes an identity and the type says so; one built on the second has
nowhere to put one, because nothing observed a rule. The report and the
document write the two apart rather than leaving a consumer to read an absent
field to know which it is holding.

**And which findings are about a rule is asked of the subject.** `About.OfARule`
answers it, so the writer does not match against a list of kinds it has to be
told about again each time one is added.

**A structural stop is a candidate, not a finding.** It becomes one where
neither the position's own declarations nor a body's rules answered — settled
once, beside the verdict and from the same pair of answers. Reported where the
producer records it, every position holding an `Option` claimed its values are
held inside something this does not reach into, which is true of the compiler
and no account of a position whose two cases were read perfectly well.

**The verdict stops carrying a cause.** `BodyCutInspection.Blocked` and
`UndividedPosition.Why.CannotDerive` say that nothing was established and no
longer say what stopped it. `notRead` is built from the findings rather than
from what the verdict was left with.

**`RuleRef.Guard` is read off the source.** It was
`CoverageSites.ComparisonRef`, so a rule's identity came from a measurement
plan — and a condition both of whose arms can record nothing is numbered
nowhere. It is now the behavior and the construct the author wrote.

**A comparison written where this compile has no file is not one of these.**
`Int.clamp`'s own comparisons stand in the tree that runs, and naming them tells
an author to edit a function they do not have. Cut per comparison on
`Citation.Elsewhere`, which is exactly "written where this compile has no file";
a helper of the author's own has a file, is cited where it is written, and stays
in. The walk still enters the expansion, because that is where a call's
arguments are bound.

**A comparison inside a fork of its own is cited by that fork.** Walked from the
fork outside it, the word came from one construct and the place from the other.
The corpus had one: a comparison written in an `if` inside a `guard`'s condition,
printed as `guard@` at the `if`'s coordinates.

**`competing_coordinates` is a cause the old word was standing in for.** A
`String` can be measured on its own order or on its length, and where its type
chooses neither and the value it sits in states an end on each, both rules are
left unread — read perfectly well, and neither choosable. That was reported as
an expression the terms do not name, which is a cause it was never observed to
have. Both rules are now named, under a word that says what is missing.

## The document

Still schema 3. The version is raised when a field is removed or renamed and
not when one is added, and a word added to an enumerated field is read the same
way — the schema's own description and `specification.adoc` both say so, and
both say outright that a closed older copy refusing a newer document is the
expected direction. This change only adds: `rule` and `ruleId` on `notRead`
entries about a rule, `competing_coordinates` to `notReadReason`, and `ruleId`
allowed on a `partition_not_read` finding.

## What the change is not

The set of rules reported is unchanged. Across the conformance corpus `notRead`
is 5 entries before and 5 after, with nothing gained and nothing lost: three
gained provenance and two are position findings that correctly have none. Where
a count moves in the suite it is a position that was carrying more than one
rule, which is the thing being fixed.

## Left for later

`Kind.PARTITION_RULES_NOT_REACHED` is reached only by the finding about a
position the axes did measure. A position with no axis whose rules the reading
never arrived at is told under `partition_not_read` with `rules_not_read_at_all`
beside it; no fixture reaches that reason at all. A recursive type, three levels
of nesting and an invariant inside an `Option` were tried, and in each the
structural reason wins first. The remaining candidate is a clause that could not
be typed.

The accounting still raises questions for the forks of a library this compile
has no file for, while `notRead` does not. That is where it stood before this
change. Whether the two should draw the line in one place is its own question,
and is written down beside the test that fixes this side of it.

### DIFF
--- a/souther-cli/src/test/java/souther/cli/MainExamplesSubcommandTest.java
+++ b/souther-cli/src/test/java/souther/cli/MainExamplesSubcommandTest.java
@@ -153,7 +153,7 @@ class MainExamplesSubcommandTest {
     void theJsonCarriesTheNumbersABuildReads() throws Exception {
         JsonNode root = JSON.readTree(run("--format", "json"));
 
-        assertEquals(3, root.get("schemaVersion").asInt());
+        assertEquals(4, root.get("schemaVersion").asInt());
         assertEquals("complete", root.get("status").asString());
         assertNotNull(root.get("compilerVersion"));
 
@@ -182,8 +182,8 @@ class MainExamplesSubcommandTest {
     @Test
     void theEmittedJsonHasTheShippedSchemaShape() throws Exception {
         JsonNode schema;
-        try (var in = Main.class.getResourceAsStream("/souther/adequacy-schema-3.json")) {
-            assertNotNull(in, "adequacy-schema-3.json ships beside the compiler");
+        try (var in = Main.class.getResourceAsStream("/souther/adequacy-schema-4.json")) {
+            assertNotNull(in, "adequacy-schema-4.json ships beside the compiler");
             schema = JSON.readTree(new String(in.readAllBytes(), StandardCharsets.UTF_8));
         }
         JsonNode root = JSON.readTree(run("--format", "json"));
@@ -304,7 +304,7 @@ class MainExamplesSubcommandTest {
     void aKeyAddedSinceIsNotDemandedOfADocumentWrittenBeforeIt() throws Exception {
         JsonNode schema;
         try (java.io.InputStream in =
-                     Main.class.getResourceAsStream("/souther/adequacy-schema-3.json")) {
+                     Main.class.getResourceAsStream("/souther/adequacy-schema-4.json")) {
             assertNotNull(in);
             schema = JSON.readTree(new String(in.readAllBytes(), StandardCharsets.UTF_8));
         }
--- a/souther-cli/src/test/java/souther/cli/MainExamplesSubcommandTest.java
+++ b/souther-cli/src/test/java/souther/cli/MainExamplesSubcommandTest.java
@@ -153,7 +153,7 @@ class MainExamplesSubcommandTest {
     void theJsonCarriesTheNumbersABuildReads() throws Exception {
         JsonNode root = JSON.readTree(run("--format", "json"));
 
-        assertEquals(4, root.get("schemaVersion").asInt());
+        assertEquals(3, root.get("schemaVersion").asInt());
         assertEquals("complete", root.get("status").asString());
         assertNotNull(root.get("compilerVersion"));
 
@@ -182,8 +182,8 @@ class MainExamplesSubcommandTest {
     @Test
     void theEmittedJsonHasTheShippedSchemaShape() throws Exception {
         JsonNode schema;
-        try (var in = Main.class.getResourceAsStream("/souther/adequacy-schema-4.json")) {
-            assertNotNull(in, "adequacy-schema-4.json ships beside the compiler");
+        try (var in = Main.class.getResourceAsStream("/souther/adequacy-schema-3.json")) {
+            assertNotNull(in, "adequacy-schema-3.json ships beside the compiler");
             schema = JSON.readTree(new String(in.readAllBytes(), StandardCharsets.UTF_8));
         }
         JsonNode root = JSON.readTree(run("--format", "json"));
@@ -304,7 +304,7 @@ class MainExamplesSubcommandTest {
     void aKeyAddedSinceIsNotDemandedOfADocumentWrittenBeforeIt() throws Exception {
         JsonNode schema;
         try (java.io.InputStream in =
-                     Main.class.getResourceAsStream("/souther/adequacy-schema-4.json")) {
+                     Main.class.getResourceAsStream("/souther/adequacy-schema-3.json")) {
             assertNotNull(in);
             schema = JSON.readTree(new String(in.readAllBytes(), StandardCharsets.UTF_8));
         }
--- a/souther-cli/src/test/java/souther/cli/WhatStrictRefusesIsWhatTheRowsDoNotCoverTest.java
+++ b/souther-cli/src/test/java/souther/cli/WhatStrictRefusesIsWhatTheRowsDoNotCoverTest.java
@@ -318,7 +318,7 @@ class WhatStrictRefusesIsWhatTheRowsDoNotCoverTest {
                                                Partitions.OmittedAxis... omitted) {
         return new PartitionEvidence(PartitionEvidence.Partitioned.of(List.of()),
                 PartitionEvidence.Bounded.of(List.of(boundary)), PartitionEvidence.PairSpace.NONE,
-                List.of(), List.of(), List.of(), List.of(omitted), List.of());
+                List.of(), List.of(), List.of(), List.of(), List.of(omitted), List.of());
     }
 
     /** What one behavior's partition makes of the whole report, with nothing else asked about. */
--- a/souther-compiler/src/main/java/souther/compiler/check/FieldDomains.java
+++ b/souther-compiler/src/main/java/souther/compiler/check/FieldDomains.java
@@ -364,7 +364,7 @@ public final class FieldDomains {
      *             own terms
      */
     public record Unread(String path, boolean measured, RuleRef.Invariant from,
-                         Core part, souther.compiler.inputs.BlockReason why) {}
+                         Core part, souther.compiler.inputs.BlockReason.AboutARule why) {}
 
     /**
      * The declaration whose clause could have moved where the coordinate at {@code path} stops, or
--- a/souther-compiler/src/main/java/souther/compiler/check/InvariantChecker.java
+++ b/souther-compiler/src/main/java/souther/compiler/check/InvariantChecker.java
@@ -1308,7 +1308,7 @@ public final class InvariantChecker {
         if (!InvariantBound.ordering(comparison.op())) {
             return;
         }
-        BlockReason why = UnreadComparison.why(sideOf(comparison.left(), at, byName),
+        BlockReason.AboutARule why = UnreadComparison.why(sideOf(comparison.left(), at, byName),
                 sideOf(comparison.right(), at, byName), quantityOf(comparison, at, byName));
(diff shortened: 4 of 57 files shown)


kawasima:
P1 and P2 are fixed in fcd3223. P3 is not a defect — the report disagrees with it.

## P1 — `findings[].ruleId`

Confirmed, and worse than reported: the schema's constraint and its description
contradicted each other.

```json
"then": { "required": ["ruleId"] },
"else": { "not": { "required": ["ruleId"] } }
```

The `else` branch *forbade* `ruleId` on every other kind, while the description
said it is written for `partition_not_read` too. Nothing caught it because the
writer never emitted the key, so the constraint held and only the prose lied.

The writer now switches over both kinds that are about a rule, and the
constraint is: required on `rule_unaccounted`, allowed on `partition_not_read`,
forbidden elsewhere. Agreed that this is as far as the schema can go without a
second discriminator, and that a producer test is the right place for the rest.

It shows up in the corpus, which is the case that makes the point:

```json
{ "kind": "partition_not_read", "subject": "指示.受付日",
  "ruleId": { "kind": "guard", "behavior": "出荷する", "ordinal": 4, "lowered": 0 } },
{ "kind": "partition_not_read", "subject": "指示.締切",
  "ruleId": { "kind": "guard", "behavior": "出荷する", "ordinal": 4, "lowered": 0 } }
```

One rule at two positions, now joinable.

Pinned by `aFindingAboutARuleCarriesTheIdentityAsWell` on `TWO_INVARIANTS`: four
`partition_not_read` findings, two distinct identities.

## P2 — the overlap test read two rules

Right, and the assertions only looked at report wording, which says a rule
apiece and nothing about whether it is the same rule.

The guard fixture you suggested does not reach both measures — an unread
comparison in a body produces the `notRead` entry and no `unanswered` one, so
there is nothing to compare. What does reach both is a clause the reading of
ends cannot turn into a line and the reading of values has no word for:

```souther
data Length = Int
    invariant min    = value >= 1
    invariant square = value * value >= 4
```

`square` is under both. The test now compares `notRead[].ruleId` against
`unanswered[].ruleId` directly.

## P3 — the accounting does raise for an expanded helper's forks

This one I ran rather than reasoned about. On the model in the test:

```
· not read: guard@10:32 — written in a form this compiler does not read, about `n`
· not accounted for: if in `Int.clamp`, reached at 10:11 — which classes a row is owed in at n
· not accounted for: if in `Int.clamp`, reached at 10:11 — where the values stop on n
```

So the PR body and the test comment are accurate: the accounting names the
helper's forks and `notRead` does not.

`AComparisonInsideAHelperRaisesNothingTest` is not in tension with that. Its
fixture is `let big (n: Int) = n <= 10 * 2` — a helper that is an expression and
has no fork, so nothing numbers a comparison inside it. `Int.clamp` has forks,
they are numbered, and the accounting raises for them. Two different cases, and
the second is the one this PR had to decide about.

Left as is, and the asymmetry stays written down beside the test that fixes this
side of it.

## Note on the `PositionReadingBlocked` javadoc

Taken: it is a resolved finding, not a raw observation, and only
`PendingPosition.reportable()` produces one. Left for a follow-up rather than
folded in here.

kawasima:
Second reading: all six taken, in 53e7496. Four of them were the same mistake —
a fact stated once in the code and again beside it.

**`findings[].ruleId`, and the duplicate entries it left.** Fixed structurally
rather than by adding the missing case. `About.OfARule` answers which rule a
finding is about, and the writer asks the subject instead of matching a list of
kinds it has to be told about again each time one is added. The constraint is
now: required on `rule_unaccounted`, allowed on `partition_not_read`, forbidden
elsewhere.

**`schemaVersion`.** You are right and I had it backwards. The rule is in the
schema's own description and in `specification.adoc`:

> `schemaVersion` is raised when a field is removed or renamed, and not when one
> is added […] every object in the schema is closed, so an older copy of it
> refuses a document carrying a field added since; a reader that has to accept
> newer documents wants the newer schema rather than a higher number.

That is exactly the argument I bumped on, considered and rejected in advance.
Nothing here is removed or renamed. Back to 3, edited in place, and
`adequacy-schema-3.json` is no longer deleted.

**`citationOf` and the nested fork.** Confirmed, and the corpus had one. In
`shipping.sou`, `急ぎか`'s body is an `if` expanded into a caller's `guard`, and
the comparison inside it printed as `guard@1:52:57` — the word from the outer
construct, the place from the inner one. `compared` no longer descends into a
fork of its own; the walk reaches that fork and reads it against itself. It now
prints `if@1:52:57`.

That also corrected what the helper gate is documented as. It is not "an
expansion happened" — a helper of the author's own has a file, is cited where it
is written, and is theirs to rewrite. `Citation.Elsewhere` is exactly "written
where this compile has no file", which is the line worth drawing: there is
nowhere to send the reader.

**`PARTITION_RULES_NOT_REACHED` with two sources.** Dropped my arm. Both Abouts
write nothing but the position, so sharing the word turned two findings a report
tells apart into two a consumer cannot. `APositionThisCouldNotRead` is
`partition_not_read` again, with `rules_not_read_at_all` beside it; the other
kind stays with the finding that owns it. That also removes a branch no fixture
reached.

**`PendingPosition`, `unreadRulesAt`, and the stranded period.** The first was
worse than a typo: `reportable()`'s javadoc went in between `complete`'s javadoc
and `complete`, so `complete` was left undocumented and its #626 rationale was
attached to the wrong method. `unreadRulesAt` still promised one line per reason
where the key is now the rule. Both fixed, and two stranded periods with them.

Full reactor green, 5726 tests. The PR body is updated — it claimed schema 4 and
described the helper gate the old way.

kawasima:
All three taken, in 199077e.

**P1 — `findings[].reason`.** Confirmed by rendering the model you described:

```souther
data R = { a: Int, b: Int }
    invariant mixed = a < b && Int.multiply(a, a) < 10
```

```
· not read: invariant R (mixed) — it relates two positions rather than dividing one, about `r.a`
· not read: invariant R (mixed) — written in a form this compiler does not read, about `r.a`
```

One rule, one position, two limits — and in `findings` one object twice, keyed
on neither. Worse, the entry beside it in `notRead` is keyed on the reason, so
there was nothing to join by.

Fixed the same way as the last one rather than by adding a field at the call
site: `About.OfSomethingNotRead` answers which limit, and the writer asks the
subject. Both shapes of `partition_not_read` implement it, so `reason` is
required on the kind and forbidden elsewhere.

Pinned by `oneRuleStoppedTwiceAtOnePositionIsTwoFindings`: two whole findings at
`r.a`, one `(subject, ruleId)` between them.

**P2 — the kind's description.** Rewritten to say there are two shapes, that
`ruleId` says which, and that `reason` is on both.

**P3 — `PositionReadingBlocked`.** Taken rather than deferred, since it is the
same class of thing as the other two: the javadoc read as a raw observation
while the type is only ever made after the resolution. It now says the candidate
is what a producer records — `StructuralInspection.Blocked`, or the value
reading's account of itself — and that this is what a candidate becomes where
nothing else answered, which is why nothing outside the package constructs one.

Full reactor green, 5727 tests. The corpus gained the `reason` key on its
`partition_not_read` findings and is otherwise unchanged.

On the schema version: agreed, and thanks for checking the project's own rule
rather than the general one. I had bumped it on exactly the argument the schema
anticipates and rejects.

Issue #891, opened 2026-08-19
A coordinate's range does not take in what a rule over several coordinates says about it


kawasima:
Split out of #871, which asked for it and did not build it.

A rule over more than one coordinate is now read and drawn as a border (#871, PR #890): `300 * straw + 600 * choco <= 4800` has its own quantity, its own four points and rows offered at them. What it still does not do is narrow the coordinates it names.

`NumericDomain` keeps such a rule as written and says so:

```java
/** A form that is neither an interval nor a difference, kept as written. It proves things
 * ({@link #entails} reads it, as a premise as well as a match) and no bound is derived
 * through it — {@link #boundsOf} does not read it, so a projection is short of what the
 * rules said. */
KEPT_UNPROJECTABLE
```

So under

```
data Strawberry = Int
    invariant value >= 0
    invariant value <= 10

data Choco = Int
    invariant value >= 0
    invariant value <= 6

guard 300 * order.straw.value + 600 * order.choco.value <= 4800
```

nothing derives that `choco >= 0` and the rule together leave `straw` at most 16 — which is wider than its own bound here and would be the binding one under other numbers. Every measure that reads a coordinate's range reads it short of what the rules state.

## Why it was left out of #871

#871 names this as its own third item and says why it is the dangerous one:

> This is the part that can be wrong rather than merely absent: a narrowing derived here refuses values the rules admit, and nothing downstream would notice.

Everything else in #871 was a line that was not drawn — absent, and visible as absent. This is a range that would be *narrowed*, and a narrowing that is too tight refuses rows the model admits, with no measure downstream in a position to catch it. It wanted its own issue and its own tests rather than to ride along.

## What it takes

A projection out of the kept forms, so `boundsOf(atom)` takes in what a multi-coordinate rule says about that atom given what the others are left. Soundness is the whole of the work: a bound derived here must be implied by the rules and never merely consistent with them.

The border side is already there to test against — a coordinate whose range this narrows should not lose a border it owed, and should lose the points the narrowing genuinely refuses.

kawasima:
Fixed by #906, merged to `develop` as `db6518db`.

What the issue asked for — a coordinate's range taking in what a rule over several coordinates says about it — is in, with the projection checked against enumerated points rather than argued for.

[...]

Issue #871, opened 2026-08-19
A border over more than one coordinate draws no line, so the partition domain testing is about is not the one measured


kawasima:
An equivalence partition in domain testing is not a set of values of one position (ISTQB CTAL-TA
v4.0 §3.1.1):

> equivalence partitions are defined using expressions that combine atomic conditions by Boolean
> operators (e.g., AND, OR, and NOT) and involve one or more interacting variables. Each atomic
> condition defines a border of the equivalence partition.

Souther draws a line where a rule compares one coordinate against a constant, and where it compares
one coordinate against another. A rule comparing an arithmetic form over several coordinates draws
nothing. That is the case the technique exists for.

```
data Strawberry = Int
    invariant value >= 0
    invariant value <= 10

data Choco = Int
    invariant value >= 0
    invariant value <= 6

let total (order: Order): Int =
    Int.add(Int.multiply(300, order.straw.value), Int.multiply(600, order.choco.value))

let price (order, budget) = {
    guard total(order) <= 4800 else TooExpensive { total = total(order) }
    ...
```

```
partition   not measured (no partition axis was derived at any position)
  · not read: order.straw (the comparison relates it to another position rather than dividing it)
boundary    0/4
  ! no row is at price/order.straw = 0   (invariant Strawberry #1)
  ! no row is at price/order.straw = 10  (invariant Strawberry #2)
  ! no row is at price/order.choco = 0   (invariant Choco #1)
  ! no row is at price/order.choco = 6   (invariant Choco #2)
```

Four borders, all of them the sides of the box. The border the model is about — `300x + 600y =
4800`, whose ON points are `(16, 0)` and `(0, 8)` and whose OFF points are one step past them — is
not among them.

## What already exists

The algebra. `NumericDomain.LinearForm<A>` is `const + Σ coef·atom`, and `assume` takes any of them.
A form that is neither an interval nor a difference is kept as written:

```java
/** A form that is neither an interval nor a difference, kept as written. It proves things
 * ({@link #entails} reads it, as a premise as well as a match) and no bound is derived
 * through it — {@link #boundsOf} does not read it, so a projection is short of what the
 * rules said. */
KEPT_UNPROJECTABLE
```

The reader. `Terms.affine(Core, Denotations, Function<Core, LinearForm<FactSubject>>)` turns a Core
expression into one, with the leaf reading handed in. It runs today: a row written past such a rule
is refused at construction, so the rule is read and enforced and only the measure cannot see it.

[...]

PR #906, opened 2026-08-20
Project a coordinate's range out of the rules over several coordinates, and derive every answer from the rules

kawasima:
A rule over more than one coordinate was read and drawn as a border (#871), and it
did not narrow the coordinates it named. `NumericDomain` kept such a rule as
written and said so: `boundsOf` did not read it, so every measure that read a
coordinate's range read it short of what the rules stated.

Measured on the domain before any of this, by probing the built classes:

```
300s + 600c <= 4800, s <= 1000    s <= 1000        ->  s <= 16
  entails s <= 16                  false            ->  true
2a - 2b <= 4, b <= 10              a unbounded      ->  a <= 12
  the same rule as a - b <= 2      lossy at a, b    ->  nothing lost
a >= 20, b >= 20, 300a+600b<=4800  not bottom,      ->  bottom
                                   a in [20, inf),
                                   proves a <= -1000
a + b = c, a,b >= 0, c <= 7        a in [0, inf)    ->  a in [0, 7]
x /= 0 ; x >= 0                    x in [0, inf)    ->  x in [1, inf)
3a = 1 over Decimal                looks satisfiable ->  bottom
```

The issue named the first. The rest came out of looking, and three of them are the
kind that is wrong rather than merely absent.

## What the cause turned out to be

Not "the kept forms are not projected". Each rule was sorted, as it arrived, into a
bound, a difference, or a bucket nothing read back — and the sorting looked at the
coefficients as they were typed, so `2a - 2b <= 4` and `a - b <= 2` went to
different places and only one of them bounded anything. A disequality was turned
into a bound or dropped according to what happened to be known at that moment, so
`x /= 0` beside `x >= 0` left `x` at nought or above depending on which was written
first — the abstract state was a function of the assertion history, which is a
contract that was already broken rather than a new one to aim at.

So what is held is the rules, canonical, and what they leave is worked out from all
of them at once when something asks.

```
canonical constraints
        |
   ClosedState          exact DBM over rationals, bottom proof
        |               sound affine reduction, round budget
   +----+----+
   |    |    |
boundsOf entails refutes

ClosedState + rules + what was written out
        |
   projection audit
```

## What is deliberately not in it

Combining two general half-spaces with different positive multipliers (Farkas / LP),
and the integer frontier — congruence, parity, integer hull. The conditions under
which each deferral expires are written down where the decision is, rather than left
to be rediscovered.

The round budget is not a fixed point and does not claim to be: narrowing over a
general sum need not settle in finitely many rounds. Stopping early only leaves the
box wider, so nothing claimed rests on the budget.

Emptiness holds one way. Where this says nothing is left, nothing is.

## How it is checked

Soundness is the whole of the work here, and it is what argument is worst at: a
bound derived too tight refuses rows a model admits and no measure downstream is in
a position to notice. So it is checked against something other than argument.

- **Whole numbers, enumerated.** Every position bounded to a small range, so the
  points in it are all the points there are — which decides "nothing is left"
  rather than failing to contradict it. The difference bounds are held to sound,
  decided and tight; the reduction and the closed state to sound.
- **Decimals, eliminated.** Fourier-Motzkin over exact ratios, sharing only
  `Rational`. It is no oracle for the strictness, since over the rationals a third
  is a value like any other; it is an oracle for the value, and a value tighter than
  the rationals allow is wrong however the decimals are counted.
- **The same rules, the same answer** — over order, repetition, and at four layers.
- Each claim mutation-checked. Reading the rest of a sum from the wrong end, a
  bound one tighter, a thousandth tighter over decimals, a cycle check missing its
  strict case, a path composition dropping strictness, differences not carried, one
  round only, the exactness check always true: each turns red the test named for it,
  with the counterexample.

Four fixtures moved, all of them asserting a limit rather than a rule. The
conformance corpus did not move.

Refs #891.

### DIFF
--- a/souther-compiler/src/main/java/souther/compiler/check/FieldDomains.java
+++ b/souther-compiler/src/main/java/souther/compiler/check/FieldDomains.java
@@ -830,7 +830,8 @@ public final class FieldDomains {
             case ProjectionEvidence.Cause.Unrepresented it ->
                     "2 " + it.rule().named() + " " + it.path();
             case ProjectionEvidence.Cause.Lossy it ->
-                    "3 " + it.rule().named() + " " + it.atom() + " " + it.losses();
+                    "3 " + it.rule().named() + " " + it.atom() + " " + it.unstated();
+            case ProjectionEvidence.Cause.Rounded it -> "4 " + it.atom();
         };
     }
 
@@ -940,14 +941,20 @@ public final class FieldDomains {
                     continue;
                 }
                 for (FactSubject atom : each.atoms()) {
-                    Set<NumericDomain.Loss> losses = constraints.numbers().lossesAt(atom);
-                    if (!losses.isEmpty()) {
-                        lossy.add(new ProjectionEvidence.Cause.Lossy(rule, atom, losses));
-                    }
+                    lossy.add(new ProjectionEvidence.Cause.Lossy(rule, atom,
+                            Set.of(each.rel() == NumericDomain.Rel.NE
+                                    ? ProjectionEvidence.Cause.Unstated.A_HOLE
+                                    : ProjectionEvidence.Cause.Unstated.A_RELATION)));
                 }
             }
         }));
         causes.addAll(lossy);
+        // And whether the ends handed over are the ends the rules drew. Asked of every position the
+        // algebra speaks of, because this is not about any one rule: the reasoning reached the edge
+        // exactly and the writing could not carry it.
+        constraints.numbers().atomsSpokenOf().stream()
+                .filter(atom -> !constraints.numbers().endsAreWrittenExactly(atom))
+                .forEach(atom -> causes.add(new ProjectionEvidence.Cause.Rounded(atom)));
         // In an order that does not move between runs. Parts are keyed by the node the tree holds,
         // which is an identity, and a map keyed on one iterates by where the addresses landed — so
         // a value with two conjuncts short of the bounds printed its two causes in whichever order
--- a/souther-compiler/src/main/java/souther/compiler/check/FieldDomains.java
+++ b/souther-compiler/src/main/java/souther/compiler/check/FieldDomains.java
@@ -927,13 +927,16 @@ public final class FieldDomains {
         });
         // And what the algebra was given and what it projects does not hold.
         //
-        // Asked of the projection and of nothing else. A form neither an interval nor a difference
-        // holds is kept as written and marked lost for exactly that reason, and asking the whole
-        // state whether it still holds such a form is asking the form to stand on itself — every
-        // rule the projection dropped comes back proven. Asked of every form the algebra was handed
-        // rather than of the losses it recorded at the time: a loss is a fact about the step it
-        // happened at, the domain holds which kinds an atom has and not how many times each
-        // happened, and a second hole at an atom that already has one is a loss nothing can see.
+        // Asked of the projection and of nothing else. Asking the whole state whether it holds a
+        // rule is asking the rule to stand on itself — every rule the ranges could not state comes
+        // back proven — so what is asked is whether the ranges alone hold it.
+        //
+        // And asked of the rules after everything has been worked out, rather than read back from
+        // marks left as each rule arrived. A mark left at that moment is a history: a rule that
+        // narrowed nothing when it was read left one behind after a later rule made it bite, so a
+        // range that had become the whole of what the rules say still carried a note that it was
+        // not. What could not be stated is a property of the rule and of what the rules were found
+        // to leave, and it is worked out from those.
         Set<ProjectionEvidence.Cause.Lossy> lossy = new LinkedHashSet<>();
         readBy.forEach((rule, byPart) -> byPart.values().forEach(read -> {
             for (Predicates.Constraint each : read.stated()) {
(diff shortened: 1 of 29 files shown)


kawasima:
All five addressed in `be29b805`. Every claim reproduced first; all four measurable ones were exactly as described.

```
entails(x+y <= 5) = true            boundsOf(x+y) = [0, 10]
entails(x+y <= 5) = true            entails(2x+2y <= 10) = false
projectionEntails(a-b <= 2) = true  projectionEntails(2a-2b <= 4) = false
a = 3 -> 1·a = 3                    -a = -3 -> -1·a = -3     equal? false
```

**Both P1s had one cause and it is the one you named.** The canonicalisation stopped at the door: rules were put into one form on the way in, questions were not. A question now goes through `CanonicalForm.of` — the same function a rule goes through. My first cut did the division inside `goalOf` instead, which was a second account of what one rule is, and that is the disease rather than the fix; corrected.

Two things fell out that the review did not have to find:

- Canonicalising the goal is right for `entails` (a sign question, where positive scaling is nothing) and **wrong for `boundsOf`** (where it is everything). Without scaling back, `boundsOf(2a - 2b)` answered `-2` for `-4` — worse than the `null` it answered before. The divisor is now carried and multiplied back.
- The difference refinement recognised `±1` only, so it was the same spelling trap one level down in the reading. It now reads `k·(a - b)` for any `k` and scales the bound.

`boundsOf`, `entails` and `refutes` all read `highestProven`. Your suggested test is in, as `whatAFormRunsUpToIsWhatFollowsFromIt`.

**P2 (EQ/NE):** identified under negation via `equals`/`hashCode`, since a comparison turns around and an equality has no direction to turn. No reader depends on which of the two is held — an equality is built from both sides of itself.

**P2 (exactness):** taken as scope rather than implemented, in #907. Your `a + b = c` example is the issue body. It also records why "one half-space with every other position bounded" is not a certificate — the residual minima need not be simultaneously attainable — so that reasoning is not rediscovered.

**P3:** both javadocs were wrong, not just `tighterLower`. An exclusive cut is the tighter on *either* side; what runs the other way is which *value* is. So my original "deliberately not Comparable" reasoning was right in its conclusion and wrong in its stated reason.

The background review found four more, all mine: `AdditiveImage.isExact` had no reader while its javadoc named the one that would have had to ask it (dropped; #907 records that the mixed case needs it back); `DifferenceBounds.canHold` asserted a coordination between sorters that do not exist; `Reduction.foundNothing` was a second way to ask what `ClosedState` decides by comparing boxes; and one of my own assertions was `assertTrue(counter >= 0)`, which checks nothing.

5927 tests green, lint green on a clean recompile. Each fix mutation-checked: reverting the goal canonicalisation, the scale-back, the single reader, and the equality identity each turns red exactly the test written for it.

kawasima:
Correction to my previous comment: that was five fixes for five findings, not a diagnosis. Pushed `6bab068f`, which finds the causes and removes them.

## The five findings were consequences of two causes

**The meaning of the six relations was written three times** — storing a rule, asking whether one holds, refuting one. Only the first was the canonical reading, so everything that reading does, the other two did not: divide a comparison through by what its weights share, move a bound onto a value the sum can take, reduce `>=` to `<=` by turning the form around.

| finding | how it follows |
|---|---|
| queries not scaled | the storing reading scales; asking is a separate switch that does not |
| `2a - 2b` not seen as a difference | asking never reaches the canonicalisation |
| `boundsOf` ≠ `entails` | asking grew its own goal type, and a second bound-reader on it |
| `a = 3` ≠ `-a = -3` | storing reduces `>=` by negating and has no such reduction for `=` / `/=`, so the sign freedom survived in exactly those two |

**"An equality is a bound above and a bound below" was written three times**, in the three readers that wanted it — and I wrote two of those in this branch.

## What replaced them

A question is not a second kind of thing. It is a comparison, which is what a rule is, so it goes through `AffineConstraint.of`. Both switches over the relations are gone; refuting is proving the opposite comparison, and which is the opposite of which is one fact stated once.

A constraint answers for itself what half-spaces it states (`halfSpaces`); the difference bounds, the reduction and the proof ask it.

This subsumes the earlier patches — the hand-rolled division inside `goalOf` is gone with them.

## Two behaviour changes that follow, stated rather than buried

**An equality can be refuted.** The old refuting excluded equalities with no reason given, and proving the opposite comparison has no room for that carve-out. Under `x <= 3`, `x == 5` is now *definitely violated* rather than *not shown*. It cannot refuse a program that was being accepted — where this is true the equality was not being discharged either — so what changes is which of the two is said.

**A question naming a position the domain was never told of is answered before the reading**, since the reading wants its spacing. Nothing here bounds such a position, so nothing here proves anything about it.

Both pinned by tests.

## On the earlier findings

`AdditiveImage.isExact`, `DifferenceBounds.canHold`, `Reduction.foundNothing` and my `assertTrue(counter >= 0)` were a third thing rather than a consequence of either cause: javadoc describing the design I meant instead of the code I wrote. Those stay fixed as they were — dropped, or corrected to say what the code does.

Exactness certification is scope rather than a fix, in #907, including why "one half-space with every other position bounded" is not a certificate.

5929 tests green, lint green on a clean recompile. The mutation checks were re-run after the restructure: reverting the single reader or the scale-back still turns red exactly the test written for it.

PR #918, opened 2026-08-21
Settle what a module emits from expanding it, not from predicting it

### LINKED ISSUE #913: A standard-library function reached only from an `ensures` is refused, because its private helper is resolved in the calling module
A module whose only reach to a standard-library function is from an `ensures` clause does not
compile, where the same call written anywhere else in the same module does. The function has
to be one whose body reaches a `private` member of the library — `List.all`, `List.any`,
`List.allDistinctBy`, anything over `List.fold`.

## Reproduction

```souther
module example.donly

data Age = Int
    invariant value >= 0

data Person = { age: Age }

data Count = Int
    invariant value >= 0

behavior countDirect : (people: List<Person>) -> Count
    ensures direct = List.all(p -> p.age.value >= 18, people) && value.value >= 0
```

`souther compile`, on `develop` at `db6518db`:

```
-- MODULE ERROR  E1506 ---------------------d_only.sou:12:22

12|     ensures direct = List.all(p -> p.age.value >= 18, people) && value.value >= 0
                         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

`List.foldFrom` is not a standard-library function. The code is written in `List.all`, which
this compile has no source for, so this is where it was reached from and not where it is.
```

`souther examples` says the same, so it is not the report's doing. `List.allDistinctBy` fails
the same way and names `List.map`.

## What tells the failing case from the passing ones

Four writings of the same call, each in its own module:

| where the call is written | result |
|---|---|
| a data's `invariant`, and nothing else in the module reaches `List.all` | compiles |
| a `let` body | compiles |
| an `ensures`, and a `let` in the same module also calls `List.all` | compiles |
| an `ensures`, and nothing else in the module reaches `List.all` | **E1506** |

The third row holds with the `let` written *after* the behavior, so it is not the order of
the source. It is whether something else in the module reaches the same library function.

An `ensures` over a kernel that reaches no private member — `value.value == List.length(people)` —
compiles on its own.

## Why it is not the author's mistake

`List.foldFrom` is `private let` in `souther/list.sou:29`, and nothing in the model names it.
It is reached from the body of `List.all`, which is the library's own code. What the author
wrote is a documented clause: spec §invariant-expressions names `List.all` / `List.any` /
`List.contains` / `List.distinct` / `List.allDistinctBy` as the quantifiers a rule may use,
and `ensures` classifies a rule the way a data's clause is classified (spec §ensures).

`souther-compiler/src/test/java/souther/compiler/query/HowMuchOfARuleTheCheckReadsIsAskedPerCaseTest.java:141`
compiles `ensures List.allDistinctBy(.name, value.value) && id.value > 0` and passes (9 tests,
green). The same source through the CLI is E1506. Whatever the unit harness does differently
is where the two answers come from.

## Where it looks like it is

`check/Resolve.java:1422`:

```java
if (Reserved.isNamespace(reachable.module()) || !Prelude.isPrivateMember(written)) {
```

A private library member is in scope only while the module being resolved is the library's
own. The comment eleven lines above says why the rest of the question is left to the check:

> asking here would tie the answer to how much of the library has been loaded, and the
> library resolves its own sources while it loads

Which is what the table above measures: the answer depends on how much of the library has
been loaded, and an `ensures` is resolved during `Resolve.ensures` (`Resolve.java:870`, reached
from the spec at `:580`) — early enough that a module reaching `List.all` from nowhere else has
not caused the library to resolve its own sources yet. Not verified past the correlation; the
mechanism is what the fix has to name.

## What it blocks

A selection rule stated as a postcondition. A behavior answering a `List` whose rule is "every
element the query returns satisfies P" is written with a quantifier and no other way, so the
clause cannot be written at all unless the module happens to call the same function somewhere
else.

kawasima:
Closes #913.

## What was wrong

A module whose only reach to a standard-library quantifier was an `ensures` clause did not compile. `List.all` is an ordinary library helper derived from the fold, so a rule naming one holds a call to `List.foldFrom` once it has been expanded — and the module was never given that method.

The cause is not a missing case in a traversal. `Prepared` **predicted** which recursions a module would have to emit, by walking the places a module writes expressions, before any of those expressions had been expanded. `ensures` was not one of the places it listed. The same `HelperInliner` instance then left the `List.foldFrom` call standing inside `ClauseHelpers.withInlinedEnsures` a few lines later — the fact was already there, and a second ledger was predicting it.

Two exits, one cause. From the library it was reported as E1506, blaming the author for a private name they never wrote and are not allowed to write. From another module's published recursion it was an `IllegalStateException`.

## What `takenOn` was carrying

One `List<FnDef>` held four independent relations:

| | now |
|---|---|
| which names a standing call can be typed against | `RecursiveCallSigs(module, policy)` — from the declarations in reach |
| which recursions this module processes | `RequiredRecursiveDefs` — from what its expansions could not remove |
| which methods it emits | the same answer |
| the definitions minted for its rows | `RowFixtureDefs` |

The first is wider than the rest and always was: what *can* stand follows from the table, and what is emitted follows from use. Answering the first with the second is what left a rule with no fold to call.

## Shape

Each expansion answers with the recursions it could not remove (`Expansion<T>`), and the requirement is made at the decision that creates it rather than recovered afterwards.

The roots are the trees that survive to run — behavior implementations, row operands, clause expansions. A helper that does not recurse is no root: it is expanded into whatever reaches it, and that expansion answers for what it left standing. A fold behind a helper nothing reaches is therefore not emitted, which falls out rather than being excluded.

The closure is the **expansion's**, taken to a fixed point — not the call graph's. A body reaches a recursion by applying it and by reading a zero-argument value whose own body applies it, and only the first is a call, so a graph of calls answers about a narrower relation than the one that decides this. Each required recursion is expanded in its turn; being required and having been read are two things, so a recursion this module declared is queued like any other.

`HelperGraph` keeps one job: whether a declaration is on a call cycle, which is what decides that a call is left standing at all.

No traversal over "where a module writes expressions" survives. Adding a construct that can hold an expression needs only the code that lowers it.

`Hir.Module.takenOn` remains as what it always was on a *lowered* module — the methods emitted beyond the declared ones — and is empty at every stage before that, which a test holds it to.

## A state's identity

A query's answer is a value, and the store decides what to re-ask by comparing the answer it recomputed with the one it held. Twice here a state gained a component while its hand-written identity went on comparing what it compared before — a settled module's standing recursions, and a prepared module's row definitions, which are built from a behavior's signature and so change while this module's text does not.

Both are fixed, and the pattern is now checked against the class rather than against a reading of it: every field a state holds must be reached by its `equals`, through the projections it compares. A field left out is named with its reason or it is a failure.

## Verified

- 6040 tests in `souther-compiler`; `CI=1 mvn -Plint clean verify` green across all modules.
- The regression test loads every emitted class, which is where the JVM verifies each method, and applies the emitted fold.
- Told apart from a silent skip throughout: a module with a plain type error in a body stays refused with E1317 at every step. That control caught two regressions while the check's input set was being moved, and the suite caught a third that the control could not (101 checks stopped firing).
- Each regression test was made to fail for its own reason before being kept.

## Notes for review

- `Prelude.entry` is the discriminator for E1506, not `isLibraryFunction`: a sugar has no declaration, so `List.fold` written with the wrong number of arguments is the author's mistake and must stay a diagnostic.
- `RecursiveHelperSigs` was not widened in place. Doing so makes `RecursiveHelperConstructs` absent and removes the behavior body check entirely, with no diagnostic. It is deleted now, along with `RecursiveHelpers`.
- A sugar's edge in the call graph comes from `Prelude.rewriteOf` rather than from the spelling `List.fold` written out in two walks.
- `sigsOf` throws only for a declaration this module wrote; a foreign one that cannot be read contributes no signature and is reported by its owner. A valid foreign declaration still gets its signature.

### DIFF
--- a/souther-compiler/src/main/java/souther/compiler/ast/Hir.java
+++ b/souther-compiler/src/main/java/souther/compiler/ast/Hir.java
@@ -332,6 +332,18 @@ public interface Hir {
      * impossible. The arity says something has to be passed; that it is this module's own is what a
      * reader of the rebuild has to see, which is why every one of them names it.
      */
+    /**
+     * A module at whatever stage the pass holding it has reached.
+     *
+     * <p>{@code takenOn} is the methods a lowered module emits beyond the ones it declared: a
+     * recursion an expansion of it could not remove, and a value minted for one of its rows. It is
+     * empty at every stage before lowering and is written there and nowhere else. It once carried a
+     * prediction — worked out by walking the places a module writes expressions, before any of those
+     * expressions had been expanded — and a tree that walk did not know to look at contributed
+     * nothing, which is how a rule reaching a quantifier came to be compiled without the fold it
+     * became. What a module emits follows from expanding it, so it is not knowable here until it has
+     * been.
+     */
     record Module(String name,
                   List<String> exposing,
                   Map<String, RetType> exposedOutputs,
--- a/souther-compiler/src/main/java/souther/compiler/ast/Hir.java
+++ b/souther-compiler/src/main/java/souther/compiler/ast/Hir.java
@@ -313,13 +313,20 @@ public interface Hir {
      *
      * <p>{@code fns} is what the source wrote, and it stays that at every stage. {@code takenOn} is
      * what the module emits as methods of its own without having written them, which is two kinds of
-     * definition: a recursive helper it reaches, which cannot be inlined and has to be lowered
-     * somewhere, and a definition minted for what a row writes at a position, which has no call site
-     * to be inlined into. The first may be declared by the standard library or by a module that
-     * published it; the second is this module's own and is no {@code let}. Only {@code fns} is
-     * declared here, and no rule reads a name to tell any of them apart — {@code List.foldFrom} is
-     * reached under the library's alias and declared in {@code souther.list}, and a row's method is
-     * spelled in no source at all ({@link FnDef#role}).
+     * definition: a recursion an expansion of this module could not remove, which has to be lowered
+     * somewhere because expanding it would not terminate, and a definition minted for what a row
+     * writes at a position, which has no call site to be inlined into. The first may be declared by
+     * the standard library or by a module that published it; the second is this module's own and is
+     * no {@code let}. Only {@code fns} is declared here, and no rule reads a name to tell any of
+     * them apart — {@code List.foldFrom} is reached under the library's alias and declared in
+     * {@code souther.list}, and a row's method is spelled in no source at all ({@link FnDef#role}).
+     *
+     * <p>{@code takenOn} is written where the module is lowered and nowhere else, and is empty at
+     * every stage before that. What a module emits follows from expanding it, so it is not knowable
+     * earlier: it once carried a prediction, worked out by walking the places a module writes
+     * expressions before any of them had been expanded, and a tree that walk did not know to look at
+     * contributed nothing — which is how a rule reaching a quantifier came to be compiled without
+     * the fold it became.
      *
      * <p>Two components rather than one list a later pass appends to. Appended, every reader asking
      * what the module declared got what it declared before {@link
@@ -331,18 +338,6 @@ public interface Hir {
      * methods its bodies call missing, which is the failure this separation is here to make
      * impossible. The arity says something has to be passed; that it is this module's own is what a
      * reader of the rebuild has to see, which is why every one of them names it.
-     */
-    /**
-     * A module at whatever stage the pass holding it has reached.
-     *
-     * <p>{@code takenOn} is the methods a lowered module emits beyond the ones it declared: a
-     * recursion an expansion of it could not remove, and a value minted for one of its rows. It is
-     * empty at every stage before lowering and is written there and nowhere else. It once carried a
-     * prediction — worked out by walking the places a module writes expressions, before any of those
-     * expressions had been expanded — and a tree that walk did not know to look at contributed
-     * nothing, which is how a rule reaching a quantifier came to be compiled without the fold it
-     * became. What a module emits follows from expanding it, so it is not knowable here until it has
-     * been.
      */
     record Module(String name,
                   List<String> exposing,
--- a/souther-compiler/src/main/java/souther/compiler/check/CallElaborator.java
+++ b/souther-compiler/src/main/java/souther/compiler/check/CallElaborator.java
@@ -577,12 +577,24 @@ public final class CallElaborator {
             }
             return applySignature(call, fn, ca, expected, env, ctx).result();
         }
-        // A library name that matched no builtin or intrinsic above is a wrong stdlib call
-        // (spec §stdlib) — reported as that, not as a missing behavior. Asked of which kind
-        // of name this reaches and not of whether the spelling holds a dot: a field read
-        // applied (`deps.count(x)`) is quoted with a dot in it and reaches a binding, and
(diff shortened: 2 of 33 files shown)


kawasima:
Pushed `5875ebeb`. Both blocking findings are one cause, and I could reproduce it before fixing it.

## Root cause

`HelperGraph` is a **call** graph — its edges come from `helperCallsIn`, which reads applications. Expansion reaches further: `HelperInliner.valueOf` substitutes the body of a zero-argument value where the value is named. So `reachedFrom` was not the expansion closure. Measured:

```
let seed  = depth(...)              // depth recurses
let reads (n: Int) : Int = seed + n // reads `seed`, never calls `depth`

calls(reads)           = []
reachedFrom(reads)     = [reads]
expansion leftStanding = [depth]
```

`followingValues()` existed for exactly that edge and I deleted it, replacing its coverage with the over-wide seed set. So finding 1 was not a separate defect — it was the compensation hiding finding 2.

Nothing was broken yet, and only by coincidence on all three paths:

| path | why it was covered |
|---|---|
| this module's declarations | every declared fn was seeded |
| imported | `closeAcross()` substitutes the value before publishing (`calls(lib.walk) = [lib.inner, lib.walk]`) |
| Prelude | 0 zero-parameter Souther-bodied helpers, of 46 |

The javadoc claim "What a required helper's own body reaches is an edge of that graph, already there" was false, and it is what made the worklist look unnecessary. Withdrawn.

## Fix

Roots are the trees that survive to run — behavior implementations, row operands, clause expansions — and each required recursion's own body is expanded in turn, to a fixed point. `required` and the work list are separate, so a recursion this module declared (required before the walk starts) still has its body read. `graph.reachedFrom` is no longer used for requirements; `graph.recurses` keeps only the recursion-classification job, and a standing name that does not recurse is now an invariant violation rather than being filtered out silently.

A helper that does not recurse is no root: it is expanded into whatever reaches it, and that expansion answers for what it left standing. The dead-helper property falls out rather than being excluded.

## Tests

The three you specified, and I checked each fails for its own reason:

- `aFoldBehindAHelperAnEmittedRootReachesIsRequired` — too narrow
- `whatARequiredRecursionsOwnBodyReachesIsRequired` — too shallow. Reaches the fold **only** through a value (`calls(size) = {size}` is asserted, so the call graph cannot see it). Reintroducing `required == processed` turns it red with `[size]` and no fold
- `aFoldBehindAHelperNothingReachesIsNotRequired` — too wide. Also asserts no method is emitted

Two existing tests were asserting the too-wide answer on modules with no behavior, row or clause; both got a root that actually runs rather than a weakened assertion.

## Also in this push

- `Hir.Module`'s javadoc: my earlier commit inserted a second block and orphaned the existing one, dropping the `takenOn` rebuild contract from the generated docs. Merged.
- `InvariantSettled.equals`/`hashCode` now include `standing`. `Db` decides `changedAt` by `answer().equals`, so without it a changed requirement set would not invalidate its readers.
- `sigsOf` throws only for a declaration this module wrote; a foreign one that cannot be read contributes no signature and is reported by its owner. A **valid** foreign declaration still gets its signature — that is what the universe is for.
- `RecursiveHelpers` and `RecursiveHelperSigs` deleted; the sugar test now holds the recursion claim of `List.fold` alone.

6039 tests green; `CI=1 mvn -Plint clean verify` green across all modules.

Issue #920, opened 2026-08-21
A comparison is numbered only where a fork tests it, so a boolean value elsewhere cannot be named


kawasima:
Split out of #908, which reads the paths to a value of `&&` and `||` and is bounded by this.

A decision is what it takes of the inputs and what a run that made it would be seen to have done, and the second half of a comparison is the place a run records it coming out one way. `CoverageSites` numbers a comparison in one place — `comparisons(iff.cond())`, called from `Core.If` and only where an arm of that fork is measured — descending through `&&` and `||` to reach the comparisons a condition is built out of.

So a comparison standing anywhere else has no `ComparisonOccurrence`, `Condition.Side` cannot be made of it, and every reading that would name it falls back to naming an arm, which places at no class.

Which is a limit on the reading of #908 rather than on the rule it reads. These are the same three comparisons and only the first is read:

```
let fee (a, b, c, d) =
    (if a > 1 && b > 2 && c > 3 then 1 else 0) + (if d > 4 then 10 else 0)

let fee (a, b, c, d) = {
    let ok = a > 1 && b > 2 && c > 3

    (if ok then 1 else 0) + (if d > 4 then 10 else 0)
}
```

The second is a fork on a name, `comparisons` is handed a `Core.Read`, and nothing under it is numbered. The same goes for a boolean written into a field, handed to a call, or matched on. `AShortCircuitOperatorsPathsToAValueAreItsOutcomesTest.aComparisonNoLineIsDrawnOnIsNotOneThisCanNameASideOf` holds the second shape so that the limit is said rather than silent.

## What widening it touches

Not only the interaction reading. The numbering is what `GuardThresholds` and `Partitions` draw boundary classes from, through `OriginRef.comparisonAt()` — so numbering more comparisons adds axes and classes to a partition, which is a decision about what the rules of a model divide and not about what a body is read as. `CoverageSites` says why the numbering is deliberately wider than what gets read and why it has to be a function of the bodies alone: the emitter builds one plan and a measurement builds another, and the two are the same numbering or the probes mean nothing.

So this is not a matter of calling `comparisons` from more places. What is owed is which comparisons a line may be drawn on, answered once, with the partition's axes and the probes both reading that one answer.

kawasima:
Done in #931 (develop `4eb5cd44`).

[...]

PR #921, opened 2026-08-21
Read the paths to a value of an operator that stops as soon as its answer is settled

### LINKED ISSUE #908: A short-circuit operator's decisions are neither read nor a way in
Split out of #904, which decided not to build it (PR #905, ADR-0108).

`&&` and `||` do not evaluate both sides: the right runs only where the left did not settle the answer. So the two sides are not consumed into one value, and the interaction reading treats such an operator three ways, all of them saying nothing:

- `Interactions.meetingAt` does not form a group from its two sides. Reading them as a meeting asks for the combinations of their decisions, and the short circuit never reaches most of them.
- `Interactions.reading` answers that the value is settled one way. A product of the two sides would say the same wrong thing one level up.
- `Interactions.descend` walks into the right side only where the left is a comparison or a read it can name a side of. Where it is not, a group in there would have a way in nobody can name, so it is not looked for.

Under `A && B && C`, three decisions that between them settle one value are read as no group at all, and what is left is the pair space.

## What the shape would be

Not a product. The paths to a value of `A && B` are the left settling the answer on its own, and the left going through with the right settling it:

```
A = false
A = true & B = false
A = true & B = true & C = false
A = true & B = true & C = true
```

Which of the left's outcomes settles the answer is *which value it comes to*, and this reading has conditions rather than values. So the work is a reading of what a boolean subexpression evaluates to per outcome — not a widening of what is here.

## The neighbouring case, which is the same shape

A fork whose condition mixes `&&` and `||` cannot say which comparison came out which way — `ControlPointId` states this and says why: under `A && B` the failing arm is reached both by a value that made `B` false and by one that never evaluated `B`. The arm therefore places at no class, and ADR-0108 drops a group any condition of which places at nothing. Two more sources of the same: the arms of an `IfConstructed`, whose fork is whether the values made the thing, and the body of a block handed to a combinator, which runs under whatever it is called with.

Each of those is a group that exists and is not offered. Under-reading is the safe direction and is where they are left; this is what it would take to stop leaving them there.

kawasima:
Closes #908.

`&&` and `||` were read three ways and all of them said nothing: the operator's value was answered as settled one way, the walk went into the right only where the left was a single comparison, and the arm of a fork on it was named by the arm — which places at no class, so ADR-0108 dropped every group inside it. Under `A && B && C` three decisions that between them settle one value were read as no group at all.

One cause. The reading held the conditions along a path and never what was at the end of one, and which of the left's paths go on to the right is which value each comes to. So the reading now answers both, and the three treatments fall out of the one answer.

## The reading

`Arrival` is a path and what this reading knows the value at the end of it to be. `Truth.UNKNOWN` is not a third truth and not a path arriving nowhere — the path is there and it comes to a value, and what is missing is this reading's answer for which of the two it came to. `outcomesOf` is a projection of `arrivalsOf` rather than a second reading, so there is one walk, one memo, and no pair of answers that have to agree.

The paths of `A && B && C` are the left's settling ones as they stand and its going-through ones extended by the right's, which is four and not the product's eight. A left path whose value is unread stops where a settling one stops, because whether the right ran at all is what the left's value would have said. Composed one operator at a time, and the same four paths come out however the run is bracketed — held by a test, because an answer that turned on the bracketing would be an answer about the writing.

`Ways` is `Known(paths)` or `Unknown`, and the type is what holds the enumeration to being complete or absent. A list with paths missing from it reads as a whole one, so `x > 10 && f(y)` answers with no list rather than with the one way to false it does know. `Known` holding nothing is a third state and a real answer: it says the value is never settled that way, which is how an arm no row reaches is told from an arm nothing can be said about.

That one enumeration names the way in for three readers — a fork's outcomes, the walk into a fork's arm, the walk into the right of the operator. `settledBy` is gone rather than kept beside it: it was this rule for the one case a single comparison makes, and leaving it would have been the same reading written twice.

An arm reached several ways is walked once with those ways carried together, and a group found inside it is offered once per way, because a way in is a conjunction. Under `A && B` a row that made the first comparison fail never evaluated the second, so the failing arm has two paths and the group in it is offered under each.

## Two bounds and they bound different things

How many ways in one position is read under is bounded separately from the outcome bound. One is a product taken at one node and the other a product along the way down, so they multiply with the nesting of the body and neither holds what the other does — four conditions of two ways each is one meeting read sixteen times.

What is counted is the ways that survive being held to what already held, not the alternatives the syntax offers. A way settling a decision the way above it settled the other way is no path, and letting it take a share would move where the bound falls with how the body is written.

Over the bound the fork that would go over names its arm, which is what every fork was named by before any of this — so going over asks for no more than was asked for then. A single way in never reaches the bound, so nothing that is read today falls back.

## One fix found on the way

`disagrees` compared what a condition is about and not how it came out, so two namings of one decision settled the same way read as a contradiction. `and()` was unchecked, so this only surfaced once the same rule was used to hold a way in to the one above it — and it took away the only path under `if a then (if a then …)`. It now asks what its own javadoc says it asks. A decision named twice at two places is one decision on the way in and two places a run passed, and both are kept.

## What has not moved

The operator is still not a meeting of its two sides: `meetingAt` answers null for it, and the combinations of the two sides' decisions are still combinations no path takes. `Outcome`, `Factor`, `Decision`, `Condition` and `Interaction` are unchanged, so `InteractionCells`, `Generator` and `Adequacy` are untouched. No measure changes; this is read for the generator and nothing else.

ADR-0109 keeps ADR-0108's reading that the operator is not a meeting and replaces the clause going on from there to "so the answer is that the operator is settled one way, which asks for nothing".

## Where this reaches, and where it stops

A comparison can be named only where the plan draws a line on it, and `CoverageSites` numbers a comparison only in the condition of a fork. The same three comparisons given a name before the fork are read as none. That is a limit of the numbering rather than of this reading, it is filed as #920, and a test holds the shape so the limit is something said rather than a silence.

Left where they are, and for a different reason: the arms of an `IfConstructed`, and the body of a block handed to a combinator. Neither is a value whose truth would help — both are conditions no class of an input names — and offering them would move ADR-0108's decision to drop a group any condition of which places at nothing.

## Verification

New: `AShortCircuitOperatorsPathsToAValueAreItsOutcomesTest`, `AWayInIsHeldToWhatAlreadyHeldTest`, `APositionIsReadUnderBoundedlyManyWaysInTest`. The first run was red on four of them — a meeting in a later operand was found not at all, and the arm of a fork on a chain was named by the arm.

- `mvn test`: green across every module, 6030 in `souther-compiler`, 0 failures
- `CI=1 mvn -o -Plint clean verify`: BUILD SUCCESS

The models are written without helpers, so what they measure is this reading and not what a helper is spliced into a body as.


kawasima:
Three findings addressed. `4e6d673f`, `194073a7`, `d4a781ac`.

## The bound counted a per-invocation product, not the contexts at the position

Real, and it cost more than an early fallback. Measured on four conditions of two ways each with a fifth that asks again about what the first one decided:

```
GROUPS = 8
  reach = [a=On, c=On, e=On, g=On, fork 26/0]
  ...  (all eight ending in the arm)
```

Eight of the sixteen contexts cannot settle the last condition the way its arm needs, so the ways they would have taken are no paths and the meeting is reached sixteen times — the bound exactly. Counted one context at a time it read as eight contexts each offered two ways, called it thirty-two, and fell back to naming the arm. An arm places at no class, so `InteractionCells` dropped all eight: nothing reached the generator at all.

Option A, as recommended. `walk` and `descend` carry `List<List<Decision>> reaches` and `contexts` is gone. The number checked against the bound is `heldTo(reaches, ways).size()`, which is the number there is, and it is checked at the one place the contexts of a position are all in hand. The bound holds by induction rather than by a running count: what an arm is read under is either the contexts above it held to the ways the condition comes out, checked there, or one per context, which is what there already were. The right of a short circuit has no arm to fall back on, so over the bound it is not walked — noted in the code.

`factors` are read once per meeting and written down once per context, and a subtree reached sixteen ways is walked once. Regression: `theContextsThatReachAPositionAreWhatIsCounted`, `expected: <16> but was: <8>` on the previous commit.

## The generic tail let an empty arrival list out

Confirmed and fixed. On the previous commit the shape below answers with no groups at all — both arms of the outer fork went unwalked, not just one:

```
let fee (a, x) =
    if (if a then x else unreachable "…") + (if a then unreachable "…" else x) > 0
        then (if a then 1 else 0) + (if x > 1 then 10 else 0)
        else 0
```

The invariant is not that arrivals are never empty. `NormalReturn` does not prove path correlation, so `product` can be empty because each part answers on its own and no run has both answering — and whether that is the body having no path or this reading not following one is a question nothing here asks. What holds is that this reader does not publish an empty enumeration as a proof of unreachability.

So the normalisation is at `arrivalsOf`, the one place a reading is answered, and the four per-branch `out.isEmpty() ? oneWay() : out` guards are gone. Three guarded themselves and the fourth did not, which is how it got out. `Ways.Known([])` keeps its meaning and now only ever says an arm's condition is unreachable. Regression: `NoArrivalsIsNotAProofThatNoValueArrivesTest`.

## NormalReturn reads a short circuit as strict — #922

Confirmed by measurement: `if a > 1 && unreachable "…" then 0 else <meeting>` answers with no groups, and the `a <= 1` run reaches that meeting. `walk` returns at the fork, so both arms go.

Separate issue, for the reason given: `CoverageSites` calls `NormalReturn`, and the truth reading is built on the plan `CoverageSites` produces. #922 records that `comparison → both possible` must not be the base rule — `1 > 2 || unreachable` never arrives while `a > 2 || unreachable` does — so value possibility and knowledge completeness have to be told apart, `Unknown` read as no claim rather than as the full set, and the plan-free half live beside `NormalReturn` so there is one reading and not two that must agree. ADR-0109 carries the limit.

## Order and cost

Groups now come out meeting-major, and ADR-0109 takes that as the order: which of them a stopped generation reaches is said by where the meetings are written rather than by which way round a fork a row goes. No compatibility claimed for the old order.

Worst case the bound allows — four nested short-circuit conditions:

| | d=1 | d=2 | d=3 | d=4 | groups (d=4) |
|---|---|---|---|---|---|
| `develop` | 16.8µs | 21.5 | 28.4 | 37.2 | 1 |
| before bundling | 32.4 | 67.8 | 137.4 | 292.8 | 16 |
| after bundling | 28.3 | 50.8 | 95.2 | 194.7 | 16 |

The structural 16× on the walk is gone; what is left is `found` and `Interaction` construction, constant in the path dimension at `MOST_WAYS_IN`. These are synthetic numbers and no model in the suite shows a change — a real reading of the cost wants the conformance corpus.

`mvn test` green across every module, 6033 in `souther-compiler`. `CI=1 mvn -o -Plint clean verify`: BUILD SUCCESS.

PR #926, opened 2026-08-21
Read what an expression comes to apart from what its ways are called

### LINKED ISSUE #922: NormalReturn reads an operator that stops early as strict in both its sides
Found while doing #908, and it stands in front of it.

`NormalReturn.of` answers whether evaluating an expression can arrive at a value, and every reader of a body asks it first. Its own javadoc says everything strict is required. An operator that stops as soon as its answer is settled is the one node kind that is not strict, and it falls in with the rest:

```java
case Core.Binary b -> of(b.left()) && of(b.right());
```

So this is answered as reaching no value:

```
a > 1 && unreachable "no large a reaches here"
```

A run with `a <= 1` settles the answer at the comparison and never evaluates the right, so the expression does arrive. Measured against `Interactions`, everything under a fork on it goes:

```
let fee (a, c, d) =
    if a > 1 && unreachable "no large a reaches here"
        then 0
        else (if c > 3 then 1 else 0) + (if d > 4 then 10 else 0)
```

`Interactions.of` answers with no groups. The `a <= 1` run reaches the meeting in the else arm, and ADR-0109 says that path is one of the operator's — the left's settling paths stand as they are. `walk` returns at the fork because `of(iff)` is false, so both arms are lost, not only the right operand.

Under-reading, so nothing wrong is reported. It is where the limit is written down in ADR-0109's Decision.

## Why it cannot be fixed where it was found

The right answer for `A && B` is that the value arrives where the left arrives and either the right arrives too or the left can come to the value that settles the answer. The last clause needs to know what the left can evaluate to, which is what #908 built — and that reading is built on `CoverageSites.Plan`, which is built on `NormalReturn` (`CoverageSites` calls it at the arm probes and at every node it decides is inside a measured region). The cycle is the reason this is not a change to `Interactions`.

Neither direction is safe to guess. Answering `true` when the left always goes through — `true && unreachable` — puts an arm no run reaches into a branch denominator, which is a measure being wrong. Answering `false` is what happens today and only loses groups.

## What is owed

A reading of what an expression can come to that needs no numbering, with two things told apart that are easy to run together:

- what values the expression can evaluate to
- whether this reading knows that, or only has nothing to say

They are not the same and the difference decides the answer. `1 > 2 || unreachable` never arrives: the left is always false, so the right always runs. `a > 2 || unreachable` arrives whenever `a > 2`. A rule that answered "a comparison comes out both ways" would get the first wrong, and one that treated "cannot say" as "both ways" would get it wrong the same way — over-reading, which is the direction `NormalReturn` must not go. So the answer wants something like `Known({TRUE})` / `Known({FALSE})` / `Known({TRUE, FALSE})` / `Unknown`, and `Unknown` has to be read as "no claim" by both callers rather than as the full set.

That is the same shape `Interactions` arrived at for `Truth`, and it is why this should be one reading rather than two. `Interactions` needs the numbering only for the conditions along a path — `Condition.Side`, `Condition.Case`, `ControlClaim` — and not for what a value comes to: a `Bool` literal, the arms of a fork, and the composition of `&&` and `||` are all readable off the tree. So the value half can live beside `NormalReturn`, where `CoverageSites` can reach it, and `Interactions` can take its `Truth` from there and add the provenance itself.

Written down here so that the fix is one reading and not a second one that has to agree with the first.

kawasima:
Closes #922.

`NormalReturn` was reading `A && B` as strict in both its sides, so
`a > 1 && unreachable "..."` was answered as reaching no value — when every run
with a small `a` reaches one. The fork below such a condition was then a fork
nothing gets to, and both of its arms went unwalked along with every group
standing in them.

It could not be fixed where it was found. The right answer needs to know what
the left can evaluate to, and the reading that had that was built on
`CoverageSites.Plan`, which is built on `NormalReturn`.

## What changed

Three facts were being carried as two: whether a run arrives at a value, what a
boolean value comes to, and whether a way to it can be written down in the
numbering's words. The third had been folded into the second — a comparison the
plan could not place was answered as a comparison with no value — so a limit of
what could be written down was making claims about what the body does.

`souther.compiler.flow` holds the reading, over `Core` alone, with the naming as
its only parameter. `NormalReturn` and `Interactions` are both projections of
it, so there is no second account of the body to keep in step with the first.
`Interactions` loses 738 lines and keeps the walk that finds the meetings and
what to call a way; `flow` has no dependency on `coverage`.

Four things hold, and there is a test for each:

- A naming decides what a way is called and never whether there is one. A
  condition it has no words for leaves the way `PARTIAL` rather than leaving the
  arrival out.
- Not knowing is never both. A value the reading worked out no truth for settles
  nothing and sends nothing anywhere.
- A value is answered as known only where a value of what it is over stands
  behind each way, asked one way at a time. `1 > 2` comes out one way and
  `a == a` comes out one way, and a rule about the shape of the node would offer
  each of them a way no run takes. Whole numbers only, read off the ends of the
  range — so a number at the end of one closes the way past it.
- A way found twice is one way. The answers are sets.

The reading is rooted at a body, because what a name reads is not a fact about
the node that reads it: a helper spliced in binds the call's argument to the
helper's own parameter, and a name bound to a number written out is settled by
that number. Every occurrence is settled once under what binds it, and
`CoverageSites`, `UnreachableClaims` and `UnreachableReasons` share the reading
rather than reading a subtree again per node.

## What this does not read

What correlates two paths. Under
`(if a > 0 then true else abort) && (if a <= 0 then true else abort)` no run
arrives, each side is read on its own and arrives, and the operator is answered
as arriving. The same holds of the model's own rules: a behavior requiring
`a > 1` has no run with a small `a`, and `a > 1 && abort` is answered as
arriving all the same. Both are the liberty the reading of what arrives has
always taken with arms of a fork, and neither is decided here.

## Measured

`mvn -Plint clean verify`: 6087 tests, BUILD SUCCESS. The conformance corpus's
checked-in answers are unchanged, which says there is no regression and not that
there is an improvement — the corpus has nothing of this shape in it.

The example from the issue now answers with one group of two two-way factors,
offered under the comparison having failed. Breaking the witness rule turns
three tests red and breaking the short-circuit rule turns two red.

ADR-0109's second limit is rewritten: it was the numbering's and not the
reading's, and it is recorded as what it was.

### DIFF
--- a/docs/adr/0109-an-outcome-is-a-path-to-a-value-and-carries-what-the-value-is.md
+++ b/docs/adr/0109-an-outcome-is-a-path-to-a-value-and-carries-what-the-value-is.md
@@ -72,17 +72,54 @@ condition of a fork is not one of those. The rule is read where it is written in
 it is given a name first. That is a limit of the numbering and not of this reading, and it is
 recorded in a test so that it is something said rather than a silence.
 
-A second limit, and it is in front of all of this rather than beside it. Whether an expression can
-answer a value at all is read before any of these paths are, and that reading takes an operator
-which stops early to be strict in both its sides. So `a > 1 && abort` is answered as reaching no
-value, when a run with a small `a` reaches one — and everything under the fork on it, both arms
-included, goes unwalked. Which paths a short circuit has is exactly the question this decides, and
-the reading that decides it here cannot be the one asked there: it is built on the numbering, and
-the numbering is built on that answer. What is owed is a reading of what an expression can come to
-that needs no numbering, told apart from what it knows — a comparison this cannot evaluate is not
-the same as one that comes out both ways, and `1 > 2 || abort` answers no value while `a > 2 ||
-abort` may. Under-reading is where it is left until then, which is the direction everything else
-here takes.
+Whether an expression arrives at a value at all is read before any of these paths are, and that
+reading was taking an operator which stops early to be strict in both its sides. So `a > 1 && abort`
+was answered as reaching no value, when a run with a small `a` reaches one — and everything under the
+fork on it, both arms included, went unwalked. It was written down here as a limit of this reading
+and it was not one: which paths a short circuit has is exactly the question decided above, and the
+reason it could not be asked there was that this reading is built on the numbering while the reading
+that answers what arrives is what the numbering is built on.
+
+That is now the same reading asked without a numbering, and what made it one reading is that three
+facts were being carried as two. Whether a run arrives at a value, what a boolean value comes to, and
+whether a way to it can be written down in the numbering's words are three, and the third had been
+folded into the second: a comparison the plan could not place was answered as a comparison with no
+value, so a fork on it lost its arms and an operator over it lost its paths. Held apart, the reading
+of what arrives is the reading of the paths with the naming taken away, and there is no second
+account of the body to keep in step with the first.
+
+Four things hold of it, and they are what the separation is.
+
+A naming decides what a way is called and never whether there is one. Swapped for another, the same
+nodes arrive and the same values are worked out; what moves is how a way is written and whether it is
+written whole. A way the naming has no words for is carried as one written in part, not left out — so
+an arm reached by a condition nothing can place is still an arm, and only the reader that steers rows
+turns the incompleteness away.
+
+Not knowing is never both. A value this reading has worked out no truth for settles nothing: it does
+not send an operator into its right and it does not stop one there, and no path is offered on the
+strength of it. A way to a value that arrives is offered where the way is worked out or where what
+follows it arrives anyway.
+
+A value is answered as known only where a value of what it is over stands behind it. Not a rule about
+the shape of the node: `a == a` is a comparison and comes out one way, `1 > 2` is a comparison and
+comes out one way, and a reading answering "a comparison comes out both ways" would offer each of
+them a path no run takes and say the expression arrives where every run aborts. So each way is asked
+about on its own, against the range of the position compared — which for a whole number is read off
+the ends, so a number at the end of the range closes the way past it. A way nothing whole stands
+behind is not offered, and where nothing stands behind either the value is answered as one this
+reading has nothing to say about. Under-read, and the direction everything here takes.
+
+A way found twice is one way. The reading answers a set for that reason: what is counted anywhere
+downstream is what the body does and not how many times the reading got there.
+
+What is still not read is what correlates two paths. Under
+`(if a > 0 then true else abort) && (if a <= 0 then true else abort)` no run arrives, and each side is
+read on its own and arrives — so the operator is answered as arriving. The same holds of the model's
+own rules: a behavior requiring `a > 1` has no run with a small `a`, and `a > 1 && abort` is answered
+as arriving all the same. Both are the liberty this reading has always taken with arms of a fork,
+which are counted without anything having asked whether the condition can come out that way, and
+neither is decided here.
 
 Two ways in that settle one decision opposite ways describe no run, so the walk does not go that
 way rather than going and leaving what it finds to be thrown out later. Which decision it is is not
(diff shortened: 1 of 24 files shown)


kawasima:
Both reviews are in. Four findings, and they were one thing said four ways plus one enumeration written out twice — so this is a fix to the structure rather than four patches.

**The naming was still reaching answers that are not its.** The invariant said a naming may not change what arrives or what a value comes to, and I kept it by hand for one of the three channels — a condition it has no words for. The other two went on moving it: `join` returning null dropped an arrival, and the arrival bound collapsed a node to one unread way. Reproduced before touching anything, with the body from the review:

```
if ((match a with | On -> true | Off -> false)
        && (match a with | On -> false | Off -> true))
        || unreachable "..."
```

The plain reading answers that the `||` arrives; the coverage naming, which sees the two arms of one decision contradict, answers that it does not. Same body, same reading, two answers about what the body does.

So it is a structure now. `Comes` says what the body does and is computed with no naming at all; `Paths` says what the ways are called and is the only half a naming touches. A naming that refuses a way, has no words for one, or will not hold a count answers that it enumerated nothing — never that nothing arrives. There is nothing left for a test to catch.

**Which arms a fork can send a run down was answered in three places.** The reading knew; `CoverageSites.armOf` and `Interactions.waysInFor` worked it out again and got a different answer. That is the same shape as #922 itself, and this PR made it reachable: `if a > MAX && unreachable "…" then 1 else 2` now arrives, so both arms were numbered and one of them is an arm no run enters — a row owed for a combination the body has no run at. Asked of the reading now, and asked in the words the question is in: an arm stands unless the reading worked out that the condition never reaches it. Which is also the answer to the second finding — that question is not `comesOut(e, want)` and no longer shares a name with it.

**The witness question was asked of a comparison rather than of a value.** A position of the input holding a truth comes out both ways and is no comparison, so `flag && unreachable "…"` read as arriving nowhere. Widened; the whole suite stays green, because the split is partial where nothing can name it and `Interactions` never sees it.

**The reading claimed every occurrence was settled while two enumerations of a node's parts were written out beside `Core.forEachChild`.** The claim went unchecked because a node the reading had missed was quietly re-read as a body of its own, with every name in it free. It raises now — and that is what found a third gap: settling a node is not descending into it, so an ancestor settling one was stopping the walk before its arms. That one was live, not latent.

Making it raise also turned up something the reading had been hiding: `UnreachableClaims` was reading claims out of subtrees nothing reaches, and could not say why an arm aborts when the `unreachable` is in a `let` above it. It carries reachability now, the same rule the numbering already had.

Also done: `Ways.Known` and `Paths.Held` copy their lists, `Comes` is unmodifiable, and the property test asks for equality over four namings — one of which tells two forks apart by the position rather than by the node, which is what the coverage numbering does and is the shape the old test could not build.

`mvn -Plint clean verify`: 6089 tests, BUILD SUCCESS.

kawasima:
Fixed in `374dfaf6`.

**Root cause.** The bound is a rule about the work, and I turned it into a predicate on the answer. It happened in the same commit that separated the two halves: the rules came out typed over `List<Arrival<P>>`, a list has no notion of "too many", so the only place left to ask was after one was finished. `MOST_OUTCOMES` went on existing and stopped meaning anything.

**Mechanism.** The combining rules are typed over `Paths<P>` now, and `Beyond` absorbs — `product` answers it before taking the pairing rather than after, and so do `through`, `fork`, `arms` and `attempted`. Every way goes in through one accumulator that stops accepting them once there are too many, so nothing can build past the bound and then classify. `Comes` is untouched: what the body does goes on being read exactly.

**Test.** Counting rather than timing, as you said. A naming with `mostArrivals() = 2` over one value made of fifteen independently forked parts — a construction and not a sum, since a sum is fifteen nested nodes and a per-node bound already stops at the first. It asserts `Paths.Beyond` and that the naming was asked to put two ways together far fewer times than the combinations would need.

Put the bound back where it was and the test goes red at **65,624 joins and 235 seconds**, against 0.7 seconds now. That is the regression, caught structurally.

`mvn -Plint clean verify`: 6091 tests, BUILD SUCCESS.

Issue #911, opened 2026-08-21
A row for a form's border is searched in the declared domain, not in the region that reaches the guard


kawasima:
A border on an arithmetic form is searched for by choosing a level and solving the form for
positions at it. The box that search runs in is handed in by the caller, and what it hands in is
each position's declared domain — what its type and that type's invariants leave. A guard draws a
threshold and an axis and narrows no domain, so a border a guard owes is searched over the whole of
what the positions could ever hold rather than over the region that reaches the guard.

Two things follow, and they are the same fault seen from either side. Too wide, and a level with an
answer in the region comes back as one the search did not reach. Not used at all, and the row
offered for a level is one a rule above the guard refuses.

### A level with an answer, reported as unreached

```
data Small = Int
data Large = Int

behavior checkout : (x: Small, y: Large) -> Order | Rejected

let checkout (x, y) = {
    guard x.value >= 0 else Rejected { reason = "a" }
    guard x.value <= 20 else Rejected { reason = "b" }
    guard y.value >= 0 else Rejected { reason = "c" }
    guard y.value <= 10 else Rejected { reason = "d" }
    guard x.value + 2 * y.value <= 16 else Rejected { reason = "over" }
    Order { total = x.value }
}
```

```
· not known to be writable: the OFF point checkout/x + 2 * y = 17 (guard@12:5)
  — nothing composed one: the search stopped before reaching x + 2 * y = 17
```

`x = 1, y = 8` is at that level and passes every guard above it.

Move the same four bounds off the guards and onto the types, and the row is found:

```
data Small = Int
    invariant value >= 0 && value <= 20
data Large = Int
    invariant value >= 0 && value <= 10
```

```
! no row is at the OFF point checkout/x + 2 * y = 17 (guard@10:5)
//     | (Small(1), Large(8)) -> <?>
```

Same equation, same region, same answer. What moved is where the bounds are written.

Coefficients of one hide it: `x + y = 11` is solved at the first value tried, so the shape shows up
on forms whose coefficients are not all one.

### A row offered for a level it cannot reach

Bounds on the types, and one guard above the form's:

```
let checkout (x, y) = {
    guard y.value <= 6 else Rejected { reason = "too many large" }
    guard x.value + 2 * y.value <= 16 else Rejected { reason = "over" }
    Order { total = x.value }
}
```

```
! no row is at the ON point checkout/x + 2 * y = 16 (guard@11:5)
[...]

PR #927, opened 2026-08-21
Carry a rule across rather than composing what each reading answered

kawasima:
Preparation for #911, and it changes no answer on its own.

## What was wrong

The rules reaching a behavior's input are read one parameter at a time, and what
a form over several positions runs between was assembled by solving each
parameter's part of it and adding the results. That composition rests on two
parameters being related by nothing — which is true of the declarations, and of
nothing a body writes. ADR-0110 says so outright: *"A day when a behavior's own
clauses relate its parameters is the day that composition has to change, and it
is the only place it does."*

#911 needs a guard's cut in the box a search walks, and a guard relates
parameters. There was nowhere for such a rule to be said.

## What this does

Each parameter's declaration-derived numeric rules are carried into one
vocabulary and said together, and what a form runs between is projected out of
that space.

* `CanonicalForm.over` / `AffineConstraint.over` / `NumericDomain.over` — a
  renaming. What a rule says is a relation between numbers and it says the same
  thing whatever they are called, so this is not a second reading of the
  declaration.
* `NumericDomain.meet` — two sets of rules said together.
* `FieldDomains.Settled.numbersOver` — hands the numeric rules over in a
  caller's names. **Every number a rule is about is carried**, including ones the
  caller has no coordinate for: a rule reaching one of those still holds two
  named numbers apart, and dropping it would leave the answer wider than it is.
* `InputAtom` — the vocabulary. `Named` is keyed by coordinate (parameter, path,
  measured) and not by `NumericTerm`, because a term carries which measure was
  written and that would make one number two.
* `ReadQuantities` — one `NumericDomain<InputAtom>` instead of per-parameter
  composition.

Non-numeric emptiness stays per parameter. Which values a position admits, which
predicates hold, where an ordering stops — none of that relates two parameters,
and it is asked where it was proved.

Removed with the composition it belonged to: `Part`, `partsOf`, `reachOf`,
`relationallyOf`, `relatedTo`, `atOneTerm`, `tighter`, and
`FieldDomains.Settled.names`. `Settled.within` stays — a test measures the
ADR-0110 property through it.

## What was measured

`mvn -Plint clean verify`, then the full suite from clean: 6130 pass, no
checked-in answer file moved.

**The corpus is not a control for this, and saying so is the point.** With
`NumericDomain.meet` deliberately made a no-op — every declaration rule dropped
from the state — `TheAnswersAboutEachConformanceCorpusAreTheOnesCheckedInTest`
stayed green. Five assertions in `inputs` went red and nothing else did. The
corpus asks no multi-term relational question: a single term is answered by
`whereOneTermRuns` and never reaches the meet.

So what holds this change is `WhatIsFixedIsAskedTogetherHoweverItArrivedTest`,
`WhatIsKnownOfOneTermSurvivesWhatIsUnknownBesideItTest`, and the algebra
properties added here — a renaming that is not one-to-one is refused, saying two
readings together is order-independent and idempotent, a contradiction leaves
nothing, one number spaced two ways is refused.

The same holds for the follow-on: a guard's region will not move a corpus answer
either, so #911 has to bring its own checked-in test.

Refs #911.

### DIFF
--- a/souther-compiler/src/main/java/souther/compiler/check/FieldDomains.java
+++ b/souther-compiler/src/main/java/souther/compiler/check/FieldDomains.java
@@ -630,15 +630,53 @@ public final class FieldDomains {
         }
 
         /**
-         * Whether a range is taken of this coordinate here, which is whether a rule about it could
-         * have been read into these at all.
+         * The numeric rules these came to, about the same numbers under a caller's names.
          *
-         * <p>Asked so that a caller can put the part of a form this can answer for on its own. A
-         * form with one coordinate this never named is one nothing can be said about as a whole, and
-         * said that way the rules relating the coordinates beside it are lost with it.
+         * <p>For a caller that holds the readings of several values at once and has to say what they
+         * leave together. What a rule says is a relation between numbers, and it says the same thing
+         * whatever the numbers are called — so what is handed over is these rules renamed
+         * ({@link NumericDomain#over}) and never a fresh reading of the declaration.
+         *
+         * <p><b>Every number a rule is about is carried, whether or not the caller has a coordinate
+         * for it.</b> A reading relates numbers at coordinates and numbers at none, and a rule
+         * reaching one of the latter still holds two of the former apart: left behind, that rule
+         * would be gone and what the rules leave would come back wider than it is. So a number with
+         * no coordinate is named by {@code otherwise}, out of the subject itself, and the caller is
+         * given something to be equal to rather than something to read — what makes two of them one
+         * number was settled here, and a caller inventing its own answer to that would put two
+         * numbers together or take one apart.
+         *
+         * <p>Only the numbers. What else a reading proved — which values a position admits, which
+         * predicates hold, where an ordering stops — relates the positions of one value and reaches
+         * no vocabulary a caller out here has, so it stays where it was proved and is asked for
+         * there ({@link #holdsNothing}).
          */
-        public boolean names(Coordinate at) {
-            return (at.measured() ? countAt.get(at.path()) : atomAt.get(at.path())) != null;
+        public <B> NumericDomain<B> numbersOver(java.util.function.Function<Coordinate, B> named,
+                                                java.util.function.Function<Object, B> otherwise) {
+            Map<FactSubject, Coordinate> where = new LinkedHashMap<>();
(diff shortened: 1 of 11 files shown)


kawasima:
Both findings are real. Root causes first, because they are different mistakes
that happen to sit in the same method.

## The fold read as a renaming

A form the caller writes is keyed by `NumericTerm`; the numbers the rules are
about are keyed by where they sit. A term carries which measure was written and a
number does not — that is exactly why the atom is keyed by coordinate, after a
first attempt keyed by term made one number into two.

So the translation is **many-to-one on purpose**, which makes it a fold, and a
fold owes a combining operation. `put` treats it as a renaming, and a renaming is
only correct where the mapping is injective. `merge(..., BigDecimal::add)`, which
is what the removed `named.merge(at, coef, BigDecimal::add)` did.

Worth naming the contrast, since I had one word for both: carrying a *rule*
across may not put two positions under one name — that says they are one number,
which nobody said, and `Renaming` now refuses it. Reading a caller's *form*
across may, because the caller wrote two spellings of a number this input has one
of.

Test: `twoSpellingsOfOneCountWeighItTwice`. A clause counts the collection at
five, and a form naming that count twice runs to ten. Under the old `put` it
comes back at five. No model reaches this today — which measure is taken of a
position is settled by its type, so a body has one to write — and the test says
so.

## The spacing asked of the position

Sharper than "a bound is dropped": a count is a whole number **whatever it is a
count of**, and `NumericTerm.carrierAt` says so with the position's type unused.
I routed the question through `byPath` anyway, so a fact the term owns came to
depend on whether the walk that reads this input reached the place the term sits
at. The walk stops two levels down, so a count under more steps than that had
nowhere to ask.

The old path never needed a granularity at all, so this is a fact the new
mechanism newly requires and I sourced it from the wrong owner.

Test: `aCountUnderMoreStepsThanTheWalkGoesDownKeepsItsFloor`. The form is a
bounded number plus a count at `p.m.inner.xs`, which the test first asserts has no
position. Its floor is nought; with the question routed through the position the
sum comes back with `min` of `null` — the widening the finding predicts.

Both were run with the fixes reverted and both go red.

Full clean suite: 6133 pass, `mvn -Plint clean verify` green, no checked-in
answer file moved.

Issue #909, opened 2026-08-20
A row composed for a cell is never checked to have reached it


kawasima:
Split out of #904, which asked for it and did not build it (PR #905).

#904 states it:

> Run the candidate afterwards and read `Probe` to confirm the intended outcomes were the ones observed, rather than assuming the values steered where they were meant to; a candidate that missed proves nothing about the cell.

`Generator` composes a row for a cell by narrowing each position to the classes the cell admits, choosing one, and building a value in it. Nothing afterwards asks whether running the behavior on that row took the path the cell names and settled each factor at the outcome it names. The row is offered, and named, for a combination nobody confirmed it sits in.

## Why this is the part to confirm

The step from a decision to a class of a position is the part that can be wrong, and it was wrong four times in review of #905:

- a decision was placed on the position a helper's parameter is spelled after rather than the one the argument names
- a group under an arm was offered without what it takes to take that arm
- a condition was written down as one class, so a second condition on the same position overwrote the first
- decisions inside an arm that aborts, and inside a function the body only makes, were read as groups

Every one of them shows up the same way at run time: the row does not reach the meeting, or reaches it at another outcome. A confirmation would have caught each of them from the row rather than from a reading of the reading. Two of the four could not be reproduced through the current front end at all, and their absence is a neighbouring mechanism's doing rather than a guarantee — which is exactly the kind of thing a check on the row would notice when the neighbour changes.

## Shape

`Probe` already records which control point a run passed through, and `CoverageSites.Plan` already mints the ids a cell's conditions are built on, so the question is asked in a vocabulary that exists.

ADR-0091 fixes what an answer may claim: a candidate refused at construction moves nothing into proven impossible, and it is reported as its own reason. The same holds here — a candidate that ran and missed proves nothing about the cell. It is not evidence the combination is unreachable, and the cell stays untried rather than being counted as offered.

Whether a missed cell is retried with other values, and how many times, is part of this and is not decided.

kawasima:
Merged into `develop` as part of #914. The issue closes here rather than on the merge because that PR targets `develop`, which is not the default branch.

PR #928, opened 2026-08-21
Search for a row where the rules on the way to the border leave one

### LINKED ISSUE #911: A row for a form's border is searched in the declared domain, not in the region that reaches the guard

kawasima:
Closes #911. Sits on #927 — based on `feature/relational-quantities`, so the diff
here is this change alone. Retarget to `develop` once #927 merges.

## What was wrong

A border on an arithmetic form is met by solving the form for a level, and the
box that search ran in was each position's declared domain. A guard states a
threshold and narrows no declaration, so a border a guard owes was searched for
over the whole of what the positions could ever hold.

Both symptoms in the issue are that one box.

* Too wide: with `x` bounded nowhere, one value of it is tried, `x + 2y = 17`
  solves to eight and a half, and a level with an answer at `x = 1, y = 8` comes
  back as one the search stopped short of.
* Not used at all: the row offered for `x + 2y = 16` is `(Small(0), Large(8))`,
  which the guard above refuses at `y <= 6`.

## What this does

The walk that already reads a body's comparisons carries what it assumed on the
way, and a border read off a comparison is searched for inside that.

* `ReachingCuts` — what a row has had to satisfy by the time it reaches one
  comparison, filed where the condition was assumed.
* `GuardThresholds` — the walk descends into the arms itself and takes each
  comparison **on its own terms** rather than off the arm it is in. Under
  `A && B` the `else` arm holds rows that failed either, so taking the arm for
  the conjunct would exclude rows that reach the guard.
* `Coverages.regionFor` — the region is beside the border, never inside it. Read
  off which kind of rule drew the line: an invariant is about the values and has
  nothing on the way to it, a guard is at a place in a body.

A condition the arithmetic has no word for leaves nothing behind, so the region
stays **wider** than what arrives. That is what a walk that exhausts it rests on
when it calls a level out of reach, and it is why this does not promise the
second symptom away in general — a row found here is a row to try, and what
shows it arrives is running it (#909).

## Where the region can and cannot reach

Obligation discovery is untouched: which borders are owed still comes off the
declarations. Narrowing the search may not take an obligation away, or a
coverage denominator would move as this compiler learned more of a body. The
type split in the previous commit is what holds that — `Cutting` takes a
`Quantities`, `LevelRealizer` takes a `SearchRegion`, and there is no way from
the second back to the first.

## What was measured

`mvn -Plint clean verify`, then the full suite from clean: 6134 pass, no
checked-in answer file moved.

The four assertions in `ARowForABorderIsSearchedForInTheRegionThatReachesItTest`
were written first and were red. Each was then run again with `regionFor`
returning the declared region, and all four went red — including the `else`-arm
one, which in its first shape stayed green either way. A search of the declared
box tries the positions in the order the form names them and so takes `y` at its
largest first, which answers a line whose arm wants a large `y` the same way
with or without the region; the model was rewritten so the two differ.

The conformance corpus stays green throughout and is not evidence here, for the
reason #927 gives.

### DIFF
--- a/souther-compiler/src/main/java/souther/compiler/inputs/EmptyInput.java
+++ b/souther-compiler/src/main/java/souther/compiler/inputs/EmptyInput.java
@@ -64,4 +64,17 @@ public sealed interface EmptyInput {
      * proved. Widening this is one edit, in the translation above.
      */
     record ProvedByTheDeclarationsReading() implements EmptyInput {}
+
+    /**
+     * The rules and what a caller took in on the way leave nothing together.
+     *
+     * <p>Apart from the reading's own proof because it is a different fact. What a declaration
+     * refuses it refuses of every value, and this is about a place — a row that would have to be
+     * both inside a region and at a level nothing in that region reaches. Read as the first, an
+     * author would be sent to a declaration that says nothing of the sort.
+     *
+     * <p>Nowhere in particular. What contradicts is the rules taken together, and naming one of
+     * them would be picking whichever the arithmetic happened to close on.
+     */
+    record ProvedByTheRulesTakenTogether() implements EmptyInput {}
 }
--- a/souther-compiler/src/main/java/souther/compiler/inputs/Quantities.java
+++ b/souther-compiler/src/main/java/souther/compiler/inputs/Quantities.java
@@ -97,6 +97,18 @@ public sealed interface Quantities permits ReadQuantities {
     }
 
     /**
+     * Where a row may be written, before anything a body says is taken in.
+     *
+     * <p>The one way over to {@link SearchRegion}, and it is one way. What the declarations leave is
+     * where the borders are read from, and a region is where a row for one of them is looked for —
+     * so a caller may go from the first to the second and nothing can come back. Read the other way,
+     * a region narrowed by the conditions on the way to one border would reach the reader that
+     * decides which borders exist, and an obligation would go away for this compiler having read
+     * more of the body.
+     */
+    SearchRegion region();
+
+    /**
      * Why the rules leave no value under what is fixed, or empty where nothing proved they leave
      * none.
      *
--- a/souther-compiler/src/main/java/souther/compiler/inputs/ReadQuantities.java
+++ b/souther-compiler/src/main/java/souther/compiler/inputs/ReadQuantities.java
@@ -348,6 +348,18 @@ final class ReadQuantities implements Quantities {
 
     @Override
     public Quantities given(Map<NumericTerm, Count> more) {
+        return fixing(more);
+    }
+
+    /**
+     * The same, answered as this reading rather than as one of the faces it wears.
+     *
+     * <p>What refining hands back is the state, and {@link Quantities} and {@link SearchRegion} are
+     * two ways of asking it. Typed by either of them, the other has to put back what it knows —
+     * which is a cast, and a cast is a check the compiler is not doing. The two faces stay apart
+     * because they answer different questions; what they refine is one thing and is typed as one.
+     */
+    ReadQuantities fixing(Map<NumericTerm, Count> more) {
         if (more.isEmpty()) {
             return this;
         }
--- a/souther-compiler/src/main/java/souther/compiler/inputs/ReadQuantities.java
+++ b/souther-compiler/src/main/java/souther/compiler/inputs/ReadQuantities.java
@@ -44,6 +44,14 @@ final class ReadQuantities implements Quantities {
      */
     private final Map<NumericTerm, Fixed> fixed;
     /**
+     * What a caller has taken in about this input beyond what the declarations say.
+     *
+     * <p>Reached only through {@link SearchRegion}, which is what keeps a reading of the
+     * declarations from acquiring one. Kept as what was said and not as what it came to: taking in
+     * accumulates, and what the whole of it leaves is worked out where everything else is.
+     */
+    private final List<Assumed> assumed;
+    /**
      * The rules read with everything fixed, worked out when something is asked and not before.
      *
      * <p>One reading per parameter and not one per question. A search asks where a position runs and
@@ -66,6 +74,9 @@ final class ReadQuantities implements Quantities {
      */
     private volatile souther.compiler.numeric.NumericDomain<InputAtom> constraints;
 
+    /** One thing taken in about a form of this input's terms: {@code form rel 0}. */
+    private record Assumed(NumericDomain.LinearForm<NumericTerm> form, NumericDomain.Rel rel) {}
+
     /** The values fixed at one term, kept as their least and greatest so that what was fixed does
      *  not depend on the order it arrived in. */
     private record Fixed(Count least, Count most) {
@@ -82,8 +93,9 @@ final class ReadQuantities implements Quantities {
 
     private ReadQuantities(Map<String, PlacedRules> byParameter, Set<String> parameters,
                            Map<TermPath, Position> byPath, Map<NumericTerm, Fixed> fixed,
-                           souther.compiler.check.Symbols symbols) {
+                           souther.compiler.check.Symbols symbols, List<Assumed> assumed) {
         this.symbols = symbols;
+        this.assumed = List.copyOf(assumed);
         // In the order the behavior declares its parameters. A proof of emptiness names one of them
         // and a report is a document compared against the one written last time, so an order read
         // off a hash would move which parameter is named between runs.
@@ -102,7 +114,7 @@ final class ReadQuantities implements Quantities {
     static ReadQuantities of(Map<String, PlacedRules> byParameter, Set<String> parameters,
                              Map<TermPath, Position> byPath,
                              souther.compiler.check.Symbols symbols) {
-        return new ReadQuantities(byParameter, parameters, byPath, Map.of(), symbols);
+        return new ReadQuantities(byParameter, parameters, byPath, Map.of(), symbols, List.of());
     }
 
     /** Each parameter's rules read with what is fixed under it, once. */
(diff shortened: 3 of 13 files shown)


kawasima:
Both findings addressed, plus the two from the background review of this PR.

## The blocker: the collection was per fork, not per comparison

The root cause is the unit, not a missing case. `assumed` was threaded per `Core`
node and every comparison of one condition was handed the same value — so what a
row had already satisfied was answered about the fork. A condition stops as soon
as it is settled, and that is a fact about each comparison.

The visible cost is the one named: a semantics-preserving rewrite moved the
answer. Four sequential guards fixed the form's line; the same five conditions in
one `&&` chain left it searched for over everything the positions could ever
hold.

`ReachingCuts.collect` now walks the condition in evaluation order and carries
what stands at each point. And what a subtree coming out one way establishes is
**one rule used twice** — for the operand beside it and for the arm below it.
`taking()` asks `ReachingCuts.stating` rather than reading flags off `Placed`;
the `trueThen`/`falseElse` fields I had added are gone. Written apart, the two
agreed by having been derived alike, which lasts until one of them learns a shape
of condition the other does not.

A conjunction coming out false and a disjunction coming out true state a
disjunction of cuts, which is not a list and is not approximated into one.

## Tests, and one of them was measuring nothing

Four of the five cases are in, and each was run with `collect` passing the
un-extended prefix to the right operand:

| | control |
|---|---|
| sequential guards ↔ one `&&` chain agree | red |
| `A \|\| B` — B's region takes `¬A` | red |
| `A && (B \|\| C)` — C's region is `A ∧ ¬B` | red |
| an unread condition above changes nothing | green — it is the widening direction |

The metamorphic one was green in its first shape. My model helper declares the
bounds as type invariants, so the chained model had them on the types *and* the
guards and the region had nothing to add. Rewritten without the invariants, it
goes red under the control with exactly the predicted message.

## `Coverages.assess` — threaded, and I could not make it matter

The per-axis path took `ReachingCuts.NONE`. It now takes `partitioning.reaching()`.

I could not build an input where that changes an answer, and the test I wrote for
it was vacuous, so it is not in. The points of a line at one position all sit
beside the line — `Criterion.Within.somewhereInside` reads from the end the line
is at — and a line the region excludes is one `PathReachability` has already
taken the obligation away for: the border is not built at all. That is a second
route to the same answer, not a reason to leave one path short of what it is
owed, and the call site says so.

## The performance finding: measured, no difference

Timed a wide bounded box (`6x + 10y` over two positions running to 200000) with
and without the aggregate `isBottom()` in `emptiness()`: 35–44ms against 35–47ms
warm. The suite itself is a useless instrument here — it varies 35s to 50s run to
run.

The leaf check is reached far less often than `STEPS_A_SEARCH_MAY_TAKE` suggests,
because the residue has to land exactly for the walk to get there. And
`emptiness()` already ran a closure per parameter on every call before this PR,
so the marginal cost is one closure over the union rather than a new kind of
work. I did not find a model where it shows.

## D5

Filed as #929, with the reason it is not folded in here: what a search is correct
to do and what a report is owed about it are two questions, and the current
silence hardening into a decision is what that issue is there to stop.

Full clean suite: 6139 pass. `mvn -Plint clean verify` green. No checked-in
answer file moved.

kawasima:
Both findings are real and both are fixed.

**The javadoc.** The `@param reaching` went in as a block of its own above the
declaration, which left the javadoc it sat under attached to nothing. Merged into
the existing block.

The root cause is sequencing rather than the edit: I ran an orphaned-javadoc
check over the touched files earlier in this PR and then made this insertion
after it. A check treated as a gate you pass once is not a check. It now runs over
every file the branch touches as the last thing before committing, and the only
hit left is one that predates the branch.

**The cast.** `ReadRegion.given` asked for a narrower reading and got back the
interface the *other* face is named by. It held only because one implementation
answers both faces today, which is not something the compiler was checking.

The root cause is that I made two faces share a state and then typed the
refinement by one of the faces. What refining hands back is the state; the faces
stay apart because they answer different questions. `ReadQuantities.fixing`
returns the state, and both `given` implementations go through it. No cast left in
the package.

Full clean suite: 6141 pass, `mvn -Plint clean verify` green.

kawasima:
P1 fixed, and the loss was larger than the finding says.

## What it costs today

Extracting one comparison of a condition into a helper:

```
inline:  borders 6, one axis on `y`, lines at `y = 6` and `y = 7`
helper:  borders 5, "no partition axis was derived at any position"
         · not read: guard@14:42 — written in a form this compiler does not read, about `y`
```

Not only the region. The **line goes away and the position comes back divided no
way** — and that last report line is the root cause showing itself: the reading
that finds comparisons anywhere in a condition sees the comparison and calls it
unreadable, while the reading that draws lines never reaches it. One condition,
two answers, depending on who asks.

## Root cause

"What a condition is made of" was answered independently by every reader, off the
`Core` node in front of it. Each knew on its own which shapes are transparent,
which combine, which are atomic, which are where this compiler stops. So a shape
one learned to see through stayed invisible to the others.

Adding `Core.LetIn` to `collect` and `stating` fixes today and leaves that. So the
vocabulary has an owner: `Condition` — `Both`, `Either`, `Compares`, `NotRead` —
read once per fork, and `comparisonsIn`/`placed`, `ReachingCuts.collect` and
`ReachingCuts.stating` fold over it. A binding is looked through and its names
taken in rather than being a shape, and `Compares` carries the reading in force
where it stands, because a comparison inside an expanded helper read against the
outer names is about nothing.

## A boundary the fix ran into, stated rather than papered over

Making the reading see through bindings made `read()` ask the plan for a site the
plan does not have: comparisons inside an expanded helper are not numbered, which
`AComparisonInsideAHelperRaisesNothingTest` already writes down as a gap of its
own.

That is the right place for a line, and it is the same split as the rest of this
PR one level down:

* **what a condition establishes** — a fact about the model, wants no
  instrumentation, so the region takes in every comparison the reading reaches;
* **what a condition owes a row at** — takes measuring that the comparison
  answered, so a line is drawn only where the plan numbered a site.

So the helper's own comparison still raises nothing (that gap is untouched), and
what it says now narrows the region beside it.

## Tests

Two metamorphic ones — inline against the same condition with one operand named,
for `&&` and for `||`. Both were run with `Condition` stopping at a binding and
both go red; the `&&` one reports `[Small(0), Large(8)]` where the inline spelling
gives `[Small(4), Large(6)]`.

## On the four readers

`comparisonsIn`/`placed`, `collect` and `stating` now share the reading.
`GuardThresholds.compared` does not, and that is deliberate: it asks which
comparisons appear **anywhere** in a condition, including inside an argument to a
call or under an operator this does not read, which is a walk of the whole subtree
rather than of the condition's structure. Reading one for the other is how a
comparison nothing could turn into a line would come back as a line. The boundary
is stated in `Condition`'s javadoc rather than left to be rediscovered.

Full clean suite: 6156 pass. `mvn -Plint clean verify` green. No checked-in
answer file moved.

Issue #912, opened 2026-08-21
A rule written about the elements of a collection reaches no position, so nothing inside one is measured


kawasima:
`StructuralInspection.of` answers `Blocked(UnsupportedTraversal(SEQUENCE_ELEMENT))` for
`Shape.Sequence`, and `TermPath` is a parameter and the fields read off it — there is no step
for an element. A position typed `List<T>` is where the walk stops, so every rule written
about what is inside one has no position to be about, and nothing the derivation does for a
bare position is done for it.

## What it produces

The same predicate twice — once on bare positions, once on the elements of a list:

```souther
behavior classify : (age: Age, status: Status) -> Verdict
let classify (age, status) = {
    guard age.value >= 18 && status == Active else Excluded
    Included
}

behavior select : (people: List<Person>) -> List<Person>
let select (people) =
    List.filter(p -> p.age.value >= 18 && p.status == Active, people)
```

`souther examples --generate --boundaries`, one row written for each:

```
  classify                 implemented   rows 1    pending 0
    partition   axes 2   equivalence partitions 2/4
      · no row is in `age/0 <= x < 18` at age
      · no row is in `Inactive` at status
    combination pairs 1 reached / 1 known reachable, 3 untried
    border      borders 3   coverage items 2/8   excluded 3
      ! no row is at the ON point classify/age = 18 (guard@15:5)
      ! no row is at the OFF point classify/age = 17 (guard@15:5)
      · no row is at an OUT point of classify/status, Active (guard@15:5)
    branch      1/2

  select                   implemented   rows 1    pending 0
    partition   not measured (no partition axis was derived at any position)
      · not read: people (its values are held inside something this does not reach into)
    border      not measured (no line was derived at any position)
    branch      2/2
```

For `classify` the rows composed include the ones that falsify one conjunct while the other
holds:

```
//     | "age=18 <= x x status=Inactive" : (Age(18), Inactive) -> <?>
//     | (Age(17), Active) -> <?>
```

For `select` there is no axis, no border, and nothing to compose a row from. `branch 2/2` is
the arms of the inlined closure and says nothing about which values reached them.

## The threshold survives its own mutation

```souther
behavior countOverThreshold : (items: List<Item>) -> Total
    constructs Total
let countOverThreshold (items) =
    Total(List.length(List.filter(i -> i.charge.value >= 21000, items)))

example countOverThreshold
    | "one under, one over" : ([ Item { charge = Charge(1000) },
[...]

PR #941, opened 2026-08-21
Read a boundary rule off the comparison, not off the fork above it

### LINKED ISSUE #932: What makes a comparison a boundary rule of a behavior?
Split out of #920, which made which comparisons a body holds one answer and left which of them a line is drawn on where it was.

`BoundaryComparisons` is that rule, now written down: a comparison in the condition of a fork whose arms a run can be recorded in, reached through `&&` and `||`. It says what it always said. What #920 changed is that it is a policy over a catalog rather than whatever a traversal happened to reach, so changing it is one edit in one file.

## A meaning-preserving rewrite moves the partition

These two state the same rule and the second divides no position:

```
if a > 10 then ... else ...
```

```
let high = a > 10

if high then ... else ...
```

Whatever the answer is, it should not be that one of them divides `a` at 10 and the other divides nothing. Naming a value is not a statement about what a model divides.

Two more shapes stand behind those, and they are not obviously the same question:

```
let high = a > 10

high
```

```
xs.filter((x) -> x > 10)
```

The third answers the behavior's own truth, so a line on it is a line on an input the caller chooses. The fourth compares a thing no input path names, and runs once per element — `Plan.mayRepeat` already says so, which is what a set of places can and cannot be asked.

## The candidate

Not `directly appears under a fork` but `the comparison's truth participates in an observable decision`. How far that reaches — a `Bool` a behavior answers with, a predicate handed to a combinator — should be settled by measuring how many axes and classes each choice adds, not argued.

The conformance corpus staying green is not evidence here. It asks no question of this shape, which is why the limit went unnoticed until #908 read the paths of `&&` and `||` and found nothing to name them with.

## What #920 deliberately did not model

A boundary drawn on a comparison may eventually want evidence that the comparison was evaluated. That evidence is relative to a particular **use** of the comparison, not to the comparison's origin: under `A && (B || C)` the arms of the fork tell those three apart — the arm the whole condition holds on proves `A` ran and proves `B` ran, and proves nothing about `C`. One comparison given a name can be consumed by two forks and by none, so a comparison has as many answers as it has uses and a rule that took one from its origin would be answering which use the rule really belongs to.

#920 removed that relation from the model rather than moving it somewhere better. It was computed, carried on the origin, folded into a line's identity, and read by nothing — forcing it to one value left the reports, the conformance answers, the borders and the branch counts unchanged, and of the 428 distinct lines the suite draws none was told from another by it. Keeping the producer would mean this design starts from how to route what it already computes.

So if it comes back it comes back as `(comparison, consumer) -> evidence`, derived from the uses. And it may not come back at all: working out from the arms whether a comparison ran was what there was before a comparison had a site of its own. It has one now, everywhere it is written.

## What a report calls it

`OriginRef.constructThatDrewIt()` answers with the construct the comparison stands in, and `if` / `guard` / `comprehension` are the only words there are for it. Every rule that reaches a report today is written in one of those, so the answer is not yet wrong. Widening the policy is what makes it wrong, and the word for a comparison standing on its own is owed with the widening.

The machine-readable report calls this kind of rule `"guard"` (`AdequacyReport.ruleWord`). Same question, and it is a schema value documents already read.

kawasima:
Closes #932.

## What was wrong

The policy asked whether a comparison stands in the condition of a fork. The question it was there to answer is whether the behavior's answer turns on the comparison at all. Those agreed only while a fork was the one thing that could measure a comparison — reaching an arm was the evidence that it ran. #908 gave a comparison a site of its own and #920 moved the policy into one file, so the measurement left the fork and the policy stayed, because nothing else had ever been asked the other question.

Measured over six spellings of one rule, five of which mean the same thing:

| body | axes | notDerivable |
|---|---|---|
| `if temp.value < 240 then Cold else Hot` | `temp: [x < 240, 240 <= x]` | — |
| `let cold = temp.value < 240` then `if cold` | none | `at=temp, why=Absent` |
| `let cold = ...` then answering `cold` | none | `at=temp, why=Absent` |
| `temp.value < 240` as the answer | none | `at=temp, why=Absent` |
| `Verdict { cold = temp.value < 240 }` | none | `at=temp, why=Absent` |
| `let cold = ...` and never reading it | none | `at=temp, why=Absent` |

Only the first divides. The other four that state the same rule are reported as positions the model draws no line through, which is a sentence about the model that the model contradicts — and they are indistinguishable from the last one, which really does state nothing.

Nothing measured it. Widening the policy to every numbered comparison a run passes at most once and running the whole suite came back **6163 tests green**, the checked-in conformance report included. The corpus asks no question of this shape.

## The rule

A comparison is a boundary rule where its truth is on a live flow to what the behavior answers with, and where one run passes it at most once. The construct it is written in is not part of that.

Which values reach an answer is `LiveFlow`'s, in one pass: every other place a value stands in a body is consumed by what it stands in, so the one way to compute a truth and read none of it is to bind it to a name nothing reads. It over-reports — a chain of dead bindings still draws its line — and that direction is chosen and pinned by a test, because a line drawn for dead code is something an author can see in their own body while a rule dropped in silence is not.

## Three ways to bear no line, kept apart

`BoundaryComparisons` was a set of the comparisons that bear a line; `BoundaryPolicy` is an answer for every comparison of the body. `NotABoundary` says which:

- `NOTHING_READS_IT` — the behavior has no boundary here, and a report saying nothing is right.
- `NOTHING_RECORDS_IT` — no run through it can be recorded.
- `REPEATED_IN_ONE_RUN` — a boundary the behavior may well draw, that no row can be shown against, because a recording holds that a place was passed and not how many times.

The last two are not the first. Folded into one `false` they become the same silence, and the day the elements of a collection can be read (#912) whoever widens it has to work out again why anything was left out.

## One join point

The unread accounting and the reaching cuts each started by finding a `Core.If`, so a comparison written anywhere else was neither a line nor a rule this could not read — it went unmentioned. Both are read off the one policy now, and a comparison outside a condition is filed under what the forks above it established rather than under nothing. Without that, the search region for a named spelling stays as wide as the declarations while the fork spelling narrows, which is the same rewrite defect one level in.

`Condition` keeps the job only it can do: what stopping early proves, and what an arm proves.

## Schema 4

The human surface already says `comparison@…` (#931). The wire said `guard`, pinned by a `const` of schema 3. Widening the policy makes that word false, so it moves with the version. The schema's own rule asks for exactly this: a change that removes or renames anything raises the version, and only a word *added* to an enumerated field leaves it alone. The `construct` enum an arm is written with is untouched — an arm does belong to a construct.

## Verification

`mvn -o clean test` from the root: 9346 tests, all modules green. `mvn -o -Plint clean verify`: BUILD SUCCESS, no new warnings on the files this touches.

The checked-in conformance answers move by 7 lines: the version, three rule kinds, and three `notRead` entries reordered because the accounting now walks the body in source order rather than in the order forks were found. Nothing is added and nothing is dropped.

Every test here was watched fail first. Four are new: the equivalence of the spellings, the dead truth that divides nothing, the reasons told apart at the policy (no report can see that difference today), the comparison outside a fork that is still accounted for, and the two surfaces agreeing on the word.


kawasima:
Addressed in `867ba0be`. The blocker was real and reproduces exactly as described; here is what it measured before the fix, counting the cuts filed against each comparison:

| spelling | cuts on the second comparison |
|---|---|
| `p.x > 0 && p.y > 10` | 0 |
| `if p.x > 0 then p.y > 10 else false` | 1 (`p.x > 0`) |
| `let high = p.x > 0` then `if high then p.y > 10 else false` | 0 |
| `p.x > 0 && Int.max(p.y, 3) > 10` | 0, at all three sites |

## Root cause

The entry point being `Core.If` is the symptom. The cause is that `Condition` was doing two jobs and one of them was a walk: it says what a boolean subtree means, and `ReachingCuts.collect` folded over it to visit the comparisons inside one. A walk of a subtree needs a root, and the only root anything ever gave it was a fork's condition. Everywhere the re-read could not enter — a boolean that is not a fork's condition, a name standing for a truth, anything under a shape it has no words for — the evaluation context was lost, though `stating` could have answered.

That is the same fault #941 fixed one layer up. Comparison discovery used to be a walk rooted at a fork's condition too, and the fix was to make it the body walk carrying context. **What evaluating before a position established is a fact about the position, and it was modelled as a fact about a condition subtree.** A subtree needs a root; a body walk does not. So the boundary moved and the assumptions did not, which is your "path assumptions の ownership だけ fork に残った" said from the other end.

The named-condition hole is the same cause seen from the other side: `Condition` re-reads a subtree with its own idea of what is transparent, and a name is transparent to `InputReads` and opaque to `Condition`.

## What changed

The traversal moved onto the body walk, with the two operators as the only place a walk of the body is not a walk of what runs: `&&` walks its right under `assumed + stating(left, true)`, `||` under `stating(left, false)`, a fork's arms under what each proves of the condition, everything else under what stood at it. `ReachingCuts.collect` is gone; `ReachingCuts.stating` was already the rule and is unchanged. `Condition.of` follows a name to the truth it stands for, which terminates because a binder's value can only mention binders introduced before it.

One thing beyond the review, for the same reason the bug existed. `BoundaryPolicy.gather` and the assumptions walk were **both threading the `let` rule through the same tree** to answer different halves of one question about one position. Two walks carrying context is what produced this, so leaving two would leave the making of it. `ComparisonReadings` is the one walk — where a comparison is written, what its names point at, what a row had satisfied to get there, and its standing — and `BoundaryPolicy` is now the decision that walk asks at each comparison, with no walk of its own.

## Regression matrix

Both halves, as asked:

- selection — `A`, `let x = A; x`, `Verdict { cold = A }`, `if A then T else F` divide their input alike, and none is left in `notDerivable` or `notRead`
- reaching — `A && B` ≡ `if A then B else false`, `A || B` ≡ `if A then true else B`, `let a = A; if a then B else false` ≡ the same, and `A && f(B)` ≡ `if A then f(B) else false`

Held between spellings rather than against cut lists written down, so they cannot all fall silent together.

## Minor

`Condition`'s Javadoc pointed at `GuardThresholds.compared`, which `44f0c545` deleted. Rewritten to say what the type is not: not a way of finding comparisons, and why that mattered.

## Verification

`mvn -o clean test` from the root: 9350 tests, all modules green. `mvn -o -Plint clean verify`: BUILD SUCCESS, no warnings on the new files. The checked-in conformance answers do not move for this commit — a narrower search region does not change what is owed.

Issue #711, opened 2026-08-13
Null analysis is enforced by an editor extension, not by the build, and that extension writes where Maven reads staleness


kawasima:
`souther-runtime` depends on JSpecify 1.0.0 and 26 files carry `@NullMarked`, `@Nullable` or
`@NonNull`. Nothing in the build reads them. The analysis that does runs inside the VS Code Java
extension: `.vscode/settings.json` sets `java.compile.nullAnalysis.mode` to `automatic`, which turns
on JDT's null analysis and points it at the JSpecify annotation names, and `.vscode/jdt-nullness.prefs`
tunes which of its problems are reported.

So the rule binds whoever has the extension installed. A checkout without it, and every CI run, gets
no null analysis at all. A violation reaches `main` unless a human happened to be looking at the
right editor tab.

## The extension also owns `target/classes`

The extension imports the modules through m2e, and m2e sets an Eclipse project's output folder to the
one in the pom. ECJ therefore writes its class files into `souther-compiler/target/classes` — the same
directory `maven-compiler-plugin` writes to and reads staleness from.

Two observations, both from this repository:

- A class whose source did not compile was present in `target/classes` with a body of
  `throw new Error("Unresolved compilation problem: Type mismatch: cannot convert from String to int")`.
  javac does not emit that; ECJ does, under `proceedOnError`.
- Editing a source file updated the corresponding `.class` one second later with no Maven process
  running.

Because the `.class` is newer than the `.java`, `maven-compiler-plugin` 3.15.0 finds nothing stale and
skips javac entirely: `mvn -o -pl souther-compiler compile` returns BUILD SUCCESS in 1.6 seconds
without compiling. What surefire then loads is whatever ECJ wrote, including the classes whose bodies
throw. This is the mechanism behind the recurring "the incremental build did not pick up the edit"
and the `Unresolved compilation problem` test failures; `clean` worked because it deleted the
extension's output, not because Maven was at fault. ECJ also runs on the extension's JRE 21 while the
pom asks for `release 25`.

## Moving the analysis into the build

Error Prone 2.50.0 with NullAway 0.13.8, configured on `maven-compiler-plugin`, runs on this
toolchain — JDK 26 running the build, `maven.compiler.release=25` — and fails the build when asked to.
It needs a `.mvn/jvm.config` exporting the `jdk.compiler` internals; without it the compile stops with
`IllegalAccessError: module jdk.compiler does not export com.sun.tools.javac.api`.

[...]

PR #946, opened 2026-08-22
Take Error Prone out of the build an edit runs

### LINKED ISSUE #944: One edit to souther-compiler recompiles 1421 sources, and what the loop pays for that is not known
Editing one file in souther-compiler and running the named test back costs 20 seconds, and
14 of those are javac. The module's staleness is decided for the module as a whole, so a
one-line change recompiles all 516 main sources and all 905 test sources, and the four
modules downstream recompile after it.

Measured on develop at `ca2de15c`, from the reactor root, warm:

| what changed | wall |
|---|---|
| nothing | 1.7s |
| one main source in souther-compiler | 14.2s |
| one test source in souther-compiler | 8.1s |
| one main source, then `-Dtest=<Class>` | 20.5s |
| the whole suite, `mvn -o test` | 79s |

The obvious move is to split souther-compiler's 905 test sources into modules of their own,
so a test-only edit recompiles a few dozen sources instead of all of them. Before designing
that, two things have to be measured, because either answer changes what is worth doing.

## Does splitting help the edit that actually dominates?

A main edit already recompiles every downstream module — the build says `Recompiling the
module because of changed dependency` for souther-build-driver, souther-lsp and souther-cli.
Test modules would be downstream of souther-compiler, so a main edit would recompile them
too and the 14.2s would not move. Splitting would only help the 8.1s test-only edit.

Measure it before assuming it: put souther-compiler's tests behind a second module in a
throwaway branch, touch one main source, and read what recompiles.

## What is javac spending the 14 seconds on?

`-Xplugin:ErrorProne ... -Xep:NullAway:ERROR` is on the compiler plugin unconditionally — the
`lint` profile only widens `errorprone.checks`, it does not turn the plugin on. So NullAway
analyses every compilation of every build, including the loop. If that is a large share of
the 14.2s, there is a cheaper lever here than restructuring modules.

Measure: compile souther-compiler with the `-Xplugin` argument removed and compare; and count
how many of the 516 sources sit in a `@NullMarked` package, since `OnlyNullMarked=true` means
the rest are walked and skipped.

## Not in scope

The other half of the loop cost was the commit guard being run per iteration, and that is
settled: `clean`, `-pl`, `CI=1` and `-Plint` are out of the loop, recorded in `CLAUDE.md`.
This issue is only about the 8-14s that remains after that.

kawasima:
Error Prone was on the compiler plugin itself, so every compile of every module ran it,
including the one an edit runs. The `lint` profile only widened which checks were enabled — it
never turned the plugin on. This moves `-Xplugin:ErrorProne`, `-XDcompilePolicy=simple`,
`--should-stop` and the annotation processor paths into that profile. A development build is
javac and nothing else.

On the build one edited source in souther-compiler costs, from the reactor root, three runs
each: 13.39 / 13.76 / 13.48 seconds with the plugin, 11.11 / 10.39 / 11.17 without.

## The nullness gate was already looking at nothing here

`OnlyNullMarked` decides where NullAway looks, and `@NullMarked` is written in two files in
this repository, both `package-info.java` under souther-runtime. Across souther-compiler's 516
sources it was analysing none of them. Moving it does not weaken a rule this module was being
held to.

## Held both ways

A `@NullMarked` package returning null from a non-null return type, and a `String.trim()` whose
value is dropped:

| | default build | `CI=1 mvn -o -Plint clean test-compile` |
|---|---|---|
| the null return | BUILD SUCCESS | `[NullAway] returning @Nullable expression from method with @NonNull return type` |
| the dropped value | BUILD SUCCESS | `[ReturnValueIgnored] Return value of 'trim' must be used` |

`combine.children` on the profile's `compilerArgs` is load-bearing: a profile's list of
arguments replaces the list the build declares rather than adding to it, so without it a lint
run loses `-Xmaxerrs` and `-Xmaxwarns` and stops counting at a hundred. Both effective poms
were read.

`mvn -o test` from the root is green, and `CI=1 mvn -o -Plint clean verify` is green on this
branch with sixteen compilations actually run.

## The other half of #944

The issue also asked whether splitting souther-compiler's 905 test sources into modules of
their own would cut the loop. Measured and dropped. Every downstream module already recompiles
in full when a souther-compiler main source changes, so a carved-out test module would too and
the main-edit path would not move. Counting 21336 builds in the local transcripts by what was
edited since the previous build: 59.3% had no source edit at all, 19.3% main only, 17.5% test
only, 3.8% both — which, weighted by what each costs, is about a quarter of the time spent
building. So the reason to leave it is not that the effect is small. It is that the effect is
confined to test-only edits and the change is a restructuring, while what this PR does reaches
every build and is a pom.

Closes #944.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -58,6 +58,21 @@
         <build.test.classes>${project.basedir}/target/test-classes</build.test.classes>
         <errorprone.version>2.50.0</errorprone.version>
         <nullaway.version>0.13.8</nullaway.version>
+        <!-- What NullAway is asked for, written once. The gate below and the `lint` profile both
+             ask, and a gate configured differently in each is a gate whose answer depends on which
+             build asked. -->
+        <nullaway.args>-Xep:NullAway:ERROR -XepOpt:NullAway:OnlyNullMarked=true -XepOpt:NullAway:JSpecifyMode=true</nullaway.args>
+        <!-- The nullness gate an ordinary build runs, carried by the modules that declare
+             `@NullMarked` — souther-runtime, today. Nothing else in Error Prone: this is the one
+             rule those modules state about themselves, and the other five hundred checks are the
+             `lint` profile's.
+
+             It is a property because the profile empties it. javac takes one `-Xplugin:ErrorProne`
+             per compilation and reports the second as a plugin it cannot find, so under `lint`,
+             which puts the plugin on every module, a module carrying its own would be handing javac
+             two. Emptying this leaves the profile's the only one, and NullAway is in both spellings,
+             so the gate does not lift when the profile takes over. -->
+        <errorprone.gate>-Xplugin:ErrorProne -XepDisableAllChecks ${nullaway.args}</errorprone.gate>
         <!-- The build protocol souther-build-driver implements. Not this project's version and not
              tracked to it: it moves when the interface between a build plugin and a driver moves,
              which is why it is released from souther-lang/souther-build-api and named here like any
@@ -267,6 +282,11 @@
                     <name>env.CI</name>
                 </property>
             </activation>
+            <properties>
+                <!-- The gate a module carries, emptied: this profile's plugin argument below is the
+                     only one javac may be given. -->
+                <errorprone.gate/>
+            </properties>
             <build>
                 <plugins>
                     <plugin>
@@ -280,7 +300,7 @@
                                      as a thing that confuses a reader, and this compiler's tests
                                      call methods the language named in Japanese: the check would be
                                      asking the tests to stop covering what Souther supports. -->
-                                <arg>-Xplugin:ErrorProne -Xep:UnicodeInCode:OFF -Xep:NullAway:ERROR -XepOpt:NullAway:OnlyNullMarked=true -XepOpt:NullAway:JSpecifyMode=true</arg>
+                                <arg>-Xplugin:ErrorProne -Xep:UnicodeInCode:OFF ${nullaway.args}</arg>
                             </compilerArgs>
                             <annotationProcessorPaths>
                                 <path>
--- a/pom.xml
+++ b/pom.xml
@@ -58,13 +58,6 @@
         <build.test.classes>${project.basedir}/target/test-classes</build.test.classes>
         <errorprone.version>2.50.0</errorprone.version>
         <nullaway.version>0.13.8</nullaway.version>
-        <!-- Which Error Prone checks a build runs. Measured on `clean test-compile`, whole
-             reactor, one run each: 8.7s with no Error Prone at all, 12.0s with NullAway alone,
-             38.0s with the default checks as well. The nullness gate is the rule this repository
-             holds itself to, and it costs three seconds; the other five hundred checks cost
-             twenty-six, which is too much to spend on every compile while writing code. So a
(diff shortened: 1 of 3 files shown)


kawasima:
Right on the blocker, and the fix is in `cdd2fe48`. The root cause is one step past where the
review put it, and it rules out the proposed fix as stated too.

## What I got wrong

The pom said the null analysis runs in the build because "a rule only some checkouts enforce is
not a rule". I read that as the rationale for a cost and not as a statement of an invariant,
measured that souther-compiler is not covered, and generalised from "this module is not covered"
to "the default build does not need the gate". #711's first Done-when is exactly the thing that
generalisation breaks.

## Why scoping to souther-runtime is not the whole fix

#711's third Done-when is:

> Adding `@NullMarked` to a package is the only thing needed to bring it under the check.

Put the gate in `souther-runtime/pom.xml` and stop there, and a `@NullMarked` added to a
souther-compiler package is checked by nothing — the module does not run the gate. That is the
same defect as mine with the sides swapped.

The root cause under both: **which packages are annotated and which modules configure the gate
are written by hand in two different files, and nothing compares them.** They agree today
because there is one of each. My change moved one set; the proposed fix moves the other.

## What is in the branch now

The gate is in souther-runtime's build, as the review asked, and
`TheNullnessGateRunsWhereItIsDeclaredTest` in souther-bench compares the two sets so that neither
can move alone. The annotated side is read off class files — souther-compiler's javadoc discusses
`@NullMarked` and its generator emits it, so the sources say the word where nothing is declared —
and the configured side off `/project/build`, never a profile, because the question is what an
ordinary `mvn test` runs. Equality in both directions: a module short of the gate states a rule
nobody checks, a module short of the annotation pays Error Prone's startup to check nothing.

One thing the review could not have predicted, found by running it: javac takes one
`-Xplugin:ErrorProne` per compilation and reports a second as a plugin it cannot find, so under
`lint` — which puts the plugin on every module — souther-runtime was handed two and the build
stopped at module 2. The gate's argument is now a property the profile empties, and NullAway is
in both spellings, so the gate does not lift when the profile takes over.

Controls, each with its opposite:

| probe | ordinary build | `-Plint` |
|---|---|---|
| `@Nullable` returned from a `@NonNull` method in souther-runtime | `[NullAway]`, FAILURE | `[NullAway]`, FAILURE |
| `String.trim()` dropped, in souther-runtime and in souther-compiler | SUCCESS | `[ReturnValueIgnored]`, FAILURE, either module |
| gate block removed from souther-runtime | the comparison fails | |
| gate block moved to a module that declares nothing | the comparison fails, both directions | |

`mvn -o test` and `CI=1 mvn -o -Plint clean verify` are green, the latter with sixteen
compilations actually run.

## On the loop cost, which survives

souther-runtime is upstream of souther-compiler, so it is not recompiled when a compiler source
changes — the build log for that edit lists compiler, build-driver, lsp, cli and bench and not
runtime. The gate therefore costs the loop nothing, and this is structural rather than a timing:
I would not report stopwatch numbers taken this afternoon anyway, since the machine was running
at load average 43 and the same configuration measured 10.7s and 51s an hour apart.

## On the ROI wording

Taking the point. 17.5% of builds carrying 25.6% of build time is the honest weighting, and the
conclusion it supports is that the split is limited to test-only edits and costs a
restructuring — not that its effect is small. I will fix the PR body.

Issue #476, opened 2026-08-08
The formatter's document does not preserve token adjacency, so a spacing rule has nowhere to run


kawasima:
Split out of #444, which needed the canonical spacing rule stated before it could compute a deviation
witness. Stating it turned up why that was hard. The first reading of this issue was that the
formatter has no function deciding what goes between two tokens; reading the formatter says the
missing function is a consequence rather than the cause.

## What is there now

The only leaf of `Doc` that carries characters is `Text(String)`, and the renderer concatenates those
strings (`Doc.java:29`, `Doc.java:93`). Nothing in a document says where one token ends and the next
begins, or which construct joins the two. Both facts are in the tree the formatter reads, and both
are gone by the time the document exists — which is before anything could apply a spacing rule to
them.

So the gap between two tokens is not a place in the program. It exists in two forms only: inside a
literal such as `text(" = ")`, where one construct spelled both sides and the gap at once, and in the
rendered text, where it can be recovered only by parsing the output again. That second one is what
`WhatGoesBetweenTwoTokensOnALineTest` does.

What is dispersed is not spacing but every decision about a token boundary, and it takes five forms:

- a spaced join spelled into a literal — 66 of them across 26 methods, `text(" = ")`, `text(": ")`,
  `text(" -> ")`;
- a tight join spelled by joining tokens into one string — `qualifiedName`, `stage`, `nameList`,
  `identPath`, and `matchCase`, which builds a whole match arm's pattern as a single `Text`;
- a breakable join whose flat form is a space — `Doc.LINE`, in `separated` and as `delimited`'s
  boundary;
- a breakable join whose flat form is empty — `Doc.SOFTLINE`, the other value of that same argument,
  chosen by the construct at 19 call sites;
- an explicitly computed join — `Formatter.spaceBetween(left, right)`, which exists in one place.

That last one is the evidence for the cause. It exists where the grammar hands the formatter a run of
tokens rather than child nodes: `CstParser.matchCase` bumps every token of an arm's pattern into
`MATCH_CASE`. That is the one place a token adjacency is exposed as an input to the formatter, and it
is the one place a function of this shape was written. Everywhere else the formatter holds child
documents, the adjacency is not exposed, and there is nowhere to write one.

## What that cost

Four defects turned up in #444's sweep, none of which any test observed. Three of them are one
defect: a match arm wrote `Some ( x )` where the `let` opening the same value writes `Some(x)`; a
[...]

PR #947, opened 2026-08-22
Write the rows of a table at one column

### LINKED ISSUE #938: The formatter takes the columns out of an example table, so a decision table cannot be formatted
An `example` table is a decision table written down, and its rows are read against each other:
which input differs between two rows is the whole point of writing them side by side. The
formatter writes one space wherever a separator goes, so a table whose columns an author lined
up comes back ragged.

## What it produces

`beer.sou` from `souther-test`, the four columns of a decision table:

```souther
example 一杯目の価格を求める
    | "通常時間・クーポンなし"       : (通常時間, クーポンなし)       -> 価格(490)
    | "ハッピーアワー・クーポンなし" : (ハッピーアワー, クーポンなし) -> 価格(290)
    | "通常時間・クーポンあり"       : (通常時間, クーポンあり)       -> 価格(100)
    | "ハッピーアワー・クーポンあり" : (ハッピーアワー, クーポンあり) -> 価格(100)
```

`Formatter.format` on the file:

```diff
-    | "通常時間・クーポンなし"       : (通常時間, クーポンなし)       -> 価格(490)
+    | "通常時間・クーポンなし" : (通常時間, クーポンなし) -> 価格(490)
     | "ハッピーアワー・クーポンなし" : (ハッピーアワー, クーポンなし) -> 価格(290)
-    | "通常時間・クーポンあり"       : (通常時間, クーポンあり)       -> 価格(100)
+    | "通常時間・クーポンあり" : (通常時間, クーポンあり) -> 価格(100)
     | "ハッピーアワー・クーポンあり" : (ハッピーアワー, クーポンあり) -> 価格(100)
```

The rows that were already at one space — the longest ones — are left alone, and the rows the
author padded to match them are pulled back. The same happens to the six rows of
`example 時間帯を求める` in the same file.

The practical effect is that `souther fmt` cannot be run on such a file at all: adding one row
to a table means a diff over every row that was aligned, so the formatter goes unused rather
than being allowed to reflow the table. `beer.sou` and `mobile.sou` were both left unformatted
for that reason.

## Where it is

`Formatter` spells its own separators: there are roughly two hundred `text(" ")` calls and no
function that decides what goes between two tokens, so there is nowhere a rule about a column
could be written. `Doc` lays out a group by whether it fits `Formatter.WIDTH`, which is a
decision about breaking rather than about padding, and nothing in `souther-fmt` computes a
width across sibling lines.

Alignment across lines is also the one thing a formatter cannot do one node at a time: the
padding of a row depends on every other row of the table, so whatever answers this reads a
whole `example` block before writing any of it. Display width, not character count, is what
has to be measured — the table above is Japanese, and the rows that line up are the ones whose
names are the same number of columns wide.

## What it takes

Two answers, and they are opposite:

- **Keep what the author wrote.** The formatter would leave the runs of spaces before `:` and
  `->` inside an `example` row alone. Cheapest, and it makes the formatter stop being total
  over its input — a canonical form that has an exception is one every later rule has to ask
  about.
- **Align the table itself.** The formatter would pad every row of a block to the widest, so
  the columns line up whether or not the author did it. Then adding a row reflows the block,
  which is a diff an author expects from a formatter, and there is one canonical form again.

The second fits what the formatter is for; it is also the one that needs a width measured
across rows, which nothing in `souther-fmt` does today.

Aside, from the same run: `fmt` also removes the blank line between a leading comment block and
the first `data` of the file. Whether that is intended is a separate question and is not what
this asks about.

Measured on `develop` at `6cf29597`.

kawasima:
Closes #938 — the formatter took the columns out of an `example` table, so a decision table could not be formatted at all.

## What was wrong

Not what the issue says. The `text(" ")` calls it describes are gone; #476 moved every separator into `Spacing.between`, and the fix is not to undo that.

`Spacing.between` reads two token kinds and the construct joining them and answers with nothing or one space. A column is a fact about the sibling rows of a block, and neither the domain nor the range of that function can hold one. Underneath it, `Doc.layout` walks the text once from left to right, and there is no point of that walk at which a width across lines is known. Losing the author's padding was the boundary rule being total and local, which is the right design.

## The rule

**The rows of a table write each of their connectors at one column.** A table is an `example` or a `fake`; a `match` is not, since its arms are read from the top down rather than against each other.

`Columns` holds the vocabulary and nothing else — which constructs are tables, and which boundaries are stops — read from the same pair `Spacing` is asked about, so no construct can make itself a table by writing one. A stop is keyed by its connector rather than by how many stops precede it, so a row with no description still writes its `->` at everybody else's column. A table is a `TableRef`, an identity like `GroupRef` and `NestRef`, so two `example`s cannot run their widths into each other.

A stop is written **after** the separator and never instead of it. `Spacing` is unchanged and still has no third answer; what a column says is where the connector begins.

## How the column is decided once

`Doc.layout` lays the document out, reads the columns off what it wrote, and lays it out again to write them, until nothing moves. Looking ahead over the table at the row that opens it would be a second walk taking the decisions the real one has to take again — the thing `flatnessOf` is careful about — and the same machine run twice cannot disagree with itself. It settles because a column depends only on the columns before it on its line. A document with no table is laid out once.

Two consequences are stated rather than left to the mechanism:

- **Padding is not measured against the width.** A group whose flatness read it would be decided by a column that is decided by which rows came out flat. `Trailing` is already handled the same way.
- **A row written down the page takes no part in the columns.** Its connector opens a line, where a column is not something a token can be at.

## `fmt --check`

`Witness.AtAColumn` has the shape of `Witness.Indentation`: the unit is the column, the canonical answer is one number, and what the source wrote is a list. The rule is evaluated once for the whole column, so a table whose five rows all miss it has departed from one decision five times.

```
4:20: the rows of a table of examples write their : at one column — column 37, and this is columns [31, 37]
4:37: the rows of a table of examples write their -> at one column — column 70, and this is columns [58, 70]
```

What the canonical form writes at a stop is the separator and then the padding, and which of the two rules answers is settled by whether the source's run comes apart the same way: that separator, and after it spaces and nothing else. So the spacing rule is not asked at a stop whose separator the source already has, and it is the only one asked at a stop where the source wrote a tab, or no space at all. One predicate, read by both families and by the repair, so they can neither both answer about the same characters — a conflict `Repair.composed` refuses — nor both stay silent, which would leave a difference no rule accounts for.

## What the tests measure

`ARowOfATableIsWrittenAtItsTablesColumnTest`, eight of them, all against the layout's own answers and display columns. None quotes a stretch of the output: a table is the one construct where adding a row moves every other row, so an expectation written as text would have to be rewritten for a change it is not about — which is the complaint this issue makes about the formatter itself.

- every row written to a column is at it, over every source in the repository
- the column is the greatest the rows need, and some row is at it unpadded
- a table settles its own columns and nobody else's — measured as a pair, not as fixed numbers
- a row written down the page changes no column — the same table with and without it
- padding is not measured against the width — lines over 100 columns and no group broken for it
- the arms of a `match` form no column
- a table that is not aligned is reported as this rule, repairs to the canonical form, and the spacing rule says nothing there
- every stop the rule lists is one some source writes

`WhatGoesBetweenTwoTokensOnALineTest` used to state that no run of spaces survives between two tokens. It now splits the run at the stop, so the separator is still nothing-or-one-space and the padding is the other rule's.

## Measured

- All nine modules green.
- The repair path is live rather than green on canonical sources: five `.sou` files in the repository hold twelve columns that are not aligned, and every one reaches `whole == true`.
- Whole test run: 1:07 on `develop`, 1:05 here.

## Left out

The five `.sou` files above are not in canonical form and `souther fmt` would now reflow their tables. Aligning only their columns is not well defined — the column rule reads the break rules' results, and those files break rows the canonical form keeps whole, so a column-only repair pads a continuation line out to a column nothing reaches. Running the whole formatter on them brings 122–358 departures that have nothing to do with this issue. Both are separable from this change.

### DIFF
--- a/souther-compiler/src/main/resources/META-INF/souther-docs/cli/commands.md
+++ b/souther-compiler/src/main/resources/META-INF/souther-docs/cli/commands.md
@@ -332,6 +332,10 @@ column, and the two answers are columns:
 - the rows of a table of fakes write their -> at one column — a `fake` has the one column, since its
   rows have no description.
 
+The column a row is said to be at is where the padding it wrote carries it, with the rest of the
+line as the canonical form has it. So a table lined up correctly and indented one column too deep is
+told about its indentation and nothing else, rather than about every row.
+
 What separates one thing from another, and what ends:
 
 - a blank line stands where the author wrote one, and under a header — how many, as lines.
--- a/souther-compiler/src/main/resources/META-INF/souther-docs/cli/commands.md
+++ b/souther-compiler/src/main/resources/META-INF/souther-docs/cli/commands.md
@@ -244,6 +244,22 @@ departure to break off: the `else` ends the guard's own line and the clauses go 
         | unique -> DuplicateProduct
 ```
 
+An `example` and a `fake` are decision tables, and their rows are read against each other: which
+input differs between two rows is what writing them one under the other is for. So the connectors of
+a table's rows are written at one column — the `:` and the `->` of an `example`, the `->` of a
+`fake` — with the shorter rows padded out to the widest. Adding a row rewrites the table where the
+new one is the widest, which is the diff a table asks for. Whatever the author lined up by hand is
+derived again like everything else, and a row too long for one line is written down the page and
+takes no part in the columns. The padding is measured in columns on the screen, so a table whose
+descriptions are Japanese lines up on a screen rather than on a character count.
+
+```
+example priceOfTheFirstGlass
+    | "off peak, no coupon" : (OffPeak, NoCoupon)     -> Price(490)
+    | "happy hour"          : (HappyHour, NoCoupon)   -> Price(290)
+    | "with a coupon"       : (OffPeak, WithCoupon)   -> Price(100)
+```
+
 A comment keeps what it was written about. On the line of the code it follows it stays there; on a
 line of its own it goes above what follows it, unless a blank line separates it from that and none
 separates it from the code above, in which case it stays under that code. A comment with nothing
@@ -301,6 +317,15 @@ How far in a line begins. The two answers are columns:
 - one level deeper is one indent further in — a step, and not a column the rule never states.
 - a line the file holds begins at column zero — the outermost level, which has no level outside it
   to be measured from.
+
+Where a table's rows line up. An `example` and a `fake` write the connectors of every row at one
+column, and the two answers are columns:
+
+- the rows of a table of examples write their : at one column — the description was padded out to
+  the widest one in the table, or was not.
+- the rows of a table of examples write their -> at one column — the same, for the input.
+- the rows of a table of fakes write their -> at one column — a `fake` has the one column, since its
+  rows have no description.
 
 What separates one thing from another, and what ends:
 
--- a/souther-compiler/src/main/resources/META-INF/souther-docs/cli/commands.md
+++ b/souther-compiler/src/main/resources/META-INF/souther-docs/cli/commands.md
@@ -289,7 +289,8 @@ What stands between two tokens the same line holds:
 - what goes between two tokens on a line — one space, or none. The two answers quote the characters,
   so a source that wrote a line break where neither text ends a line reads it back as `\n`. At a
   table's column this answers about the separator and the column rule about the padding after it,
-  which is why a tab written there is quoted here rather than counted as a column.
+  so what is written there is this rule's until the separator is what the canonical form has —
+  a tab, or no space at all, is quoted here rather than reported as a column the row missed.
 
 Whether a construct is written down the page at all. The two answers are `on one line` and `down the
 page`, and which rule is named is which one decided the form:
--- a/souther-compiler/src/main/resources/META-INF/souther-docs/cli/commands.md
+++ b/souther-compiler/src/main/resources/META-INF/souther-docs/cli/commands.md
@@ -251,7 +251,9 @@ a table's rows are written at one column — the `:` and the `->` of an `example
 new one is the widest, which is the diff a table asks for. Whatever the author lined up by hand is
 derived again like everything else, and a row too long for one line is written down the page and
 takes no part in the columns. The padding is measured in columns on the screen, so a table whose
-descriptions are Japanese lines up on a screen rather than on a character count.
+descriptions are Japanese lines up on a screen rather than on a character count, and it is written
+as spaces — a tab reaches a column too, and what a source wrote there is the rule about what goes
+between two tokens' to say rather than the column's.
 
 ```
 example priceOfTheFirstGlass
@@ -285,7 +287,9 @@ rest of the rules are about:
 What stands between two tokens the same line holds:
 
 - what goes between two tokens on a line — one space, or none. The two answers quote the characters,
-  so a source that wrote a line break where neither text ends a line reads it back as `\n`.
+  so a source that wrote a line break where neither text ends a line reads it back as `\n`. At a
+  table's column this answers about the separator and the column rule about the padding after it,
+  which is why a tab written there is quoted here rather than counted as a column.
 
 Whether a construct is written down the page at all. The two answers are `on one line` and `down the
 page`, and which rule is named is which one decided the form:
--- a/souther-compiler/src/test/java/souther/compiler/WhatIsWrittenInAnEnsuresIsQuotedOverTheRowsTest.java
+++ b/souther-compiler/src/test/java/souther/compiler/WhatIsWrittenInAnEnsuresIsQuotedOverTheRowsTest.java
@@ -76,8 +76,8 @@ class WhatIsWrittenInAnEnsuresIsQuotedOverTheRowsTest {
                 //
                 // example findTodo
                 //     | "id=0 <= x <= 0" : (TodoId(0)) -> <?>
-                //     | "id=0 < x" : (TodoId(1)) -> <?>
-                //     | (TodoId(2)) -> <?>
+                //     | "id=0 < x"       : (TodoId(1)) -> <?>
+                //     | (TodoId(2))                    -> <?>
                 //
                 // example echo
                 //     | (TodoId(0)) -> <?>
--- a/souther-fmt/src/main/java/souther/compiler/fmt/ColumnDecision.java
+++ b/souther-fmt/src/main/java/souther/compiler/fmt/ColumnDecision.java
@@ -1,0 +1,15 @@
+package souther.compiler.fmt;
+
+/**
+ * Where one column of one table is, on the screen.
+ *
+ * <p>Taken once for the whole column, as a group's is taken once for the whole group. Which column
+ * a table's {@code ->} stands at is one answer that every row of it is written to, so a decision per
+ * row would be counting the rows and calling the count decisions.
+ *
+ * <p>A display column and not an index into the text. What lines up is what a reader sees, and the
+ * rows of a table whose descriptions are Japanese line up where the columns agree and not where the
+ * characters do.
+ */
+record ColumnDecision(Columns.Unit unit, int column) {
+}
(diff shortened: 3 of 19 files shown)


kawasima:
Both blockers are real. I reproduced each before touching anything, and each new test goes red when its fix is reverted.

**Blocker 1**, reproduced: `whole=false` with **zero deviations reported** — the report says the file has nothing wrong with it and the completeness contract is false at the same time.

**Blocker 2**, reproduced with the tab in the row that *is* padded (my first attempt put it in the unpadded row and came out right, which is why the corpus never finds this):

```
    | "long description" : ("abc")    -> X
    | "short"            : ("<TAB>") -> Y
```

Both arrows are reported at `naturalColumn` 38; on a screen they are at 38 and 35. `whole=false`.

## Root causes

**Blocker 2 is the one that came first.** Not the tab. I represented the written column as a correction on the measured column — `displayCol + padded` — and a difference can express a column only while a token's width does not depend on the column it starts at. `flatnessOf`'s own javadoc, a few lines above, already says a fit is walked forward from the absolute column "because a tab advances to the next stop, which is decided by the absolute column it stands at and cannot be recovered from how much of the width remains". I reintroduced exactly that arithmetic one method below the sentence that rules it out.

Fixed as you describe: `measuredCol` and `writtenCol` are two counts, each advanced from its own value, and `ColumnStop` moves only the second.

**Blocker 1** is not really about tabs either. The separator/padding split was made on the canonical form's side and not on the source's, so one boundary was read in two vocabularies: canonical as "separator is spacing's, padding is the column's", source as "the whole run is the column's, and the answer is a number". A number the source can reach by other characters.

The split is now made on both sides. One predicate — the source writes spaces at this stop, or it does not — is read by both families and by the repair, so they are exclusive by construction: neither both answering about the same characters, which the repair's composition refuses, nor both silent, which is what left the empty report. A source whose separator is not settled yet is the spacing rule's that round and the column's the next, which is the model the rounds already run on.

```
4:16 the rows of a table of examples write their : at one column | column 16 | columns [16, 10]
4:22 the rows of a table of examples write their -> at one column | column 22 | columns [22, 16]
5:10 what goes between two tokens on a line | ` ` | `\t`
```

`whole=true`.

## Minors

Both taken.

The `contains` assertion is gone; what it was there for is that a row was not dropped, which is the two texts' code tokens. You are right that it was self-contradicting next to the class's own sentence, and right that reading the text to compute a display column is a different thing.

The corpus claim was overstated. It now reads the repository's `.sou` files as well as the formatter's rule corpus, and the javadoc says which is which and why both are there — a repository of well-written sources writes no tab and no row too long for a line, so the fixtures are not decoration.

The two fixtures are in the general properties too, so `everyRowWrittenToAColumnIsAtIt` catches Blocker 2 as well as the named test does.

## Verified red

- Blocker 2's fix reverted (`writtenCol` reconstructed as `measuredCol` plus the padding so far): `aTabWrittenAfterAColumnIsMeasuredFromWhereItStands` — `the two rows write their arrow at [38, 35]` — and `everyRowWrittenToAColumnIsAtIt`.
- Blocker 1's fix reverted (spacing skipping the whole adjacency): `aSourceThatReachesTheColumnByOtherCharactersIsToldWhatItWrote` — `the rule that quotes characters says what was written: []`.

All nine modules green.

kawasima:
Reproduced, then fixed. `fake f | (A)-> X` was reported as

```
4:10 the rows of a table of fakes write their -> at one column | column 10 | column 9
```

A table of one row, which is at its own width and has no rows to disagree with.

## Root cause

Not the empty run. `padsWithSpaces` asked whether what the source wrote there is spaces, which is a question about **what the padding may hold**, and it says nothing about the separator being there at all. The production side was made to say `separator` → `ColumnStop` → `connector`; the predicate tested the alphabet of the second half and never tested that the run comes apart at all. An empty run satisfies the half that was written and vacuously satisfies the half that was not.

Fixed as you propose — `padsAfterTheSeparator(source, before, after, separator)`, the separator taken from `CanonicalStop`, which already carried it. No new model, and it generalises to a stop whose separator is nothing without a second reading.

```
4:10 what goes between two tokens on a line | ` ` | nothing
```

`whole == true`, and the column rule has nothing to say about that run.

Held by `aRowThatWroteNoSeparatorIsToldThatAndNotAboutItsColumn`, which reads the report's rules, `whole()`, and that `Witnesses.columns` is empty — no text assertion. Reverting the predicate to the old one turns it red with exactly the old attribution. The fixture is in `sources()` too, so the general properties carry it.

The `fmt --check` section of the body is corrected.

## One thing I did not change, and want to name

`| "a": (A) -> X` still reports two rules: the `:` boundary as spacing, and the `->` as a column. The `->` stop's own separator is right; its column is one to the left only because the missing space at `:` shifted the line. A column's answer depends on everything to the left of it on the line, which is other rules' business — so this is the round machinery doing what it documents, and it goes away once the `:` is repaired. `whole == true` and it converges.

It is the same *shape* as what you found, and it differs in one way that I think settles it: the statement is true of the text as written, whereas the one-row table's was manufactured. Making the column rule silent while anything earlier on its line is unsettled would be a new and much wider rule, and no other family does it — `Indentation` reports about lines whose content is still wrong. Say the word if you want it looked at; I have left it alone on your "次 round まで厳密に固定する必要もない".

All nine modules green.

kawasima:
Reproduced both, then fixed. The second one is the sharper of the two — a table the author lined up correctly, indented one column too deep:

```
4:1  one level deeper is one indent further in | 4 columns | 5 columns
4:14 the rows of a table of examples write their : at one column | column 13 | column 14
4:20 the rows of a table of examples write their -> at one column | column 19 | column 20
```

Nothing about its columns is wrong, and the two column lines go away when the indentation is repaired.

## Root cause

The source's answer was the display column the connector physically occupies. An absolute column is the sum of everything to the left of it on the line — the indent, every separator, every column already written — so it is not a fact about the padding this rule owns. Every other family answers about the characters of its own unit.

`Witness.Indentation` met exactly this and its javadoc already contains the answer:

> Both are differences and neither is a column. A source that indents a whole construct by two writes every line inside it in the wrong column, and what it got wrong is one step.

I had copied that record's shape — unit, one canonical answer, a list for the source — without its content. So the repository had already decided A against B, and it decided B.

## The fix, and why it keeps the report speaking in columns

**B**, but the witness's shape is unchanged and so is its unit. What is read from the source is the padding; what is *said* is the column that padding comes to with the rest of the line as the canonical form has it — `occurrence.naturalColumn() + spaces the source wrote`. The padding is the quantity this rule owns; the column is the vocabulary a table is read in. Where nothing else on the line is wrong the two numbers are identical, which is every source this rule is really about.

Option A I did not take: it keeps the wrong quantity and adds a gate in front of it, and a gate is a dependency table that has to be maintained every time a rule is added. The quantity was the defect.

`Repair.columns` follows: it writes the separator and the canonical padding — what the canonical form writes there — instead of counting spaces from where the source's own line has got to. It no longer reads the source's columns at all, which is what let the earlier repair overshoot and correct itself in the next round.

Both sources now name exactly one rule, and `whole == true`:

```
one level deeper is one indent further in | 4 columns | 5 columns
what goes between two tokens on a line | ` ` | nothing
```

Held by `aColumnDoesNotReportWhatSomebodyElseMoved`, which reads the report's rules and `whole()`. Restoring the absolute-column reading turns it red with the three-rule report above. Both fixtures are in `sources()`, and so is a Japanese table — the property tests now carry the case #938 was reported against.

## The minors

- `GroupDecision.startColumn` said "the column on the screen". Past a table's column it is the column the width read. Javadoc corrected, and `AWidthIsMeasuredInColumnsNotCharacters` now holds the two apart the way it already holds the width apart from the text offsets: a group written at column 5 and measured from column 2. Making the decision carry the written column turns it red.
- The pass bound moved from `Columns` to `Doc`. You are right that `Columns` is the vocabulary and how many times a document is laid out is not part of it.
- Two javadocs still said the spacing rule is never asked at a stop.
- The `fmt --check` section of the docs says which column a row is said to be at, since that is now a thing a reader can be surprised by.

All nine modules green.

PR #958, opened 2026-08-22
A rule's rest is bounded by the relations too, and not by its positions' ends alone

### LINKED ISSUE #935: An argument that is a computed value carries neither the caller's guard nor what the helper's own body states
A helper is expanded into its caller, and a fact about one of its parameters reaches the
expansion only where the argument is a place. Where the argument is a computed value — another
helper's answer, or arithmetic over one — two things are lost at once: what a `guard` above the
call stated about that expression, and what the helper's own body establishes about the value
it answers with.

The second is the shape `AffineForms.Leaves.readsThrough` already names as #676: "a name given
a computed value is no more a place than the call it was given, and asked by the spelling it
came out one".

## The model

From `vending.sou` in `souther-test`, cut to one denomination and with the `example` blocks
removed so only the construction check runs.

```souther
let 商 (a: Int, b: Int): Int =
    match Int.divide(a, b) with
        | Int as n -> n
        | DivisionByZero -> unreachable "額面は定数で、0になることはない"

let 使える枚数 (残額: Int, 額面: Int, 手持ち: Int): Int = {
    let 必要 = 商(残額, 額面)
    if 必要 < 手持ち then 必要 else 手持ち
}

let 釣り銭を組む (額: Int, プール: 硬貨束): 釣り銭の組み方 = {
    let 十 = 使える枚数(額, 10, プール.十円玉.value)
    if 額 - 十 * 10 == 0
        then 組めた
            { 内訳 = 硬貨束 { 十円玉 = 硬貨枚数(十), ... }
            , 残り = 硬貨束 { 十円玉 = 硬貨枚数(プール.十円玉.value - 十), ... }
            }
        else 組めない
}
```

`硬貨枚数` is `Int` with `invariant value >= 0`, so two things are owed at the two
constructions: `十 >= 0` and `十 <= プール.十円玉.value`. The body of `使える枚数` states both
in both arms of its own `if`.

The caller, with the `guard` that establishes what `額` is:

```souther
guard 合計金額(自販機.投入) >= ジュース.値段.value else 何も起きなかった { 自販機 = 自販機 }
match 釣り銭を組む(合計金額(自販機.投入) - ジュース.値段.value,
                  束を合わせる(自販機.投入, 自販機.釣り銭ストック)) with
```

## What it produces

Only the two arguments at the call were changed between runs.

| `釣り銭を組む(額, プール)` | reported |
|---|---|
| `(0, 自販機.釣り銭ストック)` | — |
| `(合計金額(自販機.投入) - ジュース.値段.value, 自販機.釣り銭ストック)` | E2011 at `硬貨枚数(十)` and at `硬貨枚数(プール.十円玉.value - 十)` |
| `(0, 束を合わせる(自販機.投入, 自販機.釣り銭ストック))` | E2011 at `硬貨枚数(プール.十円玉.value - 十)` |
| as written | both |

Row one is the control: both arguments are places, and everything the helper establishes
reaches the constructions. Rows two and three are the two losses, and they are independent.

With all five denominations, as the model is actually written, it is ten E2011 — five at
`硬貨枚数(<枚数>)` and five at `硬貨枚数(プール.<額面>.value - <枚数>)`.

## What an author does about it

Restates inside the helper what the helper already says, with an `unreachable` to say it can
never be false:

```souther
guard 千 >= 0 && 五百 >= 0 && 百 >= 0 && 五十 >= 0 && 十 >= 0
    && 千 <= プール.千円札.value
    && 五百 <= プール.五百円玉.value
    && 百 <= プール.百円玉.value
    && 五十 <= プール.五十円玉.value
    && 十 <= プール.十円玉.value
    else unreachable "使える枚数 は 0 以上・手持ち以下しか返さない"
```

With that guard the module compiles clean. It is ten conjuncts restating what `使える枚数`'s
own `if` states, written at the one place the reading fails to carry it.

## Where it is

- `check/AffineForms.Leaves.readsThrough` — the javadoc names the shape and cites #676
- `inputs/InputPath.of` — what makes a spelling a place
- the seeding that carries a `guard`'s statement into an expansion

## What it takes

Not settled here. The two halves may or may not be one fix.

- **The `guard` half.** `合計金額(自販機.投入) - ジュース.値段.value` is the same expression in
  the guard and in the argument. Whether a fact is keyed on the expression or on the parameter
  the argument was bound to decides whether the two are the same subject.
- **The computed-argument half.** `プール` is bound to `束を合わせる(...)`, so
  `プール.十円玉.value` is not a place and no fact can be recorded at it. What the helper's
  `if` states is a relation between its answer and its third argument, which holds whatever
  the argument is — so what is missing is not a place to record it at but a way to state a
  relation between two expressions.
- **Whether this is what an `ensures` on a `let` would be.** A helper cannot carry one today.
  If the relation `使える枚数(a, b, c) <= c` were declarable, this model would be answered by
  declaring it, and the reading above would not have to derive it. That is a separate ask and
  probably a separate issue; noted here because it is the other way out.

Measured on `develop` at `6cf29597`.

kawasima:
Closes #935.

## What was wrong

A `guard` stating a relation between two values reached a construction where the argument was a place and not where it was computed. Measured on the issue's own model, cut to one denomination:

| argument | guard | reported |
|---|---|---|
| `値段` | `値段 >= 0` | — |
| `値段 - 10` | `値段 >= 10` | — |
| `合計金額(投入) - 値段` | `合計金額(投入) >= 値段` | E2011 |
| `合計金額(投入) - 値段` | `合計金額(投入) - 値段 >= 0` | E2011 |
| `値段 - 他` (both places) | `値段 >= 他` | E2011 |
| `値段 - 他` (both places) | `値段 >= 10 && 他 <= 10 && 他 >= 0` | — |

The issue's hypothesis — that a computed argument loses what a guard said — does not hold: the fifth row loses it with two places, and the second keeps it with a computed one. The boundary is between a fact each position carries in its own range and a fact only the two hold together.

## Where it was

The helper is where it showed and not what caused it. Binding an argument to a parameter turns a rule the body wrote about one position into a rule about however many positions the argument has, and every fact about it has to survive that.

`NumericDomain` derived a bound on a form three ways met — ends, the closed differences, and one rule taken off with the residual bounded by the ends — and called itself the one place a bound on a form is derived. `AffineReduction`, narrowing a rule one position at a time, needs the same answer about the rest of that rule and had its own: each position's ends summed. Two readings of one question, and the weaker of them builds the ends the stronger reads.

Same loss with no helper anywhere in it:

```
10z - x + y = 0     x - y >= 0     ⊬ z >= 0
```

## What changed

`FormReach` holds the rules, the ends and the closed differences as values and answers how far a form runs under them. `NumericDomain` reads it and derives nothing itself; `AffineReduction` bounds the rest of a rule through it.

Two commits, so the semantic delta is separately bisectable:

- `3ac985f7` extracts the reader and moves `NumericDomain` onto it. Behavioural no-op.
- `f386fdfb` moves the reduction onto it, which is the change.

## This widens what is derived

Composing rules is not new — `2x - y >= 0` beside `y >= 10` has always left `x` at or above five, and three chained rules have always reached what no two of them do, because each round hands its ends to the next. What is new is the channel: a rule composes through relations now and not only through ends.

The specification said two rules are never added together. It now says what the procedure does: one premise per reading, chains through rounds.

Restricting this to ends-plus-differences was measured and does not solve #935 — `合計金額(投入) - 値段` weighs 10 and −1, so the differences hold nothing about it.

## A rule does not bound the rest of itself

Sound if it did, and measured to move nothing, so it is refused to keep any bound off the rule it is a bound on. What is refused is the rule and not what the closure made of it: a difference-shaped rule is in the closed differences before this is asked anything, and those are the state.

## Tests

Small and stated, no randomness. The properties the reduction promises — sound against enumerated points, order-independent, monotone — are held where they already were, and now read this route.

Reverting to the old reading turns exactly these red, and leaves the three controls green:

```
AGuardsRelation…aRelationalGuardOverTheArgumentDischarges…    expected <[]> but was <[E2011]>
AGuardsRelation…theSameRelationStatedOfTheArgumentItself…     expected <[]> but was <[E2011]>
AGuardsRelation…aRelationBetweenTwoPlacesDischargesItAsWell   expected <[]> but was <[E2011]>
ARulesRest…aPositionIsBoundedByARelationOverTheRestOfItsRule  expected <true> but was <false>
ARulesRest…aRelationTheDifferencesCannotHoldBoundsTheRest…    expected <true> but was <false>
```

Full suite green, +9 tests, nothing else moved.

## Not in here

The vending model this came from needs two more things before its ten warnings go, both separate:

- `Int.divide`'s answer is not read as a quotient the way `/`'s is. `硬貨枚数(値段 / 10)` under `値段 >= 0` discharges; `硬貨枚数(商(値段, 10))` over `Int.divide` does not.
- A quotient is not related to what it divides. Under `a >= 0, b > 0` the relation is `0 <= a - b * (a / b) < b`; it is not unconditional, since truncation toward zero gives a remainder the sign of the dividend.

`AffineReduction.sidedBy` still reads the ends alone. Moving it onto the same reader decides disequality orientations that were undecided before, which is its own observable change and its own commit.


kawasima:
Both points fixed. The first one turned out to be wider than the sentence, so here is what I found before changing anything.

## 1. The specification sentence

You are right about which reading is meant, and the sentence could not settle it because it names a unit — "a reading" — that nothing defines.

The ambiguity is not the sentence's own. The rule is written in four places and derived in one:

| | says |
|---|---|
| `FormReach` | "two rules are never added together inside one reading" |
| `NumericDomain.proves` | "One rule, once. The residual is proven against **the box alone**" |
| `NumericDomain.highestProven` | "a reading composes at most one other rule" |
| `ClosedState` | "not inside any one reading" |
| specification | "A single reading takes at most one such premise" |

Only `FormReach` derives it; the rest are readers restating it, and the specification borrows the undefined unit from them. One of the copies had already drifted — `proves` says the residual is bounded by the ranges alone, which stopped being true once the differences were read into it. You checked that against the base and correctly withdrew it as pre-existing; I think it is mine all the same, because before this branch it was a redundant description of a derivation sitting next to it, and after it, it is a disagreeing description of a derivation in another class. That is the change turning a copy into a contradiction.

So rather than only rewording the specification, `FormReach` now states the rule once and says what it counts — one call to `of`, which takes at most one rule, with the residual bounded by the ends and the differences and nothing further — and says why the unit matters: a caller's act is usually larger than a query, since reducing a rule one position at a time asks a query per position. Every other mention points back.

The specification says it in terms of the two acts, close to your wording:

> It is also read one position at a time, with the rest of it bounded by that same question — the intervals, the differences, or one other relation of neither shape whose own residual the intervals and differences bound — so that a rule of that shape narrows each position it names rather than only answering questions put to it. Asking how far a sum reaches takes at most one such premise, which is what says none of them is added to another. What is derived is available to the next round, so longer chains come about through the ranges a round hands the next and not inside one question.

`b130741c`.

## 2. The two-place test

Caught correctly, and it was worth more than a rename. What this branch establishes is that whether the argument is a place decides nothing — the report it closes names the computed argument as the cause, and that account is wrong. A test asserting a place is asserting the refuted half, so renaming it to match what it did would have left the branch with no case that says so.

It is written with two places now and nothing computed anywhere in it: `guard 値段 >= 他` with the argument `値段 - 他`. That was reported before this change and is not now, and it goes red with the rest when the old reading is put back:

```
aRelationBetweenTwoPlacesIsLostTheSameWay                      expected <[]> but was <[E2011]>
aRelationalGuardOverTheArgumentDischargesTheConstruction       expected <[]> but was <[E2011]>
theSameRelationStatedOfTheArgumentItselfDischargesItToo        expected <[]> but was <[E2011]>
```

with the control still reported. `c93da06f`.

## Not changed

No new tests beyond the row above, and nothing randomized — agreed that `ARuleOverSeveralPositionsNarrowsEachOfThemTest`'s enumeration is the base's and stays as it is.

Full clean suite green. #962 is rebased onto this.

PR #962, opened 2026-08-22
Which side of a hole a sum falls is read off everything known

### LINKED ISSUE #961: Which side of a hole a sum falls is decided by the ends alone, and the rest of what is known is not read
`f /= v` states no half-space of its own. Which side of `v` the sum lies is something the rest of
what is known says, and `AffineReduction.sidedBy` asks the ends for it — each position's own range,
summed. Its javadoc says it is "asked of the box and so of everything known", and the box is not
everything known: the closed differences and the rules hold things about a sum that no position's
range carries.

So a `guard` writing `/=` beside a relation is read as stating nothing, while the same `/=` beside
a range on the value itself is read.

## What it produces

```souther
data Positive = Int
    invariant value >= 1

behavior f : (a: Int, b: Int) -> Positive
    constructs Positive
```

with `Positive(a)` built under each guard:

| guard | reported |
|---|---|
| `a >= 0 && a /= 0` | — |
| `b >= 0 && b <= 5 && a - b >= 0 && a - b /= 0` | E2011 |
| `b >= 0 && b <= 5 && a * 10 - b >= 0 && a * 10 - b /= 0` | E2011 |
| `b >= 0 && b <= 5 && a - b >= 0` | E2011 |

The first row is the one the ends answer: `a`'s own range starts at nought, so the hole moves it to
one. The second and third are the same argument through a relation — `a - b > b_min >= 0` over the
whole numbers puts `a` at one — and neither is read. The last row is the control: without the `/=`
there is nothing to derive and the report is right.

The second row's relation is difference-shaped and is in the closed differences; the third's is
weighed ten and minus one and is a rule. Both are held, and neither is reached from here.

## Where it is

- `check/AffineReduction.sidedBy` — takes a `Box` and sums it through `reachOf`
- `check/AffineReduction.reachOf` — the second reader of how far a form runs, and the only one left

`numeric/FormReach` is what answers that question everywhere else, over the ends, the closed
differences and one rule. `sidedBy` is the one caller that has not been moved onto it.

## What it takes

Reading `sidedBy` through `FormReach` is the whole of it, and the hole cannot reach itself by that
route because a disequality states no half-space (`AffineConstraint.halfSpaces`). Measured on a
throwaway build, the two rows above come out silent and the whole suite stays green.

It is kept out of the change that introduced `FormReach` because it is its own observable
difference: that change made bounds provable that were not, and this one decides orientations that
were undecided. A hole that was contributing nothing starts contributing a strict half-space, and
that is worth being able to bisect on its own.

Measured on `feature/issue-935` at `f386fdfb`.

kawasima:
Closes #961. Stacked on #958 — base is `feature/issue-935`, and this retargets to `develop` once that merges. The only commit of its own is `1263e906`.

## What was wrong

`f /= v` states no half-space. Which side of `v` the sum lies is something the rest of what is known says, and `sidedBy` asked the ends — each position's own range, summed. Its javadoc called that "the box and so everything known", which the existence of `FormReach` already contradicts: a sum is held by things no position's range carries.

`Positive(a)` with `invariant value >= 1`, under each guard:

| guard | before | after |
|---|---|---|
| `a >= 0 && a /= 0` | — | — |
| `b >= 0 && b <= 5 && a - b >= 0 && a - b /= 0` | E2011 | — |
| `b >= 0 && b <= 5 && a * 10 - b >= 0 && a * 10 - b /= 0` | E2011 | — |
| `b >= 0 && b <= 5 && a - b >= 0` | E2011 | E2011 |

Row one is what the ends answer on their own. Rows two and three are the same argument through a relation — `a - b` above nought and away from it is at one over the whole numbers, so `a` is at one — and neither was read. Row two's relation is difference-shaped and sits in the closed differences; row three's is weighed ten and minus one and is a rule. Both were held the whole time. Row four is the control: no hole, nothing to derive, correctly reported.

## What changed

`sidedBy` reads `FormReach`, and `reachOf` goes with it. Nothing in `AffineReduction` derives a bound on a form for itself any more.

The hole cannot side itself — a disequality states no half-space (`AffineConstraint.halfSpaces`), so it is never among what the reading may take as a premise, and nothing here arranges for that.

## Why not in #958

That change made bounds provable that were not. This one turns holes that were contributing nothing into strict half-spaces. Different observable surfaces, so they are worth bisecting apart.

## Tests

Reverting `sidedBy` alone turns exactly the two relational rows red and leaves the ends-only row and the control green:

```
AHoleIsSidedByEverythingKnown…aHoleOverADifferenceIsSidedByTheRelationThatHoldsIt      expected <[]> but was <[E2011]>
AHoleIsSidedByEverythingKnown…aHoleOverASumTheRelationsCannotHoldIsSidedByTheRuleDoes  expected <[]> but was <[E2011]>
```

Full clean suite green, +4 tests, nothing else moved.

### DIFF
--- a/souther-compiler/src/main/java/souther/compiler/numeric/AffineReduction.java
+++ b/souther-compiler/src/main/java/souther/compiler/numeric/AffineReduction.java
@@ -97,7 +97,7 @@ public final class AffineReduction {
         Map<A, RationalCut> atMost = new LinkedHashMap<>();
         for (AffineConstraint<A> each : constraints) {
             List<AffineConstraint.HalfSpace<A>> halves = each instanceof
-                    AffineConstraint.Disequality<A> hole ? sidedBy(hole, from) : each.halfSpaces();
+                    AffineConstraint.Disequality<A> hole ? sidedBy(hole, reading) : each.halfSpaces();
             if (halves == null) {
                 return new Reduction.NothingIsLeft<>();
             }
@@ -121,19 +121,27 @@ public final class AffineReduction {
      * to go above, then it goes below. And where it is known to be exactly {@code v}, nothing is
      * left.
      *
-     * <p>Asked of the box and so of everything known, rather than of what happened to be known when
-     * the rule was read. That is the whole difference: {@code x /= 0} beside {@code x >= 0} used to
-     * leave {@code x} at nought or above depending on which of the two was written first, because
-     * the answer was decided at the moment the disequality arrived and never revisited. Here it is
-     * decided against the same box every other rule is read against, and again next round if the
-     * box has moved.
+     * <p>Asked of everything known, rather than of what happened to be known when the rule was read.
+     * {@code x /= 0} beside {@code x >= 0} used to leave {@code x} at nought or above depending on
+     * which of the two was written first, because the answer was decided at the moment the
+     * disequality arrived and never revisited. Here it is decided against the same state every other
+     * rule is read against, and again next round if that state has moved.
+     *
+     * <p>Everything known is {@link FormReach} and not the ends alone. A sum is held by things no
+     * position's own range carries — a relation between two of them, a rule naming a third — and
+     * asking the ends for it left a {@code guard} that wrote {@code /=} beside a relation stating
+     * nothing, while the same {@code /=} beside a range on the value itself was read.
+     *
+     * <p>The hole cannot side itself. A disequality states no half-space
+     * ({@link AffineConstraint#halfSpaces}), so it is not among what the reading may take as a
+     * premise, and nothing here has to arrange for that.
      *
      * @return the half-spaces it comes to, empty where the sum can still fall either side, or null
      *         where the sum is pinned at the very value it is held away from
      */
     private static <A> List<AffineConstraint.HalfSpace<A>> sidedBy(
-            AffineConstraint.Disequality<A> hole, Box<A> from) {
-        Reach reach = reachOf(hole.form(), from);
+            AffineConstraint.Disequality<A> hole, FormReach<A> reading) {
+        Reach reach = reading.of(hole.form().coefs(), Rational.ZERO);
         Rational away = hole.at();
         boolean neverBelow = reach.least() != null && reach.least().at().compareTo(away) >= 0;
         boolean neverAbove = reach.most() != null && reach.most().at().compareTo(away) <= 0;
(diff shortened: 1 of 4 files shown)


kawasima:
Rebased onto `develop` now that #958 is merged (`13baeab8`), and retargeted. This is a single commit of its own, `bf515613`, and it has not been reviewed yet.

Full clean suite green on the new base.

kawasima:
Fixed in `7b689b68`. You are right that the property tests never saw a `/=`, and chasing why turned up something wider than one missing case.

## Root cause

It is not that `/=` was left out. The generator writes the set of relations out by hand:

```java
Rel rel = switch (dice.nextInt(5)) {
    case 0 -> Rel.LT;
    …
    default -> Rel.LE;
};
```

The `default` arm means it goes on compiling however far that list drifts from `Rel`, and nothing compares the two. The side that reads a relation back is a switch over the enum with no `default`, which the compiler checks. One side checked, the other a copy — and the copy drifted twice, differently:

| generator | generates | short of |
|---|---|---|
| `ARuleOverSeveralPositionsNarrowsEachOfThemTest` | LT, GE, GT, EQ, LE | `/=` |
| `OneStateAnswersForWhatTheRulesLeaveTest` | LT, GE, EQ, LE | `/=`, `>` |

The second one is where the closure's own soundness and order-independence live, and it is short two.

So I did not add `case 4 -> Rel.NE`. That leaves the same defect in place, does not touch the second generator, and goes wrong again the next time `Rel` gains a member. The relation is taken from `Rel.values()` now, in both.

## What it closes, measured

Siding every hole downward regardless of what is known, and running the two classes:

| | old generators | relations from the enum |
|---|---|---|
| `aHoleAtTheFloorLiftsIt` | fails | fails |
| `aHoleWithNothingToSideItBoundsNothing` | fails | fails |
| `noPointTheRulesAdmitIsEverRefused` (reduction) | **green** | fails |
| `noPointTheRulesAdmitIsEverRefused` (closure) | **green** | fails |

`OneStateAnswersForWhatTheRulesLeaveTest` was green throughout under the old generators. Under the new ones the rules it fails on are written with `/=`:

```
refused a point the rules admit: {a=-4, b=-4, c=2}
  under [Written[coefs={a=-2, b=1, c=3}, constant=6, rel=NE], …]
```

So the three properties do now run through `sidedBy(…, FormReach)`, and they catch it when it is wrong — which was the guarantee you were asking for.

Full clean suite green. Nothing added to the suite: this is the same generators producing one more shape.

Agreed on the other two points and left alone — no `ofWithout(hole, …)`, and the self-reference is already contracted by `Disequality.halfSpaces()` being empty.

Issue #974, opened 2026-08-22
`Int.min`, `Int.max` and `Int.clamp` answer one of their arguments and no reading says which


kawasima:
`DischargeRules.CHOOSES` writes out the cases each of these is defined in — the argument
it answers, and what holds of the arguments where it does. `Predicates.Piecewise` reads
that table to decide a clause. Nothing reads it to bound a value, so an operation that
answers one of its arguments has no range at all where a clause is not written directly
over it.

## What it produces

```souther
data U = Int
    invariant nonNeg = value >= 0
```

| the step, under `U(List.fold(step, 0, xs))` | reported |
|---|---|
| `(a, x) -> a + x.tani.value` | — |
| `(a, x) -> Int.min(a, x.tani.value)` | E2011 |
| `(a, x) -> Int.max(a, x.tani.value)` | E2011 |

Measured on `develop` at `b7b5f46a`, and unchanged by #964: the smallest of two values at
or above zero is at or above zero, and nothing establishes it. Outside a fold the same
call is silent, because a clause over it is read through `Piecewise` where it stands —
so what is missing is a range and not a reading.

## Where it belongs

These are choices. #964 settled what a choice is — a value that is one of several,
recorded as `Derivation.Chosen` and bounded by the range its arms span — and named these
as the producer it has none for. The step from the table to the arms is the same step
`Piecewise` already takes:

```
Core.If ─────┐
Core.Match ──┼──→ arms ──→ Derivation.Chosen
CHOOSES ─────┘      │
                    └───→ Predicates.Piecewise
```

What is shared is the arms and not the reading of the syntax: `Piecewise` builds its
`numeric` by substituting the clause into the case, and `Chosen` takes the answered form
as it stands. Reading `CHOOSES` a second way to build the second of those is the shape
#935 was — one table, two readers, drifting.

## What it takes

The translation from a `Choices` row to an arm — the argument answered, as a form, and
the conditions on the arguments, as constraints — lifted to where both readers ask it.
The conditions half lands where #973 puts the same half for a source-written
condition, so the two want doing together or in that order.

`Choices` states that its cases are exhaustive, which is what makes a span over them a
statement about the result. That claim is already written down beside the table and is
not one this has to make again.

kawasima:
Merged into `develop` as 917d7fa3 (PR #986).

The table in the body, measured on the merged tree with `U = Int invariant nonNeg = value >= 0`:

| the step, under `U(List.fold(step, 0, xs))` | reported |
[...]

Issue #973, opened 2026-08-22
An arm is never read under what choosing it settles — a condition, a binding, or an attempt's built value


kawasima:
A value that is one of several is bounded by the range its arms span
(§invariant-discharge-arithmetic). What choosing an arm settles holds exactly where that arm is
the answer, so an arm read under it says more than the arm alone — and no arm is read that way.

## What it produces

It is one gap and it shows through all three kinds of choice. Measured on `feature/issue-964`
at `98f9e4eb`, where the arms themselves are read.

| the step, under a fold | reported | what is unread |
|---|---|---|
| `(a, x) -> if x.flag then a + x.tani.value else a` | — | nothing is needed |
| `(a, x) -> if a + x.v < 100 then a + x.v else 100`, into a type capped at 100 | E2011 | the condition |
| `(a, x) -> match x.tag with \| Held as h -> a + h.amount.value \| Empty -> a` | E2011 | what the arm bound |
| `(a, x) -> if NonNegInt(x.v) as n then a + n.value else a` | E2011 | what the attempt built |

The neighbour of the third and fourth — the same step reading the *element* instead of what the
arm bound — is silent, so what is owed is the binding and not the branch.

`AWalkFromASeedIsBoundedByCheckingOneStepTest.anArmReadingWhatItBoundGetsNoRange` holds that
boundary today. It is the test that says this landed.

## Why it is one gap and not three

An `if`'s condition, a `match` arm's pattern and binding, and an attempt's built value are all
answers to one question: **what does choosing this arm settle?** They differ in what the answer
contains, not in whether there is one.

- a condition contributes numeric constraints and predicate facts;
- a `match` arm contributes those *and* a denotation — its binder stands for the value the
  scrutinee already was, refined to the case (`PathEngine.enteringArm`);
- an attempt contributes what its construction guarantees, bound to its `as` name
  (`PathEngine.enteringBuilt`).

So the thing to lift is not a condition reader. Reading it as one closes the `if` row above and
leaves the other two open, which is the same defect arriving twice.

## What it takes

An arm's context, as a value both readers can be handed:

```
Choice.Kind + the node + Denotations
        ↓
what choosing this arm settles
  numeric constraints
  predicate facts
  denotation / refinement of what it binds
        ↓
PathEngine          →  Entered (Known + Denotations)
Derivation.Chosen   →  the domain and environment an arm is read in
```

`PathEngine` already computes exactly this and returns it as `Entered`, through
[...]

Issue #972, opened 2026-08-22
Reading an arm under its own condition needs a cost policy, and there is none to inherit


kawasima:
A recipe is evaluated once for the reading that asks it
(`ARecipeIsEvaluatedOnceForTheReadingThatAsksItTest`). A reading is one evaluation
against one domain, so that claim holds because every arm of a choice is read against the
same domain today. Reading an arm under what chose it ends that: two arms are two
domains, and a recipe standing under both is rightly evaluated once for each.

## What it produces

Nothing yet. This is the shape the work in #ISSUE_A takes on, written down before it is
done rather than found afterwards.

```souther
if C1 then (if C2 then ... else ...) else (if C3 then ... else ...)
```

Read under the conditions, that is four domains at two levels and 2^n at n. The memo that
holds the cost flat today is keyed on the atom for one domain; it cannot be shared across
arms, and keying it on `(atom, domain)` does not help, since the domains differ.

## What it is not

It is not the split bound. `READINGS_A_PATH_MAY_COST` counts re-readings of a body, and
an arm read is a question put to a domain — different units, and sharing the number
between them would say they are one
(`HowFarSplitsAreOpenedIsBoundedByWhatARereadingCostsTest` states which is which). What
the two have in common is that neither may compound, and that is the property to keep,
not the constant.

## What it takes

A decision, stated once, about how finely arms are read:

- read every arm under its own conditions, bounded by something that says how far —
  which needs a cost model the split bound and this can both be read out of, rather than
  a second constant;
- or read an arm under its conditions only where it is the outermost choice, and read
  what stands inside it as the arms alone — precise where authors write clamps, flat
  where they nest;
- or something else, said before it is built.

Whichever it is, `ARecipeIsEvaluatedOnceForTheReadingThatAsksItTest` states what holds
today and will need saying again in the new terms. What may not happen is the claim
quietly ceasing to be true because the reading it was about stopped being the only one.

kawasima:
Settled and merged in #979 (2f953031). The policy lives in `ContextMultiplicity`: what is bounded is how far nested splits copy a reading, one limit both readers ask, and past it the copy per arm is given up rather than the reading. #973 connects `DerivedBounds` to it.

Issue #964, opened 2026-08-22
A branch anywhere in a fold's step drops the reading, even where both arms are the accumulator


kawasima:
`§invariant-discharge-reduction` reads a fold by its step: a step that takes an accumulator inside
the range and answers inside it leaves the fold inside it. A step that branches is not read at all,
whatever the branches say.

## What it produces

```souther
data U = Int
    invariant nonNeg = value >= 0

data Item = { flag: Bool, kubun: Kubun, tani: U }
```

| the step, under `U(List.fold(step, 0, xs))` | reported |
|---|---|
| `(a, x) -> a + x.tani.value` | — |
| `(a, x) -> { let v = x.tani.value  a + v }` | — |
| `(a, x) -> match x.kubun with \| Fukumu -> a + x.tani.value \| Fukumanai -> a` | E2011 |
| `(a, x) -> if x.flag then a + x.tani.value else a` | E2011 |
| `(a, x) -> if x.flag then a else a` | E2011 |

The last row is the one that says what this is. Both arms are the accumulator itself, so nothing
about the arms can be what stopped it — a branch in the step drops the reading, and what the arms
answer is never looked at. `match` is not the cause; `if` does it too, and a `let` in the step does
not. Measured on develop (b7b5f46a).

## Where it is felt

A model that writes a conditional total — sum the units of the rows in one class and skip the rest
— has to lift the condition out into a `List.filter` and fold the filtered list, which is read.
That is a workable rewrite and it is not always the honest one: the condition and the addition are
one sentence in the source document, and splitting them is a change to the model made to satisfy
the checker.

## What it takes

The rule already says what the answer is: a step is read if every arm of it takes an accumulator
inside the range and answers inside it. That is the join a branch needs and the fragment has it
everywhere else — a `guard` and an `if` in a body are read. What is missing is the reading of the
step reaching into the arms rather than stopping at the branch.

kawasima:
The reading of this landed in #975, which is the abstraction that was missing rather than a fix for branching steps: `Derivation.Chosen`, beside the product and the truncating quotient. The five branching steps in the table above are silent on that branch.

Leaving this open. The abstraction is in and two of its producers are not connected, so the same missing reading still shows through them:

- #973 — an arm is never narrowed by what chose it, so a clamp still reports
- #974 — `Int.min` / `max` / `clamp` have no producer, so they still report in a step
- #972 — the cost policy #973 needs before arms are read under their own domains

[...]

PR #975, opened 2026-08-22
A value that is one of several has a range, and a fold's step gets it for free

kawasima:
A step that branches was never read, whatever its arms answered (#964): a `match` was
reported, an `if` was reported, and `if x.flag then a else a` — both arms the accumulator
itself — was reported too, so nothing about the arms was what stopped it.

## What it was

Not the fold, and not `if`/`match`. `Terms` names a value the affine walk composes
nothing out of and files a recipe for how it was computed; `DerivedBounds` evaluates the
recipe against a domain. `Derivation` had two kinds — a product, and a truncating
quotient. **Being one of several arms is a third way a value is computed and it had
none**, so a branch was named and nothing was ever filed under the name: the atom stood
in the step's form with no range at all.

Measured on `develop` at `e004df78`, which is what said it is not about branches:

| step, under `U(List.fold(step, 0, xs))` | before |
|---|---|
| `(a, x) -> a + x.tani.value` | — |
| `(a, x) -> { let v = x.tani.value  a + v }` | — |
| `(a, x) -> match x.kubun with \| Fukumu -> a + x.tani.value \| Fukumanai -> a` | E2011 |
| `(a, x) -> if x.flag then a + x.tani.value else a` | E2011 |
| `(a, x) -> if x.flag then a else a` | E2011 |
| `(a, x) -> Int.min(a, x.tani.value)` | E2011 |
| `U(if x.flag then x.tani.value else y.tani.value)` — outside a fold | — |

Outside a closure the region walk covers this by reading the whole body once per arm,
which is why it was not felt there. **That reading cannot be made inside a step**: the
step runs once per element and the condition is about the element, so different elements
take different arms and no single arm may be assumed over the body. `bindingIn` stops at
a `Core.Block` for exactly that reason — which left a branch in a step the one value
position with no reader at all.

## What is here

`Derivation.Chosen` is the third arm and `DerivedBounds` spans it.

`InductiveBounds` is unchanged. It already reads the step's form against a domain with the
candidate assumed, and the recipe graph derives whatever that form names against whichever
domain it is handed — so the induction gets the arms for free. No fold-shaped special case
was added anywhere.

The first five rows above are now silent. A step one arm of which lowers the accumulator,
and one an arm of which is below the seed, stay reported.

### Which values a choice is one of has an owner

Review found `Core.IfConstructed` missing from the producer, and measuring it found the
cause wider than the missing case. **The question was asked in four places, each with its
own list** — `splitIn`'s search for a site, `splitOf`'s reading of the arms, `isASplit`,
and the recipe recording. All four said `if` and `match`. `splitOf` already guarded two of
them against drifting apart, so the shape was known and the guard sat between two copies of
one wrong list. An attempt answers what it built where its invariant held and what it
departs with where it did not, and it was in none of them.

`Choice` owns it: what the choice asks, and per arm the value answered and the node that
decides it. It puts **no reading** on that node — what a condition or a case or an attempt
*settles* needs an environment, and that reader is `PathEngine`, which not every caller can
reach (#973). Naming the node without interpreting it is what lets `Terms` take the values
while `InvariantChecker` takes more.

So `Terms` asks it and all three producers arrive at once; `splitOf` reads the arms off the
choice and adds only the entering; `splitIn` finds a site by asking `isASplit`, so one
answer decides which kinds this walk opens.

The forcing is on `Choice.Decides`, sealed, and not on `Kind`. A kind is inert until
something produces it; a new node needs a case in `Choice.of` — refused unclassified by
`WhichValuesAChoiceIsOneOfIsAskedInOnePlaceTest` — and a way to be entered, which is where
`choosing` stops. Measured both ways: a bare fourth `Kind` declared split-capable compiled
clean under the first attempt at this, and a producer with a new way of deciding now fails
at `choosing`.

## What is deliberately not here

**What choosing an arm settles is not read**, and it is one gap across all three kinds:

| step | reported | unread |
|---|---|---|
| `if x.flag then a + x.tani.value else a` | — | — |
| `if a + x.v < 100 then a + x.v else 100` (capped type) | E2011 | the condition |
| `match x.tag with \| Held as h -> a + h.amount.value \| Empty -> a` | E2011 | what the arm bound |
| `if NonNegInt(x.v) as n then a + n.value else a` | E2011 | what the attempt built |

The neighbour reading the *element* instead is silent, so what is owed is the binding and
not the branch. Reading any of it needs the account of what an arm settles that `PathEngine`
keeps, and `Terms` cannot ask for it — `Predicates` is constructed over `Terms`, and
`ContractDischarge` builds a bare `Terms`, so a collaborator set afterwards would weaken
that reader silently. Writing a second, smaller reader is the shape #935 was. → #973, and
the cost policy it needs → #972.

**`DischargeRules.CHOOSES` has no producer in `Choice` yet**, so `Int.min`/`max`/`clamp`
still report in a step. → #974.

Both are this abstraction unconnected, not its limit. **#964 stays open until they land.**
The spec and `Derivation.Chosen` say so in those terms rather than stating the current
reading as the rule.

## Spec

§invariant-discharge-reduction said "a step that chooses between two answers is not read",
which states a rule about steps for something that is not about steps. The rule is now
stated once beside the product and the quotient, the reduction section inherits it, and
that section's title widens to what it was already about. All three examples measured.

## Two contracts fixed where they can fail

Neither was observable before, and both were found by asking what a later change could
break quietly.

**What a recipe is read from is what a reading reaches.** `Derivation.formsRead` moves the
question to the recipe, so `Terms.reached` has one case rather than one per kind. Held
against the recipes themselves: each is built with a marker wherever a form can stand and
asked what it says it is read from; a component the test cannot build is a failure, not a
skip. Measured red three ways — dropping a component from `formsRead`, adding a bare form
field, and adding one nested inside a `Constraint` (the shape #973 will add).

**A recipe is evaluated once for the reading that asks it.** Not "nested choices are
linear", which stops being true once arms are read under their own conditions — two arms
are two domains, and a recipe asked under both is rightly evaluated twice. Measured: six
choices nested in a step cost the same per level as one; with the memo bypassed the counts
go `10, 26, 58, 122, 250, 506`.

This is not the split bound and does not share its constant. `READINGS_A_PATH_MAY_COST`
counts re-readings of a body; an arm read is a question put to a domain. Sharing the number
would say they are one unit. Measured, the two do multiply — `armReads = readings × arms`,
e.g. 16 × 12 = 192.

**That product is not bounded here and this PR does not claim it is.** How many readings
there are is the region walk's own policy, and it opens the first split on a path however
wide it is: measured, a match of eighty cases is opened and the body read a hundred and
sixty times. What is claimed is only that this adds no factor of its own — within one
reading the recipes do not multiply through nested choices, whatever produced the reading.
The path-sensitive explosion that #972 is about is a different one: it arrives when arms
are read against their own domains, which no reading does today.

`DerivedBounds.WATCHING` had to change to see any of this: it recorded the *set* of atoms
a reading reached, which cannot say one was evaluated twice, and it did not watch a walk's
own readings at all — which is exactly where a step's recipes are evaluated. A reading now
carries what it answered beside what it evaluated, counted past the memo.

`HowFarSplitsAreOpenedIsBoundedByWhatARereadingCosts` read a warning to observe whether a
split was opened. With a second route bounding the same value that warning stops
distinguishing the two, so it now reads `InvariantChecker.OPENING`. Both ends of the bound
measured red by moving the constant either way.

## Verification

`mvn -o test` from the reactor root, green — 6377 tests. Every claim above was measured on
this branch, and every new test was measured red by breaking what it is about.

Closes nothing. #964 stays open; #972, #973, #974 are its remaining work.


kawasima:
Pushed `98f9e4eb`. All four addressed; the first two turned out to be one cause.

**1. `IfConstructed` — the cause is wider than the missing case.** Measured, it reproduces #964 exactly: `if U(x.tani.value) as n then a else a`, both branches the accumulator, was reported where the plain `if` was not.

But "which nodes answer one of several values" was asked in **four** places, each with its own list — `splitIn`'s search, `splitOf`'s arms, `isASplit`, and the recipe recording. All four said `if` and `match`. `splitOf`'s `default` already guards two of them against drifting ("`splitValueIn` and this answer for the same forms"), so the shape was known and the guard sat between two copies of one wrong list. Adding a case to `Terms` would have made it five.

So: `Choice` owns the question, answering in `Core` alone. `Terms` asks it and all three producers arrive at once. `isASplit` is derived from it by a switch over `Choice.Kind` **with no default** — an attempt is a choice and is not a split, because the walk reads it with what it built bound rather than by substituting arms, and that is now a recorded decision rather than an absence. A kind added later stops that switch.

Implemented rather than deferred, since with the owner in place it is not a case to add.

**2. arm context — one gap, not three, and #973 is widened.** Your reading is right and it is broader than `match`: the same measurement shows an attempt's `as` name has it too.

| step | reported | unread |
|---|---|---|
| `if x.flag then a + x.tani.value else a` | — | — |
| `if a + x.v < 100 then a + x.v else 100` (capped type) | E2011 | the condition |
| `match x.tag with \| Held as h -> a + h.amount.value \| Empty -> a` | E2011 | what the arm bound |
| `if NonNegInt(x.v) as n then a + n.value else a` | E2011 | what the attempt built |

The neighbour reading the *element* instead is silent, so what is owed is the binding and not the branch. `anArmReadingWhatItBoundGetsNoRange` holds that boundary with both halves.

#973 is retitled and rewritten to arm context — constraints, facts, **and** denotation/refinement — with the note that lifting a condition reader alone closes one row of three and lets the same defect arrive twice.

**3. non-empty invariant.** On `Chosen` and on `Choice`. `DerivedBounds` no longer has an empty branch to fall back on — it reads the first arm and spans the rest, so there is no dead path pretending to be a policy.

**4. spec wording.** You are right, it was backwards. Now: the join is sound — a range the value cannot leave — and *wider* than the value's own range, so what is narrower is the fragment recognised, not the range.

Also measured: an attempt departs once per clause it names, and every departure counts. `everyDepartureAnAttemptNamesIsOneOfTheValuesItAnswers` puts the low value in each arm in turn; reading only the first departure makes the second case go green, measured.

**Verification.** `mvn -o test` green, 6377 tests. Each new test measured red by breaking what it is about — dropping `IfConstructed` from the owner (3 red), and reading only the first departure (1 red).

#964 stays open with the same three follow-ups; its comment records what this changed.

kawasima:
Pushed `53a0d71c`. The blocker is fixed, and measuring it found the tripwire beside it worse than the double answer.

**`isASplit` had one caller, and it was not the one that opens splits.** Sites are found by `splitIn`, which recognised `Core.If` and `Core.Match` by hand and asked nothing. So the exhaustive `Choice.Kind` switch — with a javadoc saying a kind added later is one it stops at — could be satisfied with `case A_NEW_KIND -> true` and the build stayed green, with no reader anywhere opening one. Measured: a fourth kind declared split-capable compiles clean. **That is worse than no tripwire, because the javadoc claimed it held.** Your scenario, and the reader I put the switch on was not even in it.

Three readers, one owner now:

- `Choice` carries what it asks and, per arm, the value answered **and the node that decides it** — with no reading put on that node. What a condition or a case or an attempt *settles* needs an environment, and that reader is `PathEngine`, which not every caller can reach. That is the split you asked for: `body` is Choice's, `what it settles` is #973's, and Choice now hands over the node rather than making #973 go back to the tree for it.
- `splitOf` reads the arms off the choice and adds only the entering — no index alignment across files, since each arm carries its own decider.
- `splitIn` finds a site by asking `isASplit`, so one answer decides which kinds this walk opens.

**The forcing is on `Choice.Decides`, not on `Kind`.** A kind is inert until something produces it — measured, a bare fourth `Kind` still compiles, and rightly, since nothing constructs it. What a new node really needs is a case in `Choice.of` (refused unclassified by the architecture test) and a way to be entered. Measured: a producer with a new way of deciding fails at `choosing`, where before it compiled clean.

One `Decides` arm per way an arm is chosen and not one per kind: an attempt that held and an attempt that departed settle different things, so a reader treating them alike would tell them apart again.

`splitValueIn`'s per-node specialisation is left alone, as you said — it answers where a value position is and where a different environment begins, which is not this question.

**Non-blocking 1.** You are right and it was untrue of its own members. `NOT_A_CHOICE_BY_ITS_SHAPE`, which is what it was always holding — and the javadoc now says why a shape is all this list can classify, with `PreservedCall` staying put when #974 lands.

**Non-blocking 2.** PR body rewritten to the current design: the owner, the `Decides` forcing, and #973 as arm context rather than a condition reader.

**Verification.** `mvn -o test` green, 6377 tests. The behaviour of `splitIn` is unchanged — the restructure keeps every descent in place and only moves the two site decisions behind one gate — and the suite is the evidence.

kawasima:
Pushed `420fcefd`. Both required fixes plus the three optional ones.

**1. Cost contract.** You are right and I measured it before changing the words. `opens` lets the first split on a path be opened however wide it is — its own javadoc says so — and a match of eighty cases is opened with the body read a hundred and sixty times. So "readings are capped" was false, and "the product is linear in the program" went with it.

The test javadoc and the PR body now say what this can state: **within one reading the recipes do not multiply through nested choices, whatever produced the reading.** Nothing here bounds a compilation, and the PR body says so explicitly, with the note that #972's explosion is a different one — it arrives when arms are read against their own domains, which no reading does today.

**2. `Choice` javadoc.** Also right, and it was the same fault this PR is about: the words claimed a tripwire the design had already moved off. It now says a bare `Kind` compiles and should, since a kind nothing produces is inert; what a new choice needs is a case in `of` and then a way for each arm to be decided, and it is that which fails exhaustiveness. The paragraph keeps the history, since "a tripwire that read as holding" is where this class came from.

**3. Instrumentation cost** (non-blocking, taken). The `evaluated` list is now allocated and appended only while `WATCHING` is set, matching `InvariantChecker.OPENING`. Every compilation was writing a value per recipe evaluated for nobody — and as you say, a hook a compilation pays for is one its answer could come to depend on.

**4. Positive set name** (non-blocking, taken). `A_CHOICE_BY_ITS_SHAPE`, named against its opposite, with the note that a shape can only say whether *every* value written at that node is one of several — which for a call it cannot, so a call stays on the other side whatever #974 makes of the operation.

**5. `match` binder boundary** (optional, taken). It now stands beside the other two in `anArmReadingWhatItBoundGetsNoRange`, so all three kinds of arm context have their current boundary recorded where #973 will read them.

`mvn -o test` from the reactor root, green across all ten modules.

Merging.

PR #983, opened 2026-08-23
A backend does not change a value to fit a host API

### LINKED ISSUE #976: A scale no int holds is divided at another scale, and nothing says so
`Decimal.divide(a, b, scale, mode)` takes its scale as an `Int`, which is
64 bits (§primitives). `BigDecimal.divide` takes a Java `int`. The backend
puts the one into the other with a raw `l2i`.

## Where it is

`codegen/BodyGen.decimalDivide` — `genExpr(call.args().get(2))` then
`code.l2i()`, straight into `BigDecimal.divide(divisor, scale, mode)`.

## What it produces

A scale outside 32 bits is narrowed by dropping the high bits, so the
division runs at a scale that is neither what was written nor an error.
`4294967298` (2^32 + 2) divides at scale 2. Nothing aborts and nothing
is reported, which is the shape §stdlib-int refuses everywhere else: an
`Int` operation that leaves the range it holds *aborts* rather than
wrapping.

`Decimal.round(scale, d)` and `Decimal.toInt` take a scale the same way
and want reading for the same narrowing.

## What it takes

Not settled here. The two shapes are an abort where the scale is one no
`BigDecimal` can be given, which is what an overflow already does, and a
type that cannot hold such a number in the first place. The first is the
smaller change and is what the rest of the arithmetic does.

## Why it is filed now

The invariant-discharge rule about a quotient rounded to a scale
(§invariant-discharge-arithmetic) states a range only where the scale
reads as a number the run time divides at, and it holds the number to
what an `int` is for exactly this reason: a proof over the scale as
written would be a proof about a different division. That is a rule
working around this rather than a reason to leave it — the run time and
the check should divide at one scale.

kawasima:
Closes #976.

## What was wrong

`Decimal.divide(a, b, scale, mode)` takes its scale as an `Int` — 64 bits — and `BigDecimal.divide` takes a JVM `int`. The backend put the one into the other with a raw `l2i`. Measured before the change:

```
divv scale=2            -> 2.33
divv scale=4294967298   -> 2.33          the high bits dropped; scale 2
divv scale=-4294967296  -> 2             scale 0
divv scale=2147483647   -> !! java.lang.ArithmeticException: BigInteger would overflow supported range
```

The first two are the issue: a division at a scale that is neither what was written nor an error. `Decimal.round` did the same with a `(int)` cast in the runtime. `Decimal.toInt` does not — its scale is the constant 0 — so the sites were two, not three.

The fourth is a second hole the issue does not name, and it decides the shape of the fix: `2147483647` *is* handed over exactly and `BigDecimal` still refuses. So does `add`, `subtract` and `multiply` at the ends of the scale range — `p.a * p.a` on a decoded value with scale `Integer.MAX_VALUE` left a behavior as `java.lang.ArithmeticException`, a `java.math` type named in the failure of a program that has no such type. Narrowing exactly would have left every one of those open.

And the check had already worked around the backend's silence: `placesOf` refused to state a range for a scale no `int` holds, and §invariant-discharge-arithmetic wrote down the reason — *what the backend divides at is that scale narrowed*. A rule in the specification shaped by an instruction the backend happened to select.

## What it takes

ADR-0112: **a backend never changes a Souther value to make it fit a host API.** A host operation with a narrower value domain, another representation, or failure modes that are not Souther values is mediated by a runtime kernel that owns it; conversion is exact; failure is translated to the result or abort that operation states.

The line is between knowing a representation and owning a meaning. The backend may know a `Decimal` is a `BigDecimal` — it builds literals with it, names it in descriptors, casts to it. It may not implement `+` with `BigDecimal.add`. `Int`'s operators already read this way through `IntMath`; `Decimal`'s did not, except `/`.

- every `decimal.*` kernel is one emitter, `DeclaredStatic` reading its declaration, owned by `DecimalMath` — it was four shapes, and `divide` was hand-written in `BodyGen`
- the four operators emit calls to the same runtime
- narrowing the scale and failing the operation are separate, and their messages say which happened
- `JdkVirtual.l2iArgs` is deleted: no entries, and it offered the raw narrowing in the one place the next kernel taking a host `int` would reach for it
- the specification states the range; the compiler and the runtime implement it separately, held together by a test over the ends rather than by shared code. The thousand-place reading budget stays a separate number.

After:

```
divv scale=4294967298   -> ConstraintViolation: Decimal.divide scale is outside the range the run time takes: 4294967298
divv scale=2147483647   -> ConstraintViolation: the quotient of ... at scale 2147483647 is outside the range a Decimal holds
```

## Two consequences worth reading

**`Decimal.divide` now evaluates all four arguments.** The hand-written emitter evaluated the scale and the mode only on the branch it divided on. That is a short-circuit no declaration wrote down — only `&&` and `||` decide which of their operands run (§a-condition-stops-when-its-answer-is-settled). Which of the two invalid conditions decides the answer is a separate question and is now stated: a zero divisor answers `DivisionByZero` whatever the scale, a division that does not run needing no scale to run at.

**An abort message describes an extreme value by sign, precision and scale**, not by `toPlainString()`. Spelling out a value at the end of the scale range asks for the allocation the operation just refused to make, which would have made the failure path fail.

## Tests

Regression, over both failures and both spellings: `ADecimalOperationIsHeldToWhatItCanAnswerTest` — including that `java.lang.ArithmeticException` does not reach a boundary, that the operators abort, argument evaluation, zero-divisor precedence, and message size.

Structural, so the shapes cannot come back:

- `DecimalArithmeticIsTheRunTimesAndNotTheBackendsTest` scans emitted bytecode for a Decimal operation run on `BigDecimal` by generated code, and holds every `decimal.*` kernel to one emitter on `DecimalMath`. Written as "no Decimal *operation* on BigDecimal" rather than "no `BigDecimal` in codegen", which would fail on a model that writes `1.5m`.
- `OneScaleRangeIsReadByTheCheckAndByTheRunTimeTest` asks both implementations at `MIN-1 / MIN / MAX / MAX+1`, and holds the reading budget apart from the range.
- `EveryKernelIsDeclaredTest` no longer excuses `decimal.divide`.

`Int.divide` / `Int.truncatingRemainder` stay written out in `BodyGen`: their zero divisor is a case the language defines, not a host partiality.

Full reactor: 9,659 tests, 0 failures.


kawasima:
3点、直しました (`f583c98b`)。根本原因が3つとも違ったので、それぞれ書きます。フルビルド 9,659 tests 緑。

## [P2] 仕様が範囲を定義していない

指摘のとおりです。仕様には規則の**形**（「fewer bits than that」）だけ書いて**数**を書かず、そのうえで ADR に "specification states what the run time takes" と書きました。実装2つが独立に同じ数を書いただけで、規範がどこにもない。ADR の主張が自分の変更について偽でした。

数を書かなかった理由は「JDK の都合を normative に持ち込みたくない」でしたが、これは成立しません。§primitives が `Decimal // java.math.BigDecimal` と書いている以上、scale が何でありうるかは **`Decimal` が何かによって既に決まっている**。書かなかったことでそれが抽象になったのではなく、読めなくなっただけです。むしろ書かないほうが、モデルの意味を実装から読み取らせることになる。

§stdlib-decimal に両端を書きました。

> A scale is an `Int`, which is signed 64-bit, and the scales these operations run at are the signed 32-bit numbers — `-2147483648` through `2147483647`. A scale outside those *aborts*, as an `Int` overflow does.
>
> The number is written here rather than left to whatever performs the arithmetic. […] Two things read this range, the check that says what a quotient lies between and the run time that divides, and a rule read off one of them would be a rule the other need not hold to.

`ScaleRange` の javadoc も、範囲が仕様にあることを前提にした文面に直しました。仕様に数がなければ、あれは「二つめの読み」ではなく「実装の数の写し」でしかない。

## [P2] 構造テストが negate を宣言しながら検査していない

**根本原因は fixture の欠落ではなく、テストが denylist だったこと**です。ADR が課している規則は名前ではなく性質（partial か、変換が要るか）についてのもので、名前の一覧は書いた人が思い出せたものしか禁じられません。そして #976 が起きた原因は「誰も思い出さなかった」ことでした。`negate` を一覧から外して1つ減らしても、次に `BigDecimal.sqrt`（partial）で kernel を足した人は素通りします。指摘された症状は、その形が即座に出たものです。

なので提示された二択のどちらでもなく、**allowlist に反転**しました。生成コードが `BigDecimal` に対して invoke してよいのは「値を作ること」と「値に問うこと」だけです。

```java
private static final Set<String> REPRESENTATION = Set.of(
        "<init>",       // a literal
        "signum",       // the zero test the `/` operator branches on
        "compareTo",    // the comparison operators, and a Map/Set key
        "equals", "hashCode", "toString");
```

結果として `negate` も `DecimalMath.negate` に移りました。ADR の規則は `negate` を許します（total で scale を保つ）。それでも移した理由は別で、**`BigDecimal` は自分のどの演算が total かを signature で言わない**からです。どれも checked exception を宣言しないので、「これは total か」は演算ごとに JDK を読んで答えるしかなく、その質問は `add`/`subtract`/`multiply` について既に間違って答えられていました（「Decimal does not overflow」というコメント）。だから backend には一つも残さず、質問自体をなくします。これは `BigDecimal` の性質についての判断で、二つめの一般規則ではありません — `String` のメソッドは自分が何をするか言うので、`JdkVirtual` に残ります。ADR にこの段落を足しました。

fixture に `-i.a` と比較・等値を書き、**旧実装に戻してテストが赤になることを確認しました**。宣言しただけで走らせないのが今回の指摘なので:

```
expected: <[]> but was: <[demo.Unary$Impl.apply calls BigDecimal.negate]>
```

## [P3] ADR-0053 の Status

supersede を新 ADR 側にしか書かなかったので、0053 に着いた読者から 0112 へ行く道がありませんでした。ADR-0086 / 0012 / 0036 の書き方に合わせて Status 行と README 行を直しました。本文は歴史として残しています。

`Int.divide` / `Int.truncatingRemainder` が backend に残る理由が 0053 の記述と変わった点も書きました。宣言に書けないからではなく、emit する分岐のためです。

PR #985, opened 2026-08-23
A type guarantee has one reading, and choice settlement is a second consumer of it

### LINKED ISSUE #982: A type guarantee has one reading, and choice settlement is a second consumer of it
`PathEngine.seedAt` is where what a type guarantees of a value becomes something
a path knows. #973 gives that answer a second consumer: an arm bound by a
`match` case, and a value an attempt built, are settled by what their type
guarantees, and the reading of a recipe needs it where no walk goes.

## What it produces

Nothing yet. Measured on `feature/issue-973` at `b0ab9900`, where the condition
arm of #973 is read and these two are not.

| the step, under a fold | reported | what is unread |
|---|---|---|
| `(a, x) -> if a + x.v < 100 then a + x.v else 100`, into a type capped at 100 | — | nothing is needed |
| `(a, x) -> match x.tag with \| Held as h -> a + h.amount.value \| Empty -> a` | E2011 | what `Held` guarantees of what the arm bound |
| `(a, x) -> if NonNegInt(x.v) as n then a + n.value else a` | E2011 | what the construction guarantees of what it built |

The binding itself is entered — `h` and `n` name places, not evaluations
(`Terms.choosing`). What is missing is that being one of those places says
something.

`AWalkFromASeedIsBoundedByCheckingOneStepTest.anArmReadingWhatItBoundGetsNoRange`
holds the boundary where it now stands.

## Why this is not two more cases of #973

#973 asked one question — what does choosing this arm settle? — and it is still
one question. What building it turned up is that the answer has two sources,
and they are different semantic questions:

- what a condition states, which is `Conditions`, and is in;
- what being a value of a type guarantees, which is this.

`Choice.Decides` composes them. Splitting the implementation does not split the
defect, which is why #973 stays open until both are in.

## What it takes, and what to do first

Not code. `PathEngine.seedAt` is about a hundred and sixty lines and the
traversal owns more than one thing at once: it reads a declaration's clauses,
rebases fields onto the position they govern, stops where a record holds another
of its own kind, honours `FIELDS_SEEDED` and `Reach`, reports to a `Gathering`,
collects quantified facts, and threads the `Known` it is building back into
`obligations` as it descends. So there is no line in it today where a
path-independent answer already stands.

Two ways to get it wrong, and both are ruled out before starting. Writing a
second, numeric-only traversal beside it is #935 again: the day `FIELDS_SEEDED`,
the recursion stop, or the reading of a stated clause differs between them, the
walk guarantees something the recipe does not. Reshaping `seedAt` to fit a
representation chosen in advance is how a regression gets in, because what
`obligations` is handed mid-descent is not obviously reconstructible from
outside it.

So the first work is measurement, not extraction. Classify what the hundred and
sixty lines do:

- *model semantics* — declaration traversal, field rebasing, invariant to
  relation or fact, newtype propagation;
- *reader policy* — `FIELDS_SEEDED`, `Reach`, `Gathering`, what is measured;
- *application* — accumulating into `Known`, `Held.OF_THE_VALUE`.

The question that decides everything else is whether the first group can be one
answer that no path is involved in, and where the descent stops being able to
give one. Only once that is measured is there a canonical value to name — and it
may not be `List<NumericConstraint>`. It may need predicate facts and quantified
facts beside them, or the traversal may want splitting from the interpretation
of what it finds.

What may not happen is one of the two readers coming to disagree about what a
declaration says, which is what having two of them would arrange.

kawasima:
Closes #982. Takes in the half of #973 that was left standing.

## What was measured first

`PathEngine.seedAt` threaded the `Known` it was building into every clause it read. It never read it: within `Predicates`, `k` and the read mode met at one place, `canBeDischargedFrom`, which returned true for an assumption before looking at anything. Replacing both arguments with `Known.top()` left the whole suite green, so the dependency was already vacuous.

The issue said no line of `seedAt` held a path-independent answer. That is half wrong, and the terminology needed fixing with it: what can be lifted out is not path-independent — field rebasing and field descent depend on structural position — it is independent of the *ambient* `Known`. That distinction is what the rest rests on.

`RuleRef.Invariant` was chased on the same grounds. It is the model's rule identity by its own contract, but inside `seedAt` it reached only the `Gathering`; nothing that becomes knowledge reads it, and `Derivation.Chosen.Arm` has no rule key at all. So it sits beside the answer rather than inside it.

## The shape

    model
      │
      ▼
    TypeGuarantees        at(position) → what the declaration guarantees there,
      │                                  and the positions beneath it
      ▼
    GuaranteeWalk         depth, names left out, the stop where a record holds
      │                   its own kind — and no opinion on what a stop costs
      ├──────────────► Terms.chosen → Derivation.Chosen.Arm.settles
      ▼
    PathEngine.Seeding    Gathering, assume, Held.OF_THE_VALUE, Known

`TypeGuarantees` takes a `Core` and a `Denotations` and nothing else. A scope changes which positions a reader visited, never what a declaration states at one — which is why two readers may walk to different depths and still be reading one model.

## What it fixes

The two rows the issue tabulated. An arm reading what a `match` bound, and a body reading what an attempt built, are now read against what those values' types guarantee — the same argument a seeding rests on for a parameter, since both were built through a checked constructor. Stated under the arm and not beside it: the value only exists because that arm was chosen. A departure settles nothing, having built nothing.

`AWalkFromASeedIsBoundedByCheckingOneStepTest.anArmReadingWhatItBoundGetsNoRange` held the boundary where it stood and now holds where it stands.

## Tests

`WhatATypeGuaranteesIsOneAnswerWhoeverAsksTest` asks the reading directly rather than through a consumer: what a position guarantees, a newtype at no path of its own, a record holding its own kind, a depth that is a stop and not an answer, and the two ways a scope leaves something out. Two of them were checked by mutation — breaking the newtype position rule fails two, adding a `Known` parameter fails the signature guard.

Two tests carry the contract itself. Two scopes are two walks, and where both looked they were told the same thing. And the reading's signatures mention no `Known`, `Gathering`, `PathEngine`, `Reach` or `Borne`, read off reflection rather than promised in prose.

Full reactor green: 9688 tests, 0 failures, 0 errors.

### DIFF
--- a/souther-bench/src/test/java/souther/bench/AReaderOfHowAnArmWasDecidedNamesEveryWayThereIsTest.java
+++ b/souther-bench/src/test/java/souther/bench/AReaderOfHowAnArmWasDecidedNamesEveryWayThereIsTest.java
@@ -54,7 +54,10 @@ class AReaderOfHowAnArmWasDecidedNamesEveryWayThereIsTest {
      * keep to one.
      */
     private static final Map<String, String> READERS = new LinkedHashMap<>(Map.of(
-            "souther.compiler.check.Terms#choosing", "what choosing the arm binds",
+            "souther.compiler.check.Terms#chose", "what choosing the arm binds, and the value it"
+                    + " opened. Both from one reader: a recipe wanting to say what that value's type"
+                    + " guarantees needs the second half, and working it out for itself would be a"
+                    + " second account of this node",
             "souther.compiler.check.Conditions#settledBy", "what choosing the arm states, as"
                     + " relations",
             "souther.compiler.check.InvariantChecker#opening", "what a walk over paths reads of it:"
--- a/souther-compiler/src/main/java/souther/compiler/check/Conditions.java
+++ b/souther-compiler/src/main/java/souther/compiler/check/Conditions.java
@@ -45,8 +45,9 @@ final class Conditions {
      * ({@link Derivation.Chosen.Arm}), and worth keeping straight.
      *
      * <p>A case and an attempt state nothing here. What choosing them settles is what the value's
-     * type guarantees of what the arm binds, which is a guarantee about a place rather than a
-     * relation a condition wrote, and it is read where places are seeded.
+     * type guarantees of what the arm binds, which is a declaration's answer and not a relation a
+     * condition wrote, so it is read through the one reading of a declaration there is
+     * ({@link TypeGuarantees}) and composed with this beside the arm ({@link Terms#chosen}).
      *
      * <p>An operation the library defines by cases states the relations its case is reached under,
      * which arrive already lowered to the values the call was given ({@link Choice}). Nothing about
@@ -65,8 +66,9 @@ final class Conditions {
      * and not statements the condition makes. An arm read here is read without them, so
      * {@code List.size(xs)} under {@code not List.isEmpty(xs)} comes to "not nought" here where the
      * walk has "above nought". Not two readings of one question — it is one question this reader
-     * does not ask — but the same family as the type guarantees a case and an attempt rest on, and
-     * it is answered where those are (#982).
+     * does not ask. What a case and an attempt rest on is the same family and is now asked, of the
+     * value the arm opens; this is the value a condition <em>names</em>, which no arm opened and
+     * nothing here has been given a place to read from.
      */
     static List<NumericConstraint> settledBy(Terms terms, Choice.Decides decidedBy, Denotations at) {
         List<NumericConstraint> out = new ArrayList<>();
--- a/souther-compiler/src/main/java/souther/compiler/check/Conditions.java
+++ b/souther-compiler/src/main/java/souther/compiler/check/Conditions.java
@@ -66,9 +66,10 @@ final class Conditions {
      * and not statements the condition makes. An arm read here is read without them, so
      * {@code List.size(xs)} under {@code not List.isEmpty(xs)} comes to "not nought" here where the
      * walk has "above nought". Not two readings of one question — it is one question this reader
-     * does not ask. What a case and an attempt rest on is the same family and is now asked, of the
-     * value the arm opens; this is the value a condition <em>names</em>, which no arm opened and
-     * nothing here has been given a place to read from.
+     * does not ask. It was written down as the same family as what a case and an attempt rest on,
+     * and it is not: a value built through a checked constructor carries what its declaration
+     * states ({@link TypeGuarantees}), and a list carries a length that is never negative because of
+     * what a length is, with no declaration saying so. A third source, and #988.
      */
     static List<NumericConstraint> settledBy(Terms terms, Choice.Decides decidedBy, Denotations at) {
         List<NumericConstraint> out = new ArrayList<>();
--- a/souther-compiler/src/main/java/souther/compiler/check/ContractDischarge.java
+++ b/souther-compiler/src/main/java/souther/compiler/check/ContractDischarge.java
@@ -9,6 +9,7 @@ import souther.compiler.types.TypeSymbol;
 import java.util.ArrayList;
 import java.util.LinkedHashSet;
 import java.util.List;
+import java.util.Map;
 import java.util.Set;
 
 /**
@@ -77,7 +78,10 @@ public record ContractDischarge(List<RuleDischarge> rules,
             named.add(param.binding());
         }
         named.add(rule.value());
-        Terms naming = new Terms(symbols, policy);
+        // Its own reading, over the clauses every declaration writes: a type another module
+        // declares is read off its declaration either way, and nothing here reads one at all.
+        Terms naming = new Terms(symbols, Terms.Of.THE_DISCHARGE_TREE, policy,
+                new Clauses(symbols, Map.of()));
         Denotations locations =
                 Denotations.none().locations(named, naming::placeSubject, naming::placeTerm);
 
--- a/souther-compiler/src/main/java/souther/compiler/check/GuaranteeWalk.java
+++ b/souther-compiler/src/main/java/souther/compiler/check/GuaranteeWalk.java
@@ -32,6 +32,19 @@ final class GuaranteeWalk {
     }
 
     /**
+     * How far into a value's fields a walk over a body can afford to read.
+     *
+     * <p>A type's own invariant is what its fields guarantee, and a field's type carries its own;
+     * past a couple of levels what a clause could be read against is a value the body would have had
+     * to name, and it names it by reading it.
+     *
+     * <p>A cost bound and not a rule of the model: a rule four records down refuses the outermost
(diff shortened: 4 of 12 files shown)


kawasima:
Self-check after the fact. Six things the earlier commits got working and left standing, now taken out in `4bb9270a`.

- `Terms` built a **second `Predicates`** over itself so the recipe reader could read a declaration, while `PathEngine` held its own. Two of them answer alike only for as long as neither remembers anything — which is the failure mode this branch exists to close. `Terms` owns the one now, and the engine asks it for it.
- `Terms` took a **null `Clauses`** and silently answered less when it got one. There was nothing to be null: reading every clause off its declaration is what a type another module declares is read by anyway (spec §invariant-discharge-representation), so the readings with no analysis representation pass that.
- The walk **read a declaration and then declined it** as already entered, so a record holding its own kind was read once per recursion point — into a `Terms` that memoises what it was asked. It asks the name first now, which is the order the old `seedAt` had.
- `seedAt` still took the `onPath` set the walk now keeps. Dead parameter at two call sites.
- What choosing an arm settles was **composed in a loop body**; it has a name (`Terms.settledBy`).
- `relations()` sat on `TypeGuarantee` knowing the shape of a `Predicates.Clause` — a second reader of that shape beside `assume`. It is on `Owed` now.

One test was also written badly and has been redone: `aDeclarationsOwnClausesCanBeLeftOutWithoutLeavingWhatIsUnderIt` asserted what was observed after it first failed, over a type with no rule under the one left out — so it could not have told the difference between "still read" and "nothing there". It now reads a declaration with a rule of its own and a field whose type has another.

Two things are known-and-left, not overlooked:

- `GuaranteeWalk.Reader.lostAClause` has no test. Triggering it needs a clause that types but states nothing any reader can use, and no such declaration is written in the corpus. The path is exercised in production by `InvariantChecker`, not here.
- `Terms.guaranteedBy` reads to `FIELDS_SEEDED`. That is the depth the seeding can afford, taken because it is the one the walk already had — not because the recipe reader was measured to need that depth. `Scope` is where a reader states its own, so changing it later changes nothing about what a declaration means.

Full reactor still green: 9688 tests.

kawasima:
P1 is fixed, and the root cause was one step behind where the review pointed. Rebased onto develop (which brought #974's reader check — see the third finding below). Head is `58ddbc1f`.

## The root cause of P1

`Scope` made "how far down" a mandatory `int`, and there was no value meaning *as far as the model goes*. A reader whose question has no depth in it therefore had to invent a number, and two of them already had: `Terms.guaranteedBy` borrowed `FIELDS_SEEDED`, and the construction reading in `InvariantChecker` wrote `Integer.MAX_VALUE` under a comment saying "No depth limit here". Both are spellings of "not applicable" inside a type that cannot express it. That is why an operational bound flowed into a semantic judgment — not because the wrong constant was picked, but because the API asked a question with no valid answer for that reader.

So raising the number, or switching this one call to an unbounded walk, would leave the same hole for the next reader. `Extent` is a sum now: `EveryPosition` for a question that has no depth, `AsFarAs(n)` for a reader stating what it can afford. The sentinel is gone with it.

Measured rather than argued. With the borrowed bound, the arm settlement for a record whose rule sits three positions down produced **0 relations**; with the bound lifted it produces the one the step needs. `WhatATypeGuaranteesIsOneAnswerWhoeverAsksTest.aReaderAskingForEveryPositionIsNotBoundedByWhatAnotherCanAfford` holds both halves of that.

I first wrote the regression test end-to-end, as suggested. It passes either way: with the relation dropped, the construction is still discharged by another reading, so the warning never appears. A test that cannot fail is worse than none, so it was replaced by the one above, which asserts the relation at the boundary where it is lost.

## P2

Fixed rather than deferred, because it shares a cause with the leftovers reported earlier: the extraction was done by moving code and wiring each new site with whatever it needed there, so ownership was never decided. `Terms` owns the one `TypeGuarantees` and the one `GuaranteeWalk`; `PathEngine` asks for them. `TypeGuarantees` is pure today, and its purity is not what the design rests on.

## A third thing the rebase found

#974's `AReaderOfHowAnArmWasDecidedNamesEveryWayThereIs` fired on this branch, and it was right to. `guaranteedBy` switched over `Choice.Decides` to work out which value the arm opened — a question `Terms.choosing` had already answered. That is a second account of a node with an owner, which is the failure this whole PR is about, committed inside it.

`chose` now names the ways of deciding once and answers both halves: where the reading stands, and the value the arm opened. `choosing` is the first half of that answer. The written-down reader list moves from `Terms#choosing` to `Terms#chose`, and the count stays at three.

## Correction

An earlier comment of mine said the `match` row did not depend on the arm settlement. That was wrong: the probe I read it from had not applied, and the assertion I read it from had never run — three independent facts were in one test method, and the first failure hid the rest. Measured properly, disabling the settlement puts the row back to reported. The three facts are three tests now.

Full reactor green on the rebased head: 9701 tests, 0 failures, 0 errors. The earlier CI red was `RemoveUnusedImports` on two imports only the moved code had used; the lint profile does not run in the local build, so those are now checked by hand against every declaration removed.

kawasima:
Both blockers are handled. Head is `84ae830f`.

## The third thing in scope — split out as #988, and #982's scope corrected

Not pushed into this PR, for the reason the review gives: it is a different semantic question, and answering it through `TypeGuarantees` would have that reader answering for values no declaration speaks of.

Measured before splitting, so #988 carries a reproduction rather than a note. Same fold, same value, same bound needed, and only the spelling of the guard differs:

| the step, under a fold | reported |
|---|---|
| `(a, c) -> if List.length(c.items) > 0 then a + List.length(c.items) else a` | — |
| `(a, c) -> if Bool.not(List.isEmpty(c.items)) then a + List.length(c.items) else a` | **E2011** |

The first states `length > 0` on its own. The second normalises to `length /= 0`, and "not nought" is a bound only beside "never below nought" — which the condition never said. It is what a list carries.

That also settles which side of the #981/#982 line it falls on: neither. `List` writes no clause saying a length is not negative, and an operation's answer carries what that operation declared, which is a contract rather than a data invariant. So #988 is written as a third source, and the note beside `Conditions.settledBy` now points at it instead of saying nothing answers it.

#982 has a comment moving the item out and restating that what it closes is the two rows of its table.

## The regression test

Added, and it bites. `whatAnArmBoundGuaranteesIsReadHoweverDeepTheRuleIsWritten`: a fold whose arm reads three positions under what it bound. Turning `Scope.everyPosition()` back into `Scope.asFarAs(FIELDS_SEEDED)` puts E2011 back on exactly that test and moves nothing else in the suite.

A correction to what I said earlier about this. I reported that the end-to-end test passed either way and could not be made to bite. That was measured before the rebase, and #974 changed it — on the current head the flip is clean. The review was right and my reason for dropping it was stale.

Full reactor green: 9704 tests, 0 failures, 0 errors.

Issue #997, opened 2026-08-23
A `branch` and a border point write counts beside a measurement that has no value


kawasima:
Found by a test written for #955, which asked the document to hold no number
from a measurement nobody made. The signature, the positions and the space of
combinations hold none. The arms and a border's points do.

## What is wrong

A `branch` whose measurement has no value still writes its two counts, and a
border point that was not measured still writes whether a row is at it. Both
are read as what a measure found, and no measure found them.

Measured on `0c7dc385`, `souther examples --format json` over a model with an
injected behavior and a `guard`, at `witness`:

```json
"branch" : {
  "status" : "unavailable",
  "reason" : "no_body",
  "arms" : 0,
  "covered" : 0,
  "unsettledDecisions" : [ ],
  "unreached" : [ ]
}
```

```json
{
  "point" : "on",
  "relation" : "=",
  "against" : "100",
  "hit" : false,
  "knownWritable" : true,
  "status" : "unavailable",
  "reason" : "arms_not_asked"
}
```

`arms: 0` beside `no_body` says a behavior with no body has no arms, which is
true, and it says it as a count — so a consumer dividing `covered` by `arms`
gets a measurement out of a measure that does not apply. `hit: false` says no
row is at a point nothing looked at.

## Where the mechanism is

#953 gave every measure a `Measurement` with three states that carry no value,
and the writers of the signature, the axes and the pair space ask for the value
before writing anything under it — `made().ifPresent(...)`. The two writers here
do not: they write the fields off the evidence record, which answers `0` and
`false` where there is nothing to answer.

The same rule the human report already follows. It prints `branch      not
applicable (this behavior has no body)` and no numbers; the document prints the
sentence and the numbers.

## What has to be settled

- **Whether these fields become absent, and what that does to the version.**
  The schema preamble says a change that removes or renames anything raises the
  version. A field that is present today and absent tomorrow, for documents of
  the same version, is that change for the readers that meet one — even though
  every field it would drop sits beside a `status` of `unavailable` that a
  reader is already meant to check first.
- **Whether `hit` is one question or two.** A point that was measured and missed
  and a point nobody measured are `hit: false` alike, and only `status` tells
  them apart. Either the field goes where the measurement has no value, or it
  stops being a boolean.
- **What a consumer is entitled to.** `status` is in the document from the first
  version for exactly this, so one answer is that nothing changes and the fields
  stay: a reader that ignores `status` is reading a document it was told not to
[...]

Issue #996, opened 2026-08-23
A module whose every measure does not apply reports `complete` while a row of it did not come back


kawasima:
Split out of #955, where it turned up as a test whose fixture had stopped
demonstrating what it was written for.

## What is wrong

A module every measure of which does not apply reports `measurement: complete`
while its own reasons say a row of it did not come back. The two halves of the
report read the same fact from two places and only one of them finds it.

```souther
module example.gap

data Draft = { n: Int }
data Done = { n: Int }

partial let spin (n: Int): Int = spin(n)

behavior go : (request: Draft) -> Done

example go
    | (Draft { n = spin(1) }) -> Done { n = 1 }
```

The row does not come back, so nothing about it was read. Measured on
`0c7dc385`:

```
example.gap                                              measurement: complete
  go                       injected      rows 1    pending 0
    signature   not applicable (this behavior's output is not a sum)
    partition   not applicable (the rules of this behavior divide no position)
      · not derivable: request.n
    border      not applicable (the rules of this behavior draw no line)
    branch      not applicable (this behavior has no body)
    · a row of `go` did not come back, so what it covers is unknown

1 behavior: 0 implemented, 0 unimplemented, 1 injected; 0 rows waiting for a `let`.
adequacy: undetermined
```

`status = COMPLETE`, `incompleteness = [ROW_UNDECIDED go]`, `weakenedBy = []`.

## Where the mechanism is

`AdequacyReport`'s two halves ask two different things about one run.

`status()` comes from `weakenedBy`, which is the union over the module's
behaviors of what each measure went without. That is deliberate and it is what
#953 established: a raw fact becomes a weakening in the measure that lost by it,
and never in the renderer.

> What a module could not read reaches it through the measures that lost by it —
> a source none of whose rows were seen counts against every behavior the module
> has, and every measure counted over those rows says so. Read a second time from
> the module's own list, this report was giving a raw fact its meaning as a
> weakening, which is a measure's answer and not a renderer's (issue #953).

`adequacy()` does not go that way. It reads the list:

```java
boolean unread = modules.stream().anyMatch(m -> !m.incompleteness().isEmpty());
```

So the verdict knows and the status does not. What the routing through the
measures did not cover is the module where no measure applies: a
`Measurement.NotApplicable` weakens nothing — correctly, since nothing there was
going to be measured — and where every measure of every behavior is one, there
is nobody left to carry what the run went without.

## What has to be settled

- **Whether a run that went without something is a measure's fact at all.**
[...]

Issue #953, opened 2026-08-22
A partial measure does not carry what made it partial


kawasima:
`MeasurementStatus.PARTIAL` asserts that some of a measure was not made. The measure does not
have to carry the fact that makes the assertion true.

`Unavailable.check` requires a reason exactly where there is no number, so `COMPLETE` and
`PARTIAL` are held alike: both are `counted()`, and both are refused a reason. `MeasurementStatus`
says the same in its own words — a reason is for a measure with no number. `PARTIAL` is then the
one state that is weaker than complete and states so without carrying its warrant.

## What it produces

A reader wanting to know why a measure is partial recovers it from whatever that measure keeps
beside its status: a flag on the combinations, the rules nothing took in, the rows that could not
be classified, the positions past the axis limit. Each is a second representation of the same
assertion, and each is that measure's own business to know about.

`PairSpace.truncated` is the concrete one. PR #951 stopped it disagreeing with the status —
a space too many to walk is `PARTIAL`, checked where it is built — and left it as the only thing
that says why. Before that check, `AdequacyReport.statusOf` read the flag rather than the status,
which is the reconstruction #951 was about, and it was there because the status alone did not say
enough.

The human report has the same gap from the other end: `(not all of it was measured)` names no
cause, so what a reader can act on is somewhere else on the page or nowhere.

## The invariant

A measure's own answer holds both how much was measured and whatever makes that answer weaker
than complete. No parent and no renderer reconstructs that fact from the fields beside it.

## What it takes

Not settled here, and the first question is not whether `PARTIAL` gains a reason.

- **The state space.** `MeasureReason` is today the vocabulary of why there is no number. Adding a
  partial reason to it puts three questions in one type — why a measure does not apply, why it
  could not be made, and why it was made only in part. Something closer to
  `Complete(value) | Partial(value, what weakened it) | NotApplicable(reason) |
  NotMeasured(reason)` fits what the two derivations already are.
- **One or several.** A measure can be partial for more than one reason at once — a rule nothing
  took in and rows that could not be placed are both true of one axis. Whether `Partial` carries
  one fact or a set of them should follow from an inventory of the nine measures rather than from
  whichever is convenient at the first call site.
- **What the report does with it.** `(not all of it was measured)` is where this would show, and
  what it should say is a question about the document rather than about the model.

## Where it is

- `observe/MeasurementStatus`, `observe/MeasureReason`, `query/Unavailable`
[...]

PR #998, opened 2026-08-23
A measure answers for itself whether it was made, and a verdict reads only that

### LINKED ISSUE #955: A `witness` report cannot say whether the model owes a boundary or an arm, so its verdict cannot be `satisfied` or `undetermined` honestly
Split out of #945 during review of PR#952, where a first attempt at this was written and then withdrawn.

## What is wrong

`souther compile --adequacy witness` can print `adequacy: satisfied`. It measures the cases of the signature and nothing else, while both criteria a build can be held to refuse over a line no row sits on (`E1916`) and an arm no row goes through (`E1918`). So a model is called adequate against a bar for which nobody looked.

The specification already says it should not:

> the second is `satisfied` where every measure that was asked for came to an answer and none of them found a gap, `not_satisfied` where one did, and `undetermined` where a measure that could have found one **was not asked for** or could not be made.

`AdequacyReport#requiredMeasures` reads the second half of that and not the first: at a level that does not measure arms it `continue`s, so those measures are not in the list to be short of.

## Why it cannot be fixed in the verdict

The obvious repair — count them as `NOT_MEASURED` instead of dropping them — needs to know whether this model owes anything there at all. A `>->` composition has no arms of its own, and holding every model that holds one permanently open is the defect the applicability rule exists to prevent (`spec §example-adequacy`, and `aCompositionHasNoArmsOfItsOwnAndDoesNotHoldTheVerdictOpen`).

That knowledge is not in a `witness` report. `Adequacy.Findings` and the report builder both read

```java
Map<String, PartitionEvidence> partitions =
        level.measuresArms() ? db.ask(new Coverage(name)).value() : null;
Map<String, BranchEvidence> branches =
        level.measuresArms() ? db.ask(new BranchCoverage(name)).value() : null;
```

so at `witness` both are `null`, and `null` is the level's doing. A model with no border and a model whose borders were not derived look exactly alike, which is the shape the report's own rule warns about: *whether a measure applies at all is the measure's own answer, and never the shape of what came back.*

Writing the rule against `Kind` instead — "does this criterion name a kind this level cannot produce" — is what PR#952 tried, and it quantifies over the kinds some model could owe rather than the ones this one does. It makes `witness` permanently `undetermined` whatever the model, and it puts a second decision site beside `requiredMeasures` for the same question. It was reverted for that.

## Where the mechanism is

The level gate is at the call site, not inside the measure. The measures already know how to degrade:

- `BranchCoverage#compute` reads `levelOf(db).measuresArms()` itself, falls back to `CoverageSites.Plan.NONE`, and reads `Bodies.Checked` for which behaviors have a body — which is the applicability answer, and it does not need a second run.
- `ItemAssessment.Coverage.Reason.ARMS_NOT_ASKED` already exists and already maps to `MeasurementStatus.NOT_MEASURED`. The vocabulary for "this point was not decided because the arms were not asked for" is there; a `witness` run never reaches it.

So the change is to stop pre-empting the measures and let each answer for itself how much of it was made — which is what `4bfdafc9` ("A measure answers for itself whether it was made, and how much of it") established for the measures that are asked. Then `requiredMeasures` stays the one place that decides what a verdict rests on, and applicability comes back with the evidence rather than being inferred from two enums.

## What has to be settled

- **What a `witness` run pays.** The level is documented as the one that does not generate a second set of classes or re-run every row. Asking `Coverage` for the demands without the assessment is static work, but it is work an editor keystroke would now do; whether that is inside the level's promise has to be measured, not assumed. If it is not, the answer may be that a third thing is reported — what the model owes, apart from what was covered.
- **What `--adequacy witness --warnings error` refuses.** If the findings come back at `witness`, they must arrive with a status no criterion refuses over, or the level quietly gains gates it never had.
- **What the report prints.** A `witness` report showing a `border` section with every point unmeasured is more honest and also noisier than what it shows today.

## Not urgent for #945

`souther examples` measures everything, so nothing it reports or refuses depends on this. What is affected is `--adequacy witness` and the editor at that setting.

kawasima:
Closes #955.

`souther compile --adequacy witness` measured the cases of the signature and
nothing else, and a report written at that level could print
`adequacy: satisfied` beside a border line saying eight of its points were not
measured. The other direction was there too: a line an invariant drew is met by
writing the value, so the rows the compile already ran answer it — and a point
measured in full and missed was counted in `coverage items 1/2` and never named.

The cause is not `witness`. `Level` is what a build asks to have done, and two
readers were also using it to decide what a measure's silence meant: `Findings`
dropped the border and arm measures at that level, and `requiredMeasures`
skipped them. Each measure had already answered that question. This is the
model #953 established, arriving one layer late.

## What each commit does

**A level says what work to do, not what a measure came to.** `measuresArms`
reads as "nothing about the arms is available here", and the arms a body has are
read off the checked bodies whatever the level says. `runsInstrumentedRows`
names the work.

**Whether a body owes an arm is the model's answer, not the level's.**
`BranchCoverage` took `Plan.NONE` where the arms were not asked for. The plan is
`CoverageSites.of` over the checked bodies and nothing in it waits on a run, so
that bought no work and left a body that forks nowhere looking like a body
nobody measured. It now says why it owes none — two answers over three
ways to get there — before the
level is read at all. `branch 0/0` is not a measurement and is no longer printed
as one.

**A verdict rests on what the measures answered.** The two gates go. A line a
fork drew comes back `arms_not_asked` and an arm `not_asked`; those hold a
verdict open and name no gap. A line a bound or a clause drew is measured
wherever the rows ran, and is a gap wherever it is one.

**Asking for nothing is an answer a measure gives.** `off` was a word its
callers acted on and its measures did not. It is now a state of each measure
that reads the rows and of each of their parts — `SignatureEvidence.of` is the
union of what its parts went without, and a measure nobody made went without
nothing, so a signature assembled that way over unmeasured parts came out
complete. With that, nothing outside the measures reads the level, and
`AdequacyReport` holds the bar it is read against rather than what was asked
for.

## What changes for a build

- `souther compile --adequacy witness --warnings error` refuses `E1916` over a
  line that needed no instrumentation. It does not refuse `E1918`, and it does
  not refuse a line a fork drew.
- A report at `witness` marks the points it measured and missed, which it was
  already counting.
- A behavior whose body forks nowhere reports `not applicable (this body owes no
  arm)` instead of `branch 0/0`, at every level.
- `souther examples` is unaffected: it measures everything.

## What `witness` pays

Measured over the bench corpus (`crm`, best of five in-process runs):

| | `off` | `witness` | `all` |
|---|---|---|---|
| develop `1340524c` | 934ms | 924ms | 17403ms |
| asking for the findings at `witness` | 915ms | 17232ms | 17236ms |
| this branch | 915ms | 1147ms | 17236ms |

`Boundaries` was the whole of the 18×, and it is not pathological: 380 owed
points, one decoder run each to compose a candidate. Composing a value is not
reading what the rows already established, so it is not what the level promises
— `Level.buildsValues()` is its own question beside `runsInstrumentedRows()`,
and `witness` composes none.

Nothing a search does can take a proof away, so an edge inside what every rule
reaching it leaves stays writable because the rules say so, and the row owed
there is owed at both levels. What a composed value adds is a witness for the
edges the rules cannot reach, and the row a person pastes — so at `witness` a
boundary offers no generated row, and the block says which of the two nothings
that is.

**Editor-visible**: at `witness` the "write the rows this behavior does not
cover" action is not offered where the only rows would have been at a boundary.
Composing them when the action is taken rather than on every compile is the
better long-term shape and is #1001.

## Schema

`adequacy-schema-5.json` gains `no_arm_obligations` and `not_asked`. Words added
to an enumerated field do not raise the version, as the preamble says: no earlier
document carried them.

## Specification

`every-border-has-a-row-against-its-line` said answering it takes running the
rows again. That is true of a line a fork drew and not of one a bound or a
clause drew, and the implementation has told them apart since the two were given
separate paths.

## Raised on the way

- #996 — a module whose every measure does not apply reports `complete` while a
  row of it did not come back. Pre-existing; found by a fixture that had stopped
  demonstrating what it was written for.
- #997 — `branch` and a border point write counts beside a measurement that has
  no value. Pre-existing; found by the `off` regression test in this branch,
  which is scoped around it.

## Verification

`mvn -o test` from the reactor root, green at head. Suite wall time 1:14 against
1:11 on develop; the corpus timings above were taken with a throwaway harness
and are not in the branch.


kawasima:
Both blockers are fixed, and measuring the level's promise turned up a third
thing that had to be fixed with them. Root causes first, then what changed.

## 1. `axis.reached()` in `requiredMeasures()`

The suggested filter is right about the symptom, and applying it alone fails
`aClassNoRowIsInIsReportedByDefaultAndRefusedUnderTheClassesBar`: that model's
only measure is its axis coverage, so under the default bar the required list is
empty and `required.isEmpty() → UNDETERMINED` fires — while the test expects
`satisfied`.

That is the cause. One list was answering two questions: whether anything was
measured, and whether everything the verdict rests on came to an answer. The
axis line went in unconditionally to keep the second working after
`branch 0/0` stopped propping up the first.

Both are now their own answer:

- `requiredMeasures()` is the measures that could find a gap **this bar refuses
  over**. Each entry names its kind, so a bar or a kind added later has to be
  decided about rather than inherited.
- No measure to be short of is not a doubt. A model the bar can refuse nothing
  about was asked what the bar asks and has nothing to answer for, so an empty
  list is `satisfied`. `undetermined` stays what the specification says it is: a
  measure that could have found a gap and was not made.
- A gap is answered before either, which is what `NOT_SATISFIED` already said it
  meant.

Held by a new test: the same model at `all` and at `off` gets the same verdict
under the default bar — the bar asks nothing of the one measure the model has,
so measuring it changes nothing — and under `classes` it is the answer.

## 2. `Level → emission`

The gate was not a leftover. Dropping it alone makes `--adequacy off` — the
default — ask `Findings`, and that derives the partitioning and the borders on
every compile. The level was defending against a request that contradicts
itself: `Asked.warningsAt(Level.OFF)` builds `warn = true`, which asks to be
told what the measures found and to make no measure.

So the request is what was wrong. `Asked` holds that nothing measured is nothing
to be warned about, and the emission reads `warn` alone. `off` still computes no
`Witnesses`, `Coverage`, `Boundaries` or `Findings` — checked directly.

`Level.reports` is now `readsRows`, and the three LSP call sites read it too.

## 3. What `witness` pays — measured, and it was not acceptable

The corpus number the review asked for, on `crm`, best of five in-process runs:

| | `off` | `witness` | `all` |
|---|---|---|---|
| develop `1340524c` | 934ms | 924ms | 17403ms |
| this branch, before | 915ms | **17232ms** | 17236ms |
| this branch, now | 915ms | 1147ms | 17236ms |

18×. `Boundaries` was the whole of it — 9.4s on `pipeline` and 7.1s on
`quoting` — and it is not pathological: 380 owed points, one decoder run each
to compose a candidate.

Composing a value is not reading what the rows already established, so it is not
what the level promises. `Level.buildsValues()` is its own question beside
`runsInstrumentedRows()`, and `witness` composes none.

The answer is what changes least. Nothing a search does can take a proof away,
so an edge inside what every rule reaching it leaves stays writable because the
rules say so, and the row owed there is owed at both levels. What a composed
value adds is a witness for edges the rules cannot reach and the row a person
pastes — so at `witness` a boundary offers no generated row, and the block says
which of the two nothings that is (`VALUES_NOT_ASKED_FOR`, told apart from
`NO_CLASSES`).

**This is the one editor-visible change**: at `witness` the "write the rows this
behavior does not cover" action offers nothing at a boundary. It offers them at
`all`. If that is the wrong side of the line, the alternative is the third thing
#955 predicted — reporting what the model owes apart from what was covered — and
that is its own change.

## 4. Stale Javadoc

`Criterion` no longer says a measure the build did not ask for is outside the
question; `requiredMeasures` no longer mentions `askedLevel`; `AdequacyStatus`
names the bar rather than "asked measures"; `Level.OFF` says what is acquired
rather than what is said.

## 5. `EVERY_ARM_PROVEN_UNREACHABLE` is gone

Writing the direct test for it showed nothing produces it: a fork a row gets to
has a way the row takes, so something stays owed, and a fork nothing gets to is
registered with no site and never reaches the list. A state nothing produces is
a word in a document no compiler writes and a branch no test can reach. What is
owed answers both, and the gate reads `owed`, not the numbered arms.

The branchless-body regression is in at verdict level, at both levels, as asked.

## Verification

`mvn -o test` green at head `1063d6bc`. Suite wall time 1:14 against 1:11 on
develop — no slow test added; the corpus timings above were taken with a
throwaway harness and are not in the branch.

kawasima:
Fixed at `cbad8b54`, and the root cause is one this PR has a name for.

## The Code Action

`GeneratedRows.of` already counted the rows it offered:

```java
Map<String, List<Offered>> offered = offered(asked);
int rows = offered.values().stream().mapToInt(List::size).sum();
```

and then returned a `String`. So the one caller that had to know whether there
was work to do read it off the shape of the rendered text — `block.isBlank()` —
and a block holding only the reason nothing was composed is not blank. That is
the same defect this whole branch is about, in the last place it had not been
taken out of: a caller inferring an answer from what came back instead of asking
what knows.

`of` returns the block and the count. The editor asks the count. Every other
caller says `.text()`, which is the half it wanted.

Writing the regression test turned up something worth recording: at `witness`
the block is **not** usually notes-only. `Filling` composes for the combinations
through its own construction, which `buildsValues` does not gate, so the offer
still holds rows for uncovered classes and arms. The notes-only block is the
narrower case where the only rows would have been at a boundary, and the test
uses a model of exactly that shape — one data out, no fork, one class the row
covers, two lines its invariant draws. At `all` the action appears; at `witness`
it does not.

## On-demand composition

Agreed that it is the better shape, and it is #1001. It is not a second search
beside the first — that is what `Boundaries` was written to stop — so it is the
question of a key parameterised by how much to compose, the way
`Output.Evaluated` is parameterised by what to record. Out of scope here.

## The two wordings

`Level.OFF` now says what is not made is a measurement of the rows, and that the
cases, classes and lines are derived as ever — which is what
`AskingForNothingIsAnAnswerTest` holds.

The PR description drops `every_arm_proven_unreachable`, carries the corpus
table (0.92s → 17.2s → 1.1s) as the reason `buildsValues` exists, and no longer
quotes the base SHA for verification.

`mvn -o test` green at `cbad8b54`.

kawasima:
Both blockers are closed at `ccd4f397`, and current `develop` is merged in —
head `be787ecb`, CI green.

## 1. `CLASSES` and `partitioned()`

Reproduced first with the fixture named, before changing anything:

```
SIMPLIFIED_DOMAIN = SATISFIED
RELIABLE_DOMAIN   = SATISFIED
CLASSES           = SATISFIED   ← the reading did not run out
```

The cause is an asymmetry I wrote. Finding a class no row is in takes two
answers — which positions there are to cover, and what the rows reached of each
— and I listed only the second. The lines beside them were already asked both
ways: `bounded()` for the derivation and each point for the reading. So a
reading that produced the axes it reached and no others left the positions it
never derived out of the verdict, and "no axis derived" came out looking like
"nothing to cover" — the same shape this branch is about, one level up from
where it was fixed.

Adding the derivation under the same bar condition puts `CLASSES` at
`UNDETERMINED` and leaves the other two bars where they were.
`NotApplicable(NothingIsDivided)` is dropped by `add()`, so a behavior the
reading proved divides nothing is not held open.

Held by `aPartitionReadingThatDidNotRunOutHoldsTheClassesBarOpen`, which asserts
both halves — the classes bar open, `reliable-domain` satisfied on the same
evidence.

## 2. The specification

Both sentences fixed. The arms rule now says the first and the third hold the
verdict open and the second does not, and says why: a row somebody writes would
have been counted and nobody counted it, against no row anybody writes giving
that behavior an arm. It also names a body that forks nowhere alongside the
composition, since that is now one of the three.

`satisfied` reads "every measure the bar it is held to rests on", and the
sentence carries the case the empty required set answers: a model the bar can
ask nothing of was asked what the bar asks and had nothing to answer for.

The axis-limit sentence gained "unless it is held to the classes", which is the
same fix as blocker 1 in the normative text.

## 3, 8, 9. The small ones

- schema: `first three` → `first two`, and the rest are named as a behavior that
  does owe arms.
- `NoArms` Javadoc: two answers over three ways to get there.
- `buildsValues()` Javadoc: says it is about measuring, and that an explicit
  generation request builds whatever answering it takes.
- PR description: `#999` is now `#1001` (the on-demand issue — thank you, the
  old number was one I wrote before the issue existed), and "three ways it owes
  no arm" matches the implementation.

## 10. develop

`286f5de5` merged. It touches none of `query/`, `report/`, `coverage/`, the
schema or the specification, and the suite is green locally and on CI at
`be787ecb`.

PR #1020, opened 2026-08-24
Read the positions a case of a sum declares

### LINKED ISSUE #1012: A behavior taking a sum is measured on one axis, because nothing under a case is a position
The reading of a behavior's input stops at a sum, so nothing under one is a position. Every field of
every case is unreachable to the partition, to the boundary and to generation, and a behavior taking
a sum is measured on the one axis the sum itself states.

`StructuralInspection.of` answers `Leaf` for `Shape.Sum`
(`souther-compiler/src/main/java/souther/compiler/inputs/StructuralInspection.java:141`), and
`InputDomain.walk` descends only into `Children` and `Inside`
(`souther-compiler/src/main/java/souther/compiler/inputs/InputDomain.java:236-255`). So a parameter
declared as a sum of records is read once, for its cases, and the walk ends there.

## What it produces

Measured against `develop` at `6056cc6e`. The same behavior twice — once with the sum at the input
position, once with the record that is one of its cases:

```
data Paging = { limit: Limit }
data GlobalQuery = { ...Paging, tag: Tag?, author: Username? }
data FeedQuery = { ...Paging, followees: List<Username> }
data ArticleQuery = GlobalQuery | FeedQuery

behavior readArticles : (query: ArticleQuery) -> Page
    ensures List.length(value.articles) <= query.limit.value
```

```
partition   axes 1   equivalence partitions 0/0
border      not measured (no line was derived at any position)
```

```
data GlobalQuery = { limit: Limit, tag: Tag?, author: Username? }

behavior readArticles : (query: GlobalQuery) -> Page
    ensures List.length(value.articles) <= query.limit.value
```

```
partition   axes 2   equivalence partitions 0/0
border      borders 1   coverage items 0/0   excluded 2
  · no OFF point is owed at query.limit = 1 (invariant Limit #1): excluded
  · no OUT point is owed at query.limit = 1 (invariant Limit #1): excluded
```

The two axes the record has are the presence of `tag` and the presence of `author`.
`Distinctions.ofType` already answers an optional position with `Presence(false)` and
`Presence(true)` (`souther-compiler/src/main/java/souther/compiler/inputs/Distinctions.java:59`), so
nothing about optional fields is missing — they are simply never reached through a sum. The border at
`query.limit` is the same story: `query.limit` is not a position, so no line can be drawn there.

## What it costs a model that has one

`realworld` in `souther-examples` declares `data ArticleQuery = GlobalQuery | FeedQuery`, and
`readArticles` takes it. The three optionals of `GlobalQuery` are what decide how the `WHERE` clause
is assembled — eight combinations of present and absent, a handful under pairwise — and
`souther examples --generate` proposes none of them, because the axes are not there to combine. It
proposes one row, for the case of the sum nothing covered:

```
partition   axes 1   equivalence partitions 2/2
```

The ten rows that hold the implementation to those combinations were written by hand. The row that
catches an `and()` turned into an `or()` is among them, and it is the row an author is least likely
to think of.

## The decision this asks to revisit

`InputDomain` states the current answer deliberately: "Where a value is built is not one of these
positions" — under a sum the declaration puts nothing, and a construction recipe naming one of its
cases puts the fields of that case there, which are the generator's paths and not this reading's
(`InputDomain.java:41-46`).

So this is not an omission to patch but a decision to reopen. What a case of a sum holds is declared,
not chosen by a recipe: `GlobalQuery` has a `tag` whether or not anything constructs one. A reading
that descends under each case would give those fields positions, keyed by the case they are under, and
the four measures would see them the way they see the fields of a record parameter.

What it would need settling on:

- how a path under a case is spelled, and how it stays distinct from the generator's construction
  paths that are spelled the same way today
- what `MAX_DEPTH` counts when a case is a step — whether entering a case is a level
- what a position under one case means for a row written at another case, which reads at neither the
  class nor the border of the first

kawasima:
Closes #1012.

The reading of a behavior's input stopped at a sum, so nothing under one was a position. Every field
of every case was out of reach of the partition, of the boundary and of generation, and a behavior
taking a sum was measured on the one axis the sum itself states.

`InputDomain` stated that answer deliberately, and the premise is what turned out to be wrong. What a
case of a sum holds is declared, not chosen by a construction recipe: a `GlobalQuery` has a `tag`
whether or not anything builds one.

## The decision

ADR-0114. A behavior input position may exist only under a refinement of another position; a
refinement narrows which values may stand at a position and does not move to another one. So
refinements are part of the typed path — `query@GlobalQuery.tag` is the `tag` of the same `query`,
read as the case it turned out to be — and the path is where the requirement of a conditional
position is read from, with no account kept beside it. A class may add a requirement by selecting a
refinement at its own position, and requirements are compatible exactly when they merge without
selecting different refinements of one position. Coverage and generation use that one merge.

Refinement steps do not consume structural depth: naming which case a value turned out to be does not
move into it, and counted as a level the same field would be measured or not according to whether an
author wrote a sum around the record it is in.

Structure and obligation stay distinct. The type says which refined continuations exist; the
obligation policy says which of them a behavior still owes; a position is descended only where both
permit it.

## What it produces

`realworld`'s `readArticles`, which is what the issue measured:

```
before   partition   axes 1   equivalence partitions 0/0
         border      not measured (no line was derived at any position)

after    partition   axes 4   equivalence partitions 8/8
         combination pairs 16 reached / 16 known reachable, 2 untried
         border      borders 8   coverage items 10/16   excluded 16
           ! no row is at the ON point readArticles/query@FeedQuery.limit = 1 (invariant Limit #1)
           ! no row is at an IN point of readArticles/query@FeedQuery.offset, 0 < query@FeedQuery.offset
           ! no row is at the ON point String.length(query@FeedQuery.followees[*]) = 40 (invariant Username #1)
```

The ten rows written by hand are now seen to close all eight equivalence partitions, and six rows the
model asks for at the `FeedQuery` side are named.

A record at a parameter and the same record as a case of a sum measure alike, position for position:

```
asRecord   axes 1   · rules not reached: q.tag
asSum      axes 2   · rules not reached: q@Direct.tag
```

## The commits

One change, five steps.

1. `TermPath.Step.Refine`, `Refinement`, `Requirements`; the plan and the generator address positions
   by path rather than by how a path is written. Nothing states a refinement yet, so every measure
   reads what it read before.
2. `StructuralInspection` says whether the position survives and how the reading goes on from it —
   nothing, the elements of a sequence, the branches of a sum, or a reason none could be reached. The
   walk descends the branches the obligations retain, and the rules read under a case are the case's
   own.
3. The row reader takes a refinement step. A row at another case takes it and stands nowhere below,
   which is the answer a row writing the empty list already gives at an element.
4. Pair space, cells and generation decide every combination by one `Requirements.merge`. A
   combination no value has is a reason of its own, `ONE_POSITION_CANNOT_BE_BOTH`, and it proves
   infeasibility.
5. A name a `match` arm binds denotes the scrutinee's position narrowed to the case the arm selects,
   so a `guard` inside an arm draws its line on a position the reading holds. ADR-0114 and the
   specification.

## Notes

- The optional stays unrefined. Nothing here is particular to sums, so taking it is supplying a
  branch rather than a position semantics of its own.
- `AConstructionPositionIsNotAnInputPositionTest` changes its premise: the two sets still part over
  depth, and no longer over how a position is named.
- `TermPath.depth()` had no callers and is gone.


kawasima:
Both blockers fixed, with a counterexample test each, and the two design-debt points closed. `af755fa4`.

## Root cause

The two blockers are one shape, and the small point about `and()` is the same shape again: **a constraint that reaches a reader through two channels, where the reader joins them by an order rather than by a rule.**

Blocker 1: a class of a sum contributes a requirement *and*, for a case that is not a record, a value. The position the class is a class of and the narrowed position are the same location — a refinement does not move — so the class fixed under the case is decided at a name `ConstructionPlan` reads second and never reaches. The earlier commit had moved record cases from "recipe" to "requirement" and left the other two kinds contributing both, which is why only record cases were exercised.

Blocker 2: two places turn constraints into an assignment, and only one went through `standing`. The counted product is a *preference*; nothing said where a preference becomes a legal assignment.

## What changed

**A class that narrows states the narrowing and does not also fix a value.** The value stands at the narrowed position and is chosen there, so one location is never decided twice under two names. This needed the plan to be able to write a unit value at all — `Partitions.representativesOf` could not, which is why the class had to. Two consequences fell out and are fixed with it: a position the caller narrowed is one it asked something of, so a list holding a case of a sum is built around its element rather than chosen whole; and a value chosen at a narrowed position is written under the names the position wore before the narrowing, which only a composed one had.

`AClassThatNarrowsStatesTheNarrowingAndNotAValueTest` — with the guard removed, the row for `x@Special = 10 < x` comes back as `Special(0)`, the parent class's representative, instead of `Special(11)`.

**Every assignment goes through the one place a preference becomes a legal assignment.** `standing` now takes the positions an assignment is *about* — one axis for a class, the narrowed axes for a cell — settles those first and chooses everything else beside them, answering with no assignment where the anchors themselves cannot be in one value. A cell that names positions no value has together is reported as that rather than as `NO_REASON_RECORDED`.

`AnAlternativeAssignmentIsAsCompatibleAsTheFirstTest` — with the routing removed, all four combinations of the group come back `ONE_POSITION_CANNOT_BE_BOTH`, over a conflict the search itself introduced. That was the `provesInfeasible` lie.

**`relativeTo` is gone** (it was already removed in `e596f299`, before the review landed). A path relative to a value is not a parameter-rooted path, so it is not a `TermPath`: `fieldKeyUnder(root)` and `stepsSpelledUnder(root)` answer the two questions asked of it.

**`toString`, `insideASequence`, `containingSequence` and `sequencesContainingIt`** read a step's meaning through an exhaustive switch now — `toString` shares `spell()`, and the other three share one `reachesInsideASequence`.

**`and()` refuses** a second narrowing at a position that already has one, rather than keeping the last. It adds a fact known to hold; whether two hold together is `merge`'s question.

## Verified on the real model

`realworld`'s `readArticles` still measures 4 axes / 8 equivalence partitions / 8 borders, and `--generate --boundaries` now writes the boundary row at `query@GlobalQuery.limit = 1` as a proper `GlobalQuery { ... }`.

kawasima:
The remaining point is fixed in `09acf810`, together with a second place the same cause had reached.

## Root cause

The guard was keyed on the **edge taken** rather than on the **state reached**. A termination argument is about states; keying it on the edge is right only where the edge decides the state. `SumCase(leaf)` decides it — the case names the type — so every model written in sums agrees. `Presence(true)` does not, and the outer and inner `Some` of an `Option<Option<Int>>` are the same narrowing at two different values, which is why supplying optional branches would have had to change the guard.

I have now written this guard twice and both times on a proxy — first the case's own name, then the narrowing. That is the part worth naming: the rule it keeps was nowhere written down, so each time only an instance of it was ("a sum reached through itself"), and each time a proxy that satisfied the instance looked sufficient.

## What changed

The invariant is written where the guard is and in the ADR:

> A descent that costs no level stops only where it returns to a value state already visited without a containment step.

and the guard keys on `branch.under()`. Resetting at every step into a value is unchanged and still reads correctly.

**The same cause had reached one more place.** `branchesOf(List<Case>)` took any distinction and computed what the branch carries only for sums — `Case.Presence` would have come back as the whole of a value, dropping the element type. So an optional supplying branches would have had to change that too, and the "supply a branch" contract was not yet real. It is exhaustive over the distinctions now and asks the shape, so opening the optional is the one arm in `StructuralInspection.of` that answers for its shape.

## On testing it

Neither change is observable in a model written in sums: a case and the type it carries decide each other, so no sum model separates the two identities. I have not pretended otherwise. What is held is the neighbouring fact — two narrowings with no step between them are both taken (`x@Wrap@A.n` under `data Wrap = Inner`, `data Outer = Wrap | C`) — which is what a guard that stopped too early would break. The invariant itself is carried in prose, where the next reader meets it.

## On the PR body's wording

You are right that "pair space, cells and generation decide every combination by one `Requirements.merge`" reads stronger than the code. The cells carry the body's own constraints as allowed-class sets and do not hold requirements a second time — which, as you say, is the correct layering. The merge is what turns a cell into a row. I would rather leave the sentence than restate it in the PR body now that it is on the record here; say the word if you would prefer it edited.

CI: `#1619` was green on `6fcff3e9`. `09acf810` is pushed and the run for it is what is left to confirm.

Issue #994, opened 2026-08-23
A boundary alternative space has one settled representation


kawasima:
A named sum and a behavior's output union are both a set of alternatives that has to cross the
boundary. What that set is, whether it travels as an enumeration or as a discriminated object, and
what tag each alternative wears are decided in five places, none of which reads the others.

#990 was one of them answering differently from the rest. It was reported as a duplicate variant
because that is where it showed; the duplicate is gone, and the five answerers are not.

## Who answers

1. `Deriver.deriveSum` builds `Hir.Discriminate` — a case list and a tag each.
2. `Deriver.deriveSum` builds `Hir.SumEncoder` — a second case list and a second tag each.
3. `CodecGen.generateEnumSumDecoder` takes no `Hir.Discriminate` at all. It asks
   `TypeOps.leafCases(sum, symbols)` for the cases and `emitFromNameHelper` writes `c.name()` as
   the tag. `ValueClassGen.generateSum` routes to it whenever `TypeOps.isUnitOnlySum` holds, so for
   an enumeration the HIR codec is not read.
4. `ValueClassGen.emitTagMethod` writes `c.name()` as the tag again, on the value class.
5. `CodecGen.generateResultUnionEncoder` reads no HIR codec either. It has its own
   `isEnumeration(members)`, writes the literal `"type"`, and takes the tag from `member.name()`.
   Its member list comes from `Backend.behaviorResultInterfaces`, which sorts
   `TypeOps.leafCases(sig.outputType(), symbols)` with `Collections.sort` rather than keeping the
   order `AtomSpace` answered in.

Every one of them dispatches on the first alternative that answers, so where two of them disagree,
no run reaches the disagreement. That is why #990 held its round trip and reported nothing.

## What makes it possible

`Hir.SumData` carries `Optional<Discriminate>` and `Optional<SumEncoder>` as two independent
answers about one thing. The states that pair can hold and the language cannot mean: a case named
twice on either side, a tag used twice, a case on one side and not the other, one tag on the decoder
and another on the encoder, and the two sides in different orders. `DataChecker.checkSum` fences
part of that — the encoder must cover every leaf, each variant must name a leaf — and the decoder
has no coverage check to match, and neither side checks for a repeat.

Nothing writes those two `Optional`s but `Deriver`. `AstBuilder` builds every `Ast.SumData` with
`Optional.empty(), Optional.empty()`, and there is no syntax that would make it do otherwise. So
[...]

PR #1023, opened 2026-08-24
Generate a union's members in the order its atoms are stated in

kawasima:
Stacked on #1022. Retarget to `develop` once that merges.

The last of #994. This is the only part that moves a generated artifact.

A union holds its members as a set and states no order, so one is put on it where the atoms are
answered: the members are taken in their name's order and each is descended. Codegen took that
answer and sorted it again by name. The two are the same order only while no member is itself a sum
— a sum comes out of the descent in the order its cases are declared in, and sorting moves it.

Nothing reported it, and nothing could: a value is one atom, the arms are disjoint, and each writes
the same tag whichever is tried first. The wire form does not turn on the order at all. What moves
is the `permits` of the sealed interface and the order its encoder dispatches in.

## Measured

Over the bench corpora nothing moves — no union there has a sum member. Over a corpus that has one,
exactly two classes move, the interface and its encoder, and the tags they write are the same set.

`AUnionsGeneratedOrderIsTheOneAtomSpaceStatesTest` reads the `permits` off the loaded class rather
than off the compiler, and putting the sort back turns it red.


kawasima:
Collapsed into #1022. The PR boundary was not what isolated the artifact-moving change — the commit is, and `git bisect` walks commits. Splitting also left #1022 carrying a compatibility sort whose comment pointed at a follow-up that had not merged; in one PR that intermediate state never reaches develop.

PR #1025, opened 2026-08-24
Read what a number carries off the operation that answered it

### LINKED ISSUE #1016: Whether an operation's result is ever negative is stated in two vocabularies
"The number this operation answers is never negative" is written twice, in two vocabularies, over
disjoint halves of the operations it is true of.

## Where each half is

`semantics/OperationFacts` declares it for two operations:

    about("Int", "abs", bounded(Rel.GE, 0)),
    about("Decimal", "abs", bounded(Rel.GE, 0)),

and `check/DischargeRules` is the only thing that reads `BoundsItsResult`.

`inputs/NumericTerm.ownBounds()` states it for the four size operations, by asking what kind of term
it is:

    return this instanceof SizeOf
            ? new NumericDomain.Bounds(Endpoint.inclusive(Count.ZERO), null) : null;

and `inputs/TypeBounds` and `inputs/ReadQuantities` are the only things that read it. The size
operations declare `CountsWhatItIsGiven` in `semantics` and no bound.

So the partition knows a size is never negative and does not know `Int.abs(x)` is; the discharge
knows `Int.abs(x)` is and gets a size's from somewhere else.

## Nothing is wrong today

The two halves are disjoint and each reader has what it needs. `Int.abs(x)` is not a `NumericTerm`
at all — a call stops the partition's reading (#1015) — so the bound it would want is never asked
for.

That is what dates the problem rather than removing it. The moment a call becomes a term — #1015 for
an operation that answers its argument, or stage 4 of #949 for one that states its result — the
partition asks `ownBounds()` of it, and `ownBounds()` answers null for anything that is not a
`SizeOf`. A guard at zero over `Int.abs(x)` would then ask for a row one below zero, which is the
thing the size half of this exists to prevent.

## Whose fact it is

`Int.abs(x) >= 0` and `List.length(xs) >= 0` are one proposition about two operations, and it is
intrinsic to them: it is true whether or not a discharge or a partition asks. #1011 settled where
such a fact lives — `souther.compiler.semantics`, from the first reader rather than the second — and
`ownBounds()` is a consumer's copy of one.

It is also written against the wrong thing. A `SizeOf` carries the operation it measures with
(`SizeOf(ValueName.Stdlib measure, TermPath path)`), so the term has the operation in hand and
answers from its own shape instead of asking about it.

## What would settle it

Declare the bound for the size operations in `semantics` beside the two `abs` already there, and
have `ownBounds()` ask the library about the operation the term names rather than testing which kind
of term it is. `Int.abs` then arrives already answered on the day it becomes a term, which is the
test of whether the ownership is right.

Worth checking while doing it whether `ownBounds()` has any business being on `NumericTerm` at all,
or whether its two readers should ask `semantics` directly.

kawasima:
Closes #1016.

"The number this operation answers is never negative" was written three times,
in three vocabularies, over disjoint halves of the operations it is true of.
The issue names two; the third is the discharge's own copy for sizes.

| where | which half |
| --- | --- |
| `semantics/OperationFacts` | `Int.abs`, `Decimal.abs` (and the two conditional ones) |
| `inputs/NumericTerm.ownBounds()` | every `SizeOf`, by asking which kind of term it is |
| `check/IntrinsicNumericFacts.ofSize` | every size again, as a constraint added to the atom |

A size call never reaches the reader that consults the declarations
(`Terms.atomOf` takes it to `sizeAtomFor` first), so the hardcoded line was the
only thing answering for it.

## Why it split

`Question.BOUNDS` asked only where an argument was a number too, on the
reasoning that a bound is stated against the arguments. `Int.abs` stands in the
same table and names none. An operation outside a question's range can be given
neither a rule nor a silence, so what was true of a size moved to whichever
reader wanted it.

## What is here

- The question is asked of every operation that answers a number. The twelve
  that come into range answer: the four measures, the three parts of a time,
  the three of a date, a count of minutes, and `Date.daysBetween` — which is
  silent, because the exact form it answers already decides where its result
  stands.
- `Int.compare` and `Decimal.compare` were declared silent under this subject
  beside the order fact. The order fact says which argument a positive answer
  names as the greater; that the answer is one of three numbers is a second
  statement, and it is now made.
- A count is at or above nought because it is a count. The row is generated
  from `CountsWhatItIsGiven` into the declarations, so the fifth such operation
  arrives with it. One way round: a result bounded at nought is not a count.
- `Date.year` and `DateTime.minutesBetween` are read off what a value can be
  rather than written down as numbers, with a test holding them to the
  carrier's own ends.
- A term that is what an operation answered names the operation, and
  `intrinsicBounds()` asks the library. The kind of term decides nothing, which
  is what makes the next such term arrive already answered.
- Two projections of one list of rows: a range of one number cannot hold
  `floorMod(x, k) < k`, so the projection onto a range drops it, and the
  discharge keeps the argument's name and states it. `ofSize` and `ofCall` go
  through one reader.
- `ResultBound` refuses `NE`, since a bound is where a result stops.

## Measured

`mvn -o test`, all ten modules, 6741 tests, no failures.

One answer moved. `guard Int.compare(hi, lo) >= 2` is now a guard nothing gets
through, so what stands behind it is not reached — which is
`AComparisonCallStatesTheOrderItDecidesTest` asserting that a construction was
checked at all. Its threshold is now `>= 1`, keeping what it is about, and the
unreachability is a test of its own. That is the only place the new bounds
change what a program is told.

The conformance snapshot is unchanged.


kawasima:
Three points addressed at `885b7c7`.

## The blocker: root cause, and what was done about it

The stale premise is in the comment, but the reader's own shape is what made it
unreachable. `Conditions.asOrderComparison` matches the *written* condition:

    Core zero = callFirst ? b.right() : b.left();
    if (!(side instanceof Core.PreservedCall call) || call.args().size() != 2
            || !isZero(zero)) {
        return e;
    }

`isZero` is a literal-`0` test on the syntax. So the reading never consulted
what is known about the call's result at all — not the declared bounds, and not
a name bound to nought either. Declaring `compare` to answer one of three
numbers created a second canonical fact that this reader is built not to see,
which is #1016's own division one level up. Adding `1` to the pattern would have
left the reader in that shape.

So the condition and the bounds are composed, and the domain is asked which way
the answer stands to nought:

    NumericDomain.<Object>top()
            .assuming(sign, ResultRange.of(call.operation(), ConstantArguments.NONE), spacing)
            .assume(answered.minus(LinearForm.constant(read.constant())), rel, spacing)

then the tightest of `EQ, GT, LT, GE, LE, NE` it entails is the relation the two
arguments are read as standing in. Three things follow rather than being added:

- `compare(hi, lo) >= 1` states `hi > lo`, and proves an invariant that a
  difference of nought does not satisfy.
- `compare(hi, lo) > -1` states `hi >= lo` — the step is the domain's, from the
  granularity of the call's result type. Nothing here says what `> -1` comes to
  over whole numbers.
- the number stood against is *read* (`terms.affineOf`), so `compare(a, b) > z`
  under `let z = 0` is the condition it looks like. That was not true before
  either.

`>= 2` still reaches nothing, now because the composition is bottom rather than
because a bound refuted the guard later.

The old test asserted its point with a program that no longer makes it. What it
was reaching for — a condition bounding how far from nought without saying which
way — is now written over a count: `Date.daysBetween(from, to) <= 5`.

## The `ResultOfOperation` doc, and the follow-up

Narrowed to what is true: the operation answers `intrinsicBounds()`, and
`carrierAt` and `read` are still written to `SizeOf`. Filed #1027, which covers
both — including that `read` is the one that would be wrong quietly, since a
term for `Int.abs(x)` at the path of `x` would answer with the observed `x`
rather than with its absolute value.

## The untested branches

`Decimal.toInt` is the case the `floorMod` test could not make: its rows hold
under `Always`, so what decides is only whether the argument they stand against
can be named. Under `NONE` it is `OPEN`; with argument 1 known it is
`(2, 4)` — which pins the unresolved widening and both strict ends at once.
`ResultBound` refusing `NE` now has an `assertThrows` of its own.

`mvn -o test`, all ten modules, no failures. The conformance snapshot is
unchanged.

Issue #1062, opened 2026-08-25
A border from a data's invariant is owed at every position of every behavior that carries the data


kawasima:
Found while re-measuring every module of souther-examples. Measured on `3fb34a7d6`, over
souther-examples `fc68e53`.

## What is wrong

A border derived from a data's `invariant` is owed at every position of every behavior that carries
that data, so one rule written once is asked for as many times as the model mentions the type.

Over the fourteen modules, `souther examples` marks 1504 items `!`. 1142 of them are border items,
and 969 of those name an `invariant` — from **75 distinct invariants**. In `crm` alone, 569 of the
719 `!` items are invariant-derived borders, and they come from 25 invariants; `invariant UserId #1`
is named 126 times, at 74 distinct behaviors:

```
! no row is at the ON point scheduleMeeting/String.length(draft.owner) = 1 (invariant UserId #1)
! no row is at the ON point touchCount/String.length(activities[*]@CallTask.owner) = 1 (invariant UserId #1)
! no row is at the ON point lastActivityOn/String.length(activities[*]@CallTask.owner) = 1 (invariant UserId #1)
! no row is at the ON point lastActivityOn/String.length(activities[*]@EmailTask.owner) = 1 (invariant UserId #1)
```

`UserId` says one thing: a user id is a string of one character or more. Whether the compiler
believes a row standing at length 1 is a question about `UserId`, and the answer cannot differ
between `scheduleMeeting` and `touchCount` — nothing in either behavior says anything about the
length of a user id. Discharging what the report asks for means writing 126 rows that each stand at
the same point, for a rule that was written once.

## What this costs

`--strict` is what makes the report fail a build, and no module here can turn it on. Two thirds of
what it would refuse over is this: 969 of 1504. If an invariant border were owed once where the
invariant is written, `crm`'s strict set would be 25 items and not 569 — a number somebody can work
through.

It also crowds out the rest. The 277 unreached `match` arms and the 85 unused sum cases are what a
reader wants out of the report, and they are 24% of it.

## The distinction that seems to be missing

A border comes from one of two places, and the report already knows which — it names either an
`invariant` or a comparison in the behavior:

- **From the data's `invariant`.** What it says is true of the type wherever the type is used.
  A row anywhere in the module standing at that point is evidence about the type. 969 items.
- **From the behavior's own rules** — a `guard`, an `if`, a comparison. What it says is about this
  behavior at this position, and is owed per behavior, as it is now. 173 items.

[...]

PR #1086, opened 2026-08-25
Offer an authored line one row, found by searching its readings

### LINKED ISSUE #1076: Generation is asked once per reading of a line, and a line is owed once
The second half of the split in #1062, named there as a non-goal.

A border's identity now separates the line the author wrote from the reading that met it, and everything that measures, refuses or reports reads the debt. Generation does not: a row is composed by walking one behavior's inputs, so the search is still asked per reading.

What #1074 put in is a projection — a request answers about a line from the readings *that request searched*, and says so where none of them composed a row:

```
NO_SEARCH_IS_ASKED_FOR_A_LINE_ACROSS_ITS_READINGS
```

## What is left

- A search per obligation rather than per reading, so one line is searched once however many positions carry the type.
- Retry across the readings. A row at the line may be composable at one position and not at another — a record holding a field to four characters or more cannot hold a row at length 1, and a plain field can — so a line whose first reading composes nothing is not a line nothing composes a row for.
- At most one row per obligation. Today each reading composes its own, so one authored line is offered as many rows as it has positions.
- Remove the occurrence-level generation scaffolding once nothing reads it.

## Where the seam shows now

`GeneratedRows` prints the rows a behavior's own search composed and then, for the lines the module's declarations are owed, either the row that search found or the sentence above. An author reading a block for a line owed at four positions is offered up to four rows for one thing to write.

`Analyzer` offers "write the rows this behavior does not cover" from the findings — cheap, because an editor asks on every cursor move — and resolves it by generating. The two can disagree: an offer stands where nothing composes a row. That asymmetry predates #1062 and is not a defect of it, but a cheap per-obligation predicate would narrow it, and this is where one becomes available.

## Not this

Which module owes a line another module's declaration wrote is a separate question.

kawasima:
Closes #1076. The second half of the split in #1062.

A line an `invariant` drew is owed once over every behavior carrying the type, and a row at it is composed by walking one behavior's inputs. Generation was still asked per reading, so one authored line was handed to an author as up to four things to write, and a request narrowed to one behavior could only say `NO_SEARCH_IS_ASKED_FOR_A_LINE_ACROSS_ITS_READINGS`.

## What is in it

- **A search per obligation.** `generatedForDeclarationsOf` walks the readings in the order the module declares them and stops at the first that composed a row. The work costs `BoundarySearch`, which is keyed — a reading is searched once however many lines are resolved from it, and a request about one behavior spends nothing on the rest.
- **Retry across the readings.** A reading that composes nothing is not the line composing nothing: a record narrowing the position may refuse the value the line names where a plain field takes it.
- **At most one row per point of a line.** Settled where the search is resolved, not where the block is laid out. `DeclaredRows` holds one answer per point, and the rows, the notes and the dispositions are projections of that one map.
- **The occurrence-level scaffolding is gone**, with the reason it produced.

## Three concepts, kept apart

What a reading knows is `BoundarySearch(module, behavior)` and stays per reading — a per-behavior answer that depended on other behaviors would be an answer that moves with the order the requests arrived in. What a line concludes is `DeclarationResolution`. Where a row goes is the provenance of the reading that composed it: a behavior carrying the type is a reading in whose terms the row can be written, never the owner of the line.

`GenerationScope` says which readings a request walks, and so which lines it is about — read off what a database happened to hold, the walk would answer differently depending on who asked first.

## A reading is a position, not a behavior

One behavior can carry the type at two positions, and what a search of one of them came to is a fact about that position. So a reading is `Reading(behavior, position)`, the debt keeps its readings as such and derives `carriedBy` from them, and two readings one behavior cannot tell apart are refused where they are built.

## Success and failure are asymmetric

A row is an existence claim and one reading settles it. That no row can be written is a claim about every reading, and `SearchCoverage` is total over the readings a line has — in their order — so the difference between "walked to the end" and "stopped at the first" is a state rather than an absence. The claim a reader may act on (ADR-0091) is released only where every reading was walked and every one of them proves it, asked of `Reason.provesInfeasible` and never of which scope the request was.

## The conclusion is carried, not re-derived

`DeclaredRows.unmet` passes that gate once and hands over one of three shapes — the line cannot be written and every reading that proves it; what the readings came to, each with the reading it came from; or nothing was searched. The block switches on the shape, so it cannot promote a reading's reason to a claim about the line and cannot leave a reading anonymous. `dispositions` reads the same three shapes, so a finding's answer and the block's sentences cannot come apart, and it reads the lines this request was about off what it answered rather than asking the scope a second time.

## Not this

The Analyzer's offer stays a conservative over-approximation of what resolve will compose. The issue suggested a cheap per-obligation predicate would narrow it; I looked and did not find one that is sound, compose-free and monotone under a strategy being added later — what this change produced is a resolver that has to compose across several readings to know, which is the opposite of cheap. Duplicating the composer's semantics on the cheap path would be worse than the asymmetry. Left as a follow-up, and stated here so the issue's hypothesis is not read later as an unfinished requirement.

Which module owes a line another module's declaration wrote is still separate.

## Checks

`mvn clean verify` is green, and the branch is merged up to `develop`. The resolver is a pure function and its conclusions are held by a test with no compilation in it. Mutations run and killed: reversing the declaration order, stopping at the first reading, ignoring the scope, releasing the terminal claim without walking every reading, collapsing two readings of one behavior into one, skipping the gate on the way to the block, and comparing a coverage's readings as a set rather than in order. Dropping the `owedHere` filter — the duplicate this issue is about — is killed by six tests.

The conformance snapshot loses one row: `台帳へ書く` and `出荷する` were each offered the same row for one declared line.

## Review rounds

Two rounds of review found four things, all of one root: what a search of one reading came to and what the line concludes from all of them are different kinds, and only the place that collects the evidence had a shape saying so. Every boundary the evidence crossed turned the distinction back into a convention. Fixed by giving each boundary the shape: the coverage's universe is the readings, the answer carries the conclusion, and the projection reads the domain off what was answered.

A note on this body: an earlier edit of mine overwrote it with another PR's text — a fixed temp path that already held a stale file, written without reading the target first. This is the original, brought forward.


kawasima:
Thanks — both P1s and the P2 are addressed, plus the constructor nit. One note on what the review was reading: it is based on `2bbfa4a`, and `c4c99dd` (a self-check before this) had already taken out the `first`-representative collapse. So the state now is:

## The root the three share

What a search of one reading came to, and what the line concludes from all of them, are different kinds. Only the place that *collects* the evidence — `SearchCoverage` — was given a shape that says so. Everywhere the evidence left that shape, the distinction became a rule each reader had to remember, and none of them did. That explains all three findings, and it is why fixing them one at a time would have left the fourth to be found later.

## [P1] occurrence collapse — fixed in `c4c99dd`, test added here

`readingOf` no longer folds. A reading is `Reading(behavior, position)`, the debt keeps its readings as such (`carriedBy` is derived from them rather than carried beside them), the walk asks each one and matches it back by the line itself, and two readings one behavior cannot tell apart are refused where they are built. Your requested test is in as `twoReadingsInOneBehaviorAreTwoReadings`: one behavior, two occurrences, first `THE_RULES_LEAVE_NOTHING_THERE` and second `ALL_CANDIDATES_REJECTED`, `provesTheLineCannotBeWritten() == false`. Collapsing back to a behavior-keyed universe is killed by it and by `oneBehaviorCarryingTheTypeTwiceIsTwoReadings`.

## [P1] the gate was not on the output path — fixed here

This was the live one, exactly as described. `cannot()` unwrapped the coverage into a flat list, so the gate could never be applied downstream, and `Note` carried the declaration's name beside a reading's reason.

The answer now carries the conclusion instead of the evidence to conclude from. `DeclaredRows.unmet()` passes the gate once and returns one of three shapes — `TheLineCannotBeWritten(owedBy, said, proving)`, `WhatTheReadingsCameTo(owedBy, said, came)` where each `At` carries its `Reading`, `NothingWasSearched`. The block switches on the shape, so it cannot promote a reading's reason (the arm carrying a claim about the line only exists where the walk earned one) and cannot leave a reading anonymous. `dispositions()` reads the same three shapes, so a finding's answer and the block's sentences cannot come apart.

The wording follows: a reading's reason is said `in \`held\``, like every other note in the block; the declaration's name is reserved for the licensed sentence. That returned `anEdgeNoSearchReachedIsStillSaidInTheBlock` to the sentence it held before this branch — good evidence the split is right, since my earlier change of that line to `owed by` *was* the promotion.

## [P2] scope re-judged in the projection — fixed

`dispositions()` reads the lines this request was about off what it answered (`resolved.keySet()`), not off `scope::admits`. No mutation is claimed for this one: today both spellings give the same answer, and the point was removing the second authority rather than fixing a wrong result.

## Nit — fixed

`SearchCoverage` compares its readings in order, not as a set, which also refuses a repeated reading. Killed by `aCoverageOutOfTheReadingsOrderIsRefused`.

## One thing I did beyond the list

The declaration notes were appended by the caller after the block was built, so `of(module, filling, declared, …)` claimed to be "the whole block for one module" and was not. Moved inside, so a caller rendering through it cannot print rows with nothing said about the work beside them.

`mvn clean verify` is green, and the branch is merged up to `develop` (`03941315`).

Issue #1099, opened 2026-08-26
A field every case of a sum spreads is owed once per case


kawasima:
Found while working #1096. Measured on `feature/issue-1096`.

`Shape.Sum` now carries what its cases share — the declarations every case
spreads and the fields they contribute — and the invariant reading states those
of a value standing at the sum. The input-coverage reading does not ask for that
part, and this is about whether it should.

## What the coverage reading does today

A field every case spreads is a position under each case, not under the sum. From
`APositionUnderACaseIsAPositionTest`, where `GlobalQuery` and `FeedQuery` both
spread a declaration carrying `limit`:

```
query
query@GlobalQuery
query@GlobalQuery.limit
query@GlobalQuery.tag
query@GlobalQuery.tag@Some
query@GlobalQuery.author
query@GlobalQuery.author@Some
query@FeedQuery
query@FeedQuery.limit
query@FeedQuery.followees
query@FeedQuery.followees[*]
```

`limit` is one field of one shared declaration and it stands at two positions. A
row is owed each of them, and a sum with four cases owes four. What is asked at
each is the same question about the same declaration.

## Why it is not simply a bug

The two readings differ for a reason. The invariant reading has no step that
opens a case: nothing opens one at a construction site, so what the cases share
is out of reach unless the sum states it — which is what #1096 was. The coverage
reading walks into every case, so it is already there. The same exclusive shape
(`StructuralInspection` is `Decomposed | Retained`) is a defect in the first and
is not obviously one in the second.

So nothing here is wrong today. What is open is whether owing the shared part per
case is the reading we want, and the representation cannot express the
alternative.

## What a change would have to settle

A spike that gave a shared-spread sum children turned
`APositionUnderACaseIsAPositionTest` red on all three of its position lists, and
would need answers to:

- **Where is a shared field owed?** Once at the sum (`query.limit`), or once per
  case as now. Both cannot stand: one field would have two names and a row would
  be asked for both.
- **`Partitions.axisOf` refuses a position that both divides and is made of
  positions.** A sum divides into its cases and would now have children. The
  guard says this is unreachable "today" and throws where it is not; a sum with a
  shared spread makes it reachable, so it has to become a decision rather than an
  assertion.
- **Does `--strict` refuse different models afterwards?** Moving where a field is
  owed changes which rows a model has to write, so this is a change to what a
  build says and not only to how positions are named.

`StructuralInspection` would become a product of the two facts — what the
position is made of, and how the reading goes on from it — the way
[...]

PR #1100, opened 2026-08-26
Read what a sum's cases share where the sum itself stands

### LINKED ISSUE #1096: What an invariant guarantees is carried through a record and lost through a sum
Found while working souther-examples' `businesstrip` through its report. Measured on `5e771a839`.

## What is wrong

What a data's `invariant` guarantees is carried through a record and lost through a sum. The same
value, reached the same way, is established in one shape and not in the other — so a model that
holds its amounts in a sum of expense categories gets `E2011` on every build, for a clause its own
types already answer.

```souther
module probe.both

data 金額 = Int
    invariant value >= 0

data 費目共通 = { 金額: 金額 }
data 交通費 = { ...費目共通 }
data 宿泊費 = { ...費目共通 }
data 費目 = 交通費 | 宿泊費

data 直接持つ明細 = { 金額: 金額 }
data 直和で持つ明細 = { 費目: 費目 }

let 合計A (明細: List<直接持つ明細>): Int =
    List.sum(List.map(一件 -> 一件.金額.value, 明細))

let 合計B (明細: List<直和で持つ明細>): Int =
    List.sum(List.map(一件 -> 一件.費目.金額.value, 明細))

behavior レコードから : (明細: List<直接持つ明細>) -> 金額
    constructs 金額
let レコードから (明細) = 金額(合計A(明細))

behavior 直和から : (明細: List<直和で持つ明細>) -> 金額
    constructs 金額
let 直和から (明細) = 金額(合計B(明細))
```

```
-- INVARIANT (WARNING)  E2011 ----------------both.sou:26:17

26| let 直和から (明細) = 金額(合計B(明細))
                          ^^

4|     invariant value >= 0

Constructing `金額` here may violate its invariant: nothing known here establishes it.
```

`レコードから` is silent. Every amount summed by `合計B` is a `金額`, so every one of them is `>= 0`
and so is their sum; the only difference from `合計A` is the `費目` between the line and its amount.

## The same through a spread

```souther
module probe.spread2

data 共通 =
    { 開始: Date
    , 終了: Date
    }
    invariant Date.daysBetween(開始, 終了) >= 0

data A = { ...共通 }
data B = { ...共通, 印: Date }
data 二択 = A | B
data C = { ...共通 }

behavior 移す : (x: 二択) -> C
    constructs C

let 移す (x) = C { ...x }
```

`C { ...x }` warns. Written with `x: A` instead of the sum, the same line is silent — and every case
of `二択` spreads the same `共通`, so the two dates in hand came out of a value that already answered
this clause.

## Why the hint cannot be followed

E2011 offers three ways out: guard the construction, attempt it with `... as name else ...`, or
declare the relation as an `invariant` on the input. None applies here. There is no input for which
the construction fails, so a guard is a branch nothing takes and an `else` case is an outcome the
domain does not have; and the relation *is* already an invariant on the input — that is what
`金額` and `共通` say. The model has nothing left to write, so the warning stays on every build.

## Where it was found

`businesstrip` reports four of these, and all four read through a sum:

- `金額(明細の合計(明細))` — the amounts are behind `費目 = 交通費 | 宿泊費 | 交際費 | その他費用`
- `精算額(精算する額の合計(実費用))`, twice — the same, with a `match` on `費用負担区分` deciding
  which lines count
- `申請準備中 { ...却下 }` — `却下` is `却下済み = 事前承認却下済み | 最終承認却下済み`, and both
  cases carry `出張申請共通項目`, which is where the date invariant is written

The module is otherwise at zero gaps and `--strict` exits zero, so these four lines are the whole of
what its build says.

kawasima:
Closes #1096.

## What was wrong

Four readers worked out what a value's type guarantees, each from the
declaration, and they disagreed about a sum.

- `Elaborator.elaborateFieldAccess` let a field every case spreads be read off
  the sum, through `TypeOps.commonSpreadFields`.
- `ValueClassGen` declared an accessor for it on the generated interface.
- `TypeGuarantees.at` answered that a choice between declarations states nothing
  and has nothing under it, so the invariant reading could neither state what the
  cases share nor walk into it.
- `TypeGuarantees.anyRuleUnder` descended into the cases and answered that a sum
  *does* have rules under it — a third answer, inside the class that gave the
  second.

Two more readers had the same gap for the same reason: the naming of subjects
under a parameter (`InvariantChecker.name`) and what holds of every element of a
container (`UniversalElementFacts.guaranteed`) both stopped at anything that was
not a `Hir.Data`.

Measured with the axis moved one element at a time: depth, a helper, and
`List.sum` all made no difference; the sum did.

| shape | before |
|---|---|
| `x: A` (record), `Amount(x.v.value)` | silent |
| `x: Expense` (sum), `Amount(x.v.value)` | E2011 |
| three record hops | silent |
| through `List.sum(List.map(...))`, record | silent |
| `C { ...x }`, `x: A` | silent |
| `C { ...x }`, `x: Choice` | E2011 |

## What this does

**`Shape.Sum` carries what its cases share.** `CommonProduct` is `None` or
`Shared(origins, fields)`. The origins are kept because the sharing is nominal
and the clauses are written on those declarations: held as fields alone, a reader
wanting the rules over them has only the names to find the declaration back from.
`Shape.Cases` — an unnamed union — has no common part, since a spread is shared
by being written and nothing writes those together.

**`PositionReading` is the one step off that shape.** It answers which
declarations state something of every value at a position, what is readable on
such a value, whether those are at paths of their own, and what stands below that
no reading there takes in. `TypeGuarantees`, `InvariantChecker.name`,
`UniversalElementFacts.guaranteed` and `Terms.fieldType` ask it. None of them
resolves a name to a declaration any more.

**`TypeGuarantees.At` is a product, not a choice.** A sum whose cases share a
spread states rules at the position *and* leaves its cases to the reading a match
opens. `Declared | Undeclared` could hold one of those, and an arm for the pair
would be the same shape one case later. So the answer is four independent
components:

| | |
|---|---|
| `here` | what the rules written here guarantee of this value |
| `beneath` | the positions under this one |
| `handedOn` | what a reading opened elsewhere answers for |
| `coverage` | `Complete`, or `Incomplete` naming the rules that could not be read |

`coverage` replaces the `everyClauseStated` boolean and names the rules rather
than counting them, so a reader is not left working out from a count which of
"handed on" and "lost" had happened. The two are different in kind: a rule handed
on is somebody's, and a clause lost here is nobody's and stands at run time.

**Handing on is no longer a way of stopping.** `GuaranteeWalk.Stop` is now only
what the walk itself did, and `PathEngine.leftBy` answers an
`InvariantChecker.Borne` rather than choosing between a handoff and a loss.
`NOTHING_DECLARED` and `NO_VALUE_THERE` are gone.

**Only the difference is handed on.** A case carries the shared clauses as well
as its own. Passed whole, one rule of the model would be owed twice — once at the
sum where it was read, once by whoever opens the case — and nothing below could
settle the first.

**`StructuralDescent` moves to `check`,** beside `Shape`. What is under a type is
a fact about a shape, and the invariant reading is its third reader.
`TheReadingAndThePlanTakeOneStepDownATypeTest` is what found that, by going red
on the new `product.fields()`.

## What is fixed

All four of the `businesstrip` reports the issue names, plus the container case
the measurement turned up (`List<Expense>` where the element type is the sum).
A sum whose cases share no spread still warns, which is right: the sharing is
nominal, and cases that merely declare a field of one name have not shared it.

## Tests

`ASumStatesWhatItsCasesShareAndLeavesTheCasesToAMatchTest` fixes each combination
of the two axes, and the row that matters is a sum whose cases share a spread and
one of which writes a rule of its own. Two mutations were run against it:

- stop reading what the cases share → 3 assertions red, including
  `states("capped")` inside that row
- never answer `ToAnotherReading` → 3 different assertions red, including
  `handedOn` inside that same row

Neither failure implies the other, which is the property the product exists for.

`WhatEveryCaseOfASumSpreadsEstablishesAConstructionTest` fixes the reported
symptom end to end, including the case that still warns.

`WhatEachWayOfStoppingLeavesIsWrittenDownOnceTest` no longer asserts that exactly
one way of stopping hands the rules on. That count is javac's now: nothing a stop
can answer says anybody else took them.

**Not covered:** the `coverage` axis has no mutation test. At a read position
every field is given, so `statedAt` answers null only when typing the clause
fails, and there is no source that makes that happen — so what is missing is a
witness, not a test.

## Leaving a rule out asks who wrote it

A reading is asked to leave a declaration's rules out so that the end which moves
says which declaration was holding it. What is left out is a property of each
rule, and both readers were asking it of the value the reading happened to be
opened at. The two are the same name only where a declaration writes its own
clauses and spreads nothing, so a rule a spread carried in was outside what either
reader could leave out — and that predates this branch:

```souther
data Common =
    { lo: Int
    , hi: Int
    }
    invariant based = lo >= 100
    invariant far = hi >= lo + 10

data Held = { ...Common }
    invariant near = hi >= lo + 5
```

`hi` stops at 110 and `far` puts it there, but `narrowedBy("hi", true)` answered
`[Held]` — the wrong declaration, because taking `Held` away took `Common`'s rule
with it. `WhichRuleHoldsAnEndIsAskedOfTheRuleAndNotOfWhereItIsReadTest` was red
on that before the fix.

`RulesLeftOut` is the question. A caller names declarations; a consumer can only
ask `excludes(RuleRef.Invariant)`. Held as a `Predicate`, a caller could hand one
over a position's name and it would compile — a method reference to `equals` takes
an `Object` and satisfies a predicate of any argument — and one of the tests here
was doing exactly that. So the two identities are separated by type:

| | |
|---|---|
| `stopAt`, `entered` | position identity, `TypeSymbol` |
| `withoutClauses` | rule provenance, `RulesLeftOut` |

Both readings ask it per rule now, `seedFields` included: whether the position is
read at all is asked once of the name, whether a rule is left out is asked of each
rule, and a reading short of one records a single `missed` against the position.
`Coverage.Incomplete.lost` is filtered by the same value.

## Scope of the tripwire

`TheInvariantReadingAsksWhatAPositionIsRatherThanDecidingTest` holds five files —
`TypeGuarantees`, `GuaranteeWalk`, `PathEngine`, `UniversalElementFacts`,
`StepInputFacts` — and the one place a shared part is derived. It does not cover
`InvariantChecker.name` or `Terms.fieldType`, which were two of the sites this
change moved onto the one reading: those files access declarations for other
reasons, and a check over the file would have to allow what it is meant to refuse.
Those two are held by the tests of what they answer, not by the tripwire.

`mvn -o test` from the reactor root: 10,376 tests, all green.

## Left out

The same exclusivity is in `StructuralInspection` (`Decomposed | Retained`) on
the input-coverage side, and it is deliberately not changed here. A spike that
gave a shared-spread sum children turned
`APositionUnderACaseIsAPositionTest` red, and its expectations say why: the
coverage reading already reaches those fields, once under each case
(`query@GlobalQuery.limit`, `query@FeedQuery.limit`). Decomposing the sum
position as well would give one field a second name and owe a row both.

The two readings differ for a reason. The invariant reading has no step that
opens a case — nothing opens one at a construction site — so the shared part is
out of reach unless the sum states it. The coverage reading walks into every
case, so it is already there.

Filed separately as #1099.

### DIFF
--- a/souther-compiler/src/main/java/souther/compiler/check/Clauses.java
+++ b/souther-compiler/src/main/java/souther/compiler/check/Clauses.java
@@ -141,15 +141,16 @@ final class Clauses {
      */
     StatedClauses statedAt(TypeSymbol.AtModule named, Hir.Data data, Map<BindingId, Core> given) {
         List<Stated> stated = new ArrayList<>();
-        int declared = 0;
+        List<RuleRef.Invariant> lost = new ArrayList<>();
         for (TypeOps.Declared inv : declared(named, data)) {
-            declared++;
             Core one = statedAt(inv.clause().expr(), named, data, given);
             if (one != null) {
                 stated.add(new Stated(Clause.of(inv), one));
+            } else {
+                lost.add(new RuleRef.Invariant(Clause.of(inv).ref()));
             }
         }
-        return new StatedClauses(List.copyOf(stated), stated.size() == declared);
+        return new StatedClauses(List.copyOf(stated), List.copyOf(lost));
     }
 
     /**
@@ -161,15 +162,26 @@ final class Clauses {
      * two are the same list with different things missing from it. Said here, where the dropping
      * happens, rather than counted again by whoever needs to know.
      *
-     * @param everyClauseStated whether every clause the declaration writes is in {@code clauses}.
-     *                          A caller that asked for none of them was not asked to lose any and
-     *                          is told so
+     * @param lost which clauses the declaration writes are not in {@code clauses}, named rather
+     *             than counted. A caller told only that something was lost has to find out what it
+     *             was about by reading the declaration again, and a reader that answers for the
+     *             rules it was handed would otherwise answer for a rule it never saw
      */
-    record StatedClauses(List<Stated> clauses, boolean everyClauseStated) {
+    record StatedClauses(List<Stated> clauses, List<RuleRef.Invariant> lost) {
+
+        public StatedClauses {
+            clauses = List.copyOf(clauses);
+            lost = List.copyOf(lost);
+        }
 
         /** What a reading told to leave a declaration's clauses out gets: none of them, and nothing
          * lost. */
-        static final StatedClauses NONE_ASKED_FOR = new StatedClauses(List.of(), true);
+        static final StatedClauses NONE_ASKED_FOR = new StatedClauses(List.of(), List.of());
+
+        /** Whether every clause the declaration writes is in {@link #clauses}. */
+        boolean everyClauseStated() {
+            return lost.isEmpty();
+        }
     }
 
     /**
--- a/souther-compiler/src/main/java/souther/compiler/check/DataChecker.java
+++ b/souther-compiler/src/main/java/souther/compiler/check/DataChecker.java
@@ -879,13 +879,13 @@ public final class DataChecker {
      * can see in every case. */
     private static Map<String, Type> spreadOfSum(String name, Hir.SumData sum, Type bound,
                                                  SourcePos pos, CheckContext ctx) {
-        Map<String, Type> shared = TypeOps.commonSpreadFields(sum, ctx.symbols());
-        if (shared.isEmpty()) {
+        if (!(TypeView.of(Type.ref(sum.declares()), ctx.symbols()).shape() instanceof Shape.Sum shape)
+                || !(shape.common() instanceof Shape.CommonProduct.Shared shared)) {
             throw CompileException.of(Diagnostic.at(pos)
                     .say(new DataMessage.SpreadOfASumWhoseCasesShareNothing(name, Type.show(bound)))
                     .build());
         }
-        return shared;
+        return shared.fields();
     }
 
     private static void checkEncoder(Hir.EncoderDef enc, CheckContext ctx) {
--- a/souther-compiler/src/main/java/souther/compiler/check/Elaborator.java
+++ b/souther-compiler/src/main/java/souther/compiler/check/Elaborator.java
@@ -485,11 +485,16 @@ public final class Elaborator {
             // sealed interface declares the accessor its cases already carry (issue #160). Only a
             // named sum, whose interface this compile emits — an anonymous union's cases are not
             // written together, so nothing declares their shared part.
-            if (target instanceof Type.Ref ref && ctx.symbols().declarations().declaration(ref.name()) instanceof Hir.SumData) {
-                Type shared = TypeOps.commonSpreadFields(cases, ctx.symbols()).get(fa.field());
-                if (shared != null) {
-                    return new Core.FieldAccess(targetCore, fa.field(), shared, fa.pos());
-                }
+            // Asked of the type as written and not through the names it wears: how far a read looks
+            // through a newtype is this elaboration's policy, and a `data X = S` over a sum is a
+            // value of X, whose fields are X's. `TypeView` keeps both directions available for that
+            // reason, and this reader takes the one it has always taken.
+            TypeView view = TypeView.of(target, ctx.symbols());
+            if (!view.isWrapped() && view.shape() instanceof Shape.Sum shape
+                    && shape.common() instanceof Shape.CommonProduct.Shared shared
+                    && shared.fields().get(fa.field()) != null) {
+                return new Core.FieldAccess(targetCore, fa.field(),
+                        shared.fields().get(fa.field()), fa.pos());
             }
             // A sum carries no fields of its own — its cases do, and which case it is is not known
             // until it is opened. Saying that is the difference between "this value has no such
--- a/souther-compiler/src/main/java/souther/compiler/check/GuaranteeWalk.java
+++ b/souther-compiler/src/main/java/souther/compiler/check/GuaranteeWalk.java
@@ -4,6 +4,7 @@ import souther.compiler.core.Core;
 import souther.compiler.types.Type;
 import souther.compiler.types.TypeSymbol;
 
+import java.util.ArrayList;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Set;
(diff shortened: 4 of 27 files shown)


kawasima:
Fixed in 0398db8. Taking the blocker as one predicate's domain rather than as a `GuaranteeWalk` bug was right, and it turned out to reach one step further than either of us wrote.

## The mismatch predates this branch

`Clause.Ref.Id` has always carried `declaredOn`, and `narrowers` has always been keyed by it (`InvariantChecker.relating` files `from.clause().id().declaredOn()`). Both readers tested the declaration the reading was opened at. Since `Clauses.declared(named, data)` answers the *effective* clauses, a rule a spread carried in was already outside what either reader could leave out — before any of this branch.

A red test on plain records, no sum involved:

```souther
data Common =
    { lo: Int
    , hi: Int
    }
    invariant based = lo >= 100
    invariant far = hi >= lo + 10

data Held = { ...Common }
    invariant near = hi >= lo + 5
```

`hi` stops at 110 and `far` is what puts it there. `narrowedBy("hi", true)` answered `[Held]` — not merely imprecise, the wrong declaration: taking `Held` away removed `Common`'s rule too, so `Held` looked like the one holding the end. It answers `[Common]` now. `WhichRuleHoldsAnEndIsAskedOfTheRuleAndNotOfWhereItIsReadTest`.

So this is the existing contract violation surfacing, exactly as you put it. The sum made it the normal case rather than the exception.

## `Predicate<RuleRef.Invariant>` does not close it

I did that first and the suite stayed green where it should have gone red. `named("Bounded")::equals` binds to a `Predicate<RuleRef.Invariant>` — `equals` takes an `Object` — so `WhatATypeGuaranteesIsOneAnswerWhoeverAsksTest` went on compiling and silently excluded nothing. It had been green for the coincidental reason (the position was the writer) and would have stayed green for a worse one.

So the exclusion is a type that cannot be coerced into:

```java
record RulesLeftOut(Predicate<TypeSymbol> declarations) {
    static final RulesLeftOut NONE = ...;
    static RulesLeftOut writtenOn(Predicate<TypeSymbol> these) { ... }
    boolean excludes(RuleRef.Invariant rule) {
        return declarations.test(rule.clause().id().declaredOn());
    }
}
```

Your point about the outward API is kept whole: a caller says which declarations it means, and the conversion happens once. What changed is that a caller cannot hand a bare lambda at all, so nothing can be passed a position's name by accident. The three call sites that could stopped compiling, which is how I found the `::equals` hole.

`Reach` and `GuaranteeWalk.Scope` both hold it. `stopAt` stays a `Predicate<TypeSymbol>` — it is the position question and is right as it is.

## `seedFields`

Taken as you wrote it, including the grain split. `stopAt` is asked once of the name and keeps its own `missed`; `withoutClauses` is asked of each rule; a `skipped` flag records one `missed(THE_VALUE, BY_EVERY_VALUE)` at the end. `Gathering.missed` is untouched.

`Coverage.Incomplete.lost` is filtered by the same value, so `lostAClause` reports only what the reader asked for.

## The rest

`PositionReading` carries `entering` and has one switch over `Shape` — the identity is no longer worked out a second time.

`At.HandedOn.under` is asserted directly now: `Both` hands on `Stamped` and nothing else, `Marked` hands on nothing. So it is the difference that is tested, not the fact of a handoff.

The narrow structural test is in, over the five files this touched:
`TypeGuarantees`, `GuaranteeWalk`, `PathEngine`, `UniversalElementFacts`, `StepInputFacts` resolve no declaration, and `Shape.CommonProduct.Shared` is constructed in one place. Reading `Sum.common` is left alone — that is what a consumer of a shape is meant to do.

The `match` comment and the `Borne` javadoc are updated. `CommonProduct.Shared`'s precondition now rests on a measurement rather than an assumption: `data Marker = { }` is refused with E1008, so a shared declaration has fields.

Mutations, all separately red:

| mutation | red |
|---|---|
| `RulesLeftOut` never excludes | the two provenance tests, and `narrowedBy` back to `[Common, Held]` |
| a sum states nothing of what its cases share | 3 assertions incl. `states("capped")` |
| nothing is ever handed on | 3 different assertions incl. `handedOn` |

`mvn -o test`: 10,376, green.

The one thing still uncovered is `Coverage.Incomplete` on its own — at a read position every field is given, so `statedAt` answers null only when typing a clause fails, and I have no source that does that.

Issue #1108, opened 2026-08-26
The conformance corpus never carries a requirement across a module boundary


kawasima:
Found while measuring #1103. Measured on `534dc30a2`.

The corpus declares an injected behavior and fakes it — `conformance.shipping`'s
`台帳へ書く` is declared without a `let`, `出荷する` carries it through `depends on`,
and a `fake` closes it. All of that happens inside one module. Nothing in the corpus
carries a requirement across a module boundary, and that is a mechanism of its own
rather than the same one written twice.

## What the corpus does not write

`conformance.shipping` imports `conformance.catalog` for its types and reaches
nothing else of it. There is no module that takes an injection target from another
module and carries it on.

Trying to write the smaller thing first says why the mechanism is one thing and not
two. A behavior is not a value:

```
let 値引き後 (品目: Item, 率: Rate): Price = match Catalog.discount(品目, 率) with ...
    E1023 NoValueOfThatNameInScope[name=Catalog.discount]
```

and `depends on` names a behavior nobody wrote, so a written one is refused:

```
behavior 出荷する : ... depends on 台帳へ書く, discount
    E1607 DependsOnNamesSomethingThatDependsOnNothing[behavior=出荷する, dependency=discount]
```

So there is no cross-module call to write. What the language has is a requirement
declared in one module, imported by another, carried through that module's own
`depends on`, and supplied where the rows are run. The corpus writes the last three
of those and never the first.

## What it costs

The compiler answers a question about exactly this and the corpus never puts it.
`Bodies.ImportedUnwritten` is asked over every module and comes back empty every
time, because the loop inside it runs once per borrowed behavior and nothing is
borrowed. `Bodies.Unwritten` is the question it asks of the other module, and over
the corpus it is never asked at all — one of the two questions in this compiler that
a whole-corpus run leaves unreached for a reason that is not about the batch domain.

An answer nothing asks for is an answer nothing checks. What a module says about a
requirement it did not declare is read by the assembler, by the check that a
composed behavior has everything it needs, and by what a build refuses over, and
none of those is exercised across a module boundary today.

## The shape it probably takes

`conformance.catalog` declares a behavior and does not write it — something the
catalog cannot answer on its own, in the way `台帳へ書く` is something shipping
cannot. `conformance.shipping` imports it, names it in `出荷する`'s `depends on`
beside `台帳へ書く`, takes it as a parameter of `let 出荷する`, uses it, and closes it
with a `fake` the way the other one is closed.

[...]

PR #1115, opened 2026-08-26
Answer with the way to a border, and close what an answer may hold

### LINKED ISSUE #1103: A query answer holds a region that compares by address, where nothing watches
Found while measuring #1077. Two compilations of one source answered `Adequacy.DeclaredBorders` unequal, and the value that differed was a `RegionForARow` held inside the row a boundary search composed. It has no `equals`, so it compares by address.

## What is wrong

`EquivalentDatabasesAnswerTheSameTest` exists so that a value kept in an answer and comparing by identity cannot go unnoticed: an answer that means what the last one meant has to come out equal to it, or every reader of it runs again forever. It keeps a register of what is known and not fixed, and `RegionForARow` is not in it.

It is not in it because the questions that hold one are never asked. The test reads `corpus.analyse().compilation().db().everyAnswer()`, and over the `catalog` corpus that map holds no `DeclaredBorders`, no `Coverage` and no `Findings` key at all. `analyse()` asks for `Adequacy.Asked.fullReport()`, which says how much to measure; what makes the measures be asked is somebody building the report, and nothing does before the keys are read.

So the whole adequacy family is outside what the test watches, and at least one value in it does not compare as a value.

## What it costs

`RegionForARow` reaches an answer through `ItemAssessment.Attempt.Built` — the row a boundary search composed, kept beside the point it was built for. Anything downstream that compares two `DeclaredBorders` answers, or that relies on an unchanged answer being recognised as unchanged, gets a false negative: the answer is recomputed and every consumer of it with it.

Measured rather than reasoned: a test comparing two `DeclaredDebt` lists over one unchanged source found them unequal, with `RegionForARow@6fc3e1a4` against `RegionForARow@f72203` as the whole of the difference. The test had to compare a projection instead.

## The shape it probably takes

Two things, and the second is the one that keeps the first from coming back.

`RegionForARow` becomes a value or a capability. It holds a `SearchRegion` and a list of `OnTheWay`; whether what it is is its region and its provenance, or whether it does something, is what the register asks and is what the fix records.

And the test asks the adequacy questions. What it watches is "what the corpora reach", and the corpora reach the report — so reaching it here is what makes that sentence true. What else that turns up is not known before it is run.

kawasima:
Closes #1103.

Two compilations of one source answered `Adequacy.DeclaredBorders` unequal, and
the difference was a `RegionForARow` held inside the row a boundary search
composed. The region is gone; so are the two reasons nothing noticed.

## What the region was doing there

`ItemAssessment.Attempt.Searched` kept the region a search ran in beside the row
it produced. A region answers questions about values and does not say which
conditions it was built from, so the whole reading of a module's rules travelled
in the answer behind it — the numeric domains, the scope, the vocabulary, and
`Db` itself through a captured lambda. Measured over the corpus with the report
built, every one of the twelve things that could not compare sat under

    …attempt.region.where.…

and one of them (`IdentityHashMap`) came back as a compile that had not
reproduced rather than as a container comparing by address.

Nothing read the region back off an answer. The one reader of a finished search
asks what the way to the point declined — the account, not the box. So the pair
is deleted rather than hidden: `ReachingCuts` hands back a `WayToTheBorder`, the
search narrows what the declarations leave by it and realizes against that, and
the outcome carries the account. The region is derived where it is used. The
invariant that a region and its account agree is not checked any more, because
there is one of them.

## Why nothing noticed

**The detectors.** Two walks watch this — one holds a store and asks each object
what it is, the other holds two stores over one input and asks where they come
apart — and neither sees what the other does. Both were wrong in ways a corpus
cannot show, and the failures were measured, not reasoned:

- the pair walk remembered which pairs it had visited rather than what it found
  under them, so a shared value was reported at one of its places and its holder
  was named for a defect five levels below it (`BorderAssessment$Point`,
  `Attempt$Built` and `ReadRegion` were all this artifact)
- an identity-semantic map was filed as `DIFFERENT_THINGS`, which sends a reader
  after the compiler instead of after the container
- a walk that ran out of budget, met a field it could not open, or met a graph
  that holds itself said nothing at all
- the single walk stepped into containers without asking them anything, which is
  every array, and started at an answer's value rather than at the answer —
  which is half of what `Db` compares
- neither looked inside an `Optional`: `java.base` opens nothing here, so
  everything an answer kept behind an absence went unasked

Each of those is now held by a self-test built out of the smallest graph with
the property, and each was confirmed by breaking the mechanism and watching the
right test go red.

**The register.** One register now, keyed by where a thing sits in an answer
rather than by its class. A locus is the steps of the answer's own shape, so a
component called `value` and the value side of a map are two places, and a class
already written down turning up somewhere new is a new line rather than a
known one. What a thing is and what is wrong with it are two axes. Which
detector met it in which scenario is provenance beside the entry and is compared
exactly — a detector that goes blind on something the other still finds leaves
the set of places alone and is caught by that comparison.

The scenarios are two because an answer is two things. `Answer` is what a
question came to and what was said getting there, and a corpus of valid models
exercises one of those. Adding a module the compiler has something to say about
surfaced a `Diagnostic` in an answer's reports, which means nothing by `equals`;
it is in the register.

**The questions.** `ConformanceCorpus.analyse` asked for a full report and
nothing built one, so the entire adequacy family — ten kinds of key, including
`DeclaredBorders` — was outside every check above. It builds the report now, and
what this compiler declares is discovered from the compiled classes rather than
from a list somebody keeps, so:

    declared 140 = reached 127 ⊎ 13 a batch run has no input for

The thirteen take a cursor position. `souther-lsp` asks them; whether that
vocabulary is closed the same way is not something this claims.

The operations on the reached side are what this project does — analyse and
report, offer rows, ask what a module declares, build at the level a build
measures at — not a fixture per key. `Bodies.Unwritten` was the fourteenth until
#1108 gave the corpus a requirement that crosses a module boundary.

## Reading it

Eight commits. The first three are the change as it was first written, each green on its own:

1. two projection keys leave the vocabulary
2. the answer carries the way rather than the region
3. the detectors, the register and the closed question set

The last three come from review — a self-check and two rounds — and all of them
are in the oracle rather than in the compiler:

4. a place is named by what holds it, and the scan says what it could not read
5. correspondence is established rather than assumed, and the vocabulary is
   counted before it is narrowed
6. falling short is said in one vocabulary, and a thing is judged on its own terms
7. the walk is measured against what equality denies, over graphs nobody chose
8. the readable form is asked for by name, and a walk is held to falling short
   where the shape says it does


kawasima:
Self-check found four things this had left as it found them. Three are fixed in 43d73adb9; the fourth is filed as #1117.

**A `Diagnostic` was written down under the wrong kind of name.** The register held it as `UnusedImports.reports[].diagnostic`. Measured against a model with a mistake in it instead of an unused import, the same defect comes back as `CheckedBehavior.reports[].diagnostic` — so the question in that line was an artifact of which model the scenario compiled, and adding scenarios would have added lines about one thing. An answer is what a question came to and what was said getting there; only the first is a different shape per question. A locus rooted in the reports is now keyed without a question (`*.reports[].diagnostic`), and swapping the scenario's model leaves that line exactly where it was.

The comment beside that scenario also gave a reason nobody had measured — that a model with an error would stop the compile short of the answers the register is about. It does not: 115 keys, an answer carrying a report, a complete walk. It reaches *fewer* of them, which is the real reason to prefer a warning, and that is what the comment says now.

**The scan that counts the vocabulary swallowed a class it could not load.** `catch (Throwable) { continue; }`. That is the failure this test exists to prevent arriving through the thing that counts — the declared side shrinks and `declared = reached ⊎ excluded` goes on balancing over a smaller world. It is reported now and asked about before the arithmetic. Confirmed by making one class unloadable and watching it go red.

**A locus told a record's component from a class's field, and an array's element from a list's.** Both pairs render the same way and no test could distinguish them. Collapsed to one step each.

**Two entries have never been read.** `InputDomain` and `Bodies$Elaborated` say `NOT_READ / UNCLASSIFIED`, carried over from the register before this one, where the note already said reading them was the fix. This change moved the note without answering it, so the debt is now #1117 rather than a line nobody is asked to remove.

kawasima:
Fixed in 0f0b1a09b. Before patching the six, I counted them: **four are one cause.**

This code decides five times that two things correspond, or that two occurrences are one place, and every one of those decisions was a shortcut that dropped what it could not carry and said nothing:

| where | the shortcut | what fell out |
|---|---|---|
| `Divergence` collections | paired by iteration order | a set's actual correspondence |
| `Divergence` map keys | right copied into a `HashMap` | keys that mean one thing and are two |
| `met()` | intersection of the two key sets | a question one store was never put |
| `scan()` | one package | a key declared anywhere else |
| the ledger | `Place` flattened to text | occurrences whose text agrees |

The completeness layer built in step 1 covers *going* — "I could not get there" — and I never extended the same discipline to *telling apart*. That is why finding 2 is a defect I had already removed from the walkers and then rebuilt in the test that closes the denominator, and why patching the five sites one at a time would have left the sixth to be written next time.

So: a pairing is read off the contract (`List` by position, `Set` by membership, anything else stops rather than guesses), and both collection and map correspondences must be total and one-to-one or they are a gap. The scan counts every key the compiler has wherever it is written, and "they all live in `souther.compiler.query`" is a separate claim that goes red on its own — checked by putting a concrete `Key` in another package and watching both assertions fail. `met()` requires the two stores to have been asked the same questions, per scenario, before anything is compared; the corpus-only test that used to hold that is gone because it held it for one scenario. A `Place` stays a value to the end and a member step carries its declaring type.

The other two findings share a second cause: **the mechanism's arms were not the checklist.** `BUDGET_EXHAUSTED` and `A_FIELD_THAT_WOULD_NOT_OPEN` were reported by nothing any test built, and `Cause` had lost the word for identity semantics while the class javadoc still argued the remedy differs. Both are closed the same way — the budget is injectable, both arms have graphs, and one test asserts that every arm of `Interruption.Why` is reached by a graph in that file, so adding an arm arrives with nobody having built the graph for it.

Writing that checklist paid immediately: **two of the fixtures I had just added were passing on walks that never happened.** The set fixture held its two sets inside a record whose two instances were equal, so the walk stopped above the sets and asserted "nothing found" for the wrong reason; the identity-map fixture had a single key, so no collapse could occur. Both are rebuilt, and both mutations (pair by iteration order; drop the one-to-one check) now go red on the test that names them.

kawasima:
Fixed in 34989f1dd. Five findings, **three causes.**

**A proxy answer stood in for the question it was a proxy for** (1, 2, 3). "Is this thing defective" was answered by "was anything found beneath it"; "are these the same thing" by "are they of one class"; "do these correspond" by "are they of one size". Each is now asked in its own right:

- a container whose members pair by what they say and which still says they do not is named for that **whatever its contents did** — so wrapping a defect already in the register no longer hides the wrapper
- what two things say is asked before what class they are
- a size is compared inside the branch that knows what a size is worth for that container; a collection with no rule for pairing is one whatever the two sizes are

**Identity and rendering were one string** (4). `Member` held `getSimpleName()`, so two types of one short name were one place — the collision the step was added to prevent, one level in. It holds the type's name now and shows the short form; the offender does the same, arrays included. The ledger's own entries had to be rewritten with full owners, which is what proved the two were no longer the same value.

**Falling short was said in four vocabularies** (5, and the fourth I had not reported). The pair walk had `Interruption`, the single walk had `notOpened` + `loops`, the scan had `unreadable`, the two-store comparison had `askedDifferently`. That is why one of them swallowed what another reported — twice on this branch, and I fixed both at the site. There is one `Gap` now, one list of ways to fall short, and one `Covered` with two arms and nothing above them, so **a caller cannot read what a search found without meeting where it did not look.** The scan takes the root it walks, so an unloadable class is a graph a test builds; and it counts the whole code source, because a perimeter narrower than that is the count's own bound deciding the count — at a package or at a package tree alike.

The checklist follows from the vocabulary rather than being written per detector: every arm of `Gap.Why` is reached by a graph built in one place, across all four searchers.

Mutations, one per finding: removing the container's own judgement, judging class before equality, comparing sizes before the contract, and shortening the owner name — each goes red on the test that names it, and only that test.

PR body updated for the commit count.

kawasima:
Fixed in 4d970e1a1. Before the five, I looked at the three rounds together.

**The findings are not the root cause. My way of checking is.** Round 1's findings were in code I had just written; round 2's were in what I fixed in round 1; round 3's in what I fixed in round 2. Every one of them was one of two shapes — something the oracle claims to catch and does not, something it names that is not there — and every one was met by adding the case that had been missed, by the person who had already missed it. A fixture and the mechanism under it share an author, so the two agree with each other and neither is measured against the promise. A sixth hand-built fixture would be the fourth turn of the same crank.

So the promise is written down instead: two graphs built from one recipe are one thing, so a walk over them never says `DIFFERENT_THINGS`, and what denies being what it is is countable from the recipe and `equals` alone — independently of the walk. Graphs are grown from a seeded recipe and built **on both sides separately**, so which concrete class carries a shape is not part of the shape.

**Against the walk as it stood at `34989f1d`, that finds 431 counterexamples** — `HashMap`, `LinkedList`, `HashSet`, `LinkedHashSet`, `ArrayList`, `TreeMap`, `List12`, `ListN` — which is review finding 2 arriving without anyone having thought of it. Against the walk in this commit, none in 2,000 shapes.

Two honest notes on getting there. My first generator found nothing, because I had written its alphabet from the same imagination: every list was `List.copyOf`, every map a `LinkedHashMap`, both sides always the same class. And my first property was too weak — it compared only whether anything was found, so `ArrayList` vs `LinkedList` misfiled as `DIFFERENT_THINGS` passed it. The alphabet now varies per side and the property counts what should be denied.

Under that, the findings themselves:

- **1** `Place` is no longer `Comparable`, and the ledger no longer sorts by it. How a place reads drops the owner that tells two types of one short name apart, so an order over the text put two places in one bucket of every map that sorts — which the ledger was using. Sorting happens when a failure is written.
- **2, 3** the walk answers three questions rather than letting two stand in for three: *is this one thing* (read off the shape, over each container's own contract, asked before the concrete class), *do they say so* (`equals`, the thing under test and never the evidence), and *whose defect is it* (this one's, where every part agreed with its twin). So a container comparing by address is named beside what it holds rather than instead of it. A map is also judged on the half of its equality that can be put on its own — its keys — and named once however many ways it denies.
- **4** `Covered.Partly` cannot be built saying nowhere it stopped, and both arms copy what they are given.
- **5** the tautology is gone; that test now asserts nothing is concluded about a container with no rule for pairing.

Mutations: judging class before contract, dropping the key-half judgement, and naming a thing only where nothing was found under it each make the property red.

The CI run for `34989f1d` never appeared; this push should carry one.

kawasima:
Fixed in 48948dc82. Three findings, **two causes**, and the first is the one worth naming.

**1 and 2 are one rule applied where it was found rather than everywhere it holds.** Last round I stopped ordering a place by how it reads — and left a second site that compares how it reads. Same defect, one file away, and I never counted where the rule applied. The property has the same shape: I built "a search that fell short must be accounted for" into the conformance tests and then let the property drop `Partly` on the floor.

So neither is fixed at its site.

- The trap was that `toString()` gave the short form, so anywhere a value was wanted and text got written the identity went quietly. The readable form is now `asText()`, asked for by name; what a place is by default is what it is made of. The provenance comparison keeps the values and builds its message from the difference — so there is no longer a reason to render before comparing.
- What a shape falls short on is now said **from the shape**: a collection whose equality is neither order nor membership has no rule for pairing, and a set of things that only say which object they are has members that will not line up. `Whole` where the shape says covered, `Partly` on those reasons and no others. Mutation: making every list escape into `Partly` turns the property red, which it did not before.

**3 is a branch standing for two answers.** A list and a set are both understood and neither is the other; that is two things, not a container this cannot pair. Which contract a thing answers to is read once now, and *different contract* and *unknown contract* are two outcomes. Mutation: filing a contract mismatch as unknown turns the new fixture red.

Writing the expected shortfall turned up one more in the expectation itself: a set is where the walk pairs and stops, so nothing under one is reached and nothing under one can fall short. The property was over-predicting until the model matched the mechanism.

I also counted the remaining places that render a locus rather than fixing the one you found: they are all message-side now — the text a `Gap` carries, the literals in the two self-tests, and the failure line. The registers compare typed values.

PR body says eight commits.

kawasima:
Fixed in 2a1994649 (on top of a merge of current `develop`).

**The CI failure was not flakiness and not the attribution hole.** `NarrowedBounds` arrived on `develop` in #1119, and CI builds the branch merged with its base — so the register was reacting to two new places the compiler had grown. Merging `develop` reproduces it locally. That is the register doing its job; what it wanted was a reading, not a repair. `NarrowedBounds$Held` keeps the declarations holding one end of a range, worked out once and kept, and says nothing by `equals`. A value — two of them that would name the same declarations name the same declarations — and here for the reason the standard library is, which its own header states: comparing, hashing or printing one asks for the work it exists to defer. Two places, met only by the walk that holds two stores together, because the other one stops at `InputDomain`.

**The attribution hole, and the reason it was invisible.** You are right that the rule is correct for derived equality and wrong for independent identity, and right that my property shared it. Those are one thing: an oracle that reasons the way the mechanism reasons agrees with it wherever the reasoning is wrong, which is the one thing a property is there not to do.

So neither is patched at the rule.

- Where a thing's equality comes from is now read off the language: *its parts'*, or *its address*. An array is the second — equal to another when it is the same array, whatever is in it — so it is a debt beside what it holds and not instead of it. A record holding the same thing stays the first.
- The property no longer reasons about attribution at all. It **experiments**: each shape is built again with every part replaced by something that says what it is, and asked whether it still denies its twin. A record of honest parts agrees; an array of them does not; an `IdentityHashMap` of them does not. Nothing about attribution is shared between the expectation and the mechanism now.

Both counterexamples you named are pinned: `Object[]{new Address()}` gives two findings, and `IdentityHashMap(same-meaning key → fresh Address)` gives the map for its keys and the Address for itself. Mutating the attribution rule back turns **both the property and the fixture** red — previously only the fixture would have moved, and before that neither.

One limit I will state rather than claim past: a map that compares its *values* by address, whose values also deny, is not distinguishable by any experiment the walk can run in that state — its denial is explained either way. The key half is independently askable and is asked; the value half becomes observable once the inner defect is fixed.

`mvn verify` — the command CI runs — is green locally on this head. GitHub has not started a run for `2a199464` yet.

kawasima:
Fixed in 1f1bd0b93.

**The root cause is that I wrote the instance as the rule.** Last round found the array; I encoded it as `c.isArray()`, which put the same hole in every other class — something with parts whose own equality is its address is taken for something whose equality is its parts', so it goes unnamed for as long as anything under it also denies. That is exactly what naming the array was meant to stop, one class along. Your `IdentityBox` is that class.

**And the property could not see it for the same reason it could not see the array before me.** The alphabet had a leaf that only says which object it is, and a record whose equality is its parts', and nothing with parts of its own that answers by address. Shapes nobody chose are only as wide as the grammar, and the grammar was mine. So the axis is in the grammar now — `WithOwnIdentity` — not a fixture for the case you named.

Where an equality comes from is three answers:

| | how it is settled |
|---|---|
| an array, or something that never wrote an `equals` | its address — the language says so |
| something answering to a contract this walks | what it holds — the contract says so |
| a record | asked: a twin built from the very things it holds; one that denies that is not equal by what it holds |
| anything else that wrote its own | **neither could be shown** |

Your point about identifiability is the one I had been dodging, and it decides the last row: an equality over the parts and an equality over the address answer alike while a part is denying, and nothing readable off the class tells them apart. So the walk says it cannot tell, instead of picking the optimistic one. That is the same discipline this PR already applies to correspondence — *establish it or call it a gap* — reaching the one place it had not.

Over the corpus that is exactly one place: `Resolve.Values` wrote its own equality, is not a record, and holds the store that is the line under it. It is written down beside the places rather than among them, because it is what that store costs and it goes when the store goes. A second one arriving that stands above nothing would be a walk that has stopped being able to attribute something it used to.

Two mutations, one test each is not enough to report: putting the old assumption back (`everything that is not an array is its parts'`) turns **both** the property and the new fixture red. Before this commit only a hand-written fixture could have moved; before the round before, nothing would have.

Writing the expected shortfall turned up one more in the expectation: a thing whose part gave up first is never asked the question its own equality would have answered, so the model had to learn the walk's order of giving up.

You are also right that `Cause.IDENTITY_SEMANTICS` still has no entry in the register — the category is held up by the synthetic oracle alone. That is now the oracle the grammar reaches, rather than a fixture I thought of.

`mvn verify` is green locally on this head.

PR #1116, opened 2026-08-26
A behavior's partition evidence holds only the lines it is owed a row at

### LINKED ISSUE #1078: A behavior's partition evidence holds the lines the declarations are owed
Held out of #1062 as a non-goal: storage topology rather than identity. It is worth its own issue now that the identity is settled.

## What is wrong

`PartitionEvidence` is a behavior's: "what a behavior's rows reach". Its `boundaries()` holds every border read at that behavior's positions, and two of a border's four points can be owed to the declaration that drew the line rather than to the behavior (#1062).

So every reader that measures a behavior, counts what it covers or raises a finding about it has to leave those points out. `BorderAssessment.Point.owedHere` is the one definition of the rule, and it is a rule each reader still has to remember to ask.

That is how the verdict came to keep them: `AdequacyReport.requiredEvidence` walked the points and weighed one behavior's row against another behavior having no rows, holding a build undetermined about a line the aggregation had settled. It was found by review rather than by the types.

## The shape it probably takes

```
Adequacy.Of
    signatures
    partitions      // a behavior's classes and pair space
    boundaries      // what the module's lines came to
    branches
```

`Adequacy.DeclaredBorders` is already the one production of the declarations' debts. What is left is that a behavior's evidence stops handing out what is not the behavior's, so a reader cannot ask the wrong question rather than having to remember not to.

## Careful about

A reader that only describes may still show them. The behavior's border block accounts for a border's four points by design — `excluded` exists because "a reader working to a coverage criterion counts four items per border, and a border showing two of them with nothing beside it reads as this compiler being short of the other two" — and whose debt each point is, is a different question from which lines this behavior's values are held to.

kawasima:
Closes #1078.

## What was wrong

`PartitionEvidence.boundaries()` answered two questions with one value: which lines a behavior's positions met, and what that behavior is owed a row for. Two of a border's four points can be owed to the declarations that drew the line (#1062), so the second was obtained by every reader applying the rule at `BorderAssessment.Point.owedHere` for itself. `AdequacyReport.requiredEvidence` forgot, and held a build undetermined about a line the module's own account had settled — found by review rather than by the types.

## The shape now

Raw reading → attribution → two accounts, one way:

- `PointContributions` is what settled a point and is what can be merged; `PointAttribution` is `TheReading | TheDeclarations(owners)` and is what that comes to. A value nothing contributed to is refused rather than classified, which removes the vacuous `noneMatch` the old boolean answered with for an empty set.
- The classifier is called from `Border.owes` and nowhere else; the arms are read by the two account producers and by no consumer.
- `OwedBoundaryPoint` is one thing a behavior is owed a row for. It carries the `Border`, not the `BorderAssessment`, so the other roles of the line are not reachable from a behavior's account.
- `PartitionEvidence` holds `Measure<List<OwedBoundaryPoint>> owes` and no longer the lines. The lines, and how far the reading that found them got, are `BehaviorEvidence.boundaryReadings` (produced by the new `Adequacy.BoundaryReadings` key, which is also where the level's choice between the searched and unsearched lines is made — so the coverage and the generation read one answer).
- `Measure.readAs` is the one place a measure's arms are carried over to a value read off what it found. An account is the reading's own measurement, not a second one.

Consumers now say which question they ask: findings, the verdict, the generator and the editor's lens read the account; the report's border block, the document and the module's declaration account read the lines.

## What a behavior went without

A behavior's `weakening()` is over its own account. A row owed to a declaration is short or not short in the module's account of them, which the module's status reads through `ModuleReport.declarationsWeakenedBy`. The reading that found the lines is common to both accounts and weakens both.

Measured over `souther-examples` (crm, hr, shippingfee, ordering, issuetracker — 24 modules): the account of every behavior is exactly the points its own rules settled. Four behaviors — `activity/touchCount`, `activity/lastActivityOn`, `payroll/buildPaySlip`, `inventory/stockFrom` — stop carrying `BorderValueUnreadable` reasons that belong to points only the declarations are owed; the declaration accounts of those modules carry them instead. The editor's lens over the fixture goes from `boundary 2/6` to `2/5`, the point dropped being the one an author of that behavior cannot write a row for.

## Checks

Three, each shown to go red with the rule broken:

- `WhoseARowIsIsDecidedInOnePlaceTest` — over the compiled classes: `PointAttribution.of` is called from one method, and the arms are read by the two accounts only.
- `EveryPointOwedIsInOneAccountTest` — every point a reading owes falls in exactly one account, and a behavior's holds nothing owed to the declarations. Written as a refinement rather than as a count: the declarations' side collapses the readings of one point into one debt, so no total is conserved.
- `ABehaviorsAccountDoesNotReachTheLinesOtherPointsTest` — the closure of what a behavior's evidence hands out reaches the line and never an assessment of one.

`mvn -o test` is green across the reactor.

### DIFF
--- a/souther-bench/src/test/java/souther/bench/AMeasureIsIntroducedInOnePlaceTest.java
+++ b/souther-bench/src/test/java/souther/bench/AMeasureIsIntroducedInOnePlaceTest.java
@@ -124,7 +124,14 @@ class AMeasureIsIntroducedInOnePlaceTest {
             // be built from, and unknown where a composition takes what a stage nobody could work
             // out takes. Two states and one place that chooses between them — the cases at each
             // position are that position's own answer and never this one's.
-            Map.entry("souther.compiler.query.Adequacy$SignatureEvidence#at(Ljava/util/List;)Lsouther/compiler/query/Measure;", 1)));
+            Map.entry("souther.compiler.query.Adequacy$SignatureEvidence#at(Ljava/util/List;)Lsouther/compiler/query/Measure;", 1),
+            // And the one place a measure's answer is carried over to a value read off what it
+            // found. All five arms, because carrying them over is the whole of what it does: what a
+            // behavior is owed a row for is read off the lines its positions met, and the two are
+            // one measurement. This is the only entry here that decides nothing — a projection that
+            // chose between the arms could call a reading that did not run out complete over
+            // whatever it happened to produce, which is what every other entry exists to stop.
+            Map.entry("souther.compiler.query.Measure#readAs(Ljava/util/function/Function;)Lsouther/compiler/query/Measure;", 5)));
 
     @Test
     void nothingButTheIntroductionRuleMakesACaseOfAMeasure() throws Exception {
--- a/souther-bench/src/test/java/souther/bench/WhoseARowIsIsDecidedInOnePlaceTest.java
+++ b/souther-bench/src/test/java/souther/bench/WhoseARowIsIsDecidedInOnePlaceTest.java
@@ -1,0 +1,116 @@
+package souther.bench;
+
+import org.junit.jupiter.api.Test;
+
+import java.io.IOException;
+import java.util.List;
+import java.util.Map;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+
+/**
+ * Whose it is to write a row at a point is decided once, and read by two accounts and nobody else.
+ *
+ * <p>A border owes a row at up to four points, and two of them can be owed to the declarations that
+ * drew the line rather than to the body carrying the type. Everything that measures a behavior,
+ * counts what it covers, raises a finding about it or offers it a row has to know which of the two
+ * it is holding — and while every one of them asked for itself, the one that forgot weighed a
+ * behavior's row against another behavior having no rows and held a build undetermined about a line
+ * the module's own account had settled.
+ *
+ * <p>So the question is asked where the point is made, and what comes out of it is one of two arms.
+ * This holds the two halves of that: the classifier is called from one place, and the arms are read
+ * only by the two accounts made from them. A third reader is a consumer deciding again.
+ */
+class WhoseARowIsIsDecidedInOnePlaceTest {
+
+    private static final String ATTRIBUTION = "souther.compiler.partition.PointAttribution";
+
+    /**
+     * Every way the classifier or one of its arms is reached from outside, one at a time.
+     *
+     * <p>Two rules in one answer, because one list of what is reached holds both. The classifier is
+     * called where the point is made and nowhere else: everything that settled a point has arrived
+     * there, so asked earlier the answer would be about whichever contributor had arrived, and asked
+     * later it would be asked by whoever wanted it. And an arm is named by the two accounts and by
+     * nothing else, because naming one is deciding whose a row is — which the classifier already
+     * did.
+     *
+     * <p>Which member of which arm, reached which way, and by whom. Folded to the methods that touch
+     * an arm somehow, this rule is passed by writing {@code new TheDeclarations(...)} inside a method
+     * that is already allowed to hold {@code TheDeclarations::and} — the set of names does not move,
+     * and what the name of this rule says did not happen would have happened.
+     *
+     * <p>So the answer is spelled per member and per way of reaching it, and the whole of it is
+     * compared: a reach nobody has written down is a reach nobody has thought about, whichever kind
+     * it is.
+     */
+    @Test
+    void everyWayAnArmIsReachedFromOutside() {
+        assertEquals(new java.util.TreeMap<>(Map.of(
+                        // Merging the owners of two readings of one point, which is the declarations'
+                        // account folding what it was given. A reference rather than a call because
+                        // it is handed to `Map.merge`.
+                        ATTRIBUTION + "$TheDeclarations#and REFERS",
+                        List.of("souther.compiler.query.BorderObligationPointAssessment#across"),
+                        // Which of this module's declarations owe the point, which is the arm
+                        // answering about itself rather than anybody deciding which arm it is.
+                        ATTRIBUTION + "$TheDeclarations#ownersIn CALLS",
+                        List.of("souther.compiler.query.Adequacy$DeclaredBorders#compute",
+                                "souther.compiler.query.BorderObligationPointAssessment#across"),
+                        // And the two accounts, each naming both arms because each answers for the
+                        // whole question: a producer that named one of them would be a point whose
+                        // arm neither account holds.
+                        ATTRIBUTION + "$TheDeclarations#case NAMES", ACCOUNTS,
+                        ATTRIBUTION + "$TheReading#case NAMES", ACCOUNTS,
+                        // Two points of one line are compared as values, which is what a reading
+                        // holds ({@code OwedPoint}).
+                        ATTRIBUTION + "#equals CALLS",
+                        List.of("souther.compiler.partition.OwedPoint#equals"),
+                        // And the classifier itself, which the rule above is about.
+                        ATTRIBUTION + "#of CALLS",
+                        List.of("souther.compiler.partition.Border#owes"))),
+                reachesFromOutside());
+    }
+
+    /** The two accounts made from what the classifier answered. */
+    private static final List<String> ACCOUNTS = List.of(
+            "souther.compiler.query.BorderObligationPointAssessment#across",
+            "souther.compiler.query.OwedBoundaryPoint#across");
+
+    /**
+     * Every reach to the classifier or one of its arms from outside it, by what is reached and how.
+     *
+     * <p>Sites inside {@link souther.compiler.partition.PointAttribution} are left out: what an arm
+     * does with itself — holding its own field, comparing itself, making itself in its own static
+     * initialiser — is the value being a value, and a rule about who may reach it from elsewhere is
+     * not about that.
+     */
+    private static java.util.SortedMap<String, List<String>> reachesFromOutside() {
+        java.util.SortedMap<String, java.util.SortedSet<String>> found = new java.util.TreeMap<>();
+        try {
+            for (Compiled.Site site : Compiled.sites()) {
+                if (!site.owner().startsWith(ATTRIBUTION) || site.from().startsWith(ATTRIBUTION)) {
+                    continue;
+                }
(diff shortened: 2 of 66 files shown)


kawasima:
Both findings are fixed in `2b9e8a30d` (amended, still one commit).

## Root cause

Not two defects: one. **An account has to carry the measurement it was made from, and `DeclaredBorders` was the one place that stripped it** — `read.made().orElseGet(List::of)`, with `lines == null` folded to "no debts". I counted the places the new code takes a measure's value: four. Three of them (`BehaviorReport.lines()`, the document's boundary array, `Adequacy.readingsOf`) are readers that show the lines and print the measure's status beside them; the fourth built an account. So the rule is: a describer may take the value, an account may not.

## The account is now measurement-bearing

`Adequacy.DeclaredBorders` answers `DeclaredBoundaries`, not a list:

```java
record DeclaredBoundaries(List<DeclaredDebt> owed, Map<String, WeakeningSet> reading) {
    WeakeningSet weakening();               // reading ∪ each debt's item
    DeclaredBoundaries keptFor(Set<String>); // the view a reader of some behaviors is owed
}
```

- `BoundaryReadings` absent is now `Answer.absent()` rather than an empty account.
- `ModuleReport` holds the account; `declarationsWeakenedBy()` is `account.weakening()`; `debts()` reads through it.
- The reading half is **by behavior**. That fell out of a regression rather than a guess: with one folded set, `AReasonAboutAPositionCountsOnlyAgainstItsBehaviorTest` went red — a report filtered to one behavior carried a reason about another behavior's position. Filtering both halves is the account's own answer (`keptFor`), not the view's.

Your counterexample is now a test. `EveryPointOwedIsInOneAccountTest.aReadingThatDidNotRunOutLeavesTheDeclarationsAccountShort` uses the `RULES_NOT_REACHED` shape: no debts found, and the account says the reading did not run out. Emptying the reading map makes it red.

## whole/parts knows the new part

`WhatAWholeWentWithoutIsWhatItsPartsWentWithoutTest` now checks `module.weakenedBy() ⊇ union(behaviors) ∪ declarationsWeakenedBy()`, and a second case makes the part load-bearing: a module whose behavior went without nothing and whose declaration account carries one reason. Dropping `.union(declarationsWeakenedBy())` makes it red.

On the corpus, what is only ever the declaration account's is a row owed to a declaration that could not be read at — measured over `souther-examples`, `example.inventory` carries three such causes that no behavior of it carries. The fixture is built rather than compiled because what has to be seen is an arrangement of parts, and it says so.

Also in this push, from a self-check for band-aids before your review landed: `said()` was left duplicated on `BorderAssessment.Point` and `OwedBoundaryPoint` (now one body on `Border`); `boundaryReadings` was nullable with three different readings of the null (now an invariant on `BehaviorEvidence`: a behavior has a partition and a reading of its lines, or neither); `Adequacy.Of.boundaryReadings` had no production consumer and is gone.

`mvn -o test` green across the reactor.

kawasima:
Both blockers are closed in `235a3b8e7` (amended, still one commit).

## Root cause

They are one cause, and it is the same shape this issue is about, one level up: **a value that is a view of a measurement can be assembled from parts of different measurements, because construction is open.** The relation was held by production convention — `across()` happens to be the only maker, the report happens to pass matching halves — and a convention is what every reader has to remember. My earlier reply named a place (`DeclaredBorders` stripped the `Measure`) and called it the cause; that was the instance, not the class. The evidence I was still at instance level: the filtered-view regression appeared while I was fixing it, and I patched it after seeing red rather than predicting it.

So both are closed by taking construction away, not by checking harder at the call sites.

## 1. `OwedBoundaryPoint` is made from one assessed line, not from four values

It is a `public final class` with a private constructor whose parameters are `(BorderAssessment at, PointRole role, BorderObligationPoint owed)`. The line and what became of it are taken *out of* the assessment at that role, so they cannot come from another line; the constructor also states the remaining relation (`owed.line() == line.obligation()`, `owed.role() == role`). The assessment is not kept, so the reachability rule still holds. `across()` is the only maker, as before, and now that is a property of the type rather than of the file.

`equals`/`hashCode` are written out because the next invariant asks for them.

`EveryPointOwedIsInOneAccountTest.nothingOutsideTheAccountCanMakeAnEntryOfIt` holds the type to having no public constructor.

## 2. `BehaviorEvidence` holds two views of one measurement

The constructor now refuses an account that is not what the lines beside it come to:

```java
if (partition != null && !partition.owes().equals(OwedBoundaryPoint.accountOf(boundaryReadings))) {
    throw new IllegalArgumentException(...);
}
```

"both present" was the weaker half of the same statement. `aBehaviorsAccountIsWhatItsOwnLinesComeTo` builds one behavior's account against another behavior's lines and expects the refusal, after asserting the two accounts differ so the refusal is about something.

## Smaller items

- `ModuleReport.weakenedBy()`'s javadoc said "the union of its behaviors, and nothing of its own" while the code unions the declaration account. Fixed: its parts are its behaviors and what its declarations are owed, with the reason.
- The whole/parts fixture: I tried six source shapes to reproduce the `example.inventory` topology (a cause only the declaration account carries — a row owed to a declaration that could not be read at) and did not get one; the collection-element reading that produces `BorderValueUnreadable` did not come out of a small model. The built fixture stays, and its javadoc says it is built, why, and which corpus module is the real thing. The producer half is covered source-driven by `aReadingThatDidNotRunOutLeavesTheDeclarationsAccountShort`.

`mvn -o test` green across the reactor.

kawasima:
All three are closed in `89bd7779e` (amended, still one commit). CI on this head: run 33017500355, success.

## Root cause

One cause, and it is why the same instruction keeps landing on me: **I introduced the account types without writing down what an account is, so each of its properties got settled where a test or a reviewer pushed.** An account is a fold over per-behavior readings. From that alone follow the three things you found: what is folded has to stay decomposable (1), the classification it folds has to be total (2), and the order it was discovered in is what a document promises (3). I had decided (1) for the reading half only — after a regression, not by design — and never asked the same question of the debts.

## 1. `keptFor` slices the whole account

Option B, as you framed it: a view of some behaviors is owed the account those behaviors' readings come to. `BorderObligationPointAssessment.keptFor(behaviors)` keeps the readings a shown behavior made and *makes the debt again from them* — demand, item and carriage all follow from the readings, so nothing is trimmed while an answer folded from all of them survives. A debt no shown behavior reads is dropped. Who owes the line is not re-derived: a declaration owes it wherever the type is carried, and which behaviors a reader sees is no part of that.

`aViewOfSomeBehaviorsCarriesNothingAHiddenOneWentWithout` builds a module where one behavior's row is never read (`DoesNotComeBack.everythingAboutRowsOf`), so the debt is undecided over both readings and settled over the visible one. With the old half-slice the test fails with exactly your description: the view of `seen` carries `BorderValueUnreadable` naming `stalls/d.n`.

## 2. Totality is the compiler's job now

Both account producers switch exhaustively over the arms instead of asking `instanceof`, so a third arm stops the build in both places rather than falling out of both accounts.

The classifier can no longer be reached round: `OwedPoint` is a final class whose constructor is package-private to `partition`, so an `OwedPoint` carrying an arm `PointAttribution.of` never produced cannot be made outside. The structural test now reads `CALLS` and `REFERS` for the classifier — your point about method references was right, `PointAttribution::of` passed the old rule — and a second case forbids reaching the arms by `MAKES`/`REFERS`/`READS` outside the classifier. Both were verified by writing the two bypasses and watching them go red.

I did not widen the arm-reading rule to `CALLS`: calling `ownersIn` on a value whose type is already `TheDeclarations` is using what an arm answers, not deciding which arm it is. Tried it, it flagged `DeclaredBorders#compute`, and that is a false positive rather than a finding.

## 3. Discovery order

`Map.copyOf` replaced with an unmodifiable `LinkedHashMap`, with the reason: a set of reasons keeps discovery order so that two runs of one compile write one document, and a copy that did not would hand it causes in whatever order a hash gave.

## The whole/parts fixture is source-driven now

Your non-blocking item, and item 1 handed me the shape I could not find before: a row that is never read makes the value at a declaration-owed point unreadable, and the behaviors carrying the type are owed no row there — so the cause lands only in the declaration account. `aModuleHoldsWhatOnlyItsDeclarationsWentWithout` compiles that model, asserts that something is the declaration account's alone (so the rule is about something), and that the module holds it. Dropping the union makes it red.

`mvn -o test` green across the reactor.

kawasima:
Both are closed. Head is now `b1c437c28`, rebased onto current `develop` (`e9d594b6d`, #1119 included), and CI on that head is green: run 33019931849, success. PR state MERGEABLE / CLEAN.

## Root cause of the structural-test hole

The check's evidence was coarser than its claim. The rule is about *which member of which arm is reached, and how*; I compared a flat set of **method names**, because that made the expected value easy to write. An assertion whose evidence has fewer coordinates than its proposition cannot refuse what its name forbids — which is how `new TheDeclarations(...)` inside a method already allowed to hold `TheDeclarations::and` would have passed, exactly as you said.

So the check now compares the whole map of *what is reached, how* → *who reaches it*, with nothing folded away:

```
PointAttribution#of CALLS                → Border#owes
PointAttribution#equals CALLS            → OwedPoint#equals
TheDeclarations#and REFERS               → BorderObligationPointAssessment#across
TheDeclarations#ownersIn CALLS           → DeclaredBorders#compute, BorderObligationPointAssessment#across
TheReading#case NAMES                    → both accounts
TheDeclarations#case NAMES               → both accounts
```

Every entry carries a comment saying why that reach is legitimate. A reach nobody wrote down — any member, any kind, any caller — fails the comparison, so the rule no longer depends on my having thought of the bypass in advance.

Verified against both of your scenarios:

- `new PointAttribution.TheDeclarations(...)` inside `BorderObligationPointAssessment#across` → red, with two new keys (`TheDeclarations#<init> MAKES`, `… CALLS`) even though the method name was already allowed.
- downgrading `OwedBoundaryPoint#across` from the exhaustive switch to `instanceof TheReading` → red: `TheReading#case NAMES` and `TheDeclarations#case NAMES` each lose that producer, and a new `TheReading#instanceof ASKS` appears. So the exhaustive-ness is now fixed by the test and not only by the current source.

I also dropped the two older tests this subsumes rather than leaving three assertions of one fact.

## The rebase

Rebased rather than merged, to keep the branch at one commit. `mvn -o clean test` green locally on the rebased tree before pushing, and the CI run above is on `b1c437c28` with base `e9d594b6d`.

I agree with your reading of #1118: what its transport rule changes is which contributions arrive, and the accounting pipeline downstream of `PointAttribution` does not move with it.

PR #1151, opened 2026-08-28
Own the terms a run is held to where every run reads them

### LINKED ISSUE #1067: The terms a run is held to are owned by the package that runs it
## The shape today

#971 put the questions the language can only answer by running a program behind
`ProgramExecution`, and moved what those questions answer with into
`souther.compiler.observe` — an answer owned by the one implementation that produces it is an
answer a caller still compiles against.

The ask has one thing left in that position. `ExampleExecution` carries how long a reading may take
and what a run is counted against:

```java
public Deadline deadline()
public EvaluationPolicy policy()
```

Both are declared in `souther.compiler.examples`, which is where `ExampleVerifier`,
`FixtureReader`, `MemoryClassLoader` and the rest of the JVM implementation live. Neither is a word
of the machine — a budget in steps and a wait in milliseconds are terms of the compile — but an
execution that is not the JVM's and honours a budget has to name the example subsystem's package to
read one.

## What this is not

Not something #971 refuses. What that boundary may not carry is the store this compiler answers its
own questions in, a compilation, an artifact of emitted classes, a class loader, a generated class
and a bare `Object`. A `Deadline` is none of them, and
`AnExecutionThatIsNotTheJvmsCanBeWrittenTest` was deliberately narrowed to say only what the issue
says: it listed `souther.compiler.examples` for a while, and the stand-in passed that rule by not
reading the budget rather than by not needing the machine.

Not a package move to make a test go green. That is why it was left: a rule a test finds convenient
is not a reason to move a type.

## The question

Whether the terms a compile-time run is held to are still the example subsystem's.

They were its when it was the only thing that ran anything. What reads them now is
`ProgramExecution`'s ask, which is neither the example subsystem's nor the JVM's — and what sets
them is `Compilation`, through `Front.ExampleDeadline` and `Front.Policy`.

`souther.compiler.observe` already holds `Limits`, which is close enough to be worth looking at
before inventing a home.

## Done when

- it is decided whether these two are owned by `souther.compiler.execute`, by
  `souther.compiler.observe`, or by `souther.compiler.examples` still, and the reason is ownership
  rather than convenience
- if they move, `AnExecutionThatIsNotTheJvmsCanBeWrittenTest` refuses
  `souther.compiler.examples` and the stand-in reads a budget
- if they do not, that test's javadoc says why the package is not on the list

Follows #971.

kawasima:
Closes #1067.

## The decision

`ExampleExecution` carried two things declared in `souther.compiler.examples`, so an execution that is not the JVM's had to name the example subsystem to be held to a budget. The two are not the same kind of thing, and they get different owners.

- **`EvaluationPolicy` → `souther.compiler.execute`.** Terms: `stepLimit`, `recursionDepthLimit`, `outerTimeout`. What a run is held to, whoever runs it. Not `observe` — `Limits` there is about what an observation keeps of an answer, which is the other side of the run.
- **`Deadline` stays in `souther.compiler.examples`, whole.** Not split into an interface and its implementations. `given(Work, Callable)` is a way of running work, and `Work` is the example subsystem's vocabulary (`Row`, `Fixtures`, `Table`, `With`). That `crossingBackToTheCaller` needs the package-private `Handoff` is the evidence: a design that made `Handoff` public would be lifting an examples-specific mechanism into a general abstraction.
- **`workerStackBytes` leaves the terms.** Measured: its only two readers both built a `Deadline` with it, and `withWorkerStackBytes` had no callers. A number of bytes of a thread's stack means nothing to a way of running a row that has no threads. It is now `Deadline.DEFAULT_WORKER_STACK_BYTES`, held per compilation as `Front.WorkerStack`.

`outerTimeout` stays among the terms even though today's only reader builds a deadline from it. Its javadoc now says what reading it obliges an execution to: elapsed time, not work done — bounding fuel or instructions bounds the work and still owes a clock.

## How the JVM execution gets its deadline

Not off the ask. `JvmExampleDeadlines` sits beside `JvmProgramImages` in `execute.jvm`, and `Compilation` implements it. It is asked at the question rather than handed over once, because `withDeadline` is called after the implementation has been named.

`execute.jvm` naming `examples` is not a violation — package hierarchy is not the abstraction boundary. What must not name it is the portable half, and that is what is checked.

## What closes it

`NothingCrossingTheExecutionBoundaryIsTheMachinesTest` refuses `souther.compiler.examples.` on the boundary. Verified by mutation: putting `deadline()` back on `ExampleExecution` fails it with the ten types and the path each crossed on.

`AnExecutionThatIsNotTheJvmsCanBeWrittenTest` lists the package in `THE_MACHINES`, and `RecordingExecution` now divides the step budget over the reading instead of printing the object it got — so the rule is met by a stand-in that can be held to a budget, not by one that never opened it. Both mutations were run red.

`mvn -o test` green across the reactor (7440 in the compiler, 422, 330, 2271, …).

### DIFF
--- a/souther-compiler/src/main/java/souther/compiler/Compiler.java
+++ b/souther-compiler/src/main/java/souther/compiler/Compiler.java
@@ -208,14 +208,17 @@ public final class Compiler {
                                                java.time.Duration exampleBudget, Deadline deadline,
                                                EvaluationPolicy policy, ModulePath path) {
         Compilation compilation = Compilation.ofSource(source, defaultModuleName, path);
+        // The terms first and the wait after, because the wait is one of them: said the other way
+        // round, a caller that asked for both would have the policy put the default wait back over
+        // the one it asked for, and nothing would say so.
+        if (policy != null) {
+            compilation.withEvaluationPolicy(policy);
+        }
         if (exampleBudget != null) {
             compilation.withExampleBudget(exampleBudget);
         }
         if (deadline != null) {
             compilation.withDeadline(deadline);
-        }
-        if (policy != null) {
-            compilation.withEvaluationPolicy(policy);
         }
         compilation.measure(measure);
 
@@ -380,14 +383,15 @@ public final class Compiler {
                                               java.time.Duration exampleBudget, Deadline deadline,
                                               EvaluationPolicy policy) {
         Compilation compilation = Compilation.ofSources(sources, path);
+        // The terms first and the wait after, for the reason compilingSource gives.
+        if (policy != null) {
+            compilation.withEvaluationPolicy(policy);
+        }
         if (exampleBudget != null) {
             compilation.withExampleBudget(exampleBudget);
         }
         if (deadline != null) {
             compilation.withDeadline(deadline);
-        }
-        if (policy != null) {
-            compilation.withEvaluationPolicy(policy);
         }
         compilation.measure(measure);
 
--- a/souther-compiler/src/main/java/souther/compiler/Compiler.java
+++ b/souther-compiler/src/main/java/souther/compiler/Compiler.java
@@ -9,7 +9,7 @@ import souther.compiler.diag.Diagnostic;
 import souther.compiler.diag.msg.DeclarationMessage;
 import souther.compiler.diag.Located;
 import souther.compiler.examples.Deadline;
-import souther.compiler.examples.EvaluationPolicy;
+import souther.compiler.execute.EvaluationPolicy;
 import souther.compiler.meta.ModulePath;
 import souther.compiler.query.Acceptance;
 import souther.compiler.query.Adequacy;
--- a/souther-compiler/src/main/java/souther/compiler/examples/Deadline.java
+++ b/souther-compiler/src/main/java/souther/compiler/examples/Deadline.java
@@ -36,11 +36,20 @@ public interface Deadline {
      */
     long DEFAULT_WORKER_STACK_BYTES = 64L * 1024 * 1024;
 
-    /** The stack this JVM's settings ask a worker to be given, read the way {@link
-     *  EvaluationPolicy#fromSettings} reads the terms. */
+    /** The stack this JVM's settings ask a worker to be given, on the terms
+     *  {@link EvaluationPolicy#fromSettings} states: a setting that is missing, unreadable or not
+     *  positive leaves the default in place. */
     static long workerStackFromSettings() {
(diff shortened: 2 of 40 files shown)


kawasima:
Fixed. I took the three findings back to one cause first, because fixing them one at a time is what
produced two of them.

## The cause

#1067 asked which package declares each type. I answered that and stopped. Deciding a type's owner
does not make a value single-sourced, and every finding is the same shape — two places able to
answer one question, so the boundary states one thing and the implementation does another:

| the question | who could answer |
|---|---|
| this compile's `outerTimeout` | `Front.Policy` and `Front.ExampleBudget` |
| this compile's deadline | the seam, and `Output.deadlineOf` called directly |
| how a setting is read | `EvaluationPolicy` and `Deadline` |
| the wait the JVM keeps | the term that crossed, and a second reading of the store |

One rule closes all four: **each term has one input, and the implementation is handed it rather than
fetching it.**

## What changed

`Front.ExampleBudget` and `Output.exampleBudgetMs` are gone. `withExampleBudget` writes the wait into
`Front.Policy`, so the terms are what everything reads. Three inputs remain and none of them overlap:
`Policy` (the terms), `ExampleDeadline` (the arrangement, test-only, #1152's), `WorkerStack` (a JVM
resource).

`JvmExampleDeadlines.forThisCompile(Duration outerTimeout)` — your item 2, and it is what makes the
first change hold rather than merely be true today. `Output.deadlineOf(Db, Duration)` no longer reads
the wait from the store, and is no longer public: the adapter is its only caller, so there is one
route rather than a convention that there should be.

`Compiler` now says the terms before the wait. It called `withExampleBudget` then
`withEvaluationPolicy`, so folding the budget into the terms without this would have let the policy
put the default minute back over a caller's budget with nothing saying so — the same defect moved
one call up.

`positiveLong` was already private again in `273bb415e`, which landed before your review; the reason
matches yours.

## Evidence

`AnExecutionThatIsNotTheJvmsCanBeWrittenTest` gained the case you named. With `withExampleBudget`
made a no-op it reports `given up on after 60000ms` — the boundary's old lie, to the millisecond.

`mvn -o test` green across the reactor, 7441 in the compiler.

## Not fixed here

The unit misnomer behind my earlier `steps a row`: `Prepared.Examples.rows()` answers blocks while
`Hir.Example.rows()` answers rows. That is a `check` type with readers outside this change, so it is
#1155 rather than scope creep here.

PR #1177, opened 2026-08-29
Take the snapshot bump off the pull request path

kawasima:
The snapshot bump after a release is eleven version elements written by `bin/set-version.sh` and nothing else. `docs/releasing.md` had it opening a pull request like any other change to develop; there is nothing in it to review, and the number was decided by cutting the release.

Step 6 now commits to develop and pushes. Written as the one exception, so it does not read as a general loosening: a change to develop that carries a decision still opens a pull request.


kawasima:
Folding this into the pull request that updates `Templates.java` and `BuildPlugins.java` for the 0.2.0 plugins, rather than spending a pull request on a paragraph.

Issue #1164, opened 2026-08-28
A line drawn against a field read off a sum is read once per case, and the report stops


kawasima:
Found while re-measuring souther-examples against the compiler's develop. Measured on `c3dba0cbe`.
`joboffer` reported 62 gaps on `f0fd133cc` and now ends in an internal compiler error instead of a
report.

## What is wrong

A comparison that relates one position to a date read off a sum — where every case of the sum
spreads the data carrying it — produces the same border line once per case, and the fence that says
a line is read once fires.

```souther
module probe.twice

data ベース = { 期限: Date }
data 甲 = { ...ベース }
data 乙 = { ...ベース }
data 依頼 = 甲 | 乙

data 期限切れ
data 掲載可

behavior 検査する : (依頼: 依頼, 今日: Date) -> 掲載可 | 期限切れ
let 検査する (依頼, 今日) = {
    guard Date.daysBetween(今日, 依頼.期限) >= 0 else 期限切れ
    掲載可
}

example 検査する
    | "掲載できる" : (甲 { 期限 = Date("2026-08-10") }, Date("2026-08-03")) -> 掲載可
```

```
$ souther examples probe.sou
internal compiler error: IllegalStateException: one behavior reads AtLine[line=BorderObligationId[
  line=AuthoredLine[rule=Comparison[behavior=検査する, origin=CoverageOrigin[module=probe.twice,
  ordinal=1, lowered=0, kind=BINARY]], conjunct=0, facts=LineFacts[valueBelongsBelow=true,
  holdsAtTheValue=true, singles=false], narrowedWithin=[]], at=0], role=ON] twice at 検査する/今日,
  so the two readings of it are one
```

`souther compile` on the same sources is clean; it is the report that stops.

Two cases are enough, and nothing else in the shape is needed: no second guard, no helper between the
sum and the field, no `Int.abs`. Reading the field through a `match` instead avoids it — there the
dates come back `not derivable` and no line is drawn at all.

## What it looks like it is

The line is about `今日`, and what it is drawn against is `依頼.期限` — which the walk reaches once
under `依頼@甲` and once under `依頼@乙`. Both readings carry the same `AuthoredLine`, so they are
the same line by identity while being two readings by position, and the fence reads that as one
behavior having read it twice.

## Where it was found

`joboffer`'s `応募期限を検査する`:

```souther
behavior 応募期限を検査する : (依頼: 依頼, 今日: Date) -> 掲載可 | 期限切れ | 期限が先すぎる

[...]

Issue #1161, opened 2026-08-28
Report and strict count boundary occurrences instead of obligations


kawasima:
The report and a strict build account for a boundary per reading occurrence; a generation accounts
per obligation. So a strict build can refuse over a point a generation correctly says is already
answered by another reading of the same line.

This is the part of #1149 that its first pull request (#1162) left open: the canonical relation
from occurrences to obligations exists (`Adequacy.Obligations`, `BorderObligationPointAssessment`,
`Settlements`), and the report, the findings and the strict verdict do not read it yet.

## Measured

A guard on a name the cases of a sum spread, with rows written under one case:

```souther
data Base = { deadline: Int }
data P = { ...Base, x: Int }
data T = { ...Base, y: Int }
data Req = P | T

behavior check : (r: Req) -> Ok | No
let check (r) = {
    guard r.deadline > 10 else No
    Ok
}

example check
    | "on"  : (P { deadline = 11, x = 0 }) -> Ok
    | "off" : (P { deadline = 10, x = 0 }) -> No
```

`souther examples` prints

```
border      borders 2   coverage items 2/8
  ! no row is at the ON point check/r@T.deadline = 11 (comparison@14:22)
  ! no row is at the OFF point check/r@T.deadline = 10 (comparison@14:22)
  ! no row is at an IN point of check/r@P.deadline, 11 < r@P.deadline (comparison@14:22)
  ! no row is at an OUT point of check/r@P.deadline, r@P.deadline < 10 (comparison@14:22)
  ! no row is at an IN point of check/r@T.deadline, 11 < r@T.deadline (comparison@14:22)
  ! no row is at an OUT point of check/r@T.deadline, r@T.deadline < 10 (comparison@14:22)
adequacy: not satisfied
7 gaps marked `!`: what a strict build refuses over.
```

and the same run with `--generate --boundaries` ends

```
// no row offered for `check/r@T.deadline = 11` in `check`: this line is answered by a row elsewhere
// no row offered for `check/r@T.deadline = 10` in `check`: this line is answered by a row elsewhere
```

The ON and OFF points are one obligation each, answered by the rows under `P`. The report marks
them as gaps, the strict build refuses over them, and nothing offers a row for them. A model in
this state cannot be made to pass.

## Cause

`About` has two arms for a point of a border at two grains. `APointOfADeclaredBorder` carries a
`BorderObligationPointAssessment`; `APointOfABorder` carries an `OwedBoundaryPoint`, which is one
[...]

PR #1181, opened 2026-08-30
Ask a reading's target which behavior read it

### LINKED ISSUE #1167: A reading says its behavior twice
Found while reviewing #1166, which closed #1164. Nothing here is wrong today; it is the shape #1164
was, one size down.

## What is doubled

A reading of a line is a behavior and where the line was read:

```java
public record Reading(String behavior, souther.compiler.partition.BoundaryTarget target)
```

The target already says which behavior it is of. Every quantity carries one:
`OfACoordinate` holds an `AxisId`, which is `(behavior, term)`; `Apart` and `OverAForm` each hold a
`String behavior` of their own. So a reading is

```
Reading
├─ behavior = "check"
└─ target → quantity → behavior = "check"
```

and `Reading.of(String behavior, Border line)` does not check that the two agree.

## Why it is not a defect yet

Nothing constructs the pair from two sources. `Adequacy.Obligations` walks a map keyed by behavior
whose values are that behavior's own lines, so the two halves come from one place and cannot
disagree. Every other caller reaches `Reading.of` with the behavior it is already reading.

What it is, is the structure #1164 turned out to be: one question with an answer in two places, held
together by the callers happening to ask it once. #1164 was two answers to which readings are one
line; this is two answers to which behavior a reading is of, and it fails the same way — silently,
whenever some later caller composes the pair from two places.

## What it would take

```java
record Reading(BoundaryTarget target) {
    String behavior() {
        return target.behavior();
    }
}
```

which wants `behavior()` as a common contract on `BorderQuantity` — three variants that each hold it
in their own shape and none of which answers it today. That is an API decision about
`BorderQuantity` and not about what a reading is, which is why it was kept out of #1166: the
correctness fix there is about where a line was read, and normalising the behavior would have mixed
two reasons to change into one diff.

Asserting the two agree in the constructor is the smaller move, but a `behavior()` written only to
assert against is a `behavior()` that may as well be the one answer.

kawasima:
A `Reading` of a line held `behavior` beside a `BoundaryTarget` whose quantity already names the behavior it is an input of. #1167 records this as the shape #1164 was, one size down: one question answered in two places, held together only by every caller composing the pair from one source.

Now `BorderQuantity.behavior()` is the contract for all three shapes, `BoundaryTarget.behavior()` passes it through, and `Reading(BoundaryTarget)` answers `behavior()` from the target. `across` checks that the behavior a line was filed under agrees with the line's own answer.

Observable behavior does not move; this is the prerequisite for #1161, which orders and publishes readings and needs one answer to which behavior a reading is of.

Closes #1167

### DIFF
--- a/souther-compiler/src/main/java/souther/compiler/partition/BorderQuantity.java
+++ b/souther-compiler/src/main/java/souther/compiler/partition/BorderQuantity.java
@@ -31,7 +31,8 @@ import java.util.Map;
  * probe method, an assessment path and a report arm apiece.
  *
  * <p><b>Sealed, so a quantity added is one this file answers for.</b> Sealed here and nowhere else:
- * what a variant costs is the four answers below, and nothing downstream gains an arm.
+ * what a variant costs is the answers this interface asks for below, and nothing downstream gains
+ * an arm.
  */
 public sealed interface BorderQuantity {
 
--- a/souther-compiler/src/main/java/souther/compiler/partition/BorderQuantity.java
+++ b/souther-compiler/src/main/java/souther/compiler/partition/BorderQuantity.java
@@ -100,6 +100,11 @@ public sealed interface BorderQuantity {
         }
 
         @Override
+        public String behavior() {
+            return axis.behavior();
+        }
+
+        @Override
         public String named() {
             return axis.toString();
         }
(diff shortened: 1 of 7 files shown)


kawasima:
Both points taken, the first by going one step further back.

The fence in `across` guarded a doubling rather than removing one: its input was a map keyed by behavior, and every line in it already answers which behavior read it. That is the shape #1167 names, one level down from `Reading`. So `across` now takes `List<BorderAssessment>`, `Obligations` hands the lines over flat, and the check is gone with the key it was checking — there is no unreachable branch left to pin.

The `BorderQuantity` javadoc no longer counts its answers.

Issue #1168, opened 2026-08-28
A guard boundary is dropped from an arm-reachability proof instead of the values reaching its comparison


kawasima:
A guard's border is kept or dropped by asking whether either outcome of its
comparison is one the paths prove nothing takes. That question has a different
answer from the one a border is about, and the two counterexamples below have
opposite right answers and the same reading today.

## The two cases

| model | values reaching the comparison | the line it draws | right answer |
| --- | --- | --- | --- |
| `guard a.value < 5000` above `guard a.value < 6000` | `[0, 5000)` | 6000 | **drop** |
| `invariant value >= 10`, body `if level.value < 10` | `[10, …]` | 10 | **keep** |

Both have exactly one unreachable outcome, so `PathReachability.Answers.dividesNothing`
answers `true` for both, and `Partitions.measureAt` disposes of the `LineEvidence` in
both.

The first is right. A row at 6000 takes the *first* guard's `else` and never arrives at
the second comparison, so the class `6000 <= x <= 1000000` the line would make is a
class of nothing and the two rows generated for it prove nothing. `AGuardTheGuardsAboveItRuleOutIsProvenTest.theModelAsksForWhatTheModelWithoutThatGuardAsksFor`
holds this, by comparing the generated rows against the model with that guard removed.

The second is wrong. A row at 10 is an ordinary `Level`, it does arrive at the
comparison, and it takes it the other way — so there is a row to write at the line and
the border should stand with the point on the impossible side not owed.
`ThresholdNormalizationTest.aComparisonTheDeclarationsCannotSatisfyStillDrawsItsLine`
already states this as a law and passes: it reads the intermediate `Axis`, where the
cut at ten carries both the clause's origin and the comparison's. The disposal happens
later, so the two readings of one model disagree about whether that line exists.

## What is actually being asked

> A comparison boundary remains exactly when the feasible values reaching that
> comparison reach the boundary it draws; reachability of either outcome alone does not
> decide this.

`dividesNothing` returns a boolean derived from arm reachability and drops the values
on the way. The values are what separates the two cases above, and the walk that
computes the reachability already holds them — `Known` carries a `ConstraintState`, and
`NumericDomain.boundsOf` reads an interval off it. What is missing is that they are
never published: `PathReachability.Answers` maps a control point to a `Reachability` and
nothing else.

`Border.reaches` already asks exactly this question of a `NumericDomain.Bounds`, against
what the declarations leave. Given what reaches the comparison, it is the same question
[...]

Issue #1138, opened 2026-08-27
A path environment and a reading of the input are one type, and only a runtime guard keeps them apart


kawasima:
Separated out of #1129, where the seam was opened and closed only as far as that change needed.

## What is wrong

`InputReads` answers two questions. Which location a name in a tree stands for — settled by the
parameters, the bindings on the way, and the case an arm selects — and what the reading of the input
says about that location. The first is what finds the paths a behavior's measurement names; the
second is the reading those paths are used to build. They cannot both be asked at once, and #1129
turned on separating them.

They are separated by discipline rather than by the type. The record still holds the reading as a
component with a public canonical constructor, so an environment with no reading in it is one of
these:

```java
public static InputReads ofParameters(Map<BindingId, String> parameters, ElementBindings elements) {
    return new InputReads(null, rooted(parameters), Map.of(), elements, false);
}

public InputDomain read() {
    if (read == null) {
        throw new IllegalStateException(
                "a path environment built before the reading was asked for the reading");
    }
    return read;
}
```

`InputPaths` closes the one door #1129 needed closed: `InputDemand.of` takes that interface, so the
code collecting paths cannot reach `read()`. Nothing closes the others. A caller that holds one of
these as an `InputReads` and hands it to any of the readers that ask for the model gets a
`IllegalStateException` at run time, from a call that compiles.

## What is not wrong

Nothing today does this. Both places that build a pre-reading environment narrow it immediately —
one passes it straight into `InputDemand.of`, the other assigns it to an `InputPaths` — and every
`reads.read()` in the tree is reached only from an environment built with a reading. This is a
mechanism that permits the defect, not a defect being reported.

## Measured

- `InputReads` is named in 21 files.
- 64 calls in `main` go to the environment's own methods (`pathOf`, `cameFrom`, `elementAt`,
  `meaningOf`, `insideArm`, `and`) — the half that has nothing to do with the reading.
- Two places build an environment with no reading, both in `Adequacy`. One more builds one with a
  reading, in `EnsuresThresholds`.
- Nothing checks that the two are not mixed: there is no structural test over the sources, and the
  only thing standing between them is that `read()` throws.

## Done when

- an environment built before there is a reading has no way to be typed as one that answers about
  the model, so mixing them does not compile
- the reading is not a nullable component of the environment, and `read()` has no guard to fail
- what a name stands for and what the model says about that location are asked of different values,
  and the 64 call sites above go to the first without carrying the second
[...]

PR #1182, opened 2026-08-30
Count, mark and refuse over what a row is owed for, not over where the line was read

### LINKED ISSUE #1161: Report and strict count boundary occurrences instead of obligations

kawasima:
A line is owed one row however many of a behavior's positions read it. The offering worked that way and the report did not, so a strict build refused over points the generation correctly reported as already answered — a model nobody could make pass. Measured on the model in #1161:

```
before:  border  borders 2  coverage items 2/8   → 7 gaps, and
         // no row offered for `check/r@T.deadline = 11`: this line is answered by a row elsewhere

after:   border  borders 2  obligations 2/4      → 3 gaps
           ! no row is at an IN point (comparison@14:22)
               · read as check/r@P.deadline: in 11 < r@P.deadline
               · read as check/r@T.deadline: in 11 < r@T.deadline
           · the ON point (comparison@14:22) is answered, and not at every reading of the line
               · read as check/r@T.deadline: = 11
```

## The accounting

- `Adequacy.BodyBorders` is the behaviors' projection of the one obligation relation, and account membership is one named question (`belongsToBehaviorAccount`). The findings, the header count, the strict verdict, the offering and the editor lens all read it; none re-derives it from the readings. `DeclaredBorders` is the declarations' projection of the same relation.
- `About`'s two border arms share `ABorderObligation` and one grain. What differs is whose the point is, and whether anybody wrote a word for what it is on.
- `coverage items` is retired rather than reused: `borders` stays the lines at coordinates and `obligations` is the work, and the two are not one multiplied. `excluded` counts occurrences, so it leaves that line and stays in the detail. The declarations keep their own count — a `data` clause's line is owed once for the module, so no behavior is short of it.

## What a point says, and what it does not

A point names which of the four it is and which rule drew the line. It names no quantity, because writing where a point is takes one and a quantity belongs to the coordinate that read the line — and a level written from the level alone comes out true and unreadable (the `ON` point of `from < to` is a distance of −1). Every word an author acts on is said under the mark by the reading whose word it is. A line a declaration owns is the exception the author makes: it is on something they wrote.

Where the point is is published as identity instead. `obligationId` — the authored line, which of its rule's lines, what it says about its own value, what narrowed it, the level, the role, and where the run beside it stops — is carried by both accounts' obligations and by every finding about one, so a consumer joins them. On a model whose two cases stop the run in different places: 6 distinct display keys, **7 distinct ids**, and every finding joins exactly one.

## Schema 9

`partition.boundaries` stays occurrence geometry. `obligations` is one shared definition used by `partition.obligations` and by `declarations[].obligations`, publishing every reading occurrence with its multiplicity kept — a bag and not a set, since two entries that print alike are two readings and what tells them apart is where the line was met, which this document does not carry. The identity's shapes are discriminated the way `ruleId` already was, so the schema refuses what its prose forbids, and the order a level is counted on is written in this document's words rather than as a Java record's `toString`.

## Measured, not argued

- Over every corpus checked in: 208 obligations, 91 with more than one reading, **0** where two readings would render the same detail line — including in the `BETWEEN_POSITIONS` shape the type warns about, and in a synthetic `OVER_A_FORM`. So no grouping machinery. The spec says the corpus reaches none, not that none can exist.
- **8 of those 91** had a level that differed by reading (one crm point has 25 spellings), which is why no obligation-level word is written from a quantity.
- **13 of the 191** obligations with known writability have a reading whose own is not, which is why the count and the `·` diagnosis under it read one body of evidence.

`mvn -o test` green over the reactor.

Closes #1161


kawasima:
Self-check found one thing I had patched at the site rather than at the cause, and it was live.

The header counts obligations whose writability is known; the `· not known to be writable` lines under it were said per reading. Those come apart exactly when one reading proves a point writable and another does not — the point is counted above and printed below as not known. Measured over every corpus checked in: **13 of the 191** obligations with known writability have a reading whose own is not, all in `crm`. I had reverted that line to the occurrence form to get tests passing, which is how it got in.

Both are read off the same evidence now, with each reading's search under the point. That exposed the same gap once more: a line a declaration drew is in no behavior's account, so its unpromised points were printed nowhere — the declarations now say them where they are counted.

Worth noting for the review: most of the test churn in this PR is `contains` over the rendered report. That form did not catch the defect above — measuring the corpus did — and it went red about twenty times for wording alone. What actually holds the property is that the report, the findings, the strict verdict and the offering all read one projection (`Adequacy.BodyBorders`); they cannot disagree by construction. I have not rewritten the existing prose assertions, only followed them.

kawasima:
Both P1s and the P2 are fixed. They turned out to be one cause, not three: **the obligation published words rather than its identity, and a word for a level can only be written on a quantity — which is a reading's.** That makes the words neither unique (P1-a) nor the obligation's own (P1-b).

**Measured before fixing.** Over every corpus checked in, of the 91 obligations with more than one reading, **8 have a `level()` that differs by reading** — so this was not only the synthetic `Apart` model. One crm point of one line has **25 different spellings**; `met.firstEntry()` picked one by walk order. The same 8 differ in `against(axis)`, which the review did not name and which is the declaration path (`DeclaredDebt.said()`) — I would have missed it fixing `level()` alone.

**What changed**

- `Level` writes itself from what it holds — a place on its carrier, or the number. Nothing asks a quantity, so nothing asks a reading. A distance's point now says the distance (`at 0`, `at 1`) rather than `r@P.deadline + 1`.
- `against(axis)` still needs a quantity to write a band, so it asks **every** reading and throws where they disagree, the way `asked()` and `foundAt()` already refuse two readings that are not one point. No representative is chosen, and an accidental agreement became a checked one.
- `obligationId`, built from `BorderObligationPoint` itself — the authored line (rule, conjunct, what it says about its own value, what narrowed it), the level, the role, and where the run beside it stops — published on `partition.obligations[]` and on every `boundary_unmet` / `domain_point_uncovered` finding. On your far-sides model: **6 distinct old join keys, 7 distinct ids**, and all 4 findings join exactly one obligation.
- `BorderAssessment.Point`'s javadoc no longer claims the count and the finding read that list.

**Also found while merging develop.** Schema 9 was copied from schema 8 before `#1168` added `nothing_arrives_at_the_rules_line` to it, so my copy was missing that word and its sentence. Fixed, and I checked every shared `$defs` — schema 9 now differs from 8 only in `findings` and `partition`, which are the two I meant to change.

`mvn -o test` green over the reactor; conformance documents regenerated and verified.

kawasima:
Self-check on my own last round found one more, and it was mine.

Taking the level off the readings, I left it written from the `Level` alone. That is reading-independent — and unreadable. The `ON` point of `guard from < to` came out as:

```
! no row is at the ON point at -1 (comparison@10:18)
    · read as book/d.from: = d.to - 1
```

−1 is the distance on the quantity `from - to`. No position holds it and no author can write it. I had checked that the parent no longer chose a reading and stopped there; I did not judge what it then printed.

**The cause, said properly:** writing where a point is takes a quantity, and a quantity belongs to the position that read the line. So a point has no such word — not one it can borrow from a reading, and not one it can synthesise. Your own review said this first: what cannot be written without choosing a quantity does not belong on the parent. I kept a level because it happened to be renderable.

Both surfaces now say which of the four points it is and which rule drew the line; every word with a quantity in it is said under it by the reading whose word it is.

```
! no row is at the ON point (comparison@10:18)
    · read as book/d.from: = d.to - 1

! no row is at the ON point (comparison@14:15)
    · read as check/cap: = r@P.deadline + 1
    · read as check/cap: = r@T.deadline + 1
```

One sentence shape, no branch on data, no agreement test, nothing chosen by a walk. The document keeps where the point is inside `obligationId`, where it is structure and not prose, so nothing is lost — the separate `level` string is gone from `partition.obligations[]`.

`mvn -o test` green over the reactor.

kawasima:
All five are fixed, and they were two causes rather than five.

**P1 — the declarations section was handed the findings while the page beside it was handed the account.** That is this issue's own shape, one surface over: a line a row already stands at raises no finding, so an answered obligation vanished from the document and a finding's `obligationId` had nothing to join to. Both sections are written by one writer now, from the account, with the declaration's own word for what the line is on carried alongside. Verified on the checked-in corpora: catalog publishes 10 declaration obligations and staffing 6, and every finding carrying an `obligationId` joins exactly one published obligation across both arms.

**The three schema findings were one cause: the identity's shape was written twice** — a serializer and a schema — with the agreement between them stated in prose. So the schema admitted a point against the line carrying a `region`, a `a_count` level carrying a `carrier`, and a finding of any kind carrying an `obligationId`. It is discriminated now, the way `ruleId` already was. The carrier is written in this document's own words; `"carrier": "Whole[]"` was a Java record's `toString` standing in a published identity, and is now `{"kind": "whole"}` — thank you for catching that one, it would have frozen a class's shape into the contract.

`Level.written()` is deleted. You are right that it is the abstraction the previous commit's own reasoning discarded.

Spec and this PR's description now match the implemented wording.

`mvn -o test` green over the reactor; conformance documents regenerated and verified.

kawasima:
All four are fixed. Two of them, plus one the review did not name, were the same defect — and it is the one this PR had already removed from the point, left standing at the other identities on the same boundary: **an identity published as the words a reader sees.**

- The ordinal carrier named its enumeration with `name()`, so two modules' `Status` are one carrier inside `obligationId`. You are right that this is not a collision to wait for.
- The declarations were grouped by `subject().named()`, so `Cap or Held` stood where the schema promises a declaration's name, and two owner sets a report writes alike would be one entry.
- **Not in your list:** `narrowedWithin` joined a module and a name with a dot, and a module name already contains dots — `example.line.Amount` cannot be taken apart again.

So a type is written in one place now (`typeId`: the module and the name, with the language's own types saying so), used by the ordinal carrier, its cases, and the narrowing. The declarations are keyed on `List<TypeSymbol.AtModule>` and publish `owners` structurally; `name` remains, described as a rendering and never a key. Checked in the regenerated corpus: `owners: [{"kind":"declared","module":"conformance.catalog","name":"Sku"}]` beside `name: Sku`.

**The schema findings were the other cause**, the one I only half closed last round: the writer and the schema are written twice and their agreement was prose. Now constrained — `axis`/`against` exactly where a declaration owns the line (at the two reference sites, since the shared definition cannot know whose account it is), `hit` exactly where the measurement has a value, `reason` exactly where it has none, `writableBecause` matching its verdict, and at least one reading per obligation.

On `set`: you are right that the words and the shape contradicted. The readings are a **bag** — two entries that print alike are two readings, so a consumer may count them and may not merge them — and the description says that rather than claiming losslessness the shape does not have.

Stale schema prose (`subject`'s example and its `point, level and rule`, the `partition.obligations`-only join, `declarations` as shortfall-only) is corrected.

`mvn -o test` green over the reactor; conformance documents regenerated and verified.

kawasima:
Fixed, and thank you for the ordering catch — it is the last member of the family this round has been about.

**First, a correction on what you were looking at.** The remote head was `61d611afe` when you reviewed, not `01742aaa`: the merge with current `develop` was already pushed, so the `dirty` state and the un-updated schema prose you saw were one commit behind. It is `a50ef4b98` now.

**P1 — the cause is that a set was held as a list built in arrival order.** Who owes a point is a set: two readings of one point can each name a declaration the other does not, and the point is owed to both. `TheDeclarations.and` was a first-seen union, so the order was the walk's — and that order is read: what a report calls the pair, which module keeps the account, and (since I made the document group by the owners last round) which entry two obligations are gathered under.

Fixed where the value is made rather than at each reader: `TheDeclarations` puts its owners in the declarations' own order on construction, so `ownersIn`, `FindingSubject`, the human `A or B` and the JSON grouping all read one answer. Sorting in the writer would have left the other four reading the walk.

The regression is on the value, not the page, and I checked it by mutation: without the sort it fails with `[Cap, Held]` vs `[Held, Cap]`.

One existing test had pinned `Lim or Held` — the defect itself as an expected value. It is `Held or Lim` now, with the reason written down. `Cap or Held` in `ARunIsOwedToWhoeverMovedWhereItStopsTest` was already the declarations' order and is unchanged, which is the other half of the evidence.

**The rest is prose catching up with the code**: the finding's `subject` example and its `point, level and rule` sentence, the join target (`declarations[].obligations` for a declaration finding), `set` → bag in the Java comment, and `owners`/`narrowedWithin` constrained to `kind: declared`, which is what `List<TypeSymbol.AtModule>` already means. The PR body's "losslessly" is corrected to say multiplicity-preserving.

`mvn -o test` green over the reactor, on current `develop`.

kawasima:
This review is the previous one against a stale snapshot — the head has moved twice since. GitHub now reports the PR head as `a50ef4b98752`, `MERGEABLE` / `CLEAN`, with `develop` (`ff7eed574`) merged and CI green on that head.

Checked each item against the current head rather than answering from memory:

| item | on `a50ef4b98` |
| --- | --- |
| P1 canonical owner order | `PointAttribution.TheDeclarations` sorts on construction (`owners.stream().sorted().distinct()`), so `ownersIn`, `FindingSubject`, the human `A or B` and the JSON grouping read one answer. Regression is on the value and was checked by mutation. |
| schema `subject` prose | no longer says `ON point at 11` or `point, level and rule` |
| `obligationId` join target | names `declarations[].obligations` as well |
| Java `set` comment | says bag, with why |
| `owners` / `narrowedWithin` | constrained to `kind: declared` |
| merge + CI | `develop` merged, green |

**One was genuinely still stale and is my error**: the PR body still said "publishing every reading losslessly". I claimed in a comment that I had corrected it and had only corrected the schema. It now says multiplicity-kept, in the words you suggested. Thank you — that is exactly the kind of claim that outlives the thing it describes.

On the ordering fix itself, since it is the substance of this round: the cause was that a set was held as a list built in arrival order. Sorting in the JSON writer would have left the other four readers — `ownersIn`, the finding's subject, the human wording, the generated-row notes — still reading the walk, so the order is given where the value is made. An existing test had pinned `Lim or Held`, which was the defect as an expected value; it is `Held or Lim` now and says why, while `Cap or Held` elsewhere was already the declarations' order and is untouched.

kawasima:
The `@param` is fixed, and current `develop` (`2f68bd854`) is merged. Head is `7bb96e78e`.

You were right that it is worth a commit of its own: the line said the owners are kept "in the order they contributed", which is the exact thing the change removed, and it sat directly above an implementation comment saying the opposite. A contract that outlives its mechanism is the defect class this whole PR has been about, so leaving it would have been the same mistake written in prose. It now says: in the declarations' own order and never in the order they were met, because who owes a point is a set.

**One thing from the merge worth flagging**, since it touched a call this PR changed. `Coverages.of` conflicted because both sides had removed a *different* argument — I had dropped `lines` when the account left `PartitionEvidence`, and `develop` (#1138) had dropped the `InputDomain`. The merged signature takes neither, so the resolution drops both rather than restoring either. `mvn -o test` is green over the reactor on that resolution.

PR #1188, opened 2026-08-30
Hand a behavior's examples to an output outside this compiler

### LINKED ISSUE #1121: What a behavior's examples said does not cross to an output outside this compiler
Found while designing `souther-wasm`, a WebAssembly output that reads nothing but
`CheckedProgram` — the position `souther-program-api-test` already stands in.

## What an output can do today, and what it cannot

It can emit. Declarations, signatures, implementations, `Core`, `Composition` and
the kernels all cross, and #1088 / #1094 / #1095 / #1105 closed the last places
where it had to work an identity out for itself.

What it cannot do is show that what it emitted means what the language says it
means. Its only oracle is a test it writes in its own artifact, over inputs it
picked itself. Two outputs of the same program then agree about the program to
the extent that whoever wrote the second one's tests thought of the same cases —
and neither the set that was walked nor the set that was not is stated anywhere.

## The language already decided this, twice

A behavior's `example` rows are what the language says it answers, and running
them is not a test of the compiler: a program whose rows do not hold is not
accepted (§examples, ADR-0105). The inputs and the expectation are already
written down, already bound, already run.

A value already has a form that belongs to no loader. `NeutralValue` is "one
value in the form a derived decoder reads, holding nothing that came from any
loader's classes", and `Handed` exists precisely because "the two ends of an
application are not always in one loader" — one face for an answerer holding
this compile's classes, one for an answerer whose classes are not this
compile's. An answerer that is not on the JVM at all is the same question one
step further out.

## What does not cross

`CheckedBehavior` carries a name, a signature, an implementation and what it
declares of its answer. It carries no rows, and its javadoc lists "what its
examples said" among what will grow.

`RecordedRow` is not a way in either. It is `souther.compiler.examples`, and it
holds a `Hir.ExampleRow` — `souther.compiler.ast.` is on the list
`NothingReachableFromACheckedProgramNamesTheCompilerTest` refuses.

## The part that needs deciding rather than moving

What a row states is not simply a value. `Asserted` is what the row wrote, in the
names and forms the row wrote it under; `ObservedValue` is what a value turned
out to be; `ValueMatch` decides whether the two are the same value, and its
javadoc says why that is not equality over run-time values — a written
`Set.fromList([1])` at a `List` is not the sequence a list is, and two values
differ when their types differ as much as when their contents do.

So handing an output a pair of values to compare hands it a decision the language
already makes, and it would be free to answer differently about the same row.
The shape that does not is the one #1088 and #1094 arrived at: the output states
what its answer was, in the neutral form, and asks whether the row holds. Nothing
about that needs a `Db` or anything else the snapshot refuses to carry — it is a
comparison the language owns, asked with values rather than performed on them.

## A row is more than its inputs and what it states

A row runs against a bound implementation (ADR-0105), and where the behavior it
is a row of depends on something injected, what stands in for that dependency is
part of what makes the row runnable. Handing an output the inputs and the
assertion and letting it apply the behavior does not reproduce the row: the
stand-ins are the rest of the obligation, and an output whose dependency is an
import has nothing to answer that import with.

So a row can fail to cross for a reason that is not about a value at all, and
the two must not look alike from the outside. What must never happen is a row
that cannot be reproduced arriving as no row: an output would then read "this
behavior has no rows" and count a set it never walked as one it walked and found
empty.

## What this is not

Not running rows. Nothing here executes anything; what an output does with a row
it got is the output's business.

Not the decoder / encoder external representation. That is how a value is written
for the world outside a program, and a row is not written for the world.

Not `Handed`'s two faces. The JVM asymmetry it states is the JVM's; what crosses
here is one form, not a pair.

## Done when

Every recorded row crosses. A row an output can reproduce crosses with its exact
inputs and with the language-owned way to ask whether an answer holds. A row it
cannot reproduce crosses with its identity and the reason, and is never
indistinguishable from a row that was never written.

- an output holding a `CheckedBehavior` gets the rows recorded for it, and what
  it gets names nothing on that list
- whether an answer satisfies a row is asked of the language rather than decided
  by the reader
- a value that crosses is exact: what a limit stopped and what could not be read
  back are not values, and a row carrying one crosses as unavailable rather than
  as a row holding a shortened value
- the walk in `NothingReachableFromACheckedProgramNamesTheCompilerTest` still
  passes, and reaches whatever carries a row

kawasima:
Closes #1121.

An output holding a `CheckedProgram` could emit it and could not show that what
it emitted means what the language says it means. Its only oracle was a test it
wrote in its own artifact, over inputs it picked itself. A behavior's `example`
rows are the statement it was missing: already written, already bound, already
run, and refused by the language where they do not hold.

## What crosses

    CheckedBehavior.rows()  : List<CheckedRow>
    CheckedRow              : identity() / at() / statement()
    CheckedRow.Statement    : Reproducible | NotReproducible
    Reproducible            : states() / holds(ObservedValue) : Verdict
    NotReproducible         : why() : RowStatement.NotStated
    RowStatement            : Stated | NotStated(RequiresStandIns | Incomplete | NotRead)
                              | StoppedBeforeItsValues  (a compile's own, and never handed over)
    Verdict                 : Held | NotHeld(Mismatch)

Whether an answer keeps a row is asked of the language rather than decided by the
reader — and only of a row that hands over values. A row that cannot be
reproduced crosses as itself and says why: a behavior that takes something
injected states what has to stand in for it, and a value the limits would not
keep says which of the row's values it was. What must never happen is such a row
arriving as no row, and that is counted rather than sampled.

## Where the vocabulary lives

`Asserted` moves beside `ObservedValue`, and the comparison, the places inside a
value and the verdict move with it. The compile that reads a source text and the
snapshot an output is handed are two readers of that vocabulary and are not
written in terms of each other, which the class files are now held to:

    examples -> observe          observe -X-> examples, program
    program  -> observe          examples -X-> program, program -X-> examples

The comparison names no compiler. What it needs of the declarations is one
question — where a type reads a field — and each side answers it from its own
declarations.

## What the measurements changed

**An optional was charged a node it is not.** The walk that reads a value back
charged for the `Some` envelope, so two values that are one value were bounded
differently depending on whether the position they stood at was optional. Found
by holding the walk and the predicate to the same boundaries.

**A second gathering of a module's rows.** The assembler first walked the sources
itself, which decides for itself what a source that could not be evaluated means.
That gathering is one answer, and it was inside the reading a measurement is
counted over — asked only where a build measures, so a snapshot would have
carried no rows at all at the default level. It is a key of its own now, gathers
in the place it is named for, and the measurement is one reader of it.

**Checks spelled by a name.** `product.fields()` and `rowsOf` name things other
types are also named for, and both fired on code that does not break their rule.
The first is read off the constant pool now — what a class does rather than what
it is written like — and the second names its owner. A third could not see a type
named only as a type argument, since that is written in a signature and not in a
descriptor, so a dependency inside a collection was invisible to every rule read
off the class files.

## What a published value owes

The values a snapshot hands over owe what the values beside them owe, and saying
so once is not the same as being held to it. Every type reached from what a row
states is asked whether it copies what it was handed; a part in a shape the check
has no sample of stops the walk rather than being passed over. What a statement
of values means is the statement's own — made under `Limits.DEFAULT`, not under
limits a caller chooses — so one type does not mean two things.

Claims about the language are held to checks rather than to javadoc: that a
primitive standing as a case is not a name a row can write, that the same answer
written as a value is a row like any other, and that every way of stopping a row
before its values is a way the program is refused. A reading that never reaches a
row at all is accepted, and such a row crosses saying why nothing read it.

## Verified

Full reactor green (10,963 tests). Mutations measured, not assumed: the five
limits' boundaries, the two grains a statement is compared at and the unreadable reason,
both arms of the reading that says which case an answer is, dropping a row that
states no values (4 rows became 2), a `program -> examples` reference through a
type argument, a second reader of a product shape's fields, and a published value
that keeps its collection instead of copying it.

## Left over

A row is compared, in a snapshot, against what `CheckedData` publishes of a
field; the compile compares against its own reading of the declarations. The two
should agree everywhere and are shown to agree end to end for a declaration-driven
`Set`. Holding them to each other is worth an issue of its own.


kawasima:
Thank you — four of the five are defects and are fixed. The first is not, and I have the measurement.

## [P0] measured, and the premise does not hold

A row cannot name a primitive as a bare case. `-> Int` is refused where it is written:

    13:36 E1023: I cannot find a value named `Int` in scope.

So `caseOnly()` never answers with a `TypeSymbol.Primitive`, and the old comparison never
reached one either — `NeutralForm.typeOf` reads a live value's class through
`SoutherJvmAbi.valueTypeCandidate`, which answers nothing for a boxed `Long`, so `resultArm`
was null for a primitive answer whichever side asked. `-> None` is not writable either: an
optional may not be a behavior's answer (E1402). A primitive answer written as a **value**
(`-> 3`) goes through `TheValue` and is compared by `ValueMatch`, which is untouched.

What `caseOnly` can answer with is a declaration, and `declaredAs()` answers for both kinds of
those. Rather than leave that as my argument, it is pinned:
`aRowNamingACaseThatCarriesNothingIsHeldToIt` runs a unit case end to end, and both arms of
`declaredAs` are mutation-checked (dropping either turns the acceptance suite red — the unit
one turns the *compile* red, E1905).

If there is a spelling I have not found that reaches a primitive case, I would like the model
that writes it; I could not construct one.

## The other four, at the cause rather than the site

The four are two causes, and each cause produced two of them.

**A published value did not owe what the values beside it owe.** `Asserted` kept the
collections it was handed and `Stated` could be built by anyone. Both are closed where the
values are: the collections are copied as they are taken (order kept — it decides which field a
mismatch is reported at), and `RowStatement.of(inputs, expects, kept)` is now the one place a
statement of values is made, so `Stated` cannot be written past the reading that decides what a
row states. Making it a class cost it `equals`, which the query graph stops work on — your own
`EverythingAnAnswerHoldsMeansSomethingTest` caught that within the same run, and it says what it
is equal by now.

**Where the mechanism resisted, the question moved and the design did not.** `holds` became an
operation with two states it may not be called in because a row could not hold what asking takes
while it was being built. That is the totality an algebra had, handed back as a rule a reader
must have read. `CheckedRow.statement()` is now `Reproducible | NotReproducible`, and only the
first answers `holds` — no exception, and a reader that never looked cannot get a verdict about
a row that stated no values. `RowStatement` gained the matching seam (`Stated | NotStated`) so
the two are one decision, not two parallel sums. The gathering moved to the key that is named
for it; `Adequacy` reads the answer.

Also: the reason larger than one behavior is said once rather than in every entry as well, and
what a row states of the answer is read once with the ensures check read from it.

`Side.TheReading` is gone. It restated, in a weaker vocabulary, what the stage and the phase
beside it already say — a row refused before its values were read and one that ran out of time
were spelled the same. `NotRead` carries no reason and points at the owner.

## Tests

Both you asked for: a module plus an attached `examples for` file counted together, and the
Set-typed field compared as a set through `CheckedProgram → CheckedRow → holds`. Plus the unit
case above and a check that what a text stated cannot be changed by whoever holds it
(mutation-checked against a view instead of a copy).

Full reactor green, 10,955 tests.

kawasima:
The three are fixed. P0 is fixed as a check rather than as code, and the measurement says why.

## [P0] the mechanism you named, measured

`caseOfValue` is the fixture-building side — `buildFixture`, `builtStandIn`, `admitted`. What the
old comparison compared against was `state.resultArm`, and that is `fixtures.typeOf(result)`,
which I did not touch and which answers nothing for a boxed `Long`. So the old
`arm.equals(state.resultArm)` could not have compared a primitive case either.

Above that, a row cannot write one. `-> Int` is refused where it is written:

    13:36 E1023: I cannot find a value named `Int` in scope.

What a row names as a case is resolved as a *value*, and a primitive is not one — which is the
same fact from the other side: what reads an answer's case is the declaration it is of, and every
name a row can write has one. So there is no row to write a regression test with; what there is to
hold is the pair of facts the comparison rests on, and they are checked now
(`WhatARowMayNameAsACaseIsWhatAValueCanBeReadAsTest`): the primitive case is refused, the same
answer written as a value is a row like any other, and both arms of the reading that says which
case an answer is are mutation-checked (dropping either turns the suite red; the unit one turns
the *compile* red, E1905).

If the language ever admits `-> Int`, that test goes red and whoever changed it is told that the
comparison has to learn primitives. That is the part I could give you.

## [P1] the policy is the type's own

You are right that I moved the hole one call along. The cause is that `Stated` meant what its
maker chose: under limits handed in, the same type is "values a reader may be given" where the
compiler makes it and whatever the caller asked for anywhere else. `RowStatement.of(inputs,
expects)` no longer takes limits — `Limits.DEFAULT` is the vocabulary's, which is also what a
row's inputs are observed under, so a row's two halves are held to one size by construction rather
than by two callers passing the same thing.

## [P3] the contract says where asking lives

`CheckedProgram` and the assembler point at `CheckedRow.Reproducible#holds`, and the PR body is
rewritten against the current ADT.

## And the sweep I owed

Twice now I have named a cause and then fixed the sites you pointed at. The population for "a
published value owes what its neighbours owe" is every type reached from what a row states, and
it is asked as a whole now: eight of them keep a collection, each is built with a collection its
maker keeps and mutated after the fact, and a part in a shape the check has no sample of stops the
walk. The first version of that check went red on its own witness rather than on its claim — two
values built from one set of parts both move — so it compares against the parts as they stood when
the value was made.

The population for "the name says one owner, the mechanism another" had one member left that you
had not raised: `NotRead` claimed in prose that no accepted program carries one. Measured off the
codes — E1903, E1904, E1910, E1923, E1924, none of them warnings — so the claim holds, and it is
held to those codes now instead of to a sentence.

Full reactor green, 10,963 tests.

kawasima:
All four are fixed. P0 is real — I measured the path before touching it — and the three findings
turned out to be one cause, so they are closed as one.

## The cause

A value made of parts that answer different questions, with nothing holding them together, lets a
reader take one part and drop the other. That is P0 and P2, and it is the same shape you named
last time in `Mismatch`:

* `ReadRows` is the rows **and** what stopped a reading of them. The assembler read `rows()` and
  dropped `gaps()`. A block whose classes will not link produces no `RowOutcome` at all — I read
  the `catch (LinkageError)` to confirm, and it records only an `Incompleteness` — and `Acceptance`
  accepts. So: three rows written, zero rows crossed, exactly as you said, and my own sentence
  broken.
* `Mismatch` for a missing field named the field in its path and carried the values from around it.

So the fix is not "carry the gaps too". The gathering is over **what was written**, and what came
back is attached to it: `ReadRow = Ran(outcome) | NotRun(identity, at, why)`, enumerated from
`Prepared`. A row nothing came back for is listed as itself. And where a row becomes what an output
holds, both sums are switched over exhaustively, so dropping one is a compile error rather than a
shorter list.

For the difference: what differs when one side holds a place the other has not got is *which places
the two hold*, which is a fact about the values — so it is reported at the value whose places they
are. Every part of a `Mismatch` is now about one place, and that is the contract line.

`RowStatement.NotRead` carries the reason now (`LINKAGE_FAILED`, `OBSERVATION_ABSENT` …), which is
what it needs to say when it crosses. The javadoc claiming accepted programs hold none of these is
gone — it was true only of the paths I had enumerated.

## The regression test, and why the count did not catch it

You were right that the happy-path count does not catch this: I put the drop back in and
`EveryRowWrittenIsARowThatCrosses` stayed green, because in a healthy compile every written row has
an outcome. There is no seam to inject a `ProgramExecution` through `CheckedProgram.of`, so the
degraded state is exercised where it is produced: a real compilation, its real outcomes with one
taken out, and the gathering asked for the rows. Both rows are listed, the missing one says
`LINKAGE_FAILED`, and a whole-reading gap gives `OBSERVATION_ABSENT` to every row of the source.
With the drop back in, both go red.

## P1 and P3

`Expectation` is data again — `compare` is off it. Whether an answer keeps a statement takes a
reading of the declarations, and whoever supplies that decides part of the answer, so a comparison
hanging off `row.states().expects()` is one a reader can make with declarations of its own. The
comparison is where the two bound faces reach it (`Comparisons.verdict`) and nowhere a traversal
from a row arrives.

`Stated`'s javadoc says what the visibility gives: the one admission rule cannot be gone round,
because `of` is the only way to make one.

Full reactor green, 10,981 tests, on `ff7eed5`.

kawasima:
Fixed, and the root you named is the one I closed it at.

## The reason is taken where it happened

The two findings are one: I gathered the sources into `stopped` / `everywhere` and then read a
reason back out for a row that had already lost which source it is from. So the fix is not to pass
the row's `SourcePos` into the lookup — it is to put a source's rows and what became of reading
that source together **while both are still that source's**, which is what
`Prepared.forExamplesWrittenIn(sourceId)` is for. `readOneSource` does that per source and the
behaviors are merged afterwards.

Your two-source case is a test now: the model's own source stops with `LINKAGE_FAILED`, the
attached file is not observed at all, and the three rows say `LINKAGE_FAILED, LINKAGE_FAILED,
OBSERVATION_ABSENT`. Putting the old behaviour back (any recorded reason will do) turns it red,
along with the whole-source case.

## Neither an outcome nor a reason is a broken invariant

You are right that `OBSERVATION_ABSENT` there was a fabrication, and worse than the defect it was
covering: the whole point of this reading is that a row cannot go quiet, and a consumer inventing a
reason is exactly how it goes quiet again one layer up. It throws now, saying which row in which
source, and there is a test for it.

## `NotRead` said two things

Split, as you suggested:

* `StoppedBeforeItsValues()` — there is an evaluation of the row and it says how far it got; no
  reason here, because the stage and the phase beside it are that reason. A program the language
  accepted holds none of these, and that claim is still checked.
* `NotRead(Incompleteness.Code why)` — nothing read the row at all, so there is no evaluation to
  say why and it says so itself. Only `ReadRow.NotRun` makes one.

The `RowState` default is the first, which is what it always meant.

## Stale wording

`WhatARowMayNameAsACaseIsWhatAValueCanBeReadAs`'s third test is narrowed to what it actually holds
— those five codes stop a row *before its values*, and a reading that never reaches a row is
accepted — and the PR body's "every way of leaving a row unread" is corrected the same way.

Thank you for the `Comparisons` judgement; I agree it is a different kind of hole from the one it
replaced, and I have left it as it is.

Full reactor green, 10,983 tests.

kawasima:
Fixed. The root, in my own words: I put a state that belongs to one phase into the vocabulary of
another, because that is where the arm it was split from happened to live.

## What changed

`StoppedBeforeItsValues` is beside `NotStated`, not among its arms:

    RowStatement = Stated | NotStated(RequiresStandIns | Incomplete | NotRead)
                 | StoppedBeforeItsValues

`CheckedRow.NotReproducible` takes a `NotStated`, so what an output is handed cannot be a compile's
own progress — by the static type, which is what you asked for, rather than by no accepted program
happening to have one. The javadoc's "three ways" is true again as a consequence of the type rather
than by being edited to match.

## And the guarantee is asserted where it is relied on

    case RowStatement.StoppedBeforeItsValues _ -> throw new IllegalStateException(...)

in the assembler, naming the row and where it is written. Acceptance guarantees it; the boundary
says so out loud rather than trusting it silently.

## The pairing is held where it is written

The statement and the stage are one evaluation read apart, and they are assigned a line from each
other — so `RowOutcome` now refuses the two combinations that say different things about one
reading: values read but the statement saying the reading stopped before them, and the other way
round. Measured: taking the assignment out turns three tests red at construction with
`COMPARED with StoppedBeforeItsValues`.

## Wording

The PR body carries the third arm now, and `Expectation.compare` is gone from the Verified list —
the comparison is `ValueMatch.verdict` behind `Comparisons.verdict`.

Full reactor green, 10,983 tests, head `d8c9ccc63`.

kawasima:
Not waiting for the next one this time. I swept for what the five rounds were each an instance of,
and there was one left.

## The families, and what I found

Five rounds have been five instances of four patterns, so I went looking for the rest of each in
what this PR adds.

* **A published value owes what its neighbours owe** — the check over the crossing vocabulary walks
  eight types and refuses a part in a shape it has no sample of. Nothing left.
* **The name says one owner, the mechanism another** — no stale link to a moved or removed API
  remains, and `Front.ExampleSources` has two callers: the gathering, and `Compilation`'s
  passthrough for which sources exist, which is the other question.
* **A state of one phase in another phase's vocabulary** — the crossing sum reaches
  `NotReproducible` through `NotStated` alone, which the static type now says.
* **A value made of parts that answer different questions** — this one had a member left, and
  nobody had raised it.

## What was left

What a row handed over is read from two places: a measure reads `RowOutcome.inputs()`, an output
reads `Stated.inputs()`. They are written from one list, so they could not differ — except on the
path where a row is given up on, which publishes the statement its worker reached while saying the
row handed over nothing. A row that timed out after its values were read said it had none, beside a
statement holding them.

The values come from the statement there, and the two are held to each other where the row is made:
a row that states values states the ones it handed over.

That invariant found three fixtures building rows that cannot exist — an input the observation
stopped in, beside a statement that says every value is there in full. They ask the one place that
decides what such a row states, and get `Incomplete` where they were keeping `Stated`.

Full reactor green, 10,983 tests, head `a803d0fb1`.

kawasima:
Fixed. The root is mine and it is worth naming plainly: **I wrote "by the static type" and did not
read the static type.**

`Statement.states() : RowStatement` handed back the whole sum, so the case the arms exclude was one
every reader had to consider — one call from the arms that leave it out. Narrowing at the arms and
widening on their interface is not narrowing. It is gone; each arm answers what it has
(`Reproducible.states() : Stated`, `NotReproducible.why() : NotStated`), and there was no use for
the common one.

## And the claim is checked now

Deleting the method fixes today. What stops the next convenience from re-admitting it is that the
claim is a check: the walk that already reads what an output can reach goes down a sealed type's
arms, so it can say that `RowStatement$StoppedBeforeItsValues` is **not** reachable from
`CheckedProgram` — and that `NotRead` is, so the check is looking at something.

Measured both ways: putting a `default RowStatement states()` back on `Statement` turns it red,
naming the case in the reached graph.

That is the part I should have done when I made the claim, and it is the same lesson as the last
few rounds — a sentence about a type is worth what the type says, and here the mechanism to hold it
to was already in the file.

Full reactor green, 10,984 tests, head `9770ee24b`.

PR #1192, opened 2026-08-30
Ask what stopped a reading at the place the answer is filed, not at a side

### LINKED ISSUE #1148: Unread comparison reasons are classified per comparison but filed per position
Opened to characterize whether one `BOUNDARY` question can stand for two published words. It cannot, today — and working out why turned the question into a defect underneath it, which is what this is now about.

## The defect

A reason for a comparison that drew no line is decided once per comparison and then filed at every position the comparison names.

```java
// InvariantChecker.noLineDrawn
BlockReason.RuleWithoutLineReason why = UnreadComparison.why(left, right, quantity, ordered);
for (Coordinate each : coordinatesIn(comparison, at, byName)) {
    out.add(new FieldDomains.NoLine(each.at(), from, comparison, conjunct, why));
}
```

The guard reader has the same shape written as a record rather than a loop —
`ComparisonAssessment.Unread(RuleReadingStopped why, List<FilingCoordinate> filedAt)` and
`ComparisonAssessment.CutsNothing(List<FilingCoordinate> filedAt)`.

What `why` is read off is one side of the comparison — `UnreadComparison.speakingSide` picks the
left where both name a position — so every other filing coordinate is told a fact about a position
it is not. `GuardThresholds.filedAt` already writes the case down in its own javadoc: `a * a + b - b
<= 9` is filed at both positions, and the arithmetic that would have cancelled `b` is the part that
stopped.

So the subject a reason is about and the subject its evidence is filed under are two different
things, and nothing holds them together.

## Three keys where there is one question

```
question = (rule, coordinate)        Owed.Boundary, raised from ClauseStates.ABound.line
reason   = (rule, conjunct)          one call to UnreadComparison.why per comparison
filing   = (rule, conjunct, coordinate)   one row per position the comparison names
```

`FieldDomains.boundaryAnswered` then scans the filing table and de-duplicates to recover the
question's answer. There is nowhere that decides "the reason at this coordinate"; there is only the
comparison's reason, copied.

## Why BOUNDARY converges today, and what the convergence rests on

It converges. `ClauseStates.ABound` is built in one place (`InvariantChecker`, the arm that runs
where the comparison is an ordering, one whole side names a coordinate, and the other side names no
coordinate at all), and `Required.ofInvariant` raises `Owed.Boundary` only from `ABound.line`. So a
conjunct carrying a boundary question is `c <op> (expression naming no coordinate)`, the comparison
names exactly `c`, and the single row filed for it has its word read off `c` — whether through
`Quantity.OverOne(c)` or through `speakingSide` landing on `IsAPosition(c)`. The word is
`ordered(c) ? UnreadComparisonForm : UnreadComparisonDomain`, a function of `c` alone, so every
conjunct raising the question at `c` comes to the same one.

Measured: `x <= 10 * 2 && x <= Int.abs(x)` files two rows and the question stands with
`[UnreadComparisonForm]`; `stage >= Prospecting` over a sum's case stands with
`[UnreadComparisonDomain]`. Both are one word.

That the answer is one word is therefore not something the mechanism holds. It is the product of
four facts in three files, and the load-bearing one is the reachability guard on `ABound`. The day a
comparison naming two coordinates can carry a boundary question, `c` receives a word read off the
other position.

## Before implementation

Measure, over every checked-in model, each comparison whose arithmetic reaches a quantity and whose
filing coordinates contain a position outside that quantity's support. Record the model, the filing
coordinate, the internal reason it carries now, and the word a document publishes for it. The
implementation accounts explicitly for every change this list predicts.

Not a Done condition. An empty list is a successful measurement, and the property it establishes is
about the change rather than about the code afterwards.

### Measured

Every row `noLineDrawn` files over the whole suite: 12838, of which 11012 carry a quantity the
arithmetic read to the end and 1826 do not.

```
support known, filing position outside it        0
support known and empty (CutsNothing)          108   ComparisonCuttingNothing
support known, filing position inside it     10856   280 OverOne, 10576 OverSeveral
support unknown (NotRead)                     1826
```

The gate is empty. No checked-in model files at a position outside a non-empty support.

It is reachable from source all the same, through the reading rather than through a hand-built
value. On `data N = { x: Int, y: Int }`:

```souther
invariant said = x + y - y <= 10 * 2     -- support {x}, filed at {x, y}
                                         -- y is told UNSUPPORTED_SYNTAX
```

and with a third position, where the sentence `y` is told is further off still:

```souther
invariant said = x + y - y + z <= 10 * 2 -- support {x, z}, filed at {x, y, z}
                                         -- y is told UNSUPPORTED_PARTITION_SHAPE
```

Both are read to the end. Neither cuts `y`, and neither relates `y` to anything — so `y` is a place
the rule is filed at and is no subject of it. The guard reader already files a read rule at its
quantity's coordinates and says why (`ComparisonAssessment.filedAt`), and the clause reader takes
its places from the walk over the comparison; these two readings are the controls that tell the two
apart, one per shape of quantity, since the words `y` is given today differ.

The unread half is where the words move. Split by what a per-position classifier would need to
decide them:

```
UnreadComparisonForm      ordered  not under an operation   1319
RuleAboutADerivedValue    ordered  under an operation        272
RuleAboutADerivedValue   !ordered  under an operation         96
UnreadComparisonForm     !ordered  not under an operation     81
UnreadComparisonForm      ordered  filed at the speaking position   46
UnreadComparisonDomain   !ordered  filed at the speaking position   12
```

The 96 are why the carrier is not asked first: under carrier-before-provenance they become
`UNSUPPORTED_DOMAIN`, and the rule is about what an operation answered.

The 81 are the rows a per-position rule would move from `UNSUPPORTED_SYNTAX` to
`UNSUPPORTED_DOMAIN`, and none of them should move:

- `example.deepbound.Root`, `invariant ordered = n < l1.l2.leaf.x`, filed at `l1.l2.leaf` while the
  reason is read off `n` (20 rows). The misattribution above, observed.
- `example.deepbound.N` and `example.deep.N` at `l1.l2.leaf` (56 rows). A position the walk met
  inside something it did not take apart, whose own carrier the rule says nothing about.
- `demo.V`, `data V = { a: A }` with `invariant describe(a) > 0`, filed at `a` (5 rows). A rule
  about what a call answered of `a`.

The last one settles how a classifier may ask whether a filing is about the position's own values.
It cannot read that off `ValueOrigin`: `Terms.readThrough` reads a call through to its body, so
`describe(a)` arrives as `Composed` holding `IsAPosition(a)` and carries no `Applied` to recognise.
The reader has to say what its filing coordinate means.

And the second meaning is not "a number taken of the position". It covers a position met inside
something the reader could not take apart, which `GuardThresholds.filedAt` already writes down in
prose. What the two share is that the rule is not about the position's own values, which is what the
carrier question needs to be gated on.

Guard side, `ComparisonAssessment.CutsNothing`: 48 occurrences, filed at 0, 1 and 2 coordinates. Its
support is empty by construction, so every one of them keeps `ComparisonCuttingNothing`.

## Done when

- a reason is asked for at the subject it will be filed under, and a reader cannot hand one reason
  to several subjects. `UnreadComparison.speakingSide` has nowhere left to stand
- the two readers reach the classification through one entry. `UnreadComparison` exists to keep an
  invariant's clause and a `guard`'s comparison at one answer (ADR-0090), and per-position on one
  side alone would put two meanings inside it
- once a non-empty quantity has been read, where the rule is filed comes from that quantity, so a
  position outside its support receives no evidence for the rule at all. Not a reason of its own:
  the guard reader already holds this and writes down why (`ComparisonAssessment.filedAt`), and a
  word for a position outside the support would fix an impossible state in a published vocabulary.
  `CutsNothing` stays the deliberate exception — an empty quantity is filed at the positions the
  comparison named, because the model states a rule there and that rule cuts none of them
- whether the carrier is asked at all is decided by the subject a filing stands for, and not by the
  order of tests inside the classifier. The reader says which of two a filing is — the rule is about
  the position's own values, or it is not — and the second arm has no way to reach the carrier:

  ```java
  sealed interface RuleAt<K> {
      record AboutOwnValues<K>(K position) implements RuleAt<K> {}
      record NotAboutOwnValues<K>(K position) implements RuleAt<K> {}
  }
  ```

  Named for what it denies rather than for what it holds, because the second arm has no positive
  ontology in common: a length taken of a string, a call read through to its body, and a position
  met inside an expression the reader did not take apart are one arm only in that none of them
  constrains the position's own carrier. `QuantityOf` claims a quantity that two of the three are
  not, and `SomethingOf` drops the one thing the type exists to say
- `boundaryAnswered` no longer recovers a question's reason by joining rows filed under a finer
  key. A boundary answer carries one reason and the conjuncts that stand behind it, so "several
  conjuncts" and "several reasons" stop being one axis
- `FieldDomains.noLines` stays. Evidence filed per position and an answer held per question are two
  projections of one reading for two different readers, not a duplication

## Not this

Whether an `ADMITTED_VALUES` question can carry two published words. It can, and PR #1146 checks it:
`invariant both = a /= b && String.matches("x+", a)` reports `unsupported_partition_shape` and
`unsupported_syntax`, in the order the author wrote them.

kawasima:
Closes #1148.

## What was wrong

A reason for a comparison that drew no line was decided once per comparison and filed at every
place the comparison named. What it was read off is a side — `UnreadComparison.speakingSide`, which
takes the left where both name a position — so at every other place it stated a fact about a
position that place is not.

The same shape stood in both readers: a loop in `InvariantChecker.noLineDrawn`, and
`ComparisonAssessment.Unread(why, filedAt)` with the loop in `GuardThresholds.publish`.

Three keys where there is one question: a question is `(rule, coordinate)`, a reason was
`(rule, conjunct)`, and evidence is filed at `(rule, conjunct, coordinate)`.

## What it is now

Two questions, and which is asked turns on whether the arithmetic reached a quantity.

**Read to the end — the quantity is the subject.** It says what the rule leaves, and
`UnreadComparison.filedAt` says which of the places the walk met are the rule's at all.
`x + y - y <= 10 * 2` is `x <= 20`, so `y` gets nothing. The guard reader already held this and
wrote down why (`ComparisonAssessment.filedAt`); the clause reader took its places from the walk
over the comparison, which is where the two disagreed.

`CutsNothing` keeps its places, deliberately: its support is empty and what is worth saying is that
the model names those positions and cuts none of them.

**Stopped — each place is asked for itself.** `UnreadComparison.whereItStopped` takes a `RuleAt`,
which is the reader's answer to the one question a classifier of expressions cannot see: whether the
rule constrains the values at this position or something taken from them. `ValueOrigin` reads a call
through to its body, so `describe(a) > 0` arrives as arithmetic holding the position with no
operation left to recognise — the distinction has to come from what the reader knew when it chose
the place.

Both readers answer it the same way: a whole side of the comparison is that position and nothing
else, and the coordinate there is the position's own value.

`RuleAt.NotAboutOwnValues` is named for what it denies because its members share nothing else — a
length taken of a string, a call read through, and a position met inside an expression the walk did
not take apart. What they have in common is that the carrier is not the rule's subject, which is the
one thing the type exists to gate.

`speakingSide`, `whyItStopped` and both copy loops are gone. What each place is left with is one
answer of the assessment (`whatEachPlaceIsLeftWith`), and the copy survives only inside
`sameAtEachPlace`, where the places are one quantity's coordinates.

## Not added

A reason for a position outside a non-empty support. It was on the table, and reading the guard side
settled it the other way: filing a read rule at its quantity is already the design there, so the
state that word would name stops being reachable rather than needing a name. A published vocabulary
is the wrong place to fix an impossible state.

## Measured

Before the change, every row `noLineDrawn` files over the whole suite: 12838, of which 11012 carry a
quantity read to the end.

```
support known, filing position outside it        0
support known and empty (CutsNothing)          108
support known, filing position inside it     10856
support unknown (NotRead)                     1826
```

The 1826 unread rows split by what a per-place classifier needs to decide them, and every one of
them keeps the word it had:

```
derived, whatever the carrier                    368
a whole side is this position, ordered            46
a whole side is this position, not ordered        12   (the only Domain rows)
not a whole side                                1400
```

No checked-in report moves. Full suite green: 7715 (7710 before, plus five new).

## Controls

- `ARuleIsFiledAtWhatItsQuantityIsAboutTest` — a cancelled position is filed at by nothing, and the
  surviving ones still are. Both halves, so the assertion cannot pass over a reader that files
  nothing. Plus `filedAt` over the three shapes of quantity.
- `AStoppedReadingSaysWhatStoppedItAtEachPlaceTest` — one comparison, two places, two words: `s`
  carries no order and is told so, `v` was met inside a call read through to a `match` and is told
  the form. Under the old rule both came back as the carrier's answer about `s`.
- `ARuleInABodyIsFiledAtWhatItsQuantityIsAboutTest` — the same source shape in a body, so the
  agreement between the two readers is written down rather than held on one side.

Both mutations go red: `RuleAt` always `AboutOwnValues` moves `v` to the carrier's answer, and
`filedAt` passing every place through fails all three of the first test's assertions.

## One arm without a control

Where a line could not be realized on a quantity that was read
(`Cutting.realized`), the places now come from the quantity as well. That arm is reachable —
`guard a > b` over two cases of a sum — but the model that reaches it names the same two places
either way, and no model distinguishes the two rules, because a form on a sum's order has nothing to
cancel. Kept because the alternative is `Read.Stopped` taking its places from a different authority
depending on which producer built it, which is the shape this removes.

### DIFF
--- a/souther-compiler/src/main/java/souther/compiler/check/FieldDomains.java
+++ b/souther-compiler/src/main/java/souther/compiler/check/FieldDomains.java
@@ -66,7 +66,7 @@ public final class FieldDomains {
      */
     public static final FieldDomains NONE =
             new FieldDomains(Map.of(), Map.of(), Map.of(), Map.of(), Set.of(), List.of(), List.of(), Map.of(),
-                    Map.of(), new ReadingEvidence(), Map.of(), Set.of(THE_VALUE), Set.of(),
+                    Map.of(), Map.of(), new ReadingEvidence(), Map.of(), Set.of(THE_VALUE), Set.of(),
                     NO_POSITIONS,
                     ConstraintState.<FactSubject>top(), null, null, null, null, Map.of(), Set.of(THE_VALUE),
                     Map.of(), Map.of(), Map.of(), Map.of());
@@ -83,6 +83,9 @@ public final class FieldDomains {
     /** The same per part of each clause. A reader that found one conjunct wanting names what that
      *  conjunct is about, and not what the conjunct written beside it raised. */
     private final Map<RuleRef, Map<Core, Required>> raisedByPart;
+
+    /** What the reading answered for each boundary question it raised and left standing. */
+    private final Map<BoundaryQuestion, BoundaryStanding> standing;
     /** Which readings took each clause in, as each of them said so. */
     private final ReadingEvidence took;
     /** The accounting, worked out once. Every position of a value asks the same question of it. */
@@ -152,7 +155,8 @@ public final class FieldDomains {
                          Set<String> notSeparatedByField,
                          List<InvariantChecker.Direct> directs, List<NoLine> noLines,
                          Map<RuleRef, Required> raised,
-                         Map<RuleRef, Map<Core, Required>> raisedByPart, ReadingEvidence took,
+                         Map<RuleRef, Map<Core, Required>> raisedByPart,
+                         Map<BoundaryQuestion, BoundaryStanding> standing, ReadingEvidence took,
                          Map<String, List<TypeSymbol.AtModule>> narrowers,
                          Set<String> notGathered, Set<String> handedOn,
                          SequencedMap<FactSubject, String> positions,
@@ -172,6 +176,7 @@ public final class FieldDomains {
         this.noLines = noLines;
         this.raised = raised;
         this.raisedByPart = raisedByPart;
+        this.standing = standing;
         this.took = took;
         this.narrowers = narrowers;
         this.notGathered = notGathered;
@@ -396,7 +401,8 @@ public final class FieldDomains {
                 named(seeded, field).forEach(term -> placeOf.putIfAbsent(term, field)));
         return new FieldDomains(Map.copyOf(out), Map.copyOf(holds), Map.copyOf(admitted),
                 Map.copyOf(unread), Set.copyOf(notSeparated), seeded.reading().directs(), seeded.reading().noLines(),
-                seeded.reading().raised(), seeded.reading().raisedByPart(), seeded.took(),
+                seeded.reading().raised(), seeded.reading().raisedByPart(),
+                seeded.reading().standing(), seeded.took(),
                 seeded.reading().narrowers(),
                 seeded.notGathered(), seeded.handedOn(), placeOf,
                 seeded.constraints(), named, data, symbols, policy, settled,
@@ -471,6 +477,55 @@ public final class FieldDomains {
         /** Where in the value the end was to have been placed. */
         public String path() {
             return at.path();
+        }
+    }
+
+    /**
+     * One rule, one coordinate: the question of where the values there stop.
+     *
+     * <p>What a reader of an accounting asks about, and the key its answer is held under. A rule is
+     * read a conjunct at a time and any number of them may draw the same line, so the parts are not
+     * this — held under them, an answer had to be put back together from whatever rows a finer key
+     * happened to hold.
+     */
+    public record BoundaryQuestion(RuleRef.Invariant from, Coordinate at) {}
+
+    /**
+     * A boundary question the reading of ends did not answer, and everything behind it.
+     *
+     * <p><b>One reason and several conjuncts, which are two different multiplicities.</b> Which
+     * limit stopped the reading is read off the coordinate — a carrier lines are drawn on wants a
+     * reader for the form, and one nothing draws a line on wants the carrier
+     * ({@link UnreadComparison#whereALineWouldFall}) — so every part of the rule that raises this
+     * question comes to the same word. How many parts are standing behind it is what an author has
+     * left to lift, and that is its own count.
+     *
+     * <p>Made where the question is raised and not gathered from the findings afterwards. Gathered,
+     * the answer was whatever the rows filed under a finer key came to, and the day two of them
+     * differed a reader would have been handed both with nothing saying which the question's is.
+     *
+     * @param conjuncts the parts of the rule still standing behind it, in the order they are
+     *                  written
+     */
+    public record BoundaryStanding(
+            souther.compiler.inputs.BlockReason.RuleReadingStopped why, List<Integer> conjuncts) {
+
+        public BoundaryStanding {
+            if (why == null) {
+                throw new IllegalArgumentException("a question left standing says why");
+            }
+            if (conjuncts.isEmpty()) {
+                throw new IllegalArgumentException(
+                        "and says which part of the rule is standing behind it");
+            }
+            conjuncts = List.copyOf(conjuncts);
+        }
+
+        /** The same, with {@code one} standing behind it too. */
+        BoundaryStanding and(BoundaryStanding one) {
+            List<Integer> both = new ArrayList<>(conjuncts);
+            one.conjuncts().stream().filter(each -> !both.contains(each)).forEach(both::add);
+            return new BoundaryStanding(why, both);
         }
     }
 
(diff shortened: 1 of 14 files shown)


kawasima:
The one arm without a control is now #1193.

kawasima:
Self-check pass, second commit.

Seven things read as patched-at-the-site rather than decided. Four are fixed here.

**The law was spelled twice.** `ruleAt` and `isExactly` stood in both readers — the one thing they have to agree about, written once each. `UnreadComparison.subjectAt` decides it now, and a reader supplies only what it calls the place.

**A null stood for a second reason.** The guard reader told "a number taken of the position" apart from "not a whole side" with a null, which is one arm reached two ways with nothing saying which.

**That second condition had no witness, so it is gone.** Neither reader can make it change an answer: a side naming a number is a call and never a position the walk stops at, and where a clause names one the carrier asked about afterwards is that number's. A control written to reach it did not, which is how this was found — and the test that could not go red was deleted rather than kept.

**`filedAt` tested one arm and filtered the rest.** It switches over `Quantity.Read` now, so an arm added is a compile error. `support()` went with it: no reader once each arm is named, and for `CutsNothing` it answered a question about places that it is not the answer to.

**`RuleAt.NotAboutOwnValues` carried a position nothing read.** It carries nothing now.

**Two helpers built one thing.** A quantity read with no line realizable is its own arm of `Cutting.Read` rather than a reading that stopped, so its places are the quantity's like every other read comparison's, and one helper builds "this answer at each of these places".

Full suite green, 7714. The mutation on the remaining law still goes red.

kawasima:
Both findings addressed. They share a cause, and it is the one #1148 named.

**Root cause.** An answer that belongs to one subject was re-derived by a consumer walking a table keyed by something finer, so whatever that table held decided the answer. The branch fixed it for reasons and left it in three more places.

1. `boundaryAnswered` rebuilt the `(rule, coordinate)` answer out of `(rule, conjunct, coordinate)` rows and deduped the result.
2. `filedAt` said the quantity decides membership and computed `met ∩ support`.
3. And one I found while checking whether (1) could be closed honestly: the derived test ran on the stopped expression before anything else, so a global fact about a comparison answered at a place whose own values the rule plainly states something about.

**(3) first, because (1) depends on it.** I could not type the boundary answer as one reason until the reason was a function of the coordinate — and it was not, because the derived arm sat in front of the carrier arm. Moved inside `NotAboutOwnValues`, both words an `AboutOwnValues` place can be left with are read off its carrier. `whereALineWouldFall` is that single reading, called by the boundary answer and by both classifier arms.

I first wrote the derived test as a per-position walk. `WhatStoppedAComparisonIsReadOffWhatItIsTest` refused it, correctly: a rule bounding `String.length(s)` is not a rule about a value made from `s` just because `s` reaches it through the length. The place there is a term, and `ValueOrigin`'s key cannot name one. Reverted to the top-node test, now inside the arm.

**(1).** The reading makes the answer where it raises the question. `FieldDomains.BoundaryQuestion` → `BoundaryStanding(one reason, the conjuncts behind it)`. `boundaryAnswered`, the walk over the findings, and `whatStopped` are gone; `TheEndReadingSays` carries the standing answer. `noLines` is untouched — the per-place projection is a different reader's.

No runtime distinct-size check anywhere. Two conjuncts raising one question compute the same word because both read the same coordinate's carrier.

**(2).** The walk orders, the quantity says which, and the two disagreeing is refused rather than intersected away.

**Measured before designing.** 34 boundary questions stand across the suite, all with one word (33 Form, 1 Domain); Derived has never reached one. I tried to build a witness for two words and could not — which is why the reason had to become a function of the coordinate rather than be asserted to be one.

Full suite green, 7716. Mutations: dropping the second conjunct from the evidence goes red; making the carrier stop deciding goes red in four tests.

kawasima:
Blocker closed, and not by checking the mismatch.

**Root cause.** The reason belongs to the question, and the producer was still able to make a whole answer — reason included — per conjunct. That is what left the merge holding two words with a decision to make, and `first-wins` was the shape that decision took. Refusing a mismatch would have kept the producer able to come apart and turned the consequence into a throw; the type should not let a second conjunct carry a reason at all.

So `BoundaryStanding.and(int conjunct)` takes the part and nothing else, and the producer reads the reason once:

```java
standing.compute(new FieldDomains.BoundaryQuestion(from, stated.line()),
        (question, had) -> had == null
                ? new FieldDomains.BoundaryStanding(
                        UnreadComparison.whereALineWouldFall(about.carrier() != null),
                        List.of(part))
                : had.and(part));
```

`reason multiplicity = 1` and `conjunct multiplicity = N` are now the shape of the API, as you put it. There is no second word to reconcile and nowhere one could be dropped for being second.

Worth writing down why the single read is the question's and not the first conjunct's: `about.carrier()` comes from `byName`, which is built once per path from `typeAt`, and the question's key is the coordinate. Two coordinates equal by `at()` are the same path and kind, so the carrier is a function of the key — one read, not a first read.

**Negative control added**, as suggested — both shapes of quantity:

```java
assertThrows(IllegalStateException.class, () -> UnreadComparison.filedAt(
        new UnreadComparison.Quantity.OverSeveral<>(Set.of("x", "z")), List.of("x", "y")));
assertThrows(IllegalStateException.class, () -> UnreadComparison.filedAt(
        new UnreadComparison.Quantity.OverOne<>("z"), List.of("x", "y")));
```

Mutations: removing the refusal goes red; making `and` drop the part goes red.

Full suite green, 7717. Head is `f11e8ad8`.

PR #1197, opened 2026-08-31
Ask whoever settled a value's shape what a field of it holds

### LINKED ISSUE #1191: A field is read two ways, and only one of them is what a row is compared against
Found while closing #1121, and left out of it deliberately: what a field of a data is declared to be
is read from two places, and a row's comparison uses whichever one is asking.

## The two readings

A comparison asks one question of the declarations — where a type reads a field — and each side
answers it from what it holds.

* The compile answers from the syntax it read: `NeutralForm` implements `ValueTypes` over
  `DeclaredTypeEvidence.fieldTypes`, which walks the `Hir` declaration and follows what an
  `...includes` brought in.
* A checked program answers from what it publishes: `DeclaredFields` reads
  `CheckedData.Product.fields()`, which is the `ValueShape` the checker settled.

Both should say the same thing about the same field. Nothing holds them to it.

## What it costs where they differ

The answer decides part of a verdict. A field declared a `Set` is compared without its order, and
one nothing declares is compared in the order it stands — so a field the two readings disagree
about is a row that holds at compile time and does not hold in the snapshot, or the other way
round. That is the one thing #1121 exists to prevent: a row means one thing wherever it is read.

## What is already known

The end-to-end case is checked for a declaration-driven `Set`
(`AnOutputReadsWhatABehaviorsExamplesSaidTest.andWhatAFieldHoldsIsReadFromWhatThisProgramDeclares`),
so the two agree there. Nothing says they agree for a newtype's base, a generic instantiation, a
field an include brought in, or a field of a declaration read off the path.

## Done when

Either the two readings are one — a comparison asks whoever owns what a value is made of, and the
compile and the snapshot reach that one owner — or they are held to each other by a check over
every declaration a compile resolved, so a field they answer differently about is a red line rather
than a row that quietly means two things.

kawasima:
Closes #1191.

## What was wrong

A field of a data was read from two places, and a row's comparison used whichever was asking.

* The compile answered from the syntax it had read: `NeutralForm` implemented `ValueTypes` over a walk of the `Hir` declaration, following what an `...includes` brought in.
* A checked program answered from what it publishes: `DeclaredFields` read `CheckedData.Product.fields()`, the `ValueShape` the checker settled.

The answer decides part of a verdict, so a field the two disagreed about is a row that holds at compile time and does not hold in the snapshot. Nothing held them to each other.

## What was measured before changing anything

The two walks do not disagree today, and the issue's example of a disagreement is unreachable:

* `Hir.TypeTerm` is `TypeRef | FnType`, so the field the examples-side walk dropped is exactly a function-typed one. The surface parses it, but `Deriver` runs `CodecShape.of` over every field of every data and `Type.FnOf` is refused there, so no accepted program has one.
* An include collision is refused where the declaration is checked before either walk sees it, and a newtype's base, a generic instantiation and an included field all reduce to the same `denotes()` on both sides.

So this is an architecture defect rather than a live wrong verdict, and the tests below fix the boundary rather than reproduce a failure that does not happen.

Two facts settled the shape of the fix. A product owner is always a `TypeSymbol.AtModule`, which names the module whose `Shapes.ValueShapes` holds its shape — so a field's authority is reached by identity and needs no search order. And the language declares no product data, so nothing is lost by making `ValueShape` the only accepted answer.

## The rule now

A declaration's effective value shape is decided once by checking. Every reader of a value's parts projects that checked shape; none reconstructs it from the HIR. Where a program was accepted, a missing shape is an invariant violation and never an absent field.

* `observe.FieldTypes` is the question. `ValueTypes` stops being an interface with two implementations and becomes the one step from that answer to a `Position`.
* `check.CheckedFieldTypes` answers it for an accepted program, `program.DeclaredFields` for a snapshot, `check.ResolvedFieldTypes` for an editor reading a text that has not checked.
* `DeclaredTypeEvidence` keeps the walk that carries evidence through an expression and stops deciding what a field holds. Crossing a `.` asks the field authority of the world the reading is being made in, so a fixture's projection is typed by the answer its row is compared against.
* A newtype is a product of the one field it is written with, everywhere. The three places that branched on it no longer do.

## Tests

Seven, each checked by a mutation that makes it red: routing the `FieldAccess` edge back through the declarations, answering an unsettled shape as empty on either side, dropping a newtype's `value` from the checked projection, and leaving a declaration out of the assembled program.

Not covered: the two projections are compared over a single module's declarations and the language's, not over a module read off the path.


kawasima:
Pushed `c95e1bb2`. The first finding is closed from its cause rather than at the site, so the shape of the fix differs from the one suggested — the third throw is not there.

## P2 (1) — closed by construction, not by another guard

The suggested patch is right about the states, but adding a third `throw` would have left the cause standing: **each accepted reading decided for itself whether an owner is a data a value is built out of, and wrote that decision as a predicate whose negative swallows what it cannot reach.** `!(declaration(x) instanceof Hir.Data)` is false for a sum and false for a declaration out of reach. The snapshot only told them apart because I had happened to give it a membership set. That is the same defect this PR exists to remove, one level below where it was removed — and a third site-local guard would have been the third place deciding the same thing.

So the classification is no longer a predicate and no reader derives it:

* `observe.Composed` — `OfFields(fields)` or `OfNothing`. Two answers, not one answer with a hole.
* `observe.Declarations` — what a world answers, total over what it should hold, refusing what it should hold and does not.
* `FieldTypes.over(Declarations)` — the one step from a declaration to the fields under it. What the language gives has no fields, said once here instead of in each world.

Each world now writes the classification as a switch over a sealed type with the absence as an arm of its own (`case null ->` for the compile's registry lookup, the `CheckedData` arms for the snapshot). A declaration kind added later breaks compilation rather than falling into "no fields". The three states you list are preserved, and so is the fourth (`known Product, no shape`).

## P2 (2) — path owner

Added, with one thing narrower than asked. The test compiles two modules, one declaring `Foreign` and one holding a value of it, and fixes that the reading made in `down`, the reading made in `up`, and the snapshot all answer with the same `ValueShape`. **A module read off the path is still not covered** — this is a second module of the same compilation.

It does close the risk named in the review: replacing `ValueShapes(declared.module())` with `ValueShapes(symbols.module())` was run as a mutation, and it fails this test (as a refusal now, not a wrong answer).

## P3 — the javadoc

Fixed as documentation, behaviour unchanged. Note the cause is the same family as the rest of this PR: I moved the walk and wrote a precedence into its javadoc that the walk does not have. The text now says what a repeated spread field comes to there is not a contract and nothing may read an answer off it.

## Also found in a self-check, not by the review

* `FixtureReader` still compared a newtype's field name against a `"value"` literal beside the constant this PR introduced for exactly that name.
* One `assertEquals(true, …)` where every neighbour uses `assertTrue`.

Both fixed here. `mvn -o test` is green across the reactor.

kawasima:
Pushed `204b5e31`. The finding is real and I measured it before changing anything, but I did not fix it as a call-site choice.

## Measured first

A data whose clause does not elaborate is visible as a declaration and has no settled shape, so the strict reading refuses it:

```
SHAPES={}
DECLARED=Data[written=Bad, declares=demo.Bad, ...]
ASKED=java.lang.IllegalStateException: `demo.Bad` is a data this compile resolved
      and the check said nothing about what a value of it is made of
```

for

```
data Bad =
    { n: Int }
    invariant broken = n
```

So the chain holds at the authority, and the unreachability argument does not.

## The cause is not the call site

Switching Adequacy to the tolerant authority is right, but it is the second finding of the same shape in this PR, so the fix cannot be the argument I pass.

`Shapes.fieldTypes(db, symbols)` was a public factory, and **`db` plus a scope is not evidence that the caller is downstream of a check** — which is the whole warrant for the one thing that reading does. I gave a total contract to a projection of a deliberately partial answer (`ValueShapes` drops what has no meaning) and put the thing that makes it total — acceptance — nowhere in the type. Adequacy was the first caller to wander in; the next one would have been written the same way.

So the factory is gone. The checked reading is built in `ExampleExecutions`, which establishes that the module checked before it builds anything, and the snapshot's own is built by the assembler after the language's verdict is asked for. Adequacy names `ResolvedFieldTypes` — not as a fallback from a missing shape, but because it is measuring a text; a measure that raised where it could not measure would report a compiler fault for a mistake the author is already being told about.

`TheCheckedReadingIsMadeWhereItsWarrantIsTest` reads the class files for who constructs `CheckedDeclarations`. Putting the strict authority back into Adequacy was run as a mutation and fails it. The snapshot side needs no such rule — its reading is package-private and only the assembler can build it.

## Tests

Two of the readings you would want measured are fixed:

* a reading that is not downstream of a check answers `Int` for `Bad.n` where the strict one refuses — the two worlds, side by side, on the same unsettled declaration;
* every test that needs the checked reading now takes it the way a run of the rows does, `ExampleExecutions.of(db, module).fieldTypes()`, which is also what says the gate is there.

I did not drive the adequacy generation query itself over that model. What is fixed is the authority and who may hold it, not that particular query's traversal.

`mvn -o test` is green across the reactor; the unused-import hook passes.

Issue #1206, opened 2026-08-31
A claim that says the operator is no comparison is answered for six times below where one is required


kawasima:
Left from #1193 (PR for it), where the half that produced a published word was closed and the type
was not.

## What is written

`ComparisonClaim` has three cases, and the third says the operator is not a comparison at all:

```java
static ComparisonClaim of(BinOp op) {
    if (!op.compares()) {
        return new Nothing();
    }
    ...
    case AND, OR, ADD, SUB, MUL, DIV, CONCAT -> new Nothing();
}
```

Everything under `Cutting.read` is reached only for an operator that compares — `AffineReading.read`
sends the rest to `CutsNothing(Set.of())` before a realization is attempted — so `Nothing` cannot
arrive there. It is spelled all the same, in six places:

```
ComparedLine.of                   case Nothing -> null
ComparedLine.fromTheForm          case Nothing -> null
ComparedTerms.fromTheForm         if (... instanceof Nothing) return null
Cutting.holdsAtTheValue           case Nothing -> false
ReachingCuts.relOf                case Nothing -> null
AffineReading.turned              case Nothing -> nothing
```

Each is an arm of an exhaustive switch over a sealed type, so javac requires them. What it does not
require is that they say anything true: `holdsAtTheValue` answers `false` for a claim that holds at
no value because there is no comparison, and `relOf` answers `null` for a relation that does not
exist.

## Why it is not a defect today

The one arm that turned an unreachable case into something a reader is told was
`overAForm`'s, and the PR for #1193 removed it — a `null` there became
`UnreadComparisonDomain` at `ComparisonAssessment`. What is left produces nulls that the shape
readings treat as "not this shape", which is a dispatch and reaches no report.

## Why it is worth closing

Nothing observes these arms. They cannot be tested, so nothing would notice one of them becoming
wrong, and a reader cannot tell an arm that cannot happen from an arm that was thought about. The
same shape has already cost this area once: `overAForm` refused `Nothing` beside two conditions that
could happen, and the three were read as one absence.

Six sites is also enough that the next reader of any one of them has to work out the reachability
argument again, and that argument lives in `AffineReading.read` rather than beside them.

## What it would take

A claim narrower than `ComparisonClaim` for everything below the point where an operator has been
established to compare — `Cut | Singled` and no third case. `AffineReading` would carry that, and
the six arms would stop existing rather than answer for a case nobody can construct.

The place the two meet is `ComparisonClaim.of`, which is where an operator is asked what it states.
[...]

PR #1207, opened 2026-08-31
Ask what the general form needs where it needs it, and name the answer for what was found

### LINKED ISSUE #1193: A realization that produced no line says which reading declined, not why
## The investigation changed the question

This was filed as a filing-authority issue: PR #1192 made `Cutting.realized` file at the quantity
rather than at a walk over the comparison, and no model distinguishes the two rules, so the choice
had no control. That hypothesis is rejected. Reading what makes `realized` reach that answer at all
turned up the defect underneath it, and the filing set is a consequence of it rather than the
subject.

Kept here because it is the evidence for why the subject moved.

## The defect

`realized` tries three readings and concludes from their combined silence:

```java
Cutting drawn = atAPosition(behavior, ComparedLine.fromTheForm(read, inputs, symbols), quantities);
if (drawn == null) {
    drawn = apart(behavior, ComparedTerms.fromTheForm(read, inputs, symbols), read.claim(), quantities);
}
if (drawn != null) {
    return new Read.Cuts(drawn);
}
Cutting form = overAForm(behavior, read, inputs, symbols, quantities);
if (form != null) {
    return new Read.Cuts(form);
}
return new Read.NoLineOnTheQuantity(AffineReading.filedAt(read.form().coefs().keySet()));
```

`ComparisonAssessment.of` then writes `UnreadComparisonDomain` for that answer — the values compared
are not ones a line can be drawn on here.

Nothing established that. It is read off three nulls, and those nulls do not mean one thing. Two
different questions are collapsed into the same absence:

- **which shape this quantity is.** `ComparedLine.fromTheForm` returns null where the form is not
  over one coordinate, `ComparedTerms.fromTheForm` where it is not a pair. That is a dispatch saying
  "not this one", and it carries no fact about the model.
- **whether the shape chosen can be realized.** The same two return null where the order is absent
  or does not count, and `made` returns null where the order has no place at the level the rule
  wrote. Those are facts about the model, and they are not the same fact.

The last of them is where the word goes wrong. `LevelSpace` writes it down itself: a rule holding
two strings three apart "asks for a line at a distance the order has no place for at all, which is
not a line this compiler could not find — there is none". Reported as `UnreadComparisonDomain`, a
model that states something with no line is described as one this compiler was not able to read.

This is #1148's shape one layer down: a reason recovered by a consumer from an absence, rather than
said by what holds the fact.

## Measured

Over the whole suite, which reading answers, against whether every term of the form has an order
that counts (`AffineReading.carriers`):

```
atAPosition          carriers present    1960
overAForm            carriers present    1285
apart                carriers present     360
apart                carriers ABSENT        7
NoLineOnTheQuantity  carriers absent        1
```

Three things follow.

**`apart` answers 7 times with no counting order.** `ComparedTerms` is deliberately wider than
counting: two strings stand no measurable distance apart and are still one above the other, so the
place they meet is a line. `aCarrierOfStringsDrawsOneToo` fixes that as a contract. Any design that
asks for counting orders before `ComparedTerms` runs loses those lines.

**`overAForm` refused nothing in 1285 opportunities.** Its own preconditions are the claim not being
`Nothing` and the carriers being present, and past those `made` refuses only on `canCutAt` — which
the sole override belongs to the meeting-point order, and that order does not count, so the carriers
gate has already excluded it.

**The one answer reached has no order at all**, on `guard a > b` between two cases of a sum. The
`ComparisonClaim.Nothing` route is dead: `AffineReading.read` sends a non-comparing operator to
`CutsNothing` before `realized` is called.

## What to build

**Ask a recogniser first, and let its answer decide which reading owns the form.** A shape that was
recognised and then could not be realized is two facts about one shape; falling from the second to
the next reading is how the first came out as the last one's shortfall. And the prerequisite of the
general form is asked where it belongs, which is not the front: the narrower readings keep the world
they are meant to cover.

```java
private static Read realized(...) {
    ComparedLine line = ComparedLine.fromTheForm(read, quantities);
    if (line != null) {
        return new Read.Cuts(realizedAt(atAPosition(behavior, line, quantities),
                "a line on one position", read));
    }
    ComparedTerms pair = ComparedTerms.fromTheForm(read, quantities);
    if (pair != null) {
        return new Read.Cuts(realizedAt(apart(behavior, pair, read.claim(), quantities),
                "a distance between two positions", read));
    }
    Map<NumericTerm, TermOrders> on = read.carriers(quantities);
    if (on == null) {
        return new Read.NoOrderToCountOn(AffineReading.filedAt(read.form().coefs().keySet()));
    }
    return new Read.Cuts(overAForm(behavior, read, on, quantities));
}
```

A recogniser's null is not free of facts about the model — an order the shape would need may be
missing, and that is worth saying. What it is not is this reading's to classify: a reading further
down may still answer, and what a form on an order that counts nothing is left with is settled at
the one place that knows it has run out of readings.

The two nulls above the gate do not establish the reason `realized` may publish. The answer below it
does, and it is the only thing `NoOrderToCountOn` is ever made from.

**`overAForm` becomes total.** Handed the orders, it has nothing left to refuse: the claim was
settled before `realized`, the carriers are the parameter, and `canCutAt` cannot be false for an
order every term counts on. Left nullable, the next refusal added to it arrives at
`NoOrderToCountOn` under a word that does not describe it.

**`canCutAt` stays, and stops being a route to a published word.** It is a real question and `Seam`
and `movedTo` ask it for their own reasons. What goes is `made` turning a false into a null that
becomes `UnreadComparisonDomain`: at that point the producer's contract has been broken, and it
should say so rather than be reported as a limit of this compiler. If the language later admits an
offset on a non-counting order, this stops at the contract rather than reusing a word silently.

**The arm is named for the fact, not the outcome.** `NoLineOnTheQuantity` says only that no line
came of it, which the next four reasons would also satisfy. Its contract is not "there is no order"
but "there is no order this affine quantity can be realized as a count on" — `carriers` refuses both
an absent order and one that does not count, and the javadoc has to say so.

## Done when

- `realized` does not infer a reason from several readings having returned null. A recogniser is
  asked first and its answer decides which reading owns the form; a line is built only after that
- a shape that was recognised and could not be realized does not fall to the next reading. There is
  one way that happens — an order with no place at the level the rule wrote — and it is a fact about
  the model, so it may not come out as whatever the reading below was short of
- `NoOrderToCountOn` is made at that one place and nowhere else
- `ComparedLine` and `ComparedTerms` still run before the gate, so a line on an order that does not
  count is not lost. `aCarrierOfStringsDrawsOneToo` is the control on that side
- `aComparableTypeWithNoCarrierDrawsNone` is the control on the other: no order at all, and the
  answer says that rather than being what is left when three readings declined
- `overAForm` returns a `Cutting` once the orders are in hand
- no route leads from `ComparisonClaim.Nothing` to a published word. Making it inexpressible below
  `realized` wants a claim type that carries no such case, across five files that had just been
  reworked, and is #1206
- `canCutAt` returning false does not become `UnreadComparisonDomain`. Unreachable in the language as
  it stands, it is a broken contract where it is met, and not silently reused if it ever is
- the word for the case that is reached is established by what `realized` holds, so
  `ComparisonAssessment` writes it rather than infers it

kawasima:
Closes #1193. Splits off #1206.

## What was wrong

`Cutting.realized` tried three readings and concluded from their combined silence, and
`ComparisonAssessment` wrote one word for that silence — `UnreadComparisonDomain`, the values
compared are not ones a line can be drawn on here.

Nothing established that. The nulls it read do not mean one thing:

- **which shape this quantity is.** `ComparedLine.fromTheForm` answers null for a form that is not
  over one coordinate, `ComparedTerms.fromTheForm` for one that is not a pair. A dispatch, carrying
  no fact about the model.
- **whether the shape can be realized.** The same two answer null where an order is absent or does
  not count, and `made` where an order has no place at the level the rule wrote. Facts, and not the
  same fact.

`LevelSpace` writes the last one down itself: a rule holding two strings three apart "asks for a
line at a distance the order has no place for at all, which is not a line this compiler could not
find — there is none". Reported as `UnreadComparisonDomain`, a model that states something with no
line is described as one this compiler was not able to read.

This is #1148's shape one layer down — a reason recovered by a consumer from an absence rather than
said by what holds the fact.

## What it is now

```java
// Which shape it is, narrowest first, and each answered before anything is built.
ComparedLine line = ComparedLine.fromTheForm(read, quantities);
if (line != null) {
    return new Read.Cuts(realizedAt(atAPosition(behavior, line, quantities),
            "a line on one position", read));
}
ComparedTerms pair = ComparedTerms.fromTheForm(read, quantities);
if (pair != null) {
    return new Read.Cuts(realizedAt(apart(behavior, pair, read.claim(), quantities),
            "a distance between two positions", read));
}
// And what the general form needs, asked once and here.
Map<NumericTerm, TermOrders> on = read.carriers(quantities);
if (on == null) {
    return new Read.NoOrderToCountOn(AffineReading.filedAt(read.form().coefs().keySet()));
}
return new Read.Cuts(overAForm(behavior, read, on, quantities));
```

**A recogniser is asked first, and its answer decides which reading owns the form.** A recogniser
answering nothing is this form not being that shape, and the next shape is asked; a recogniser that
answered and a line that then could not be built are two facts about one shape, and the second stops
at `realizedAt` rather than falling to a reading whose own shortfall would become the reason.

A recogniser's null is not, in itself, free of facts about the model — an order the shape would need
may be missing, and that is worth saying. What it is not is this reading's to classify, because a
reading further down may still answer. A form on an order that counts nothing is turned away by both
narrower shapes and is settled at the one place that knows it has run out of readings.

**That place is after the shape readings, not before.** They reach orders that do not count — two
strings stand no measurable distance apart and the place they meet is a line — so asking for
counting orders first refuses lines the model draws.

**`overAForm` is total.** Its preconditions were the claim and the carriers; both are settled before
it is called. Past them `made` refuses only where an order that counts has no place at a level,
which is two readings disagreeing about one order rather than a rule this compiler is short of, and
it says so where it happens.

**`canCutAt` stays.** `atAPosition`, `apart`, `movedTo` and `Seam` ask it for their own reasons, and
a false there is a real answer for a distance on the meeting-point order. What is gone is `made`
turning one into a null that becomes a published word.

**The arm is named for what was found.** `NoLineOnTheQuantity` held for every way of failing to
build a line, so the next one added would have gone out under the word this one earned.
`NoOrderToCountOn` is made at one place, from the answer that decides it.

## Measured

Which reading answers, over the whole suite, against whether every term has an order that counts:

```
atAPosition           carriers present   1994
overAForm             carriers present   1285
apart                 carriers present    360
apart                 carriers ABSENT       7
NoOrderToCountOn      carriers absent       1
```

The 7 are why the gate cannot come first. The 1285 are why `overAForm` is total: `made` refused none
of them, and the only `canCutAt` override belongs to an order that does not count, which the gate
has already excluded.

## Controls

`aCarrierNoOrderCountsOnSaysThatAtEachPositionOfTheQuantity` is new, and it is the one the route was
missing: before it, changing the published word for this answer to anything else left the whole
suite green. It pins `a: UNSUPPORTED_DOMAIN, b: UNSUPPORTED_DOMAIN` for the model that reaches the
arm.

`aCarrierOfStringsDrawsOneToo` is the same question from the other side. The pair holds this to "is
there an order this quantity can be realized on" rather than "is this domain numeric".

Mutations: moving the gate ahead of the shape readings turns 5 tests red across 4 classes, that one
among them; changing the word for the reached route turns the new control red.

And for the boundary above it, one mutation shows both halves. Make the meeting-point order refuse
the place two values meet, then run `aCarrierOfStringsDrawsOneToo` twice: with the shapes falling
through to the general form it reports `UNSUPPORTED_DOMAIN`, and with the recogniser asked first it
stops and names the shape it recognised. A recogniser succeeding and `made` then refusing happens 0
times in this suite for either shape, so what stands there is a tripwire on a language change and is
written as one.

Full suite green, 7788.

## Split off

#1206 — `ComparisonClaim.Nothing` is still spelled six times below the point where an operator has
been established to compare. The arm that turned it into a published word is gone here; making it
inexpressible wants a narrower claim type across `AffineReading`, `ComparedLine`, `ComparedTerms`,
`ReachingCuts` and `Cutting`, which is the reason it is not in this PR.

### DIFF
--- a/souther-compiler/src/main/java/souther/compiler/partition/ComparisonAssessment.java
+++ b/souther-compiler/src/main/java/souther/compiler/partition/ComparisonAssessment.java
@@ -268,10 +268,12 @@ sealed interface ComparisonAssessment {
             case Cutting.Read.Stopped stopped -> stopped.why().isEmpty()
                     ? aboutNoPosition(comparison, reads, read.symbols())
                     : new Unread(stopped.why());
-            // And where the quantity was read and no line could be built on it, the carrier is
+            // And where the quantity was read and stands on no order this counts, the carrier is
             // what a reader is owed — at the quantity's own coordinates, because the quantity is
-            // what such a rule is about.
-            case Cutting.Read.NoLineOnTheQuantity over -> over.over().isEmpty()
+            // what such a rule is about. The word is what the reading established and not what is
+            // left when several answers were absent: it says the values here carry no order to
+            // draw a line on, and that is exactly what was found.
+            case Cutting.Read.NoOrderToCountOn over -> over.over().isEmpty()
                     ? aboutNoPosition(comparison, reads, read.symbols())
                     : new Unread(atEachOf(over.over(),
                             new BlockReason.UnreadComparisonDomain()));
--- a/souther-compiler/src/main/java/souther/compiler/partition/Cutting.java
+++ b/souther-compiler/src/main/java/souther/compiler/partition/Cutting.java
@@ -191,12 +191,14 @@ record Cutting(BorderQuantity of, Level at, ComparisonClaim claim,
      * values, two positions held apart, and a form over several. Nothing here reads the comparison
      * again — what each of them is handed is the form the reading already came to.
      *
-     * <p><b>Which shape it is, and what that shape needs, are two questions and only the second is
-     * a fact about the model.</b> The two narrower readings answer nothing for a form that is not
-     * theirs, and read as an answer that told a rule whose shape was simply not a line on one
-     * position that its values carry no order. So the shapes are tried first and say nothing when
-     * they decline, and what the general form needs — an order that counts, under every term — is
-     * asked once, between the last of them and it.
+     * <p><b>A shape declining and a recognised shape failing to draw are two different things.</b>
+     * A recogniser answers nothing where the form is not its shape and where the orders that shape
+     * would need are not there, and neither of those is an answer about the model — the next shape
+     * may still draw the line, which is exactly what a form on an order that counts nothing does
+     * one reading further down. What must not happen is the other one: a shape that was recognised
+     * and then could not be realized falling to the next reading, where whatever that reading is
+     * short of becomes the reason. So a recogniser is asked first and its answer decides which
+     * reading owns the form, and only then is a line built.
      *
      * <p>Asked in that order and not before them. The narrower readings reach orders this counts
      * nothing on: two strings stand no measurable distance apart and the place they meet is still a
@@ -204,18 +206,20 @@ record Cutting(BorderQuantity of, Level at, ComparisonClaim claim,
      */
     private static Read realized(String behavior, AffineReading read,
                                  Quantities quantities) {
-        // Which shape this quantity is, asked narrowest first. Both of these answer null for a form
-        // that is not theirs, and that is a dispatch and no fact about the model — read as one, a
-        // rule whose shape simply was not a line on one position came out as a rule about values
-        // nothing draws a line on.
-        Cutting drawn = atAPosition(behavior, ComparedLine.fromTheForm(read, quantities),
-                quantities);
-        if (drawn == null) {
-            drawn = apart(behavior, ComparedTerms.fromTheForm(read, quantities), read.claim(),
-                    quantities);
+        // Which shape this quantity is, asked narrowest first, and each answered before anything is
+        // built. A recogniser answering nothing is this form not being that shape, and the next
+        // shape is asked; a recogniser that answered and a line that then could not be built are
+        // two facts about one shape, and falling from the second to the next shape is how the
+        // first came to be reported as the last one's absence.
+        ComparedLine line = ComparedLine.fromTheForm(read, quantities);
+        if (line != null) {
+            return new Read.Cuts(realizedAt(atAPosition(behavior, line, quantities),
+                    "a line on one position", read));
         }
-        if (drawn != null) {
-            return new Read.Cuts(drawn);
+        ComparedTerms pair = ComparedTerms.fromTheForm(read, quantities);
+        if (pair != null) {
+            return new Read.Cuts(realizedAt(apart(behavior, pair, read.claim(), quantities),
+                    "a distance between two positions", read));
         }
         // And what the general form needs, asked once and here. Not before the two above: they
         // reach orders that do not count — two strings stand no measurable distance apart and the
@@ -227,6 +231,31 @@ record Cutting(BorderQuantity of, Level at, ComparisonClaim claim,
                     AffineReading.filedAt(read.form().coefs().keySet()));
         }
         return new Read.Cuts(overAForm(behavior, read, on, quantities));
+    }
+
+    /**
+     * The line a recognised shape draws, which is one it draws.
+     *
+     * <p>What is watched here is a shape this reading recognised and could not put a line on. Only
+     * one thing refuses past recognition — an order with no place at the level the rule wrote
+     * ({@link LevelSpace#canCutAt}) — and it is a fact about the model rather than a limit of this
+     * compiler: a rule holding two strings three apart asks for a line at a distance the order has
+     * no place for at all, and there is none to find.
+     *
+     * <p>Unreachable while the language admits no arithmetic over an order that counts nothing, so
+     * a distance on such an order is only ever the place the two meet, which every such order has.
+     * That is a fact about what can be written and not about these types, so it stops here rather
+     * than falling to the reading below — where a shape that was recognised would have been
+     * answered for by whether the general form's orders were there, and gone out as values nothing
+     * draws a line on.
+     */
+    private static Cutting realizedAt(Cutting drawn, String shape, AffineReading read) {
+        if (drawn == null) {
+            throw new IllegalStateException("this reading recognised " + shape
+                    + " and its order has no place at the level the rule wrote: "
+                    + read.form() + " at " + read.cut());
+        }
+        return drawn;
     }
 
     /**
--- a/souther-compiler/src/main/java/souther/compiler/partition/Cutting.java
+++ b/souther-compiler/src/main/java/souther/compiler/partition/Cutting.java
@@ -48,11 +48,16 @@ record Cutting(BorderQuantity of, Level at, ComparisonClaim claim,
     /**
      * What reading {@code comparison} as a line came to.
      *
-     * <p>Three answers, and the two that are not a line are not one absence. A comparison whose
-     * positions cancel was read from end to end and there was no line in it; one this compiler could
-     * not take apart leaves whatever it states unknown. Told apart by a {@code null}, whoever asked
-     * had to work out which — and worked it out by reading the comparison a second time, which is
-     * how a rule read in full came to be described as one whose spelling defeated this compiler.
+     * <p>Four answers, and the three that are not a line are not one absence. A comparison whose
+     * positions cancel was read from end to end and there was no line in it; one this compiler
+     * could not take apart leaves whatever it states unknown; and one whose quantity was read and
+     * stands on no order this counts is neither. Told apart by a {@code null}, whoever asked had to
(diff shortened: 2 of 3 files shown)


kawasima:
Self-check pass, second commit.

Five things read as patched at the site. Four share a cause.

**The prose that governs the mechanism was left describing the old one.** `Read` said three answers and there are four; `read` said "which of the three"; `realized` named three shapes and never mentioned the question now asked between the second and the third, which is the whole of what this PR changes; `overAForm` still claimed a gate of its own that its caller holds.

Not four slips. One habit: I wrote the explanation into the body at the place I edited and left the javadoc — what a reader meets first — saying what the method used to do. That is the same site-or-owner mistake this branch is about, one layer up. Two of the counts had already been stale since #1192 and I walked past them twice.

**The fifth: the refusal in `overAForm` explained its reasoning but not what would fire it.** It now names that: `OverAForm.levels()` answers `steppingBy` or `overFiniteDecimals` and neither names a single place, the order that does is the one two values meet on, and every term counting has already excluded it — so an override added to either of those two is the change this watches. That is the difference between a tripwire and a throw put there in case, and it is why I kept it rather than removing a branch nothing can reach.

One thing I looked at and left alone: `ComparisonAssessment`'s class javadoc counts "six ways a comparison leaves the positions nothing" and lists six while `NoFeasibleInput` is a seventh record that also draws no line. It predates this branch, the file was reworked by #1201 last week, and my change neither added nor removed an arm — so I have not touched it. Say the word if it should go in here.

Full suite green, 7788. Head is `795d5457`.

kawasima:
Blocker closed at the boundary it was left at, and the two follow-ups with it.

**Root cause.** I applied the principle to `overAForm` and not to the two readings above it. The reason is how I derived the sweep: from the symptom I had measured — the arm reached through `carriers() == null` — rather than from the invariant, which covers every place `realized` turns a null `Cutting` into "try the next thing". Three such places, one closed.

**What it is now.** Each recogniser is asked first, and its answer decides which reading owns the form.

```java
ComparedLine line = ComparedLine.fromTheForm(read, quantities);
if (line != null) {
    return new Read.Cuts(realizedAt(atAPosition(behavior, line, quantities), "a line on one position", read));
}
ComparedTerms pair = ComparedTerms.fromTheForm(read, quantities);
if (pair != null) {
    return new Read.Cuts(realizedAt(apart(behavior, pair, read.claim(), quantities), "a distance between two positions", read));
}
Map<NumericTerm, TermOrders> on = read.carriers(quantities);
if (on == null) {
    return new Read.NoOrderToCountOn(AffineReading.filedAt(read.form().coefs().keySet()));
}
return new Read.Cuts(overAForm(behavior, read, on, quantities));
```

A recogniser answering nothing is this form not being that shape. A recogniser that answered and a line that then could not be built are two facts about one shape, and the second stops rather than falling.

**The mutation that shows it.** Make the meeting-point order refuse the place two values meet, and run `aCarrierOfStringsDrawsOneToo` both ways. With the fall-through it reports `UNSUPPORTED_DOMAIN` — the silent reuse the Done condition forbids. With the recogniser asked first it stops: "this reading recognised a distance between two positions and its order has no place at the level the rule wrote". Same mutation, two behaviours.

Measured first, before designing: over the whole suite, a recogniser succeeding and `made` then refusing happens 0 times for either shape. So this is a tripwire on a language change, and it is named as one rather than a throw put there in case.

**The comment was wrong and is fixed.** `ComparedLine.fromTheForm` and `ComparedTerms.fromTheForm` do not answer null only for a form that is not theirs — they also answer null where the orders that shape would need are absent. That is exactly why falling through is right when nothing was recognised and wrong when something was, and the javadoc now says that. The PR description above has the same sentence and I have left it standing so the correction is visible; read the code comment as the current one.

**#1193's Done when is realigned.** The `ComparisonClaim.Nothing` item now reads "no route leads from it to a published word", with the type-level inexpressibility recorded as #1206. Two items were added for the boundary this commit closes.

Full suite green, 7788. Head is `f480a0c5`.

Issue #1205, opened 2026-08-31
A name a clause may write and a position a row writes at are one word


kawasima:
`PositionReading.positionsUnder` answers a sum's shared names as positions under the sum:

```java
Map<String, Type> positionsUnder() {
    return atOwnPath ? fields : Map.of();
}
```

`fields` at a sum whose cases share a spread is what those cases spread, so `依頼.応募期限` comes back
as a position under `依頼`. The reading of a behavior's input says the opposite about the same
declaration: a sum is retained with branches, and the positions are under the cases —
`依頼@プロジェクト依頼.応募期限` and the rest.

Neither is wrong. `InvariantChecker` is the reader here, and it walks these to seed the values a
clause of the type may name; a clause may name `q.limit`, so a name to walk is exactly what it
wants. What is doubled is the word. "Position" means a place a row writes a value in one mechanism
and a name a clause may write in the other, and a reader crossing between them has nothing to tell
it which it has.

## Why this is not #1137

#1137 separated what is readable off a value from what a value is composed out of, and gave each of
the four questions about a shape an owner. This is the layer above: `field`, `name`, `position` and
`path` are used for overlapping things across `check`, `inputs` and `partition`, and settling that
means saying which of them a value can be written at, which a clause can name, and which a report
counts — for every mechanism that spells one, not only for the two this method sits between.

Filed rather than folded in because the termination condition is different: #1137 ends when four
answers have four owners, and this ends when a vocabulary is settled and the mechanisms that use it
agree.

## What to look at

- `PositionReading.positionsUnder` and `atOwnPath`, whose other reader (`TypeGuarantees`) uses the
  flag for a different purpose — a newtype's contents at no path of their own;
- `InvariantChecker`, the caller, and what it actually needs: names to seed, which it could ask for
  as names;
- `InputDomain` and `StructuralInspection`, which already keep "the positions under this one" apart
  from "the names readable on a value here" (`ReadableFields`);
- whether anything else reads one word where it means the other.

## Done when

- what a clause may name and what a row writes a value at are asked as two questions, each by a
  name that says which it is;
- no mechanism reads one of them where it means the other;
- the vocabulary the report and the reading publish agrees with what the mechanisms hold.

kawasima:
Settled by #1209, merged into develop as 2d6e5095.

[...]

PR #1214, opened 2026-08-31
A claim that says the operator is no comparison is answered for below where one is required

### LINKED ISSUE #1206: A claim that says the operator is no comparison is answered for six times below where one is required

kawasima:
Closes #1206.

## The cause

The issue names six arms that answer for `ComparisonClaim.Nothing` under
`Cutting.read`, where an operator has been established to compare. Reading for
the invariant behind them rather than for the list found fifteen, in two shapes
and under two gates:

| what established it | what still answered for it |
|---|---|
| `AffineReading.read`'s `op.compares()` | `AffineReading.turned`, `ComparedLine.fromTheForm`, `ComparedTerms.fromTheForm`, `ReachingCuts.relOf`, `Cutting.holdsAtTheValue`, `LineFacts` (a constructor check and `holdsAtTheValue`), `Border` ×4, `GuardThresholds`, `EnsuresThresholds`; and `Cutting.asWritten`, which asked the operator again |
| `ComparedNumber.drawsALine()` | `ComparedLine.of`, `NumberWays.leaves` |

They are one defect: **a `boolean` cannot carry what it established**, so every
reader below the test still held the wide classification. What each of them
answered was invented — a `null` for a relation that does not exist, a `false`
for a rule that holds at no value because there is no rule, and, in the readings
#1205 added, a `throw` for a state nothing can build. `LineFacts` checks in its
constructor exactly what a type can refuse.

`Nothing` is not removable. Measured, not assumed: a `throw` in
`ComparedNumber.onOneSide` for a non-comparison that names a number fails 16
tests, reached as `ValueArrivals.comparedOut → NumberWays.comesOut →
ComparedNumbers.of` with `SUB` and `MUL`. That reading is above the boundary and
the wide answer is what it is for. The same probe in `AffineReading.read` leaves
the whole suite green, so the gate that was there was already dead — an arm for
"not a comparison" added there would have been the same defect one level up.

## The mechanism

`ComparisonPlacement` is what any operator places, `ComparisonClaim extends
ComparisonPlacement` is what a comparison placed, and only the first has an arm
for an operator that places nothing. Between them is `Comparison` — a node and
its claim, made only by `Comparison.of` and so itself the evidence that the
operator compares. `Core.Binary` no longer crosses the boundary: what recognised
a comparison mints one and hands it down, and the readings take that.

```
BinOp ──▶ ComparisonPlacement.of ──┬── Nothing ──▶ ComparedNumber only
                                   └── ComparisonClaim
                                            │
                              Comparison.of (private constructor)
                                            │
   Condition.Compares · ComparisonCatalog.Catalogued · declaration · ensures
                                            │
        ComparisonAssessment ─▶ Cutting ─▶ AffineReading ─▶ the readings
```

The fifteen arms are gone because javac no longer requires them, not because a
`default` was added. So are `ComparedTerms.ordersStrictly` (a third spelling of
an operator set already spelled twice), `ComparedNumber.mirrored(BinOp)` (a swap
is now the claim turned round, so there is one way from an operator to a
meaning), `AffineReading`'s static `turned`, and `LineFacts`'s constructor check.
`ComparisonCatalog.Comparison` becomes `Catalogued` and holds one.
`CoverageSites.requireIsACatalogued` keeps its two diagnostics and asks
`Comparison.of` for the first: what is checked there is a plan against a
catalog, and "not a comparison" and "not in this catalog" are different
breakages.

`ComparisonPlacement.Nothing` is now written in one place in the compiler:
`Comparison.of`.

## What holds it up

The arms cannot be tested — nothing reaches them, which is why they went on
answering wrongly. Two things a test can say hold the arrangement instead.

Four laws: what an operator places is a claim exactly where `BinOp.compares`
says so; turning a placement round twice leaves it (asked of every operator,
`Nothing` included); reading the operator a swap would have been written with is
turning the reading round — which is the theorem that lets `mirrored(BinOp)` go.

Two counts off the compiled classes: who may read `ComparisonPlacement.of`
(`Comparison.of`, `ComparedNumber.of`, and the boolean spelling beside them) and
who may ask `orders` (`InvariantBound.ordering`), each with why.

Neither is idle. Three mutations, each red at the test meant for it:
`Cut.turned()` returning `this` (the swap law), `NE` classified as `Nothing`
(the membership law), one added `ComparisonPlacement.of` call in
`ComparedTerms.asWritten` (the count, naming the method).

`Comparison.equals` reads the claim as well as the node, though the claim is a
function of the node. Left out, `AnAnswersIdentityCoversEverythingItAnswersWith`
fails, and adding an entry to its exception list would be silencing a rule that
exists for a real failure. Reading it costs nothing: one node has one claim.

## Left for a follow-up

`InvariantChecker.arithmeticOf` gates on `compares()` and
`Arithmetic.holdsOfEveryRow(BinOp)` answers `false` for the operators it
excluded. Same shape, different question — that one evaluates a comparison at a
residue rather than saying what it placed, and deriving it from the claim is its
own piece of algebra. `ComparisonPlacement.orders` and `InvariantBound.ordering`
go with it: a clause of a `data` is read off the syntax tree, so it has no
recognised comparison to carry a claim, and that boundary is a design of its own.

`mvn -o test`: 7815 tests in the compiler, all green.

### DIFF
--- a/souther-compiler/src/main/java/souther/compiler/check/Comparison.java
+++ b/souther-compiler/src/main/java/souther/compiler/check/Comparison.java
@@ -1,0 +1,79 @@
+package souther.compiler.check;
+
+import souther.compiler.core.Core;
+
+import java.util.Optional;
+
+/**
+ * A binary the language reads as a comparison, together with what its operator placed.
+ *
+ * <p><b>The two are one value because they have to agree.</b> What a rule places is a function of
+ * the operator it was written with, and every reader below the point where a node was recognised as
+ * a comparison needs both: the node to read the two sides from, and the claim to say what the rule
+ * states about the line. Passed as two arguments, the pair is only as true as each caller made it;
+ * held here, there is nowhere for a claim of one operator to arrive beside the node of another.
+ *
+ * <p><b>Made in one place, which is what carries the recognition.</b> {@link #of} is the only way
+ * in, so a value of this type is itself the evidence that the operator compares — a reader holding
+ * one has no case left for an operator that placed nothing, and no reason to go back to the operator
+ * and ask again. A {@code boolean} answering the same question leaves the fact behind at the
+ * {@code if}, which is how the readings below came to carry the wider classification and answer for
+ * a case none of them could be handed.
+ *
+ * <p><b>Equal where the nodes are equal.</b> The claim is read off the node's operator, so it adds
+ * nothing to tell two of these apart, and a record holding one compares as it did when it held the
+ * node alone. This is a comparison's value and not an occurrence's identity: which occurrence of a
+ * comparison a body is at is a question about the tree, kept where the tree is
+ * ({@link souther.compiler.coverage.ComparisonCatalog}), and a reader wanting that asks about the
+ * node.
+ */
+public final class Comparison {
+
+    private final Core.Binary at;
+    private final ComparisonClaim claim;
+
+    private Comparison(Core.Binary at, ComparisonClaim claim) {
+        this.at = at;
+        this.claim = claim;
+    }
+
+    /** {@code at} as a comparison, or nothing where its operator compares no values. */
+    public static Optional<Comparison> of(Core.Binary at) {
+        return switch (ComparisonPlacement.of(at.op())) {
+            case ComparisonPlacement.Nothing _ -> Optional.empty();
+            case ComparisonClaim claim -> Optional.of(new Comparison(at, claim));
+        };
+    }
+
+    /** The comparison itself, for a reader of what its two sides name. */
+    public Core.Binary at() {
+        return at;
+    }
+
+    /** What its operator placed on the values. */
+    public ComparisonClaim claim() {
+        return claim;
+    }
+
+    /** Everything this holds, which is what an identity is of. The claim adds nothing to the
+     *  answer — it is read off the node's operator, so two of these over one node carry one claim —
+     *  and it is read here all the same, because a field left out of an identity is a field a
+     *  reader of the identity is told has not changed. */
+    @Override
+    public boolean equals(Object other) {
+        return other instanceof Comparison that
+                && at.equals(that.at) && claim.equals(that.claim);
+    }
+
+    /** Off the node alone, which spreads these exactly as the claim would: one node has one
+     *  claim. */
+    @Override
+    public int hashCode() {
+        return at.hashCode();
+    }
+
+    @Override
+    public String toString() {
+        return "Comparison[" + at + "]";
+    }
+}
--- a/souther-compiler/src/main/java/souther/compiler/check/ComparisonClaim.java
+++ b/souther-compiler/src/main/java/souther/compiler/check/ComparisonClaim.java
@@ -1,9 +1,7 @@
 package souther.compiler.check;
 
-import souther.compiler.types.BinOp;
-
 /**
- * What a comparison places on a position's values, read off the comparison and nothing else.
+ * What a comparison placed on a position's values, read off the comparison and nothing else.
  *
  * <p>One classification, asked wherever a rule compares a position to something: a clause of an
  * invariant, a comparison in a body, a clause of an {@code ensures}. A rule is read the same way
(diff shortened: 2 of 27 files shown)


kawasima:
Filed the two the description leaves out: #1215 for the invariant side, where a clause is read for its operator again after `InvariantBound.ordering` has settled it, and #1216 for `Arithmetic.holdsOfEveryRow`, the one reading under a recognised comparison that still holds an operator.

Two and not one because the answers differ. #1215 wants what a clause placed, which is `ComparisonClaim`, and needs a recognition for `Hir.Expr`. #1216 wants the operator evaluated at a residue, which is derivable from the claim and is its own piece of algebra — and it runs on `Core.Binary`, where `Comparison.of` already exists.

kawasima:
Both findings had one cause, so they are closed by one change.

**The cause.** `ComparedNumber.of` was one door for two worlds: the reading of any binary a walk met, and the only way to the number a recognised comparison is about. A narrow caller could only enter through the wide door, so it dropped the warrant one call short and the door read the operator again; and the `Nothing` inside `line()` was reachable or not depending on which caller had come through. Finding 2 is the same thing seen from the check: the licence in the test is given to a *method*, and while one method serves both worlds, naming it necessarily licenses both. Counting invocations alone would not have said `ComparedLine.asWritten → ComparedNumber.of` was wrong.

**The fix.** Two doors. `ComparedNumber.lineOf(Comparison, ...)` takes the claim from the comparison, reads no operator, and answers with a `DrawnLine` or nothing; `of(Core.Binary, ...)` stays what it was and is now **package-private**, which leaves `ComparedNumbers` as the one way to it from outside. `ComparedLine.asWritten` takes a `Comparison` and `Cutting.asWritten` stops calling `.at()` for it. Which side of a binary the reading is about, and whether saying so turned the two round, is factored into one `sideOf` both doors use — asked twice, a rule with its number on the right would be about one position for one caller and another for the other.

Verified the two ways it can come back:

- the old routing no longer compiles — restoring `ComparedNumber.of(comparison.at(), ...)` in `ComparedLine` fails with `of(...) はpublicではありません`, so this is javac's now rather than a test's;
- the count bites — a second `ComparisonPlacement.of` inside `ComparedNumber.of` fails the tripwire with `ComparedNumber.of=2` against the declared `1`.

The test now declares who may ask, **how many times**, and why, and says in its own words what a count cannot say: who came through a door is the package boundary's answer, not the count's.

One more from a self-check while going over the diff: `Cutting.holdsAtTheValue` had no reader, and narrowing the claim had turned it into a delegate rather than into a question about who wants it. Removed in `d8a258b03`.

`mvn -o test` green (7815), Checkstyle clean.

Issue #1223, opened 2026-09-01
The interner reads the operator for which canonical term a comparison is


kawasima:
Found by the inventory #1217 added, which lists everything still carrying an operator.

## What is written

`Term.Interner.operator(BinOp, Term, Term)` reads the operator to say which canonical term a
comparison is:

```
EQ -> Shape.EQ
NE -> not Shape.EQ
LT -> OP LT (left, right)
GT -> OP LT (right, left)
GE -> not OP LT (left, right)
LE -> not OP LT (right, left)
```

Its javadoc says what it is doing: "Six comparisons are three: `>` is `<` the other way round, and
`>=` and `<=` are the denials of the other two."

That is the same fact `Conditions.canonical` now reads off what a comparison placed — which sides the
canonical form wants is which class the value named is in, and whether the assertion turns over is
whether the comparison holds at the value. Two spellings of one normalisation, one of them in the
operator's own words.

## Why this is not "call the other one"

The two do not produce the same thing. The reading of a guard produces a `Core` comparison and a
polarity, and the interner produces a `Term` under a `Shape`. What they share is the normalisation,
and the caller — `Terms.written(BinOp, Core, Core, Denotations)` — holds an operator because the
interner takes one.

So the question is how far up `Terms.written`'s entry can be narrowed. It is reached from
`Terms.theOneOf` with a `NumericMeaning.Operator`, which holds an operator of its own, and that is
reached from a walk over `Core.Binary`. Which of those is where a comparison should be recognised is
what has to be measured before anything is moved.

## Done when

- the six-into-three is read from what a comparison placed, wherever it is written
- `Terms.written` and `Term.Interner.operator` either take what was placed, or say in one sentence
  which of their questions the operator is still needed for
- `WhereAnOperatorMayStillBeHeldIsWrittenDownTest` loses the lines that say otherwise

kawasima:
Merged to develop as 03bda103b (PR #1240). `Closes` does not fire on a merge into develop, so closed by hand.

Issue #1222, opened 2026-09-01
A comparison's occurrence is identified by the node it was recognised from


kawasima:
Left from #1217, which listed the readers of `Comparison.at()` rather than closing them.

## What is written

`Comparison` holds the `Core.Binary` it was recognised from, and hands it out. Ten methods take it,
for two different reasons:

- the two sides the rule names — `ComparedNumber.lineOf`, `AffineReading.read`,
  `ComparedTerms.asWritten`
- which occurrence of a comparison in the tree this is — `ComparisonCatalog`, `CoverageSites`,
  `ComparisonReadings`, `Condition.Compares`, `Cutting`, `EnsuresThresholds`, and
  `ComparisonAssessment`, which asks for both

The first is a value question and the second is an identity one, and the node answers both because it
is what a coverage site is keyed by.

## Why this is worth closing

While the node is handed out, the operator is one call away from every reader that holds a
comparison, and what stops one being read is that nobody has yet.
`WhereAnOperatorMayStillBeHeldIsWrittenDownTest` says nobody does, which is a check on a habit rather
than on a mechanism.

Handing out `left()` and `right()` closes half of it and leaves the other half, because the identity
readers need something that tells one occurrence from another — and today that is the node.

## What it would take

Decide what identifies an occurrence of a comparison. The catalog keys on the node; a coverage site,
a finding's place and a reading's citation all follow from it. Give that its own value, hand the
sides to the readers that want them, and `Comparison` has no node to hand out.

## Done when

- no reader takes a `Core.Binary` out of a `Comparison`
- what identifies an occurrence is a value of its own, and the catalog is keyed by it
- the list of readers of the node is gone from
  `WhereAnOperatorMayStillBeHeldIsWrittenDownTest` rather than shortened

kawasima:
Closed by #1281 (merged into develop as a3868803).

The node stops being what identifies an occurrence. A comparison is what its operator placed and the two sides it placed it on; which comparison it is, is a name the catalog issues — module, behavior, and where it stands among that body's comparisons; where a run through it is recorded is an address the plan hands out. Three questions, three answers, and no way back from any of them to the binary the recognition was read off.

Issue #1221, opened 2026-09-01
A composed comparison is written back into the tree as a node


kawasima:
Left from #1217, which made this the only reason the way back from a meaning to an operator exists.

## What is written

`Conditions.asOrderComparison` reads a comparison of what an operation answering an order answered,
works out in the numeric domain which way that answer stands to nought, and then writes the result
back into the tree as a `Core.Binary` — composing the operator with
`ComparisonWriting.operatorStating`. Its callers (`Predicates.assumeCond`, `Predicates` at the clause
reading, `Conditions.stating`) compare the answer with what they passed in, and where it differs they
read the new node from the top: negation, conjunction, recognition, both sides into forms.

So a fact the reasoning already holds — a claim and two expressions — is spelled as syntax and read
back as a fact.

## Why this is worth closing

`ComparisonWriting` exists for this one caller. Nothing else in the compiler needs a relation written
as an operator, and while it is there it is a way from what a rule means to how one is written that
any later reader can reach for.

The re-reading is also work that cannot say anything: a composed comparison is never a negation and
never a conjunction, and it was recognised before it was written.

## What it would take

Have `asOrderComparison` hand back what it read — the claim and the two sides — rather than a node,
and have the callers take that tail directly instead of recursing. Then `ComparisonWriting` has no
caller and goes, and the crossing between how a comparison is written and what it means is one
arrow.

## Done when

- nothing composes a `Core.Binary` out of a relation
- `ComparisonWriting` is gone, and `AnOperatorIsAskedWhatItPlacesInOnePlaceTest` says so by having
  nothing left to license for writing
- what the callers do with an order comparison is what they do with any other one

kawasima:
Closed by #1250 (merged into develop).

PR #1224, opened 2026-09-01
Ask what a comparison placed, in the words each reader answers in

### LINKED ISSUE #1217: Three readers turn a comparison by rebuilding its operator
Left from #1215, which named these as readers it did not move.

## What is written

`ComparisonClaim.turned()` is what turning a comparison round means, and `ComparisonPlacement.of` is
the one place an operator is read for what it places. Three readers turn a comparison by rebuilding
the operator instead, each with its own copy of the same table:

```
check/Conditions.mirrored              LT->GT LE->GE GT->LT GE->LE default op
flow/Witnessed.mirrored                the same six lines
codegen/InvariantConstraints.mirrored  the same six lines, with `default -> op // == and /= read
                                       the same either way`
```

They agree with `ComparisonClaim.Cut.turned()` and with each other today. An operator added to the
language lands in four tables, and only one of them is next to the classification that says which
operators compare.

## Why this is not "delete three switches"

The three sit in readers that hold a `BinOp` for their own reasons, and each asks a different
question of the turned comparison:

| reader | what it answers after the turn |
| --- | --- |
| `Conditions` | which `NumericDomain.Rel` the comparison is |
| `Witnessed` | whether a value satisfying it exists in the domain |
| `InvariantConstraints` | which decoder constraint it becomes |

Replacing `mirrored` alone in any of them leaves the operator being read for meaning one line later,
which is the shape #1215 was about, moved rather than closed:

- `Conditions` also has `relOf(BinOp)`, `opOf(Rel)` and a normalisation switch that rewrites `NE`,
  `GE`, `GT` and `LE` into `LT` forms. The boundary here is not only operator to claim but claim to
  `Rel`, and there is a bidirectional conversion already.
- `Witnessed` reads the operator directly for what it means: `against(BinOp, long, boolean)` and
  `compares(BinOp)` stand beside `mirrored`.
- `InvariantConstraints` has five `of*(BinOp, Hir.Expr, Hir.Expr)` methods — `ofInt`, `ofDecimal`,
  `ofListSize`, `ofMapSize`, `ofStringLength` — that interpret the operator into a constraint.

So this is one question asked of three readers, not one repair applied three times.

## What it would take

For each reader, measure how far up its entry can be narrowed to a `ComparisonClaim` — `Comparison.of`
for a `Core.Binary`, `ClauseComparison.of` for an `Hir.Expr` — and move only the ones where the
operator is not needed below that point:

- `Conditions`: how much of it needs a `BinOp` at all, and whether `Rel` is derivable from the claim
  rather than from the operator.
- `Witnessed`: whether `against(BinOp, ...)` can be asked of the claim, which makes it a question
  about a `Cut`'s satisfying side and a `Singled`'s value.
- `InvariantConstraints`: whether each constraint mapping follows from the claim, so that the walk
  reads `Hir.Binary` once through `ClauseComparison`.

## Done when

- each of the three readers either takes what the comparison placed rather than its operator, or
  says in one sentence which of its questions the operator is still needed for
- no `mirrored(BinOp)` remains in a reader that could have held a claim
- `AnOperatorIsAskedWhatItPlacesInOnePlaceTest` names every mint that is left, with its count

kawasima:
Closes #1217.

## What was wrong

Not that three readers had a copy of one operator table. `ComparisonClaim` said what a comparison
placed and answered none of the questions a reader of a comparison actually asks — which relation it
states, how it comes out at a value, which end of a range it bounds — so every reader went back to
the one value that can answer all three, the operator, and wrote its own table. Turning the
comparison round is what a reader has to do on the way to its own answer, which is why `mirrored`
appeared once per reader. Replacing `mirrored` alone would have left the operator being read one line
later.

Measured, the same shape was in more places than the issue named:

| question | tables |
| --- | --- |
| which relation | `Conditions.relOf`/`opOf`/`negateRel`/`mirrored`, `ReachingCuts.relOf`/`negated`, `NumericDomain.opposite` |
| how it comes out at a value | `Witnessed.against`, `InvariantChecker.holdsOfEveryRow`, `AffineConstraint.settledByConstantAlone` |
| which end it bounds | `InvariantConstraints`, five mappings each spelling out that a bound refusing the number it names admits what the next one along admits |
| whether it compares at all | `BinOp.compares`, `Witnessed.compares`, `Conditions.relOf(op) != null` at five sites |

## What it is now

```
BinOp --ComparisonPlacement.of--> ComparisonClaim --statedRelation--> Rel --holds/denied--> readers

Rel --ComparisonWriting.operatorStating--> BinOp        (composition only, one caller)
```

`ComparisonClaim` is the partition vocabulary and `Rel` the proposition one; the six of each line up,
so the crossing is written once rather than wherever a reader wants the other words. Nothing crosses
back from a claim to an operator: the way back takes a relation, because what is composed is composed
out of what the rules proved.

Each reader keeps the fact that is its own — the residue, the constant, which signs the whole numbers
hold something for, which of Raoh's constraints can carry the answer — and asks for nothing else.

## What holds it

- Three laws over all six comparisons: every relation is written as a comparison stating it; an
  operator written back from what it placed is itself; denying before and after the crossing is one
  relation; and turning a comparison round is the same statement about the other difference. The last
  is the only one that catches a reader taking the difference the wrong way about — read at nought
  the two conventions agree exactly.
- The three readings were written down as data first, with both directions of every operator, and are
  unchanged by the move.
- Two tripwires. One counts the crossings, in both directions. The other is where an operator can be
  at all, and its rule is the machine's rather than a list of the ways anybody thought of: to use a
  value as an operator the type has to be established, which is written down as a symbolic
  reference, and which instructions carry one is a closed set the class file format decides — held
  to that set by javac, since the kinds are sealed and the switch has no default. So the list is
  every method whose code names the operator, beside every method whose signature does and every
  field of that type.

  Neither works out what was done with it. A switch, a constant compared against, a set asked for
  membership and a map asked for an answer are four spellings of one act, and a check that knew
  three would be a check somebody could write the fourth past. What each holder does with an
  operator is the reason written beside it — the copiers included.

  What it does not see is written down: an operator reached by name through reflection is a string,
  and what is read is this module's classes.

## Measurements

- Full suite green throughout, and after taking `develop` in: 11,288 tests.
- The inventory named four the compiler already had, none of them visible to a descriptor.
  The reading of what values a rule admits took `==` and `/=` off the node and worked out which of
  the two it stated under the denials it stood beneath, which is what a comparison placed and what
  denying it comes to; it asks for those now. And the reading of whether a rule stands one position
  against another held the six comparing operators in a `Set` of its own, which is a membership the
  enum already answers for; it asks the enum now. Beside them it named two that stay and now say so:
  a library fact's operator composed into a comparison everything downstream reads back as one, and
  the parsed tree's operator translated into this one through the name each is spelled with.
- Eleven ways round have been written and run against the list: a switch, a `Map` read with `get`, a
  `List` through a cast, the enum's own `values` array, an array of arrays, a constructor, an
  anonymous class, a static initialiser, an operator handed over as an `Object`, one named through
  the class itself, and a table keyed by the name an operator is spelled with. All eleven are named.
- And the list of instruction kinds is javac's: taking one arm out of the switch was run, and the
  build refuses it.
- Every characterization was shown to see its subject: negating the residue's sign, negating the
  constant's sign, making `mirrored` the identity, opening the `GT` arm, turning a claim's
  `turned()` into the identity, and flipping the end a bound is at — each was run and each went red
  where it should.
- The one place with a comparison's node and an operator both to hand reads a disjunction it has not
  recognised as a comparison, and passes the node it did recognise to where a finding is filed.

## What is left, and named

- A composed comparison is still written back into the tree as a `Core.Binary` and read again from
  there, which is the only reason the way back exists at all (#1221).
- A comparison's occurrence is identified by its node, which is why `Comparison.at()` exists and why
  the operator stays one call away (#1222).
- The interner still asks which of the canonical terms a comparison is in the operator's own words,
  which is the six-into-three the reading of a guard now takes from what was placed (#1223).
- Three memberships of operators are spelled outside the enum, where the enum's own answers are
  exhaustive switches (#1225).


kawasima:
Fixed in `fab7155`. The blocker was right, and the cause was one step above the two accessors.

## The cause

The inventory named where an operator can be **stored** — a descriptor is what a declaration says it
carries — and the defect this change is about is a **read**, which is a call site. A reading that
takes the operator out of a node and switches on it there puts one in no signature and holds one in
no field, so it arrives in nothing that was declared.

## What is checked now

Every method that both gets hold of an operator from a `Core.Binary` or a `Hir.Binary` and asks it
something: names one of the constants, switches over them, or puts a question to the enum. 29 of
them, each with what it decides.

Taking is not deciding, which is why the list is 29 and not 111: a pass that copies a tree reads
every operator it meets and writes each one back into the node it is rebuilding, and there is nothing
to say about that. `Core.atSlots`, `Hir.withRegion` and `HelperInliner.rename` are all of that kind.

Asking the enum its own questions is in the list rather than out of it, so
`Predicates.atomsNamedBy` is named as you asked — as "asks the enum whether the operator compares:
both sides of a comparison are named whatever it states". A reader that stops consulting `compares`
and spells the set out again is then a line that changes rather than one that appears.

## It found one on its first run

`AdmissibleReading.leaf` took `==` and `/=` off the node and worked out which of the two it stated
under the denials it stood beneath:

```java
if (e instanceof Core.Binary b && (b.op() == BinOp.EQ || b.op() == BinOp.NE)) {
    return comparison(b, (b.op() == BinOp.EQ) == positive);
}
```

That is `Singled` and its denial, derived from the operator. Its signature is
`leaf(Core, boolean)`, so nothing about it reached the descriptor list, and it spells neither
`mirrored` nor `relOf` so no sweep of this change's vocabulary would have found it either. It asks
the claim now.

## One thing found while reading the 29

`BodyGen.orderingOf` spells `switch (op) { case LT, LE, GT, GE -> true; default -> false; }`, which
is a third membership written outside the enum beside `Terms.isArith` and `Condition.combines`.
Added to #1225 rather than to this change: an ordering is a comparison that is not an equality, and
where that belongs on `BinOp` is a decision about the enum's surface rather than about this reading.

Full suite green at 11,237.

kawasima:
Fixed in `377aa51`. The `Map` counterexample was right, and it is a symptom rather than the gap.

## The cause

The check was working out from the class file whether a reading **decided** anything on the operator
it took. A switch, a constant compared against, a set asked for membership and a map asked for an
answer are four spellings of one act; the check knew two, so there was a third and a fourth to write
past it. Any predicate of that kind has a complement.

So the taking is what is fixed now. It is one instruction, every reading of an operator begins with
it, and it has no fourth spelling.

## What changed

`everyOperatorTakenOutOfATreeIsWrittenDown` is now the full set of callers of `Core.Binary.op()` and
`Hir.Binary.op()` — 49 methods, each with what it takes one for. The heuristic list is gone rather
than kept beside it: two populations where one is a guess at a subset of the other is a second thing
to keep in step, and the guess is the half that was wrong.

The copiers are in the list, as you said, with lines like "copies it into the node it is rebuilding".
Anyone making one of them do more has to edit its line.

Run against your counterexample — a `Map<BinOp, Rel>` in a static field, read with `get` from a
method whose descriptor names no operator — the list names `Conditions.relationOf`.

## It found a second real one

`Relates.twoPositions` held the six comparing operators in a set of its own:

```java
private static final Set<BinOp> COMPARES = Set.of(BinOp.EQ, BinOp.NE, BinOp.LT, BinOp.LE, BinOp.GT, BinOp.GE);
```

Erased to `Set` in the descriptor, and `contains` rather than a switch — your hypothetical, in the
compiler already. It is a membership `BinOp.compares()` owns, so it asks the enum now rather than
going to #1225: nothing had to be decided to fold it, unlike the three there that have no owner yet.

## #1225

Title and Done when now say three, and the fourth is recorded as folded where it was found.

Full suite green at 11,237.

kawasima:
Fixed in `290606a`. (The head you reviewed, `377aa51`, has since been followed by `6464dae`, which
had already widened the taking to every way an operator arrives, and by this one. `develop` is
merged in as well.)

## The cause

The exclusion was written for the first version of this list, when the population was the enum's own
members and `values`/`valueOf` had to go. It was then carried through three rewrites without being
asked again, and against the population it ended up guarding it said something else entirely: no
constructor and no static initialiser is looked at.

That is the third time in this change that something survived a rewrite because I did not re-derive
it from the rule. The rule is "a method that has an operator", and both of those are methods.

## What changed

`<init>` and `<clinit>` are in both lists. Six record constructors are declared, each with what it is
handed one for — a value is handed its operator when it is made, which is the one call that decides
what it holds for as long as it lives.

What is skipped instead is a class the compiler *wrote*, told by `ACC_SYNTHETIC` rather than by the
shape of its name: the holder javac emits to switch over an enum holds an operator on behalf of the
method that does the switching, and that method is in the list under its own name. An anonymous class
is not one of those.

Re-derived beside it: a field written to is not a field read. Matching every `FieldInstruction`
counted `putfield` as an arrival, when what is stored came from somewhere else in the same method —
which is where the list already had it. Only `getfield` and `getstatic` count now.

## Measured

Written and run, each one named by the list:

- the table read in a constructor — your shape, `OperatorProbe.<init>`
- the same read inside an anonymous class — `OperatorProbe.1.get`, so the synthetic skip hides no
  hand-written table
- the table built in a static initialiser — `OperatorProbe.<clinit>`

Beside the four from the round before: a `Map` read with `get`, a `List` read through a cast, the
enum's own `values` array, and a switch on a fetched operator.

Full suite green at 11,288, with `develop` merged in.

kawasima:
`ca1b10c` — and the fix is not the exclusion.

## The cause

Four rounds, four versions of this list, and each one was the ways of having an operator that
somebody had thought of: what a descriptor shows, then a guess at what counts as deciding, then two
accessors a review named, then every source I could enumerate. Each version said there was no other
way. None of them had any ground for saying it, and each time there was one.

The `<...>` exclusion is the same thing once more: written for the first version, where the
population was the enum's own members and `values`/`valueOf` had to go, and carried through three
rewrites without being asked again.

## Ground for the rule

To use a value as an operator the type has to be established, and that is written down as a symbolic
reference. Which instructions can carry one is a closed set the class file format decides — a call, a
field, a constant, a cast, a made object or array — not something anybody works out from what a
program might say.

So the list is every method whose code names the operator, beside every method whose signature does
and every field of that type. `<init>` and `<clinit>` are methods and are in it, with no exclusion
left to carry. What is skipped is a class the compiler wrote, told by `ACC_SYNTHETIC`.

It is also one list now. Handed in and fetched stopped telling two things apart the moment a method
could be both, and a split nobody needs is a way to put a line on the wrong side of it.

## Shown rather than claimed

Ten ways round, each written and run:

| what | named as |
| --- | --- |
| switch on a fetched operator | ✓ |
| `Map<BinOp, Rel>` read with `get` | ✓ |
| `List<BinOp>` through a cast | ✓ |
| the enum's own `values()` array | ✓ |
| the table read in a constructor | ✓ |
| the same inside an anonymous class | ✓ |
| the same in a static initialiser | ✓ |
| an operator handed over as an `Object` | ✓ |
| `Enum.valueOf(BinOp.class, name)` | ✓ |
| a table keyed by `op().name()` | ✓ |

## And what it does not see

Written into the test rather than left to be found: an operator reached by name through reflection is
a string, and a string is not a reference to the type; and what is read is this module's classes, so
a reader compiled elsewhere is not there. Both are to be told about rather than widened to.

One thing measuring turned up that I had not known: the parsed tree has an operator enum of its own,
`Ast.BinOp`, and `Resolve.expr` bridges the two by the name each constant is spelled with. That site
is in the list with that as its reason.

Full suite green at 11,287, `develop` merged in.

kawasima:
`c27fade`. `multianewarray` is in, and it is not the fix.

## The cause

The last version said the set of instructions that can name a type is the class file format's, and
then transcribed that set from memory — with a `default -> false` to swallow whatever the
transcription missed. Claiming the machine's authority while typing my own list is the same mistake
this list has made in every version of it, and the missing kind is what it looks like.

## The fix

The kinds are sealed, so the `default` was the only thing making a memory of them necessary. Without
it the switch is javac's to check. Every kind is answered for; an arm left out is a compile error,
and an instruction the format grows later is a compile error the day the runtime is raised.

Shown rather than said: taking `ThrowInstruction` out of the switch was run, and the build refuses
it — `switch式がすべての可能な入力値をカバーしていません`.

The kinds that carry no reference are answered together in one arm, since Java 25 takes several
patterns in a label: they work on what is already on the stack, on a number, or on where to go next,
and where what they work on was put there is one of the arms above.

## Your counterexample

```java
BinOp[][] ops = new BinOp[1][1];
ops[0][0] = BinOp.LE;
return TABLE.get(ops[0][0]);
```

Written and run: the list names `OperatorProbe.relation`, beside the ten from the rounds before.

Full suite green at 11,287.

Issue #1215, opened 2026-08-31
An invariant's clause is read for its operator again after it has been established to order


kawasima:
Left from #1206 (PR #1214), which named this as the boundary it did not move.

## What is written

`InvariantBound.ordering(BinOp)` is the boolean form of what an operator places, and the reading of
a `data`'s clauses is built on it. Four callers ask it and then read the operator again:

```
InvariantChecker.endOf         !InvariantBound.ordering(bin.op()) && bin.op() != BinOp.EQ
InvariantChecker.boundOf       found == null || !InvariantBound.ordering(op) ? ... : ...
InvariantChecker.narrowedBy    about != null && InvariantBound.ordering(op) && ...
OrderedReading                 if (!InvariantBound.ordering(said)) ...
```

`InvariantBound` itself does it twice more. `of` swaps the two sides where the value is on the right
and turns the operator round with a table of its own:

```java
private static BinOp mirrored(BinOp op) {
    return switch (op) {
        case LT -> BinOp.GT;
        case LE -> BinOp.GE;
        case GT -> BinOp.LT;
        case GE -> BinOp.LE;
        default -> op;
    };
}
```

and `ordered(BinOp, Place, Carrier)` switches on the operator a third time, with
`default -> NO_END` for the cases `ordering` has already excluded.

That is the shape #1206 was about: a `boolean` settles what an operator states, the fact is left
behind at the test, and every reader below holds the operator and reads it again. #1214 closed it
for a body's and a declaration's comparisons and left this one, because the two are not the same
boundary — `InvariantBound` reads `Hir.Expr`, and `Comparison` refines a `Core.Binary`.

## Why it is not a defect today

`ordered`'s `default` is unreachable while `ordering` guards every call, and the two operator tables
agree. Nothing observes either of them being wrong, which is what makes it worth writing down: an
operator added to the language lands in three tables, and only two of them are next to each other.

## What it would take

An `Hir`-level recognition that carries what the clause's operator placed, the way `Comparison`
carries it for a `Core.Binary` — minted where a clause is recognised as a comparison, so
`InvariantBound.of` receives the claim instead of asking for it, `mirrored` becomes the claim turned
round (`ComparisonClaim.turned`), and `ordered` takes a `ComparisonClaim.Cut` rather than an
operator with a `default`.

Whether the two representations share one type or the `Hir` side gets its own is the design question
this issue is for. What they must share is `ComparisonPlacement`: the answer to "what does this
[...]

PR #1240, opened 2026-09-01
Derive what a comparison states from what it placed

### LINKED ISSUE #1223: The interner reads the operator for which canonical term a comparison is

kawasima:
Closes #1223.

## The defect

The six ways to compare two values state three things between them, and which of
the three was worked out in two places:

- `Conditions.canonical`, in the words of what the comparison placed;
- `Term.Interner.operator`, in the operator's own words.

The interner could do it because `Terms.naming`'s `Core.Binary` arm handed it a
raw `BinOp` — the walk never asked whether the binary was a comparison — which is
the boundary #1215 set: `BinOp` does not flow below the point where what a
comparison states is settled. #1215 also said the derivation from a claim's facts
is not the consumer's to write, and both of these wrote it: `exchanged` from
`valueBelongs`, and the denial from `holdsAtTheValue`, which the two shapes read
opposite ways.

Removing the operator table alone would have left that derivation in two places.

## What was measured before deciding

Every call into `Term.Interner.operator` over the whole suite, with its caller:

| entry | operators reaching it |
|---|---|
| `Terms.naming`'s `Core.Binary` arm | 134 — all six comparisons, and the arithmetic, the connectives and `CONCAT` |
| `Terms.written` | 1, and it was `DIV` |
| `Terms.repeating` | `MUL`, written there |

So the issue's question — how far `Terms.written`'s entry can be narrowed — has a
different answer than the issue supposed: it is not on the comparison path at
all. Its operator comes from a `DIV` it names itself and from
`NumericMeaning.Operator`, which is guarded at `Terms.numericMeaningOf` by
`answersANumber()` and built from library rows that are `ADD`/`SUB`/`MUL`. Nothing
said so in a type, which is what the first commit fixes.

`NumericMeaning.Operator` never carries a comparison either, so the walk over
`Core.Binary` was the only place a recognition could go.

## The shape

```
Core.Binary ──Comparison.of──▶ ComparisonClaim ──canonical(l,r)──▶ CanonicalComparison<A>
                                                                        ├──▶ Polar  (Conditions.AsPolar)
                                                                        └──▶ Term   (Term.Interner.AsTerms)
```

`CanonicalComparison` is what a claim states, over two values of whatever a reader
holds them as. Nothing about it can be taken apart — a reader that switched on
which of the three it is and forgot the denial would state the comparison that
holds exactly where this one does not — so the only way to read one is
`expressedAs`, which takes the reader's own `theSameValue` / `below` / `denied`.

Each reader keeps what is its own. A condition writes the comparison as a node
standing where the one it read did, with the denial carried by the polarity
beside it; the interner writes it as a term, whose equality has its two parts
unordered. The node a condition needs is held by its `Expression` and not by the
statement: where a comparison came from, what it answers and where it stands are
the representation's business.

`Term.Interner.operator` now knows nothing about comparisons and refuses one. No
name of any of the six appears in its body.

## Invariants

| what | how |
|---|---|
| what a comparison states is derived where what it placed is held | `OnlyAClaimSaysWhatAComparisonStatesTest`: only `Singled.canonical` and `Cut.canonical` call `theSameValue` / `below` / `denied` |
| who may ask a claim for it | the same test: only `Conditions.canonical` and the walk that names a binary |
| the operator has no way in | `Term.Interner.operator` refuses an operator that compares |
| arithmetic is keyed by an operator answering a number | the two constructors |

Both tripwires read the compiled classes. `CanonicalComparison` is this package's,
which keeps it out of the vocabulary anything outside reads — it does not decide
who calls what, since a method whose answer nothing outside can name is still a
method anything can call, so both questions are answered by the tripwires rather
than by the language.

`AnOperatorIsAskedWhatItPlacesInOnePlaceTest` is unchanged: the new recognition
goes through `Comparison.of`, which already holds its licence.

## Tests, and what each catches

Three layers. The first fixes the derivation against an algebra of its own, so a
term and a node that both got it wrong the same way would still not agree with
what is written down.

| mutation | what went red |
|---|---|
| `Cut.canonical` stops exchanging the sides | layer 1 (3), layer 2 (2), and 15 existing tests |
| `Cut.canonical` denies the other way round | layer 1 (2), layer 2 (1), **and nothing else** |
| `Singled.canonical` denies the other way round | layer 1 (2), layer 2 (2) |
| the interner stops refusing a comparison | layer 3 (1) |
| `answersANumber()` weakened to `!compares()` | red at `AND` |
| an unlicensed `denied()` call | construction tripwire |
| an unlicensed `canonical` call | canonicalization tripwire |

The second row is the one worth reading. A polarity keys a fact and is read from a
guard, so inverting it everywhere at once leaves the matching intact — before this
change the same error in `Conditions.canonical` would have been caught by nothing.

Layer 2's expected values were pinned as literals against `develop` and run green
there before the implementation was swapped, so the two implementations do not
share a rule.

The turn law is stated of the orders only. An equality's claim does not turn, and
`canonical` puts the sides down in the order it was given, so turning both gives
the same statement with its two sides written the other way round. That those two
are one thing is answered where the equality is written (`Shape.EQ` is unordered),
not by what it states.

`mvn -o test`: 7948 tests, green. The first commit alone is green too (7939).

### DIFF
--- a/souther-compiler/src/main/java/souther/compiler/check/CanonicalComparison.java
+++ b/souther-compiler/src/main/java/souther/compiler/check/CanonicalComparison.java
@@ -4,18 +4,18 @@ package souther.compiler.check;
  * The canonical statement a comparison comes to, over two values of whatever a reader holds them
  * as.
  *
- * <p>Six ways to compare two values state three things between them: that the two are the same
- * value, that one stands below the other, and the denial of either. Which of the three a comparison
- * states is decided from what it placed ({@link ComparisonClaim}), and it is decided once — a
- * reader that worked it out from the operator would be a second answer, and the two agree only for
- * as long as somebody keeps them so.
+ * <p>Two statements carry the six ways to compare two values: that the two are the same value, and
+ * that one stands below the other. Each of them and its denial is four, and the two sides of an
+ * order are six. Which of them a comparison states is decided from what it placed
+ * ({@link ComparisonClaim}), and it is decided once — a reader that worked it out from the operator
+ * would be a second answer, and the two agree only for as long as somebody keeps them so.
  *
  * <p><b>A proposition and not a record of one.</b> Nothing here can be taken apart. Which of the
- * three this is, which side each value is on, and whether it is denied are what a reader would have
- * to put back together, and putting them back together is the derivation this type exists to hold:
- * a reader that switched on the first fact and forgot the last would state the comparison that
- * holds exactly where this one does not. So the only way to read one is {@link #expressedAs}, which
- * hands the reader its own three constructors and assembles the answer here.
+ * two it is, which side each value is on, and whether it is denied are what a reader would have to
+ * put back together, and putting them back together is the derivation this type exists to hold: a
+ * reader that switched on the first fact and forgot the last would state the comparison that holds
+ * exactly where this one does not. So the only way to read one is {@link #expressedAs}, which hands
+ * the reader its own way of writing each and assembles the answer here.
  *
  * <p><b>Over any two values.</b> A comparison of two terms and a comparison of two expressions
  * state the same thing about their sides, and what differs is only what a side is and what a
@@ -24,7 +24,7 @@ package souther.compiler.check;
  */
 final class CanonicalComparison<A> {
 
-    /** How a reader writes the three statements in its own representation. */
+    /** How a reader writes the two statements, and a denial, in its own representation. */
     interface Expression<A, T> {
 
         /** The two sides are the same value. */
--- a/souther-compiler/src/main/java/souther/compiler/check/CanonicalComparison.java
+++ b/souther-compiler/src/main/java/souther/compiler/check/CanonicalComparison.java
@@ -1,0 +1,95 @@
+package souther.compiler.check;
+
+/**
+ * The canonical statement a comparison comes to, over two values of whatever a reader holds them
+ * as.
+ *
+ * <p>Six ways to compare two values state three things between them: that the two are the same
+ * value, that one stands below the other, and the denial of either. Which of the three a comparison
+ * states is decided from what it placed ({@link ComparisonClaim}), and it is decided once — a
+ * reader that worked it out from the operator would be a second answer, and the two agree only for
+ * as long as somebody keeps them so.
+ *
+ * <p><b>A proposition and not a record of one.</b> Nothing here can be taken apart. Which of the
+ * three this is, which side each value is on, and whether it is denied are what a reader would have
+ * to put back together, and putting them back together is the derivation this type exists to hold:
+ * a reader that switched on the first fact and forgot the last would state the comparison that
+ * holds exactly where this one does not. So the only way to read one is {@link #expressedAs}, which
+ * hands the reader its own three constructors and assembles the answer here.
+ *
+ * <p><b>Over any two values.</b> A comparison of two terms and a comparison of two expressions
+ * state the same thing about their sides, and what differs is only what a side is and what a
+ * statement about two of them is written as. So the values are the reader's, and so is what comes
+ * back.
+ */
+final class CanonicalComparison<A> {
+
+    /** How a reader writes the three statements in its own representation. */
+    interface Expression<A, T> {
+
+        /** The two sides are the same value. */
+        T theSameValue(A left, A right);
+
+        /** The left side stands below the right. */
+        T below(A left, A right);
+
+        /** What holds exactly where {@code statement} does not. */
+        T denied(T statement);
+    }
+
+    private final Statement<A> statement;
+
+    private CanonicalComparison(Statement<A> statement) {
+        this.statement = statement;
+    }
+
+    /** The statement that the two sides are the same value. */
+    static <A> CanonicalComparison<A> theSameValue(A left, A right) {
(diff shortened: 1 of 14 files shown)


kawasima:
Self-check, and what it found. Four, three of them real.

**Six copies of a narrowing, in the very change that removes one.** Three tests
reached `ComparisonPlacement.of` and cast to `ComparisonClaim` — the shape a reader
below a recognition stopped having, written six times in the tests of the change
that closes it. Narrowed once now, in `AsPlaced`, where the refusal says what
happened rather than arriving as a class cast in the middle of an assertion.

**The denial law carried a rule of the test's own.** Its expected side wrapped what
the claim stated, and for `NE`, `GE` and `LE` that is a denial of a denial — a shape
neither `canonical` implementation makes, so the law only held because the test
cancelled it back. Both readers do cancel (`Interner.not`, and `Polar` by flipping a
flag) and `CanonicalComparison` does not, which nothing said. The law now reads its
expected side off the same table as the other, with the six refusing pairs written
out beside it.

**A pair used as neither half.** `Terms.binary` held a `Comparison` and read its
claim while taking the two sides from the node it already had. `Comparison` is one
value so that a claim of one operator cannot arrive beside the node of another, and
holding it that way gives that up without gaining anything. It takes the claim, the
way the reading of a condition does.

**A counted claim.** `CanonicalComparison`'s javadoc said "six ways state three
things between them" and then counted a denial as one of the three. Two statements,
each of which may be denied, and an order that has two sides — which is where the
six come from. Reworded in the three places that had said "three".

`mvn -o test`: 7948, green.

kawasima:
Both findings are real, and they are one defect. `151a8442d`.

## Measured first

The concrete-receiver call does slip through. Adding one and reading the class
file back:

```
invokevirtual  souther/compiler/check/ComparisonClaim$Cut.canonical
```

and the tripwire stayed green.

## The cause the two share

Neither is a hole in what the check asserts — both are holes in what it watches,
and both because I wrote that set by hand. One named a single owner; the other
named three method names. A watched set written by hand has a complement, and
what falls in the complement is answered by silence. Fixing an owner comparison
and appending to a name list would leave that, and the next thing to arrive is
neither an arm nor a method on the value.

## What it does now

Nothing is named. Every method the compiler declares whose return type is
`CanonicalComparison` is read off the classes, and the two questions are the two
halves of that split by where the method is declared:

- declared on the value — how one is assembled;
- declared anywhere else — how one is asked for.

`ComparisonClaim$Cut.canonical` and `Singled.canonical` are declared methods, so
they are watched whichever way a receiver is typed. An operation added to the
value is watched the day it is written.

A third assertion says what the enumeration is: the six ways in, each with what
it is for. A way in nobody has written down is a finding before anything asks who
may call it — which is the half of #2 that appending to a list would not have
given.

## And the way past that enumeration

Keying on a return type cannot see a method that makes one and hands it back
wrapped, since the type argument is erased. Every one of these values comes from
the one constructor, so a fourth assertion holds the makers and the ways in to
making one to being the same set, both sides read off the classes.

## Mutations

| mutation | what went red |
|---|---|
| `ComparisonClaim.Cut cut; cut.canonical(...)` from an unlicensed method | the asking half, and the enumeration |
| a fourth operation `reversed()` on the value, called from a licensed reader | the assembly half, and the enumeration |
| a maker returning `Optional<CanonicalComparison>` | the closure assertion, alone |

The first of those was green before this commit.

`mvn -o test`: 7950, green. CI on `767a2c8a` and `868654542` both finished green
(runs 33489445898, 33490139602); the run for this commit is queued.

No production code changed — `git diff --stat` against the previous commit touches
only the test.

Issue #1258, opened 2026-09-01
An explanation's order is still an enum's declaration order on one half of a sentence


kawasima:
An explanation's order is the report's own, and two of them are still an enum's declaration order.

## What is wrong

The report says a point nobody can decide in two halves, and they now follow two rules about
the order they say their reasons in.

```
? undecided whether a row is at the ON point value = 0 (invariant Amount #1), and the observation
  of it was stopped by a limit, so where it stands could not be read
? nothing could show a row can be written at the ON point (comparison@280:15) — nothing was
  composed for it: this compiler stopped at how many elements a total is spread over, how many
  ways a total is decomposed
```

The first half is a `ReadingReasons`, whose order is written down in a `switch` over `ReadingGap`
and `Incompleteness.Code`: the reasons are said in an order that follows from what they are, and
arranging either declaration is not a change to what a report says.

The second half is a `WritabilityKnowledge.Prevented`. It holds `EstablishmentGap.Observation`
with an `EnumSet<Incompleteness.Code>` and `EstablishmentGap.Composition` with an
`EnumSet<CompositionBudget>`, and both are joined into the sentence in `EnumSet` order — which is
ordinal order, which is the order the constants happen to be declared in. Reordering
`CompositionBudget` for tidiness changes the second line of that sentence.

So one sentence, two orderings, and the difference is invisible from the report.

## Why this is not what #1247 closed

Two properties are being asked for and they are not the same.

*Permutation invariance* — the order the readings were walked in does not reach the explanation.
`Prevented` already has this: `EnumSet` iteration does not depend on the order gaps arrived in,
and `Prevented.normalised` folds them to one gap per kind before anything reads them. This is the
property #1247 was about and it holds on both halves.

*Independence from a declaration* — arranging an enum is not a change to what a report says.
`ReadingReasons` has this and `Prevented` does not.

## Why it was left out of #1247

The rule cannot be applied to half a sentence. `why(Prevented)` renders both `EstablishmentGap`
arms, so making the observation codes canonical without doing the same for `CompositionBudget`
would leave the same split one level down. `Incompleteness.Code` has seven constants and
`CompositionBudget` has fourteen, and every one of the fourteen would need an order and a reason
for having it — which is a decision about what a reader is shown, not a mechanical edit.

There is likely more of it. Every enum the report joins into a sentence is a candidate, and the
question of how many there are has not been asked.

## Done when

- the population is counted: every place a report sentence is built by joining a set or an enum
[...]

PR #1259, opened 2026-09-01
An open question carries what it is open on

### LINKED ISSUE #1247: An undecided obligation repeats its reason once per reading
Found while re-measuring souther-examples' `businesstrip` on `493668218`, the day the `?` mark
arrived.

## What is wrong

The reason on an undecided obligation is joined once per reading, so a point read at several
coordinates says the same clause several times in one sentence.

```
? undecided whether a row is at the ON point value = 1 (invariant 使用人数 #1), and the walk reached
no value there to read, and the walk reached no value there to read, and the walk reached no value
there to read, and the walk reached no value there to read, and the walk reached no value there to
read, and the walk reached no value there to read, and the walk reached no value there to read
    · read as 事前承認する/申請.予定費用[*].費目@交際費.使用人数: = 1
    · read as 事前承認を却下する/申請.予定費用[*].費目@交際費.使用人数: = 1
    · read as 最終承認する/申請.予定費用[*].費目@交際費.使用人数: = 1
    · read as 差し戻す/却下@最終承認却下済み.予定費用[*].費目@交際費.使用人数: = 1
```

Seven copies of `and the walk reached no value there to read`. The readings under it are what the
sentence is about, and they are already listed one per line.

## Where

`businesstrip` at `2269e58` — one obligation of `使用人数`, read at seven coordinates across four
behaviors. The behavior-level `?` lines in the same report carry the clause once, so it is the
declaration block's join rather than the sentence itself.

Since the mark is a day old this is presumably a loose fold over the readings rather than a decision
about what to say.

kawasima:
Closes #1247.

## What was wrong

A point read at seven positions and undecided at all of them said why seven times in one sentence.

```
? undecided whether a row is at the ON point value = 1 (invariant 使用人数 #1), and the walk reached
no value there to read, and the walk reached no value there to read, and the walk reached no value
there to read, ...
    · read as 事前承認する/申請.予定費用[*].費目@交際費.使用人数: = 1
    · read as 事前承認を却下する/申請.予定費用[*].費目@交際費.使用人数: = 1
    ...
```

The clause was joined once per reading of the line, and the readings it is about were already
listed under it, one to a line.

## The cause

The set it was joined from is right. `Weakening.BorderValueUnreadable` is keyed on the border each
reading was made at, because a module counting what it could not read counts one fact per line.
What was missing is the other question about the same readings — what an author is told about this
one point, where the border is no part of the answer and a clause said once per reading counts the
paths a fact arrived by.

Nothing owned that crossing, so every reader made it again. `ObligationDisposition.Undecided` named
which questions are open (`Uncertainty.COVERAGE`, `Uncertainty.WRITABILITY`) and carried neither
answer, and its own doc said to read it rather than what is beside it — which the report could not
do, since the answers were beside it. Two readers went back to the evidence and came out
differently: the writability half landed on `WritabilityKnowledge.Prevented`, a value normalised at
construction, and said what stopped it once; the coverage half landed on the raw `WeakeningSet` and
said its reason per reading. Both halves are printed by one `switch`, two lines apart.

The same reach-around dropped whole arms in silence: the coverage half picked
`BorderValueUnreadable` out with `instanceof`, so a way of weakening a measurement that is not that
one left the sentence with no reason at all and no compile error to say so.

## The change

The question carries its answer. `Uncertainty` is a sum whose arms hold what they are open on, and
`ObligationDisposition.of` — already the one place the pair is read — makes the crossing in a
`switch` over every `Weakening` arm: an arm is classified as a reason of the readings' or as one
said where it happened, and cannot arrive as neither.

`ReadingReasons` is the explanatory value: each reason once, in an order that is a written sequence
of the reasons rather than a number per reason, so two runs over one model write one sentence and
arranging a declaration elsewhere is not a change to what a report says. A value is made by keeping
what was met from that order, not by putting what was met in order, so nothing compares two reasons
and a walk's order has no way in. The same shape holds the order of the questions an obligation can
have open. It checks that invariant and never repairs it — the one place a canonical value
is built is the fold, and a constructor that quietly sorted whatever it was handed would put that
boundary wherever somebody called it.

`AdequacyReport.undecidedBecause` is handed the disposition and not the assessment it came from, so
the evidence is out of reach of the sentence rather than in reach and left alone.

`specification.adoc` §example-report-vocabulary gains the rule, since the report's vocabulary is
where a reader looks a word up: an obligation marked `?` names each distinct reason once however
many readings met it, says them in an order that follows from what they are, and does not say which
reading met which — that pairing is a word about a reading, and the readings are named under the
mark one to a line.

## How it is held

Three fault models, each caught at the level it belongs to.

The crossing is tested against duplicate readings (1, 2, 3 and 7 of them), against borders swapped
underneath it, against permuted discovery order, and against reasons that must stay apart —
including several readings of each of two reasons, which separates saying each reason once from
counting the readings.

The block a reader sees is fixed whole in the CLI. Not as the absence of a repeat: a report that
says no reason at all repeats nothing, and is the other way to get this wrong.

Extension safety is held in two ways, and which one holds what is deliberate. A `switch` with no
`default` holds every place a thing has to be *classified*: the crossing from what weakened a
measurement to what a point is told about, and both renderings of a reason. A written sequence holds
every place things have to be *ordered*, and what a sequence cannot say for itself — that it holds
everything — is one check per order, reading its population off the sealed type and the observation
codes rather than off a second list.

| adding | what stops |
| --- | --- |
| a `Weakening` arm | the crossing does not compile |
| a `ReadingGap` arm | both renderers do not compile; the order's coverage check goes red |
| an `Uncertainty` arm | the report does not compile; the order's coverage check goes red |

`ObligationDisposition.of` is deliberately not among these — it is a producer, and a question
produced elsewhere is not a contradiction.

The `Weakening` arm that arrived with `develop` while this was open — `InputNotRead` — is the first
thing the crossing stopped in earnest. It is classified beside `BoundaryNotDerived`, which comes out
of the same enum by the same call and carries the behavior's name and nothing else: the fact is
about the behavior, so it is said on that behavior's own measure line rather than under every point
the behavior has.

Measured rather than argued. Three mutants that compile: dropping the deduplication, collapsing an
observation's code, and re-expanding the clause per reading in the renderer. The first two are
killed by the crossing's tests, the third by the CLI block and by nothing else — which is what says
the two levels are not testing the same thing.

## Left out

`WritabilityKnowledge.Prevented` still says its reasons in `EnumSet` order. It has the property
this issue is about — the order the readings were walked in does not reach it — and lacks the
weaker one, that arranging an enum is not a change to a report. Applying that rule to half a
sentence is not possible: `why(Prevented)` renders `CompositionBudget` as well, and every one of
its fourteen constants would need an order and a reason for having it. Filed as #1258 with the
population to count first.


kawasima:
All three addressed in `7d3945b`. Point 2 is fixed at the cause rather than as proposed, so the reasoning is below.

## 2 — the rank collision, and why not `>=`

`>=` is a correct check, and it is a check standing beside the wrong carrier.

A total order over a finite set of reasons is the reasons written out in order. Both orders here
were a function into `int`, whose codomain is far wider: every collision and every gap in it is a
value that is not an order. The answer was in something wider than itself with a check beside it to
catch the difference — the shape this branch exists to remove, one level down. Tightening the check
keeps the carrier and keeps the hole one edit away.

It also has a cost. With `>=` alone, a wrongly written table throws `IllegalArgumentException` the
moment the two reasons meet in one obligation, so the failure mode is a compiler that dies while
writing its report. Printing two reasons the wrong way round is the better of those.

So the order is now the sequence. `ReadingReasons` holds every reason a reading can meet, written
out once in the order a report says them in, and `of` keeps what was met *from that order* rather
than putting what was met *in* order:

```java
List<ReadingGap> out = EVERY_REASON.stream().filter(found::contains).toList();
```

Nothing compares two reasons, so collisions, repeats and a walk's order are all unrepresentable —
there is no `Comparator` and no `sort` left. The constructor's one condition, that the position
moves forward, refuses a repeat, a pair out of order, and a reason the order does not hold.

The same table existed for the open questions (`ObligationDisposition.orderOf(Uncertainty)`, two
arms). Applying the rule to one order and not the other is the half-application this branch already
filed as #1258, so both are sequences now.

### What moved on the extension axes

A sequence cannot say for itself that it holds everything, so that is the one property a check
carries, and it reads its population off `ReadingGap.getPermittedSubclasses()` and
`Incompleteness.Code.values()` rather than off a second list. Measured again with a dummy arm on
each axis:

| adding | before | now |
| --- | --- | --- |
| a `Weakening` arm | the fold does not compile | unchanged |
| a `ReadingGap` arm | the order and both renderers do not compile | the renderers do not compile; the order's coverage check goes red |
| an `Uncertainty` arm | the order and the report do not compile | the report does not compile; the order's coverage check goes red |

The order's totality moved from the compiler to a check. That is the trade for the sequence, and it
buys three failure modes becoming unrepresentable instead of one being checked.

The collision was real, not theoretical: giving `LINKAGE_FAILED` the number already on
`ROW_UNDECIDED` — a pair no test names — left the whole suite green.

## 1 — spec

Narrowed to what the value quantifies over, and the second half of the rule is now written down
rather than left to the type's doc:

> An obligation marked `?` MUST name each distinct reason **its readings met** once, however many
> readings met that reason.
>
> What left the measurement short of an answer without being a reading's own is said where that
> measurement stopped, and is not repeated under each obligation. […]

## 3 — constructor tests

`new ReadingReasons(List.of(A, A))` and a reversed pair are refused, held against the constructor
directly rather than through the fold. `Undecided` has the pair as well: `[writability, coverage]`
and `[coverage, coverage]`.

## Run

`mvn -o clean test` from the reactor root: 11,495 tests, 0 failures, 0 errors.

Issue #1270, opened 2026-09-02
A weakening set keeps the order its causes were found in, and says a document prints it


kawasima:
`WeakeningSet` keeps the order its causes were found in, and nothing reads it.

## What is wrong

`WeakeningSet` holds its causes in a `LinkedHashSet` and says why:

> **Order is kept anyway.** Equality is the set's, so the order is not part of the value; the order
> the readers found things in is what a document prints, and a run that produced them in a different
> order each time is a report that cannot be compared with the last one.

The second clause is no longer what happens. What a document prints is a set of words from two
vocabularies, and those words are now written in an order of their own — the walk that found the
weakenings does not reach the array. Every other reader folds the causes into something else: a
membership question, a projection to one word, a classification. None of them is a reader of the
sequence.

So the carrier keeps a property with a reason attached that is not the reason it is kept for, and
the property is one a reader has no way to see.

## The question

Is the order the readers found the weakenings in something a person is owed, or is it the walk's?

Three answers, and they are different work:

* *Provenance.* It says which reader met what first, and a person debugging a measurement wants it.
  Then it should be a sequence that says so — a list, in a type that carries what it is a sequence
  of — and the equality that ignores it is the thing to fix, because two runs that met the same
  facts in two orders would be one value while carrying two answers.
* *Diagnostic.* It is worth having while somebody is looking at this compiler and not worth
  publishing. Then it belongs where such things are said, and the carrier can be a set with nothing
  claimed about its order.
* *Accidental.* A `LinkedHashSet` is what got written. Then the sentence goes and the type says
  what it is: a set of facts, unioned, with no order.

Nothing in the type decides between the three today, and the sentence quoted above reads as the
first while the code is the third.

## Why it is not what #1258 closed

#1258 is about a plurality of a kind of reason acquiring an order from a declaration — an
`EnumSet` iterating in the order somebody declared the constants in, which is nobody's decision
about what a reader is shown. It closed that by writing an order down per kind and making the
crossing to a published sequence the one place an order is put on.

This is the other direction: an order that *was* decided by something, and the question of whether
what decided it is something a reader may read. Answering it by giving the causes a canonical order
would say the encounter order is worth nothing, which is one of the three answers above and not a
mechanical consequence of #1258.

## Done when

* which of the three the encounter order is, is decided and written where the carrier is;
* the type matches that answer — a sequence that says what it is a sequence of, or a set with no
  claim about order;
[...]

PR #1283, opened 2026-09-02
Fold a WeakeningSet on the declared fact, not on the value that carries it

kawasima:
Part one of a **three-part stack** for #1270. **None of the three merges independently** — see the end.

`RuleWithoutALine.sameAs` and `Incompleteness.identity` each say what makes two findings one, and each leaves something out: the handle an author is sent to, and the place a reason was met at. `StandingQuestion` says the same of its handle in prose, and of what a reading was short of, and writes down no identity at all. The carrier that unions them folded on Java value equality — a finer relation than the one those types declare — so one fact carried by two values was two.

## What this claims

> When `WeakeningSet` and `MeasureClosure` union, the atom is the declared fact identity and not the Java value, and what a carrier holds beside the fact is accumulated rather than picked.

That and no more. **It is not a claim about the system.** Six producers upstream still fold the same rules on the same identity with `noneMatch(sameAs)` and keep the first representative, so a second handle for one rule is dropped before this union ever sees it. The downstream union no longer depends on producer uniqueness; the upstream normalization still does, and closing that is the third change in the stack — deliberately, because the handles it makes plural need a canonical witness, and that machinery is the second.

## What it costs today

Nothing observable. Running the suite with the merge instrumented to throw whenever it accumulates anything shows that no production path reaches it: of 11,581 tests, only the algebra tests added here exercise a merge. A rule's handle is a function of the rule at both producers, and the observation whose place is not a function of its fact — a linking that failed, cited at each example — cannot be raised in process. So this is a correction to the representation and not to an output.

No structural check is added for the property that made it harmless: it is the thing this change makes unnecessary at the carrier, and checking it would put an implementation fact back where an invariant goes.

## What changed

- The fact is a value at each of the three, and the evidence is accumulated commutatively beside it.
- `ClosureGap` answers what one of its gaps is and how two come together. `MeasureClosure` and `WeakeningSet` are two union boundaries that both ask it, rather than two places that decide it.
- Two hand-written first-representative folds go, in `ModuleReport.incompleteness` and in `Coverages`.
- `WeakeningSet` keeps no encounter order.

## Why it is not mergeable alone

Nothing decides what order a document prints these in any more. `AdequacyReport` writes one citation out of the set it now has, marked `TEMPORARY`, and the tests that pinned the walk's order now ask what an array holds rather than what order it holds it in. On its own this tree has better semantic identity and worse publication determinism, which is the property #1270 is about.


kawasima:
Folded into #1290 with #1284 — none of the three merged independently, and three branches to keep level with develop was not worth the split. The review here is addressed there: the claim was narrowed to the carrier and the upstream folds are closed in that PR's third commit.

PR #1284, opened 2026-09-02
Put an order on the arrays whose order this compiler decides

kawasima:
Part two of a **three-part stack** for #1270, on top of #1283. **None of the three merges independently.** Review #1283 first.

Two of the document's arrays came out in the order the account handed them over: the facts a verdict is open on, and what a module could not read. Nothing decided either. #1283 took the encounter order out of the account, which left them with no order at all; this is where one is put on.

## A place crosses before an order is asked for

A citation is a sum whose arms are not all places — a position in a text no reader holds is not one, and neither is code out of sight with no position in it — so there is no order over citations to be had, and asking for one would be asking which of two things that are not places comes first. `PublishedAt` is the projection onto the four things the schema has room for, and it names the source without resolving it: what a document calls a source is recorded by writing it, so a comparison that asked would have made the sorting decide the order of the table of sources. A fact met at several places is written at one of them, chosen from the places.

## `CanonicalArrangement` is beside `CanonicalSelection`, not a widening of it

A selection is a finite list of places with at most one member to a place, which is what a plurality of a kind of reason is. These arrays are as long as the model has facts. What the arrangement rests on is that a tie is not a choice — two entries a key cannot tell apart would come out in whichever order the caller handed them over — so an entry that compares equal to another it is not equal to is refused. The keys are over what a document writes and over nothing that went into deciding it.

The two arrays disagree on purpose and both are held to it. `keptOpenBy` counts facts, so two facts one word covers are two entries and neither is folded away; `incompleteness` says which source each was about, so the two are told apart there.

## Three surfaces, one order

The first version of this put the order on the document's arrays and left the other two surfaces that publish those facts walking the account: the page a person reads, and the comment block a generator writes. Both had been deterministic before #1283 stopped keeping an encounter order, so two compiles of an unchanged model wrote a different page and different generated files. The population was wrong rather than the fix — it came from the schema's arrays, which is one surface out of three. The order is now put on where the sequence is made, so there is no set left for a surface to walk.

## The population is declared, not inferred from the schema

That a field holds an array says nothing about where its order comes from — a module's behaviors are in the order somebody wrote them — so a check reading the schema would demand an order of arrays that have one, and would be met by sorting them. Which fields have an order of this compiler's is written down beside the orders, and the schema is asked whether it agrees rather than asked to produce the list.

## What holds it

- **Order** — the arrangement over every one of the twenty-four orders four facts could be met in; ties refused; witness selection independent of the order the citations arrived in.
- **Wiring** — a model two of whose sources were not read, with the two arrays and the source table written out; and the page held against the document, so a surface put back on the account is caught.
- **Bypass** — a bytecode guard: a field whose order this compiler decides is written by a method that writes that field and no other array, and that takes its sequence from a crossing. Held with a negative control so it cannot grow into "every array must be sorted". All four mutations measured.
- **Separation** — nothing in `publish` reaches anything in `report`, so what is decided cannot depend on what has been written so far.

## Two things found on the way

`AdequacyOpeningWord` moves to `publish`, and `NotMeasuredWord` is new: the schema's `notMeasuredReason` enum had no type behind it, and ordering by the raw strings made `String` a kind with an order, which the existing census caught.

`Adequacy.RowReading.NotAsked` carries a fourth word the schema does not allow. It is unreachable — `EverySchemaWordIsAccountedForTest` says so in prose and excludes it — so the crossing refuses it rather than naming it, and that prose is now a guard.

The specification's rule said the order of what a report says together is one of two. It is now three: the first two are about a plurality of a kind, and an array said once per fact needs an order over what is written of each entry.

## Note on the stack

The commit after the first one here also carries fixes to defects review found in #1283's own tree — the refusal `QuestionUnanswered` lost, a rule written twice over, and two unchecked casts. They sit here rather than below because the three merge together; say if you would rather they moved down.


kawasima:
Folded into #1290 — see the note on #1283. The findings here are addressed there: the non-JSON surfaces, the vocabulary census, the order-authority check, and the merge preconditions.

PR #1286, opened 2026-09-02
Hold a construction's origin where it can be decided, and let only a crossing change it

### LINKED ISSUE #1241: The parsed tree offers to say a construction was carried, which no parse decides
Found while closing #1235, which said that a form of the parsed tree is one the frontend
writes. This is the same question a granularity down, where it is about a component and not a
form, and the check that closed #1235 enumerates forms.

## What is written

`ConstructionOrigin` stands on two forms of the parsed tree, `Ast.NewData` and `Ast.Apply`,
and says where a construction came from: written here, carried in by a published body, or
carried in by a value this body named.

The frontend builds those two at four places — `AstBuilder` lines 849, 896, 902 and 1066 — and
passes `ConstructionOrigin.own()` at every one of them. The other two answers are made by three
members of the parsed tree: `Ast.NewData.publishedBy`, `Ast.NewData.carriedByValue` and
`Ast.Apply.carriedByValue`. Nothing calls any of them, asked of the method references in every
compiled class of every module and its tests. The pass that does carry a construction into a
reader, `HelperNames`, calls `Hir`'s three.

Nothing on this side reads the value either. `Resolve` copies it into the `Hir` node, and
`Ast.withRegion` and one rebuild in `AstBuilder` carry it along; no reader asks what it says
until the tree below.

`CoverageOrigin`, which stands on five other forms, is not like this: the frontend mints one
per construct with `CoverageOrigin.written(module, counter, kind)`, so that component varies
here and says something the parse decided.

## What it costs

A component with one value, and three ways to make the other two that nobody uses, reads as a
distinction the parsed tree draws. It does not draw it: whether a construction was carried is
decided by inlining, which runs below. A reader takes the parsed tree for something that knows
where a construction came from, and a writer of a new frontend pass has three methods offering
to say something the frontend has no way to know.

Nothing is wrong at run time, and the value that reaches the tree below is the right one.

## Where to start

Whether the parsed tree carries the component at all. `Resolve` is the one reader, and it could
answer with `own()` where it builds the `Hir` node, which is where the question first has more
than one answer. The other way is that the component stays and the three members go, which
leaves the parsed tree saying a construction is the reader's own — true of every one it holds,
and then said by a component that can hold nothing else.

Which way it goes decides what `Ast.Apply`'s and `Ast.NewData`'s components are; it is not a
question about `ConstructionOrigin` itself, which the tree below needs as it is.

## Done when

* the parsed tree does not offer to say where a construction came from unless a parse can
  decide it
* what a construction's origin is where inlining has not run is said in one place
* no member of the parsed tree makes a value no pass over the parsed tree can reach
* construction origin is held only where a later reader distinguishes its answers, and only the
  operations that create or carry a construction can change it — which is the rule the three above
  are each a case of, found by reading what else held one

kawasima:
Closes #1241.

`Ast.NewData` and `Ast.Apply` held where a construction came from, and no parse decides
that: the frontend wrote `own()` at all four places it built them, and the three members
that say otherwise had no caller in any module. The pass that carries a construction into
a reader works on `Hir`.

Reading it found the same shape twice more, so what closes here is one rule rather than
three fixes.

> Construction origin is held only where a later reader distinguishes its answers, and only
> the operations that create or carry a construction can change it.

## What each commit does

1. **The parsed tree stops holding one.** `Resolve` answers `own` where it builds the `Hir`
   node, which is where the question first has more than one answer.
2. **`ValueName.OfType` stops holding one.** Nothing read it — the permission check collects
   no unit data (spec §constructs-excludes-unit-data) — and two of its three producers passed
   `null`, which a reader would have failed on.
3. **The type moves beside the two forms that hold one, and its arms are declared where only
   `souther.compiler.ast` can name them.** `Own` / `Published(module)` / `ByValue` replace the
   `(String, boolean)` pair; the fourth state that pair could spell is the value's, which is
   how `carried` already read it. The forms are built through constructors that take no origin
   and rebuilt through `with` / `withArgs` / `withFunction` / `rewritten`, which carry it.
   `DataChecker` asks the node rather than reading the value off it.
4. **The door only the mark on a name went through is taken out** — `Hir.Var.Denoting.withReachedAs`
   had no caller left, and the frontier of positions that may hold a whole `ReachName` narrows
   by one.
5. **The two ADRs that describe the old shape say what the tree holds now.**

## What holds it

Two mechanisms, and the checks divide along them.

- **javac** refuses an origin made, asked or changed outside `souther.compiler.ast`. The type is a
  marker; the arms and the operations on them are `Origins`', package-private. `new Origins.Own()`
  and `nd.origin().carried(...)` from another package do not compile.
- **What javac cannot refuse** is a pass taking an origin off a node and handing it back in, so a
  check over the whole reactor's compiled classes forbids the crossing: from outside the package,
  naming anything inside it that takes an origin. It knows nothing about which forms hold one, so a
  method beside them taking an origin is not a way past it. It found `NewtypeDesugar` on its first
  run.
- **Where an origin may live** is the compiler's own question and is asserted in its tests: no form
  of the parsed tree holds one, below it exactly `Hir.NewData.origin` and `Hir.Apply.origin` do, and
  the type has no public protocol at all.

The repo-wide check lives in `souther-architecture-test`, a module that depends on every other so
that Maven builds them before it. That every module is depended on is asserted of this pom's own
`<dependencies>`, which is what Maven orders the reactor on — not of the build order, and not of the
classpath, which says how the build was invoked rather than what was declared.

Each was mutated and seen to go red:

| mutation | what caught it |
|---|---|
| the component back on `Ast.NewData` | the parsed-tree check |
| a `ConstructionOrigin` component on `Hir.Binary` | the carrier check |
| `carried` back on the interface | the no-protocol check |
| `nd.origin().carried(made)` in `DataChecker` | javac |
| `new Origins.Own()` in `partition` | javac |
| a constructor call handed an origin, from a module built after the compiler | the crossing check |
| a `Hir.copyWithOrigin` bridge in the package, called from outside | the crossing check |
| the dependency on `souther-cli` dropped, and moved to `<dependencyManagement>` | the population check |

`mvn -o -B verify` green on the final state — 12 modules.


kawasima:
Two things this PR touched but does not close, filed apart: #1288 (a rewrite used to restate which of a construction's fields had to be written, and nothing tells the two answers apart) and #1289 (`NewtypeDesugar` writes the construction that replaces an application as the reader's own, whatever the application said).

kawasima:
Self-check over the diff found four things, now in `4638d6c` and `d26f976`.

- The origin transitions were public default methods on the interface, so a pass holding one off a node could derive another — which the type's javadoc denied. They moved to `Origins`, package-private beside the arms; javac now refuses `nd.origin().carriedByValue()` from another package.
- `Apply.rewritten` asked a caller to list every slot but the origin, which is what `withArgs` exists to avoid. Replaced by `with(function, args, pos, region)` and `standingIn`.
- The class-file check claimed its population was every module. In a reactor build it is this module's main and test classes, which is all that names either form today.
- The parsed-tree check attributed a case to its neighbour that the neighbour does not measure.

Plus one comment: `publishedBy` now says why a call left standing takes no mark where the value walk marks one.

kawasima:
All three in `7c87510`, with the root cause first.

**Blockers 1 and 3 are one cause: the check spelled the set it watched, so I spelled the same set when I closed the package.** `HOLDS_ONE` assumed the two carriers and the predicate was narrowed to constructors, so "every way an origin can be answered" was never enumerated — and the interface's own members were the ones I missed. Both are now read off the trees: the forms are the record components whose type is an origin (asserted to be exactly two, so a third carrier fails where it is written), and the members forbidden from outside are every member of those forms whose descriptor takes one.

**Blocker 2 is a different cause: when the mechanism could not support the claim, I narrowed the claim instead of moving the mechanism.** That is what `d26f976` did. The check is now `souther-architecture-test`, a module that depends on every other, which is what makes Maven build them before it; the population comes from `RepositoryLayout` rather than a second reading of the root pom — the repository already refuses that, and its check caught my first attempt.

On 1: `publishedIn` and `carriedByValue` had already moved in `4638d6c` (after the head you read). `carried` and `viaValueReference` moved now, so the interface has no members. `Published` refuses a null module in its compact constructor.

Mutations, each seen red:

| | mutation | result |
|---|---|---|
| 1 | `DataChecker` back to `nd.origin().carried(made)` | does not compile |
| 2 | forbidden constructor call in `souther-program-api-test` | named by the check: `souther/program/api/AProbeThatHandsAnOriginOverTest` |
| 3 | `ConstructionOrigin` component on `Hir.Binary` | `[Apply.origin, Binary.construction, NewData.origin]` |

`mvn -o test` green: 12 modules, 1561 test classes.

kawasima:
All four in `0235a5f`, and one correction to the plan.

**1. The type is held to having no protocol.** Read off the type — no declared method, sealed, arms named in `souther.compiler.ast` and not public — rather than as names that may not appear, since a member here is public whatever it is called. Mutation: `carried` back on the interface gives `expected: <[]> but was: <[carried]>`.

**2. The module stays, but listing it first does not do what it was meant to.** I dropped the dependency on `souther-cli` and built: this module was still twelfth of twelve. The edges it still has — `souther-program-api-test` among them, written last — send it past every module declared after the one it forgot, so an omitted dependency is only visible when the module it forgot happens to be written after all the ones it kept.

So the completeness of the dependencies is asserted directly instead, of the classpath these tests run on: Maven puts one `target/classes` on it per dependency, and that set must be `RepositoryLayout.modules()`. The same mutation now fails, naming `souther-cli`. The root pom's order does no work, so the module is back at the end where it was.

**3. Test-scoped**, and Maven still orders on them: twelfth of twelve.

**4. Narrowed.** Which forms hold an origin is asserted in the compiler's own tests; the control here only needs a form to watch. The module's description says its population is the compiled output of the whole reactor and nothing else.

`mvn -o test` green: 12 modules, 1561 test classes.

Issue #1304, opened 2026-09-03
A held fact is a check that ran, not a value that carries what it settled


kawasima:
`OperationFactBinder` walks every declaration in `OperationFacts` and holds it to the library:
the operation is resolved against `Stdlib`, and a fact naming an argument has that position
checked against the declaration's parameters. What it hands back is the authoring declarations
themselves.

So the binding is a check that ran, not a value that carries what it settled. A reader below it
goes back to `OperationFacts` for the authoring fact — a `ValueName` and an `ArgumentRef` — and
puts the same question a second way. The two agree for as long as nobody changes either.

#1279 closed one of these. `MeansTheSameAsASizeOfNought` names another operation, and what a
reader does with it is write a call of that operation where a call of this one stands. That is
now held where both declarations are in hand and comes back as `BoundOperationFact`, so the
reader that rewrites has nothing left to check. Every other kind of fact was left as it was,
because `DischargeRules.holdToTheDeclaration` already refuses a position the declaration has not
— and a mutation confirms it does.

## What is still open

`ArgumentRef` is read below the binding, so `CallArguments.positionIn(ArgumentRef, ValueName)`
exists: a reader holds the word a fact wrote and the name of an operation, and works out the
position for itself. Bound, "which argument of which declaration" is one value and that function
has no reader.

And the binding runs as a side effect. `DischargeRules.Bound` asks for it during class
initialization, and its accessors read the authoring index afterwards — but several readers
reach `OperationFacts` without going through that holder at all (`Accumulations`,
`NumericAnswers`, `NumericMeasures`, `NumericReadings`, `NumericTerm`, `RunReach`,
`ElementLineage`, `ResultRange`, `Question`). For those, nothing says the binding has run. That
is what a value carrying the answer fixes and an initialization order does not.

## Done when

- `OperationFactBinder` no longer returns authoring `OperationFacts.Declared` as its bound
  result
- every proposition that needs the library to be settled has a domain-specific bound
  representation
- readers below the binding boundary do not query the authoring indexes
- `ArgumentRef` is not named below that boundary
- `CallArguments.positionIn(ArgumentRef, ValueName)` has no reader
- no reader reaches a fact by a path on which the binding may not have run
- adding an `OperationFact` arm forces the binder to state what bound value, if any, it produces

## Not this

A generic carrier — `Bound<T>`, `Certified<T>` — is the wrong shape. What a value of that kind
says is that something was settled, not what. The types this repository already reaches for name
[...]

PR #1305, opened 2026-09-03
Read a name against its declaration before a call stands for it

### LINKED ISSUE #1279: A synthesised call is not held to the signature it names
`Core.PreservedCall` does not say whether its arguments were held to the operation's
signature, and two of the places that build one never asked.

One of them types an application against what it names:

```
CallElaborator.preservedCall
    arity(call, params.size())          refuses another number of arguments
    new Core.PreservedCall(callee.denotes(), ca.cores(), ...)
```

Three more rebuild a call that came from there, keeping its operation and the number of its
arguments (`CallArguments.replacedIn`, and the two walks in `Core`). The other two build one
out of something else:

```
Conditions.asSizeComparison
    new Core.PreservedCall(DischargeRules.sizeMeantBy(call.operation()),
                           call.args(), Type.INT, call.pos())
                                        another operation, the same arguments

Elaborator (a library value kept standing)
    new Core.PreservedCall(v.denotes(), List.of(), ...)
                                        no arguments, whatever the name declares
```

Neither consults a signature. Nothing is wrong today: the first substitutes a size
(`List.length`, `Set.size`, `Map.size`, `String.length` — the range of
`MeansTheSameAsASizeOfNought`) into a call of one argument, which is what a size takes; the
second bears a `ValueName.Helper`, which no reader of library operations asks about.

## What it costs

A reader that wants to say "a call standing for this operation has the number of arguments
the library declares" cannot say it about `PreservedCall`. It has to say it about the calls
it happens to see, and work out for itself that nothing synthesises one of those.

#1276 has one. `StringPredicates.statedByChecked` refuses a call whose arity is not the
table's, because a call bearing one of its operations with another number of arguments means
the table and the library have come to disagree — and read as no predicate at all, that
disagreement would go on quietly costing every rule written with the operation. What holds
that up is not a mechanism. It is that the range of one table does not meet the domain of
another, checked by hand, in a comment.

That is the shape to close. A synthesis site added tomorrow that substitutes an operation of
its own is not required to consult a signature, and the reader that finds out is whichever
one had written a claim about arity.

## What it wants

Not a test that the two tables stay disjoint — that pins today's coincidence and says
nothing about the next synthesis site. The question is where a synthesised call's arguments
are held to what the operation declares, and the answer wants to be somewhere a site cannot
skip: the constructor, a factory the synthesis sites go through, or a `PreservedCall` that
carries which of the two it is.

Worth measuring before choosing. How many synthesis sites there are, what each of them
knows about the operation it names, and whether a signature is reachable where they stand
are all questions this has not asked yet.

## Left out of #1276 deliberately

The throw there stays. While this is open it is what turns a broken disjointness into a stop
rather than a rule that quietly says nothing, which is the failure mode #1267 was about.

kawasima:
Closes #1279.

`Core.PreservedCall` held the operation and the arguments as separate components, so "these are
the arguments the declaration takes" was a property of whichever site built one rather than of the
node. Its readers stood on it anyway. `Combinators.handedTo` indexes the closure and the container
straight into the arguments; `CallArguments.of` indexes whatever position a rule about the
operation names; `StringPredicates.statedByChecked` indexes the subject and the written text. What
licensed all three was prose saying a kept call had been typed against the signature it names —
and two of the sites that build one substitute an operation of their own, consulting no signature.
What kept those from reaching those readers was that two hand-written tables did not overlap.

## The operation is a name read against its declaration

`DeclaredOperation` is a `ValueName` where a declaration has been read for it, the way
`TypeSymbol` is a `TypeKey` an authority resolved. It is minted on a `CompleteSignature` coming
to be and nowhere else, and `CompleteSignature` stopped being a record so that its own two ways in
are named and counted rather than published as a constructor taking whatever its components are.

`PreservedCall` carries one and refuses a call standing with another number of arguments, so the
passes that rebuild a call to replace an argument are held to it as well — they carry the
operation across rather than looking it up again.

A value kept standing is the same statement rather than a second node. Being written with no
parameters is what makes it a value, so a reference to one is an application of what its
declaration takes, which is nothing. `HelperTyping` is where both halves are in hand — the empty
parameter list it just read, and the type checking the body just answered — so the signature is
made there instead of downstream from a type.

## An emptiness check and the size it means are held to each other

`MeansTheSameAsASizeOfNought` names another operation, and no arm of the binder held it: the arm
it passed through said these facts name no argument, which is true of the others there and not of
this one. What a reader does with it is write a call of the size where a call of the check stands,
keeping the arguments, so what has to hold is that the two take the same argument and that a size
is what one compares against nought. Asked where both declarations are in hand, and answered with
a `BoundOperationFact` the reader applies.

The node's own invariant does not cover this. A check on strings rewritten to a length of lists
satisfies it — both take one argument — and the binding is where it does not. There is a test
saying exactly that, with the arity agreement asserted beside it so the two layers can be told
apart.

`Conditions.asSizeComparison` no longer picks a name or counts arguments, and
`StringPredicates.statedByChecked` no longer carries the throw #1276 left in it.

## What holds the tables up

The positions a table names have to be positions the declaration has, and that is a fact about the
table and the library rather than about any call: where they disagree, every rule written with the
operation is read at the wrong argument, whichever call arrives first. `StringPredicates` is held
to the library once. The argument positions in `OperationFacts` were already held by
`DischargeRules.holdToTheDeclaration`, and a mutation confirms it — nothing was added there.

## Left as it is

Every other kind of operation fact is still read from the authoring index below the binding, so
what holds those readers up is that the walk ran rather than a value carrying what it settled.
That is #1304, and the boundary this change is the first cut through.

## Notes on what the checks found

The census of what may read a declaration into a signature is written the way the one over
canonical comparisons is: the ways in are enumerated from the class files and each is declared
with what it is for, so a way in nobody wrote down is itself the finding rather than a mismatch
against a list. It corrected the declarations in this PR when it was first run.

A mutation that made a rebuild put back a different number of arguments did turn the build red,
but not through the refusal — `InvariantChecker` is fail-open, so the exception was swallowed and
the discharge result changed instead. The test that claims the rebuild sites are covered now goes
through `CallArguments.replacedIn` rather than through the constructor directly.


kawasima:
Both blockers are the same defect as the one this PR removes, one level up: two things were made one value, and the container went on stating one of them separately. So the fix is not a check on the pairing but the removal of the half that could be stated twice.

**Blocker 1.** The node was checked at construction and then went on holding the caller's list, so what it says about itself was true for as long as the pass that built it left that list alone — and every reader reads the call afterwards. It takes the arguments over now. A mutation removing the copy turns the new test red on the count.

**Blocker 2.** Not a validation in the constructor: that moves the convention into a runtime convention, which is what the branch is trying to stop doing. `Preserved` is built out of signatures and files each under the name its own signature carries, and the lookup a value's check fills in records rather than files, so there is no key left to disagree with. That also retired `SettledValues` — a public callback supplying signatures, which is the shape rejected earlier for the declaration authority and had no business surviving here.

The census caught the change on its own: one way in disappeared and a private accessor took its place, and it failed until that was written down.

**A regression of mine, found by the review agent.** Deleting the authoring index for what an emptiness check means also deleted what it was refusing — one operation declared to mean two sizes. It answered two questions and only one of them was a lookup. The refusal is back where the bound facts are collected, with a test that is red without it.

Also fixed: the shape comparison claimed a leaf holds no variable, which is false of the two that stand for one — a variable of an application reaching it is refused as what it is rather than reported as two operations disagreeing; the null-result message was written for a declaration and raised for a value as well, so each factory says its own; and the fully qualified names in the bodies I wrote.

The fail-open finding is #1306. It is not about `InvariantChecker`'s contract, which is right, but about a representation refusing to be built arriving as the same exception as a limit.

Issue #1336, opened 2026-09-04
A clause's topology is recognized once, while each reading decides how far to descend


kawasima:
Found while closing #1333, and deliberately left out of it.

`ClauseExpr` says of itself that it is the one mapping from a clause tree to a shape, and everything downstream is an evaluator over it. `Predicates.read` is not one of those evaluators: it recognises `&&` and a restated condition for itself, walking the same clause with its own recursion.

That is what made #1333 two symptoms rather than one. A binding under a clause was invisible to both walks, so the reading that asks what a position admits and the reading that asks what a construction owes each had to be taught about it separately. #1333 applied the binder decision to both — the binder now has one owner (`Terms.inside`, ADR-0106) — and left the connectives with two.

## Why it is not a move

`Predicates` does not split what the fold splits. Its recursion descends only where

    ConditionJoin.of(op).map(join -> join.under(positive)) == BOTH

so a denied disjunction is descended (the denial makes it a conjunction) and an **asserted disjunction is not**: it is handed whole to `Conditions.comparisonsStatedBy`. Read through `ClauseReading`, `either` would compose what each branch owes, and what a disjunctive clause owes would change. That is a different question from #1333 and has to be asked on its own.

The fold does not carry what `Predicates` carries either — `PartsToRead`, `PerPart`, `decidesFalse`, the `Discharge` budget, and the loop that takes the first reading in reading order that answers.

## What would close it

Probably not "make `Predicates` a `ClauseReading`". `ClauseReading.over` evaluates both children of a connective before composing them, and what is wanted here is a traversal in which the shape is recognised once while each reading decides whether to descend into a node or take it whole. An evaluator that recognises `Either` and declines to descend it is the shape `Predicates` needs.

The order that keeps it reviewable: pin today's obligation semantics in a characterization test first — asserted `AND` reads both, denied `OR` reads both, asserted `OR` distributes nothing to its branches, the `PerPart` attribution, `decidesFalse`, and the budget and first-answer ordering — and only then give the topology one owner.

## Boundary

Only evaluators that read the same clause extensionally are in scope. `Predicates.assumeCond`, `PathReachability` and the `partition` walks change knowledge along control flow and short-circuit order; they are not proposition folds and unifying them here would be a different claim. That boundary belongs in the ADR alongside the decision.

kawasima:
There is a third reader, and #1333 has now measured what it costs.

[...]

Issue #1335, opened 2026-09-04
An answer-level reading limit is reported as if no reading interpreted the rule


kawasima:
Found while closing #1333, and deliberately left out of it.

A reading may take every rule of a position in and still be unable to build the exact answer they come to between them. Two patterns each small on its own have a meet the size of their product, and past the allowance the position comes back holding every value.

That loss is about the answer and about none of the rules — the same rules met in another order would have been built — which is why `ReadingEvidence.stoppedBy` refuses to file such a reason under a rule. It is recorded at the name instead, as `UnreadReason.EXACT_VALUES_TOO_COSTLY`, whose `About` is `THE_ANSWER`.

## What it is called on the way out

The question the rule raised then stands with no reason filed under it, and what carries that state says the opposite of what happened:

- `RuleAccounting.Why.NothingTookItIn` — a reading took it in.
- `BlockReason.NoReadingTookItIn` — the same.
- `UndividedPosition.Reason.RULE_NOT_INTERPRETED_HERE` — the rule was interpreted; what could not be built is what the interpretations come to together.

An author reading the document is told that this compiler established no interpretation of their rule, and is left to look for a form to rewrite. Nothing about the form was the matter, and what would lift it is an allowance.

## Where it stands

Not a rename. `About.THE_ANSWER` already tells the two apart inside the compiler, and the question is which published word it should reach — whether an answer-level limit is a word of its own out there, or the same word `exact_values_too_costly` already carries at a position, and what a reader is owed either way.

#1333 left the names alone and wrote the state down beside them, so the account is at least true. It also holds the three causes apart in a test, so a change here has somewhere to be measured: `WhatAStandingQuestionIsAccountedForByTest`.

PR #1337, opened 2026-09-04
Read a clause under the binding a helper's expansion made

### LINKED ISSUE #1333: A clause that reaches its rule through a helper is read less than the same rule written out
Found while measuring #1312, and deliberately left out of it.

A newtype states its rule by calling a helper its module defines:

```souther
let atLeastOne (s: String) = String.length(s) >= 1

data Named = String
    invariant atLeastOne(value)

data NamedInline = String
```

The two say the same thing about the same values. Asked what the position admits, `NamedInline` comes back with the rule read to the end and a line drawn on it. `Named` comes back as a clause written in a form this compiler does not read — in the module that declares it, with the helper in scope and the expansion done.

So the reading gets to the clause and stops on its shape. What the expansion leaves where the call stood is the helper's body under a binding for its argument, and that binding is what the reading has no rule for. The rule itself is one it reads perfectly well written the other way.

## What it costs

Nothing is reported wrongly: a rule this could not read widens what a position admits and says that it did, which is what a limit is supposed to do. What it costs is that the limit is reached by writing the rule in the way the language offers for writing it once and using it twice. An author who names a rule loses the reading that an author who repeats it keeps, and nothing about the two spellings says so.

It shows up in the construction verdict too. A construction the guards prove must fail is refused where the clause is written out (`E2010`) and only warned about where the same clause reaches through a helper (`E2011`), in one module, on one model.

## Where it stands

Not what #1312 was about. That one was about which representation a reader gets, and it closed: a clause reaches every reader as what its own module expanded, so this limit is now reached from every module rather than from the declaring one alone. The tests there hold the helper case to agreeing with itself across modules and say in as many words that they do not hold it to being read — because that is this.

What would close it is a rule for the binding an expansion introduces, wherever the reading meets one. `spec §invariant-discharge-representation` already says a `let` is read as its body and that a helper is expanded by binding each argument and reading the parameter through that binding, so the language says the two spellings are one thing; it is the reading that does not.

kawasima:
Closes #1333.

A helper call is expanded by binding each argument and reading the parameter through that binding (spec §invariant-discharge-representation), so a clause stating its rule through a helper is that rule under a binding. Measured at the start: with the rule written out, a construction the guards prove must fail is refused (`E2010`); with the same rule reached through a helper, the compile said nothing at all.

## What a clause reading is

What a clause states is answered upward and what its names mean is handed downward. `ClauseExpr.Scoped` is where the second changes and nothing else — it composes no statements — and `ClauseScope` answers what a binding is entered as, so no clause evaluator is handed a binding **as the clause's shape**. For a reading carrying `Denotations` the answer is `Terms.inside`, which is the one place a `let` is entered (ADR-0106). The readings hold no environment of their own any more: a leaf is read at where it stands.

**A binding nested inside a leaf stays part of that leaf**, and the decision reaches it there rather than by taking the shape further in. Binding transparency is not a property of the clause's topology — it is a property of every semantic question a reading asks about the inside of a leaf, and a question that answers from the tree alone is one the decision has not been applied to yet. So each of them crosses the binding through the environment: which position an operand is, what written value it is, which positions a part names, and which authored rule placed which end.

`Predicates` reads the body of a binding under the same environment, asking that same one place for it. Its own recognition of the connectives stays where it is — an asserted disjunction is not split there, so moving it onto this fold would change what a disjunctive clause owes. That is #1336.

## What the accounting had resting on the defect

The account of a question a rule leaves standing has an arm naming no reading, documented as what a rule about a position none of the readings knows comes to — a helper reading a field of the value it was handed being the example. That example is gone.

The arm stays, because it was never its only cause. A reading may take every rule in and still be unable to build the exact answer they come to within its allowance; that loss is about the answer rather than about any rule that paid into it, and `ReadingEvidence.stoppedBy` refuses to file such a reason under a rule. So a standing question is accounted for by a reading's own words about the rule, or by a reason at the name that no rule could be answerable for — and neither is the accounting disagreeing with itself, which is refused where the account is made.

The name says no reading took it in and the reading did take it in. Left as it is with the state written down — what an answer-level limit should be called in the published vocabulary is #1335.

## Held

The corpus is the spellings of one rule — written out, through a helper, through a helper of a helper, denied two ways, joined inside the helper, standing beside a written-out rule, handed to a helper as a value, given more than it reads, standing in an operand, and reaching a bare parameter the clause's shape already crossed. Every one refuses the construction the guards refute and admits the one they allow, and what the position admits is the same value, the same set and the same format whichever spelling it arrived by. The rule the issue is written about is among them.

The lines are held on their own, because a construction reads the rule through one more reader and two spellings can reach one verdict with different lines under them.

The fold is held apart from any of that: a binding is a shape and not a leaf, no evaluator is handed one, each binding on the way down is entered in turn, a connective under a binding is read inside it, a binding under a connective reaches its own side alone, a denial is carried through, a closure keeps the place its parameter is bound at, and a leaf is handed on as the very node it was — which is the control saying the fold changes the environment and nothing else.

The three causes of a standing question are told apart over their own fixtures, so closing the helper one is not read as the answer-level one having gone away.

## What closes the issue, and what it hands on

Every difference a reading's inability to cross a binding made is gone: what a position admits, where its values stop, which values and which formats it holds, what a construction owes, what a standing question is accounted for by, and which authored rule placed which end — each answers alike whichever spelling the rule arrived by.

One shape is left, and its cause is not a binder. A helper whose body joins two rules is one authored conjunct, and the line reader crosses the binding, arrives at the conjunction and reads neither end: reading its halves as two would number them as parts the reading beside it does not have. Telling a clause's semantic decomposition from its authored-part decomposition is what a third reader of the topology costs, which is #1336 — where `InvariantChecker.direct` is now named alongside `ClauseExpr` and `Predicates.read`, with this case and the characterization it needs.

### DIFF
--- a/docs/adr/0106-a-binders-meaning-belongs-to-the-environment.md
+++ b/docs/adr/0106-a-binders-meaning-belongs-to-the-environment.md
@@ -112,10 +112,23 @@ every rule an author stated by naming it rather than writing it out.
 
 The fold now hands an environment downward beside the answer it carries upward. `ClauseExpr.Scoped`
 marks where that environment changes and composes nothing; `ClauseScope` answers what a binding is
-entered as, and for a reading carrying `Denotations` the answer is `Terms.inside`. No evaluator sees
-that a binding was there, and none of them holds an environment of its own any more: a leaf is read
-at where it stands. `Predicates` reads the body of a binding under the same environment, asking the
-same one place for it, though it still recognises the connectives for itself.
+entered as, and for a reading carrying `Denotations` the answer is `Terms.inside`. No clause
+evaluator is handed a binding **as the clause's shape**, and none of them holds an environment of its
+own any more: a leaf is read at where it stands. `Predicates` reads the body of a binding under the
+same environment, asking the same one place for it, though it still recognises the connectives for
+itself.
+
+**A binding nested inside a leaf stays part of that leaf**, and the decision reaches it there rather
+than by taking the shape further in. Every question asked about the inside of a leaf crosses the
+binding through the environment instead of interpreting the binder: which position an operand is
+(`Terms.subjectOf`), what written value it is (`Terms.asWrittenValue`, and `Terms.folded` over it),
+and which positions a part names (the descents in `AdmissibleReading`, `StatedByClauses` and
+`Predicates`). What the shape settles is where the environment changes; what the leaf language reads
+is unchanged by any of it.
+
+So binding transparency is not a property of the clause's topology. It is a property of every
+semantic question a reading asks about what is inside a leaf, and a question that answers from the
+tree alone is one this decision has not been applied to yet.
 
 A helper-expanded clause is therefore read under the same denotation environment as the equivalent
 inline clause. In particular, passing through a helper is no longer a way for a rule to reach an
--- a/docs/adr/0106-a-binders-meaning-belongs-to-the-environment.md
+++ b/docs/adr/0106-a-binders-meaning-belongs-to-the-environment.md
@@ -103,6 +103,31 @@ readers and is a separate gap (#866); every binder-introducing form is a place w
 has to be applied rather than restated, and where it has not been applied yet the reader in question
 still owns an account of its own.
 
+### Applied to the clause readings
+
+The readings that ask what a declaration's positions admit were such a place. They folded a clause
+into one answer carried upward and had nowhere for a binding to be, so a clause under one arrived as
+a shape none of them had a word for — and since a helper call is expanded as a binding, that is
+every rule an author stated by naming it rather than writing it out.
+
+The fold now hands an environment downward beside the answer it carries upward. `ClauseExpr.Scoped`
+marks where that environment changes and composes nothing; `ClauseScope` answers what a binding is
+entered as, and for a reading carrying `Denotations` the answer is `Terms.inside`. No evaluator sees
+that a binding was there, and none of them holds an environment of its own any more: a leaf is read
+at where it stands. `Predicates` reads the body of a binding under the same environment, asking the
+same one place for it, though it still recognises the connectives for itself.
+
+A helper-expanded clause is therefore read under the same denotation environment as the equivalent
+inline clause. In particular, passing through a helper is no longer a way for a rule to reach an
+admitted-values question without any reading recognising its positions.
+
+This does not make every standing question attributable to a rule-level reading failure. A reading
+may consume the rule completely and still be unable to construct the exact answer its rules come to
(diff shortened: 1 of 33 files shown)


kawasima:
Rebased onto `develop`, so the diff is this issue's commits and nothing else. Review findings addressed:

**The `unnamed` guard bypass** — real, measured: written out gave no warning, through a helper gave one, which is the disagreement this issue set out to remove reappearing the other way round. The cause is one level below the descent: which values a site hands over was matched against the clause by walking the tree, and a binding standing between the clause and the value is invisible to that. It is asked by subject now, which works because a binding is named by what its body names — the same fix, one commit earlier.

**`quantifiedBy` stopping at a binding** — closed here rather than deferred, since the asymmetry is one this change introduced. Both readings of a clause reach the same depth under one selection of its parts.

**The `Resolve` javadoc** — from the lint backlog rather than this issue, and it left the diff with the rebase.

Also in since the review: the refusal asks whether the answer accounts for a standing question rather than whether no rule does, with the reasons that account for nothing enumerated beside the one that does; and a helper standing in a comparison's operand is held beside a helper standing where the clause does, which is what turned up the naming defect.

kawasima:
Rebased onto current `develop`, so the Error Prone work is under this. No conflicts; the gate found three `ExposedPrivateType` warnings in a parameterized test of mine, now fixed. `CI=1 mvn clean verify` passes across every module locally.

The two P1s from the last round, each with the mutation that shows the test holds it:

**A written value through a binding.** Measured first: `value >= 1` and `value >= itself(1)` already agreed on where the end falls, but `value == itself(1)` and `String.matches(itsSelf("[a-z]+"), value)` did not — the finite-value reading and the string predicate both came back with the position wide open and the rule unread. One cause: `Terms.asWrittenValue` asks what written value an expression is and reads the tree alone, so a binding is not one, and all three readings ask it. The rule is given there rather than by teaching `ClauseExpr` to descend into an operand: a binding is the value its body is, with the binder standing for what it was given, which the node already holds. Dropping the rule again turns the two new equivalences red.

**A closure's parameter.** Naming a binding by its body under the entered environment lost the other environment: what a name denotes is the environment's, and what it is *called* is the naming walk's own, where a closure's parameter is called where it is bound so that two lambdas written alike are one term. Both are entered now. With only the first, `x -> let y = x in y` names `\1(<a value at 1:1>)` where `x -> x` names `\1(#0.0)` — which is the test, and the mutation reproduces it exactly.

**P2.** Whether a reason may stand in for a rule's own account is a named question now, with its three answers given by what the reason is about. Putting `NEITHER` back on the standing-in side turns the test red, which the previous shape did not.

PR #1340, opened 2026-09-04
Record what a standing question stands on, not that nothing read the rule

### LINKED ISSUE #1335: An answer-level reading limit is reported as if no reading interpreted the rule

kawasima:
Closes #1335.

A question a rule raises can stand because a reading gave up on that rule and because the answer its position waited for was never built. The two are recorded in different places — the first under the rule, the second at the position, since no rule is answerable for it — and the account took the first alone. Where a rule had none of its own, the answer-level limit went out as a rule nothing interpreted: about a rule, met again however much a run allows, and asking an author to rewrite a form nothing complained of.

## The two halves

What a rule is answerable for stands at a place an author wrote. What its position's answer was short of stands at no place at all: the rules of a position come to their answer together, and the same rules met in another order would have been built. So the two are two carriers, and a question that stands on both holds both. Held as one, the half with no written place would be offered to a reader as something to go and look at.

`FieldDomains` asks the position beside the rule rather than instead of it, and refuses only where neither has a word. Two accounts of one question agree about what the rule left; what the position's answer was short of is taken from whichever met it, because a reading whose neighbours left the answer buildable met none and is not disagreeing with one that did.

## The order

A reason of a rule stands somewhere, and reasons written in one text stand in the order that text puts them in. Reasons written across texts stand in no order anybody wrote — nothing an author did says which file comes first — so the carrier says which of the two it holds, and only the first is an order a reader may act on.

The claim is made where the places are still in hand, which is the last call that holds them. Past it the reasons are words, and a word says nothing about where it was written; that is how an order arrived at by walking came to be published as the author's, and was right only while one walk produced every member. Which order it is, is decided before anything is compared, since what may be compared is what the answer turns out to be. Sorted before folded, so an entry is the earliest place a reason stands on rather than the first the walk met.

A choice carries the operator an author wrote it with, beside the identity that tells it from every other. A leaf always knew where it was; a choice knew only that it was itself.

## What a document says

`stopped` and `answerStopped` are two fields, because the two are two facts and no order runs between them. Where every rule-side reason was written in one source text, the array is in the order those places were written; where they were written across texts, it is steady from run to run and means nothing else. Schema 13 drops `rule_not_interpreted_here`; documents of schema 12 may carry it and that version is unchanged.


kawasima:
Waiting on #1341. Marked draft rather than merged as it stands.

The rule-side half of what a question stands on is published here in the order the author wrote the parts that raised it, and the carrier that claims that order is made downstream of anything that saw a clause. Moving the claim to where it is true turned out not to be possible yet.

`AdmissibleValues.metAll` is where it becomes true: the work is scheduled by what it costs and the reading's own order is put back afterwards. What travels from there is the author's order. But what reaches a question is assembled in `StatedByClauses.aRuleIsAnswerableFor`, which appends the reasons a shared allowance recorded at the position to whichever parts happen to name it — and those carry no part to be placed among the author's, so the list is the author's order with an unplaced tail.

Typing that tail into the same carrier would publish an order that comes from how this compiler records reasons rather than from anything the model says, which is the defect this branch set out to remove one layer up.

So the dependency runs the other way from how the two issues were split: #1341 is about keeping a reason's provenance, and the rule-side of a question cannot be ordered honestly until it is kept. Everything else here stands — the capability split, the merge of what a rule left with what its position was short of, the deletions, the two published fields, the run-sensitivity over both.

kawasima:
Now waiting on #1352 rather than #1341, which is merged.

Working out where the rule-side reasons of a question get their order turned up what has to be true first. The order is the one the parts that raised them were written in, so every such reason has to stand at an authored place — and one of them cannot. `ALTERNATIVE_NOT_READ` is written into a reading's standing by a join that holds two readings and a flag, knowing nothing about which choice, and again by the part reading, which does know. Both land in one map, and an account of a rule is sifted out of it.

That map is the position's own account and is right to hold what it holds. What is not right is reading it back the other way, which is the shape #1341 closed at a branch's fate met again where the map is made. So the split goes there first and this waits for it.

What stays here: the sites a rule-side reason can stand at, the source order over them, and the one place that claims it.

PR #1414, opened 2026-09-06
A name reaching a declaration is some reference of it, and an application says why it is here

### LINKED ISSUE #1377: A reference of the source is not optional, so the type should say so
`Ast.Var` carries a `SourceReferenceOrigin` and refuses neither a null one nor a reader that assumes there is one. `AstBuilder` mints one at every name it reads, written or desugared, so nothing in the tree is without one — and the fact lives in the builder rather than in the value.

What that leaves is a check standing in for an invariant. `HelperInliner.etaOf` throws where `function.origin()` is null, saying "a name no source wrote, reaching a declaration, expanded as a value". Nothing can reach it, and a reader cannot tell from the type whether that is because it is impossible or because it has not happened yet.

The fix is to hold the value to it: refuse a null origin where a `Var` is made, and drop the throws that stand in for it. A reader is then told by the type what it may assume, and a construction that forgot one is refused where it was made rather than wherever it is first read.

kawasima:
Closes #1377.

`HelperInliner.etaOf` threw where a name expanded as a value carried no reference, saying "a name no source wrote, reaching a declaration, expanded as a value". A reader could not tell from the type whether that was impossible or merely had not turned up.

The issue read that as an `Ast.Var` problem: hold the surface tree's references to being non-null and the throw becomes droppable. That is not the mechanism. `etaOf` reads a `Hir.Var`, and null there is a documented, deliberate state — a pass writing a name of its own wrote a reference, and the passes below mint those. Holding the surface tree to something says nothing about the tree beneath it.

What the throw stood in for is a reader working out from an application's shape what kind of thing it was. An application no source wrote looks the same whichever pass wrote it and for whatever reason, so `siteOf` guessed: written where a construct came back, and otherwise the block a name used as a value is expanded into. The guess was right for the common case and was the only answer left for every other — a library operation reached for and a value composed for a fixture were both read as expansions of a name — and where it ran out, the refusal appeared.

So the producers say it where they write one.

A name reaching a declaration is some reference of it, whoever wrote it: an author's is one the source counted, one a pass wrote for what a construct means is named by that construct, and one the generator composes begins where it is composed. A name reaching anything else carries none, because what such a name is, is what it reaches.

An application says why it is here: the construct the author wrote, the name a block was expanded from, the construct a pass derived it for, or that it was composed for a fixture. Every application answers that; only one that is actually expanded is asked which expansion it is, so a composed value is not made to carry an identity nothing wants.

A collection written in brackets gained a construct of its own, because the operations a body holds for it are references derived from it and had only the place to be told apart by — and one helper expanded at two calls puts two of them at one place.

`etaOf`'s refusal and `siteOf`'s shape reading are both gone. Adding a fifth kind of application stops the compiler rather than being read as the common case.

Each invariant stands on its own and none is argued from the others; they are separate commits for that reason.


kawasima:
The escape hatch this touched repeatedly is filed as #1419: `Core.withoutItsPlace()` answers with a comparison value that is still a `Core`, so a reader that requires provenance can be handed one it was erased from. Closing it means giving the comparison value a representation of its own, which redesigns the boundary #831 settled — not a further tightening of what is null here.

PR #1460, opened 2026-09-08
Measure what a caller assumes, and what an edit to it costs the callers

### LINKED ISSUE #1459: Nothing that is measured reads what a caller may assume
`Bodies.Assumptions` answers what a caller of a behavior may take as holding of its answer, and the discharge analysis reads it at every call that has matched a case (`PathEngine.assuming`): it substitutes the call's arguments into each rule and reads what they say. None of that runs while this repository is measured.

Where the conformance corpus states a relation, nothing calls the behavior that states it, so the reading is built and never read. No model in the bench corpus states one at all. What exercises the path is hand-written fixtures — a term built in a test and handed to a `PathEngine` — and a fixture says what its author already believed.

## What follows from it

An answer's identity is what decides whether everything downstream is asked again. `Bodies.Assumptions` compares a reading whenever the module that declares the behavior is edited, and what that comparison costs is a question no measurement here can answer.

A change to the reading, to the substitution, or to what the predicates make of one cannot be seen to regress. The nightly population run does not reach it either, so there is nowhere a regression would show.

And work nothing asks for goes unnoticed. A reading built while checking a call was walking its term to hold a hash that nothing on that path ever reads; it was found by reading the code, because measuring could not have found it.

## What would close it

A model in the corpus the benchmarks compile that states a relation about its answer, and a body that calls it, so that compiling the corpus reaches the substitution and the assumption reading rather than stopping at the declaration. The editing benchmark wants the same model for its own question: an edit to a rule is an edit to what every caller of that behavior assumes, and that is the shape of recomputation this analysis is written against.

## Done when

Compiling the benchmark corpus reaches the reading a caller depends on, and an edit to a declared relation is measured at the bodies that call it.

## What it touches

The bench corpus, and whichever conformance corpus is the natural home for a caller of a behavior that states something. Noticed while reviewing #1453, which rewrote this path and could not measure it.

kawasima:
Closes #1459.

Nothing measured here read what a caller may assume. `Bodies.Assumptions` answers what a behavior states about its answer, and the discharge analysis reads it at every call that has matched a case; the bench corpus stated no relation at all, and the one the conformance corpus states is on a behavior nothing calls. So the reading was built and never read, the substitution of a call's arguments into a rule ran nowhere a number is taken, and a change to any of it could not be seen to regress.

## A relation the model already wanted

`findAccountByDomain` answers an `Account` or says it found none. The type cannot say the account it answers is the one on the domain it was asked about — one `Account` is the same shape as another — and answering some other account is the failure the behavior exists to avoid. The clause says it, through `accountHasDomain`, which is the predicate the duplicate screen already matches the account book with.

`linkLeadToAccount` calls it and matches the cases, so compiling the corpus now reaches the reading, hands the call's arguments to it, and lets the predicates make something of the result. The `fake` and `example` rows written for that lookup are held to the clause as they are anywhere else.

## An edit that leaves the declaration where it is

The edit measurements added a definition to a module or a comment to the end of one. Adding a definition changes what the module declares, so what is re-asked is everything that reads the declaration list. No shape left a declaration alone and changed only what a caller of it may assume, which is the recomputation this analysis is written against.

A fifth shape rewrites a rule of the clause. It writes the conjunct the rule already states a second time and takes it away again, alternating, so that no round writes what the round before it wrote — two rounds writing one text is no edit, and the second would find every answer holding and time the floor.

Restating the rule with a rule it did not state is what the first writing of this did: it added a bound on the parameter, the way an author tightening a rule would. A comparison in an `ensures` draws a line on what it compares (spec §a-clause-draws-a-line-on-what-it-compares-an-input-against), so that edit moved the adequacy reading with it and the round was timing an edit to two readings under a name that says one.

## What holds the shape up

A timing says nothing about whether it reached anything, so what the measurement needs is asked of the corpus rather than read off the number: that it writes the rule the edit rewrites, that a body of it reads what the behavior states, that restating the rule changes what that body may assume, and that it leaves the other two readings the corpus is held to — what the compiler says about it, and how well the module's rows cover it — where they were. Each was mutated and seen to go red, the last of them by putting the bound edit back.

Renaming the clause instead of the rule changes the file and leaves what a caller assumes exactly as it was: the clause name does not cross into that reading. That is why the edit had to be to the rule, and it is the mutation the third of those refuses.

## Notes for review

- The test carries the population tag, so it runs on the merge into develop rather than in this PR's checks. Its subjects are the models this repository holds and it compiles the corpus to ask, which is what the tag is for. It has been run here.
- Left out: giving the conformance corpus a caller for the behavior that states something there. The measurement is what this issue closes, and moving that corpus moves the documents written beside it.
- The relation edit and the definition edit come out close together, and so does an edit that changes no reading at all. Whether that is the store re-establishing more than the edit reached, or the spans below an edited line moving, is not separated here — the definition edit appends and moves no span, the clause rename does — so it is left as something now watchable rather than as a cause named.

### DIFF
--- a/souther-bench/src/main/java/souther/bench/Incremental.java
+++ b/souther-bench/src/main/java/souther/bench/Incremental.java
@@ -29,10 +29,11 @@ import java.util.Set;
  * <p>Restating a relation is a fifth, and it is the one the other four cannot stand in for. Adding a
  * definition gives the module something it did not declare, so what is re-asked is everything that
  * reads the declaration list. Rewriting a rule of an {@code ensures} leaves the declaration where it
- * was and changes one thing: what a caller of that behavior may assume. The bodies that call it have
- * to be checked again and nothing else has learnt anything, so this is where an answer held per
- * behavior rather than per module would show, and where losing one would show as well. The two edits
- * are the same keystroke to an author, and what separates them is only visible if both are timed.
+ * was and changes what a caller of that behavior may assume, while what the compiler says about the
+ * corpus and how well its rows cover it come back as they were. So the bodies that call it are what
+ * has anything to learn, and this is where an answer held per behavior rather than per module would
+ * show, and where losing one would show as well. The two edits are the same keystroke to an author,
+ * and what separates them is only visible if both are timed.
  */
 final class Incremental {
 
@@ -51,6 +52,9 @@ final class Incremental {
      */
     static final String STATED_RULE =
             "    ensures onTheDomainAsked = Account -> accountHasDomain(domain, value)";
+
+    /** What the rule already says, said again — see {@link #restated}. */
+    private static final String RESTATED_CONJUNCT = " && accountHasDomain(domain, value)";
 
     private Incremental() {}
 
@@ -102,17 +106,22 @@ final class Incremental {
     }
 
     /**
-     * {@code source} with the rule restated, differently in each round.
+     * {@code source} with the rule restated, alternating between rounds so that no round writes what
+     * the round before it wrote — two rounds writing one text is no edit, and the second would find
+     * every answer holding and time the floor.
      *
-     * <p>A conjunct is added rather than the rule replaced, so what the behavior states is what it
-     * stated and one thing more: the rows the corpus writes go on satisfying the clause, and a
-     * corpus that stopped compiling would be timed as it stopped early. The bound moves with the
-     * round because two rounds writing one text is no edit at all — the second would find every
-     * answer holding and time the floor.
+     * <p>The conjunct the rule already states is written a second time and taken away again. It says
+     * what the rule said, which is the point: what a caller may assume is a different value and every
+     * other reading of the corpus is the one it was. An edit that added a relation instead — a
+     * comparison of the parameter against a bound, which is what an author tightening a rule writes —
+     * draws a line on what it compares (spec §a-clause-draws-a-line-on-what-it-compares-an-input-against)
+     * and moves the adequacy reading with it, so the round would be timing two things and the number
+     * would be attributable to neither. Both halves of that are held by
+     * {@code TheRelationAnEditRestatesIsOneACallerReadsTest}.
      */
     static String restated(String source, int round) {
-        return source.replace(STATED_RULE,
-                STATED_RULE + " && String.length(domain.value) < " + (100 + round));
+        return round % 2 != 0 ? source
+                : source.replace(STATED_RULE, STATED_RULE + RESTATED_CONJUNCT);
     }
 
     private static String added(String source, int round) {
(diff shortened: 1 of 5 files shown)


kawasima:
## P1 — fixed

The cause is that the check read what the analysis is *given* rather than what it *did*. `ContractsForBody` is the input to `PathEngine`; the walk that follows the call to what produced the answer, matches the arm against the rule's guard, pairs the declaration's parameters with the call's arguments and substitutes is on the other side of it. Asserting on the input is asserting on what stands beside the fact.

It was in reach during the work and thrown away: a temporary print at the substitution point is how the corpus was confirmed to reach it in the first place. What went in as the durable check was the cheaper thing one step earlier.

So a contract now records that a rule of it was substituted into a caller's arguments, at the point the substitution is made — `AssumedContract.TAKEN_IN`, beside `PathEngine.SEEDED` and for the reason that one gives: taking a rule in is silent, since it narrows what the caller may hold of the answer and a body correct either way is checked either way. Public because the corpora it is asked over are in another module; recorded by the contract itself so that one place decides what the record is, and `PathEngine` only says when.

What is recorded is the behavior whose contract it is, not where it was taken in. A `Core.Call` carries no occurrence, so naming the body would mean threading it through this analysis for a reader that has the corpus in front of it and can ask which bodies call what — which the assertion beside it does.

The mutation is the one named in the review. With the hand-over cut in `answeredBy`, the three checks that were here stay green and only the new one fails:

```
compilingTheCorpusTakesInWhatItStates: compiling the corpus substituted no rule of
example.crm.findAccountByDomain into a call, so the reading a caller depends on is
built and not reached. Taken in: []
```

`aCallerOfItReadsWhatItStates` is kept as the dependency-graph assertion and renamed to say only that — `whatThatCallerIsCheckedAgainstIncludesWhatItStates`.

## P3 — already changed, and the suggested form was measured

This was found in a self-check before the review arrived and is fixed in `8e9fcc2` — the review is against `cfd39d1`.

The cause is the same as P1's in a different place: the edit was chosen for being easy to vary rather than derived from what it had to isolate. A comparison in an `ensures` draws a line on what it compares, so the bound moved the adequacy reading as well as what a caller may assume, and the round was timing an edit to two readings under a name that says one. Diagnostics were identical either way, which is why reading those alone would have missed it.

The lower-bound form was tried here, and it does not close it:

```
&& String.length(domain.value) >= 0 - round
  → restating the rule moved how well the module's rows are said to cover it
```

It is true of every `DomainName`, so it fixes the false postcondition — but it is still a comparison, so it still draws a line and still moves the adequacy reading.

What is there now writes the conjunct the rule already states a second time and takes it away again, alternating. It states nothing the behavior did not state, so there is no false postcondition to fix, and it moves no other reading. Both halves are asked of the corpus rather than claimed in a comment: what a caller may assume comes back different, and the diagnostics and the module's adequacy come back as they were. Putting the bound back is what the second of those refuses.

Issue #1472, opened 2026-09-08
Source positions make declaration cuts propagate location-only edits across modules


kawasima:
A declaration is already a cut at the module boundary: a reader in another module asks for the declaration it names rather than for the module-wide answer it came from.

That cut currently carries where the declaration was written as part of its value identity. Moving a declaration without changing what it says therefore looks like a changed declaration to every module that reads it.

## What happens

Every HIR node carries a `SourcePos`. A resolved declaration is answered as a `Hir.Def`, and the normalized declaration answers below it ultimately carry the same authored nodes.

So an edit that only moves declarations — for example, writing or removing a comment line above them — changes those declaration answers by equality even though none of the declarations says anything different.

The change then crosses the module boundary.

In the benchmark corpus, a comment line written in `example.crm` caused answers in `example.pipeline` to be worked out again through chains such as:

```text
Adequacy.Inputs[example.pipeline]
  <- CheckedBehavior[example.pipeline, openFromLead]
  <- NormalizedDef[example.crm.LeadConverted]
```

and:

```text
TypesWithNoValue[example.pipeline]
  <- ResolvedDeclaration[example.crm.BlockReason]
```

The importer is not responding to a changed declaration here. It is responding to the declaration having moved.

A same-width edit that changes declaration meaning does not have this shape: the declarations whose meaning changed propagate, and the rest do not.

## What was isolated

The store was probed by changing only how selected declaration keys decide whether their newly computed answer changed.

The new answer was still kept in full, including its current source positions. Only the comparison used to advance `changedAt` ignored source location.

Applied to the declaration cuts involved in the cross-module chains, a location-only edit stopped at those cuts. The expensive `Adequacy.Inputs` path into the importing module disappeared, while edits that actually changed declaration meaning behaved as before.

The comparison had to descend through the normalized declaration wrappers as well as records. `Normalized.Data`, `Normalized.Sum`, and `Normalized.Unit` delegate equality to the HIR node they hold; treating a non-record wrapper as an opaque value leaves the position in the comparison and makes the experiment a false negative.

## Why this is a boundary problem

The declaration still needs its location.

After the same location-insensitive change test, asking the declaration itself returns its new position after the edit. The store keeps the newly computed answer and only leaves `changedAt` where it was when the declaration's meaning is unchanged.

The problem is therefore not that HIR has source positions.

It is that the answer used as a semantic cut across a module boundary uses source location as part of the identity that decides whether semantic consumers have changed.

[...]

PR #1475, opened 2026-09-08
Separate invariant judgments from where their clauses are written

kawasima:
A judgment held the clause together with the place it is written at, so an edit that moved a clause and changed nothing it states produced a judgment that was not the judgment before it — same clause, same verdict, same status, a different value.

**Before:** a judgment carried `Clause`, and `Clause` carried a `DiagnosticPlace`.

**After:** a judgment carries `Clause.Ref` — which clause it is, and what a sentence calls it. A diagnostic that needs to point at the clause asks `Shapes.ClauseLocationsFor`, answered from the declaration's own text.

`Clause.merge` moves to `Clause.Ref.merge` and loses its comparison of places. That is not a check relaxed: two readings of one clause can no longer hold a place, so there is nothing left to disagree about. Where the clause is written is asked once, of the declaration, wherever the reading came from.

## Why merge this on its own

It does not reduce incremental work, and it does not fix a report this compiler gets wrong today. It establishes the location dependency that a semantic cut at the module boundary needs.

A boundary that let a reading stand because the imported clauses still mean what they meant would leave the judgment cached — and a judgment carrying a place is a cached report pointing at where the clause used to be. Asked of the declaration instead, the place is worked out again whether or not the judgment was.

Worth stating plainly, because an earlier note on #1472 had it the other way round: the stale report is not an existing defect on `develop`. It is a latent dependency the intended optimisation would expose. The two halves have to land in this order.

## What holds it

`WhatAClauseMeansAndWhereItIsWrittenAreTwoAnswersTest` fixes both halves against one edit — a line inserted above an imported declaration:

- what the construction was judged against is what it was
- where a reader is sent to find the clause has moved

Run against `develop`, the first half fails, and the two judgments it prints differ in the place and in nothing else.

`AReportAboutAnImportedClauseNamesWhereItIsNowTest` is the integration guard for the cut that comes next: a store that absorbed the edit agrees with a compile that never saw the old text. It passes on `develop` too, which is right — it is what will catch the boundary cut going wrong, not something this change repairs.

Toward #1472.

### DIFF
--- a/souther-compiler/src/main/java/souther/compiler/check/Clause.java
+++ b/souther-compiler/src/main/java/souther/compiler/check/Clause.java
@@ -1,73 +1,36 @@
 package souther.compiler.check;
 
-import souther.compiler.diag.DiagnosticPlace;
 import souther.compiler.types.TypeSymbol;
 
 import java.util.Objects;
 import java.util.Optional;
 
 /**
- * One invariant clause, as what identifies it and what a diagnostic knows about it.
+ * What a reading of a declaration's invariants says about one clause of it.
  *
- * <p>Four questions, kept apart because none of them answers another. Which clause it is, is
- * {@link Id}. What a sentence can call it, is {@code name} — a clause MAY be written without one.
- * Where a reader can be sent, is {@code at} — a clause of a published module was written somewhere
- * this compile has no source for. What was proved of it is neither of these and is
+ * <p>Three questions, kept apart because none of them answers another. Which clause it is, is
+ * {@link Id}. What a sentence can call it, is a name — a clause MAY be written without one. Those
+ * two are {@link Ref}, which is what a reading carries. What was proved of it is neither and is
  * {@link InvariantChecker.Judgment}'s, which is why a clause is put on one of its sides rather than
- * carrying a verdict here.
+ * carrying a verdict.
  *
- * <p>{@code name} is optional and {@code at} is not. An absent name is a fact about the
- * declaration: the author wrote none, and every reading of that declaration finds none. Where the
- * clause is written is always answered — a clause of a published module is
- * {@link DiagnosticPlace.Unavailable}, which says where the code came from rather than being the
- * absence of an answer. It used to be absent, and absent is what every reader turned into silence.
- *
- * <p>{@code at} is the same for every reading of one clause, which is what {@link #merge} rests on.
- * Where a clause is written is settled where its text became positions and follows from the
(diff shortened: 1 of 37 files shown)


kawasima:
The grain is fixed. From the cause first.

**The cause is that the key's grain was picked from what suited the producer rather than from what a
consumer means by one.** One walk of a declaration has the place of every clause of it in hand, so
one answer per declaration was the easy thing to write. `Db`'s javadoc says an edge is what its
consumer means, and that was applied where the decision and the place were separated and then not
applied a step further down.

The same shape was swept for. `ClauseLocations` takes a `Clause.Id` at its mouth already, which is
the consumer's grain; the key that had been added was the one that did not.

```java
ClauseLocation(Clause.Id clause) implements Key<DiagnosticPlace>
```

What `stillHolds` is asked about is now the atom that was measured.

## The absences are closed here too

At the coarser grain, "no such declaration" came back absent and "an ordinal past the end" was
thrown by the capability — two places for one thing. Asked a clause at a time, the ways there is no
such clause — nobody declares the name, that kind writes no invariant, no clause is written that far
along — are one answer, because to a reader they are one fact. Which of them it was is a question
about the declaration, and this is a question about a clause.

`Shapes.clauseLocations` now only turns the absence into `NoSuchClauseIsWritten`.

## The test

A declaration writing more than one clause, with the later one moved.

```java
assertEquals(clauseLine(TWO_CLAUSES, 0), clauseLine(SECOND_MOVED, 0));
assertNotEquals(clauseLine(TWO_CLAUSES, 1), clauseLine(SECOND_MOVED, 1));  // the control
```

At the coarse grain this test cannot be written: the answer is a list, and the list moves.

Said plainly, **what is not shown end to end is that a reader of the first clause stands when the
later one moves.** The report of this fixture points at both clauses, so `CheckedBehavior` depends
on both locations and is worked out again, rightly, when either moves. What is shown is that the
edge is a clause at a time and that the answer for the clause that did not move is equal; that
`changedAt` does not advance follows from `Db`'s value-based early cut-off.

## The second and third findings

Neither is a defect of this change, and I agree with that reading. Lost, the upper half is written
and the chain goes on anyway, so they are
[on #1472](https://github.com/souther-lang/souther/issues/1472#issuecomment-5582625543):
`ResolvedDeclaration` keeps its positions, and another cut is wanted under `CheckedBehavior` between
its reports and its semantic projection. The second goes into what the upper half is done when.

Issue #1462, opened 2026-09-08
An author is sent to a choice for what it opened in the values and never for what it opened in the ends


kawasima:
`StatedByClauses.Part.either` files a shortfall against the choice — the entry that tells an author
to go and look at it, under `ALTERNATIVE_NOT_READ` — out of `byTheLeftGoingUnread` and
`byTheRightGoingUnread`. Both are read off the account of the values:

```java
one.byValues().hasUnreadPart() ? reachedBy(other.byValues()) : Set.of(),
other.byValues().hasUnreadPart() ? reachedBy(one.byValues()) : Set.of(),
```

So a choice whose unread alternative stands beside a branch that placed an end, and named no
position the reading of values reached, sends an author nowhere. The reading of order has an
account of the same clause and knows both halves of the question — which alternative it had no word
for, and which positions the branch beside it reached — and neither reaches this.

## Why it is not part of the correction that named it

The correction is about a fact derived twice by two rules, one of them false, and its whole
evaluation is that the set of findings does not move. Routing the ends' shortfalls here adds
findings, and mixed into the same change there would be no way to tell an entry that appeared
because the false rule went away from one that appeared because a new road was opened.

`leftOpenBy` is now spelled `leftOpenByValues` and says in its own javadoc that the other half is
not routed, so the asymmetry is written down rather than implied.

## What to decide

Whether an author is owed the same sentence about a choice that took back where a position stops.
An `ensures` clause whose branch bounds a position, beside an alternative nothing reads, is the
shape to write the answer against — the position comes back with no line and nothing says which
choice is why.

Beside that, whether the two roads produce one entry or two where a choice opened a position in
both languages. They are the same choice and the same clause, so a reader lifting one and finding
the other still there is the failure to avoid.

kawasima:
An attempt at the ends' half of this, withdrawn from #1473. What it found is worth keeping.

## Where the seam is not

The classification was written as a fold over the clause an author wrote: which names a choice leaves the line undecided at, composed connective by connective and handed to `Required` so that `BOUNDARY` came out undetermined rather than not raised.

That is a second algebra for choices. The readings already compose them — where the branches are settled, and where which branch nobody can be in is known. A fold over the shape has to work that out again and cannot: a branch nothing satisfies constrains nothing, and read off the shape it looks like a branch that constrains nothing.

```souther
data R = { s: String, x: Int }
    invariant said = s < "" || Int.abs(x) >= 2
```

[...]

PR #1477, opened 2026-09-08
Do not call a choice's boundary absent where its alternative was not classified

kawasima:
A soundness correction found while working on #1462, and independent of it. #1462 stays open.

## What was said, and what was true

The walk that classifies a rule goes into a conjunction and stops at a choice. So a comparison written under one was never read for the line it states, and a clause holding nothing that could be classified came out as a clause raising no question about where the values stop — which is what a rule read to the end and stating no line comes out as.

A behavior whose rules raise no boundary question has a border that is not applicable, and what a document says there is that the rules of this behavior draw no line. So a limit of this compiler went out as a sentence about somebody's model, and an author reading it was told to stop looking for a line their own clause may well draw.

The same comparison standing alone says what it is, and so does one standing under a conjunction. Under a choice it said nothing.

## What this does not do

It does not decide what a choice means. A choice whose alternatives are read to the end and hold a name nowhere between them draws no line, and a reading that composes the branches can say so; this one cannot, and a name under such a choice is one it declines to speak for. That is weaker than the answer there is, and it is not a claim about the model — which is the difference this is about.

Recovering the precision, and saying which choice is why, is #1462's, and it belongs where the branches were composed rather than where a rule's questions are raised.

## What moves

Two shapes, both of which said `not applicable` and now say `not measured` with the account the rule already had:

- a comparison this compiler cannot follow, offered as an alternative
- the same beside a branch nobody can be in, where the rule is its other half

The corpus does not move. Beside them, controls holding a choice both of whose alternatives were read, and a rule that draws a line, to what they said before.

### DIFF
--- a/souther-compiler/src/main/java/souther/compiler/check/InvariantChecker.java
+++ b/souther-compiler/src/main/java/souther/compiler/check/InvariantChecker.java
@@ -2039,6 +2039,13 @@ public final class InvariantChecker {
         // would have to change: written over the other, whichever ran second would be the only one
         // an author was sent to.
         runs.undecided(byName).forEach((name, why) -> stopped.merge(name, why, InvariantChecker::alsoSaying));
+        // And the same of the comparisons a choice offers, which nothing else here reaches. The
+        // walk above this goes into a conjunction and stops at a choice, so a comparison standing
+        // under one was never read for the line it states — and a clause holding nothing this could
+        // classify came out as one raising no question about where the values stop, which is what a
+        // rule read to the end and stating no line comes out as.
+        unclassifiedUnderAChoice(clause, at, byName)
+                .forEach((name, why) -> stopped.merge(name, why, InvariantChecker::alsoSaying));
         // What a rule about the strings at a position came to, where this conjunct is one. Such a
         // rule states where the values stop exactly when the strings it admits run between places
         // the order does not already hold them, which is the reading's own answer and not something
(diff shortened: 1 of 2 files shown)


kawasima:
Self-check found the same regression the review did, and my first fix for it does not work. **This should not merge as it stands.**

## The regression, reproduced

Measured on this branch against `develop`, in throwaway worktrees:

| clause | develop | this branch |
| --- | --- | --- |
| `s >= "a" \|\| s <= "z"` | not applicable | **not measured** |
| `s >= "a" && (s <= "z" \|\| s >= "m")` | borders 1 | **borders 1 (not all of it was measured)** |
| `s == "a" \|\| s == "b"` | not applicable | not applicable |

## The cause

The same shape as the change itself: one reader answering for a fact another owns.

`states` asks the affine arithmetic what a comparison's form left undecided. That reader is the authority for where a *number's* values stop. It is not the authority for a string's — it stops at every string literal, so `s >= "a"` arrives from it as a rule whose form was not read, and the walk files that as an undecided line. What such a rule states is read from the carrier, by the reading that places ends.

For a leaf this never showed, because the classification for a leaf does not rest on the arithmetic: `direct` places the end through `InvariantBound`, and the clause is promoted to one that states a bound before anything looks at what the arithmetic left. Under a choice there is no such promotion, so the arithmetic's stop is all there is — and it is the wrong reader's.

## What I tried and why it is not the fix

Skipping names the runs reading spoke for. Instrumented, `RunsRead.byPosition` is empty for these clauses: the runs are not who answers a string ordering either. The reader that does is the one holding the carrier, which is exactly what `states` may not reach — the guard on this classification names what that reader produces as something answering produces, and reaching it from here is what the withdrawn half of #1462 was caught doing.

So the honest position is that this change cannot be completed where it is written. Either it narrows to the comparisons the arithmetic is the authority for — and what says which those are is the arithmetic's own account of where it stopped, which is a rule about reasons and drifts — or it moves to where the end is in hand, which is the seam #1462 needs.

## Also worth narrowing, whether or not the above lands

`andAChoiceReadToTheEndDrawsNoLineAsBefore` states its reason as "both alternatives were read, so what they leave `n` is the model's answer". The mechanism does not check the second half: `n >= 2 || n >= 5` says the rules draw no line on both `develop` and this branch, and that choice does draw one. The control is right that the answer does not move; its stated reason claims more than what holds, and the shape it claims about is #1462's.

kawasima:
Closing. The central claim of this PR does not hold.

It was written as a soundness correction independent of #1462: a comparison under a choice keeps the question it raises on its own. Measuring it showed that `states` cannot decide that question. Whether a comparison under a choice is undecided is answered by the reader holding the carrier, and for a leaf the classification never rests on the arithmetic at all — the end is placed through `InvariantBound` and the clause is promoted before anything looks at what the arithmetic left. So this is not independent of #1462; it needs the same seam.

Not reworked, and not narrowed. Limiting the walk to the comparisons the arithmetic is the authority for means reading that reader's own account of where it stopped and deciding from the kind of reason it gave — which is the same move this whole line of work keeps being caught making, one size smaller. And the tests cannot be kept either: their central assertion is that a choice over an unclassified comparison comes back unmeasured, which the string counterexample denies.

What is worth keeping is above, in the self-check comment: the string regression measured against `develop`, why `RunsRead` does not answer it, and why `states` may not reach the reader that does. Beside it, `n >= 2 || n >= 5` reports that the rules draw no line while drawing one — a second place where that sentence is said without the mechanism to say it, and #1462's to answer.

The counterexamples are on #1462.

Issue #1504, opened 2026-09-09
A declaration's clauses cross the module boundary as the tree they were written in


kawasima:
`Shapes.ClausesExpandedFor` answers a declaration's clauses in the representation its module expanded them into, which is an authored tree. Moving the declaration without changing what it states makes a different answer, and every reader of what a declaration states reads it — so an edit that says nothing crosses into every module that imports the declaration.

This is what is left of #1472 once the declaration cut is in place. What a declaration says is answered apart from where it is written (`Shapes.MeaningOf`), and asking it before and after a location-only edit gives an equal answer; the readers that carry the move have not been moved onto it. `WhatCrossesAModuleBoundaryWhenADeclarationOnlyMovesTest` holds the present state and names this answer as what carries it.

## Why it is not one more reader to move

The clauses a body is checked against become `Core` in `Clauses.typed`, and that is the rung at which a clause has a form saying what it states rather than how it was written. What a published declaration already carries is that form, as `ClauseMeaning`. So the input is there; what is not is a consumer that can take it.

`Clauses.statedAt` answers a `Core`, and what reads it works over one: the clause a reading carries, which of the declaration's own fields it reads, and the parts its author wrote. `TermMeaning` does not hand a term back, and the reason it does not is the reason a boundary answer may not: a value that can be unwrapped is one two of which compare equal and can then be told apart. The discharge reads positions off what it is given, so that is not a hazard in the abstract.

Beside it, `Clauses` keeps two memos keyed by object identity — the typed clause by the expression it came from, and which fields a clause reads by the `Core` it produced. Both hold while one reading is in flight and one representation is in hand. A term arriving from a boundary answer is another object, so the lookup misses; a miss on the second answers with no fields read, and a clause whose fields are not all given is left to its run-time check. That is fail-open, and nothing that runs today would go red.

## Done when

- What a body is checked against, for a declaration another module wrote, comes from what that declaration publishes rather than from the tree it was written in.
- No `Core` is handed out across a store boundary.
- Nothing about what the discharge concludes depends on which object a term is.
- Where a guard or a clause is written is answered as a fact of its own, and not read off the term a reading was given — the same separation this issue's parent made for declarations, at the grain of a clause.

## Where to start

[...]

PR #1528, opened 2026-09-09
Let a world say which parts it leaves out by where they stand

kawasima:
A reading asked what one conjunct of a rule was holding is a reading of the
declaration in a world the author did not write. That world kept the roots of
the parts it leaves out and matched them against the nodes the walk was over.

A clause reaches a value through the tree the step that brought it there built,
and a substitution that changes anything builds afresh. So a world built from
one tree left nothing out of a walk over another: the reading was handed the
conjunct it had been told to leave out, the end that conjunct and its neighbour
placed came back owed to both of them, and the same line was written twice.

Which parts a world holds is now said by where in the clause they stand. The
occurrences are the clause's own, handed out once where the shape is read out of
the tree, so they name the same places whatever tree says them.

Found by reading every clause over a structurally equal tree rebuilt for it:
that leaves the whole suite green now and left two reports of a line drawn twice
before. The regression is stated where the world is asked — a world built from
one tree, a walk over another — and the mechanism it replaces turns it red.

The fixture that read each part into a shape of its own is read the way parts
are made instead. Apart, each of them is the first occurrence of a clause of its
own, and a world told to leave one out is told a place both of them stand at;
the one thing that makes parts descends a single shape.

Part of #1504.

### DIFF
--- a/souther-compiler/src/main/java/souther/compiler/check/ClauseView.java
+++ b/souther-compiler/src/main/java/souther/compiler/check/ClauseView.java
@@ -1,11 +1,10 @@
 package souther.compiler.check;
 
-import souther.compiler.core.Core;
 import souther.compiler.semantics.ConditionJoin;
 
 import java.util.ArrayList;
 import java.util.Collections;
-import java.util.IdentityHashMap;
+import java.util.LinkedHashSet;
 import java.util.List;
 import java.util.Set;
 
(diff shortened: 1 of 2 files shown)


kawasima:
Added the duplicated-part case as `andTwoPartsThatSayTheSameThingStandAtTwoPlaces`.

It asks the world directly rather than through a reading, for the reason you name: the fold takes the first side that is out and reads the other, so a world that had lost both hands back the reading of a world that had lost one. The case builds `same && same`, leaves the first conjunct out, and asks the world about a shape read from a tree built again.

Checked against the shortcut you were guarding: keying the omitted parts by what they say rather than where they stand leaves the rebuilding case green and turns this one red on the second conjunct alone.

PR #1543, opened 2026-09-09
Name a condition on the way to a border, and ask where it is where the sentence is written

### LINKED ISSUE #1486: A condition on the way to a border is told by what it does, except where it is not
Prerequisite of #1472, found by the census #1479 added.

`OnTheWay` has three arms and they do not agree about what their place is for. Measured over the conformance corpora and the benchmark corpora together:

* `OnTheWay.TakenIn.at` — reached, and no two of them differ only in the place. What tells one from another is the cut it came to.
* `OnTheWay.Narrowed.at` — reached, and no two of them differ only in the place. What tells one from another is the position it narrows.
* `OnTheWay.Declined.at` — reached, and two of them *do* differ only in the place. Taking it away would merge conditions this compiler tells apart.

The first two are handles: a reader is sent to the condition, and the value beside the place already says which condition it is. Both are reachable from an `Adequacy.Finding`, so each is a way a location still crosses into an answer — and a helper whose conditions move, saying the same thing, moves them.

## Why the third stays, and why that is the interesting part

`Declined` says a condition could not be turned into a cut, and what it carries beside the place is only why it was declined. Two conditions declined for one reason are then one value, and this compiler does tell them apart — so the place is doing identity's work because nothing else is.

That is worth saying out loud rather than leaving as an exception: what `Declined` is short of is a way to name which condition was declined, which the other two arms have and it does not. Closing that is the same question as the one behind an occurrence identity for an expression, and this issue is where the three arms are put on the same footing.

## Done when

`TakenIn` and `Narrowed` name what they are about and no place, with the place asked where a sentence is written; and `Declined` either does the same, having been given something to be about, or says in its own words why it is the one arm of three that cannot.

`WhatStillHoldsAPlaceUnderAFindingIsOneOfThreeThingsTest` is where that is counted: what this closes is its `StillToSplit` entries for this family, and it is what will say whether `Declined` is still read as a discriminator afterwards.

kawasima:
A condition on the way to a border carried where it is written. For two of the three arms that
place was a handle: what tells a `TakenIn` from its neighbours is the cut it came to, and what tells
a `Narrowed` from its neighbours is the position it narrows. For `Declined` it was identity — what
it carries beside the place is only why it was declined, and two conditions declined for one reason
are one reason — so the place was doing identity's work because nothing else was doing it.

So the reading names every condition it meets, and an answer holds the name and which question
places it rather than the place. `Declined` is the one arm that carries the name, because it is the
one arm whose other half does not tell it from its neighbours; giving all three the same field would
make two of one cut, taken twice, into two things.

Which question places a condition is settled where the condition is read, and it is a switch rather
than a first choice with a fallback. A construct the source wrote is placed by the module that wrote
it, filed beside the forks it already files and under an identity a copy cannot change — so a
condition inside a helper spliced into three callers points into the helper's file. A condition of a
shape the reading has no words for is any expression at all, and the source wrote no construct there
to name one by, so it is placed by the reading that met it, which is the only thing that did. Asked
the other way round, a condition written in a file this compilation has stopped holding would
quietly be reported wherever a reading met a copy of it.

What the reading of a body divides and where that reading met each condition it places itself are
now two questions over one reading, so an edit that moves a condition without changing what it says
leaves the measurement alone and moves only where a sentence points.

Closes #1486

### DIFF
--- a/souther-cli/src/test/java/souther/cli/AReportOfOneBorder.java
+++ b/souther-cli/src/test/java/souther/cli/AReportOfOneBorder.java
@@ -270,14 +270,16 @@ final class AReportOfOneBorder {
                                 .filter(point -> point.belongsToBehaviorAccount("weigh"))
                                 .toList()),
                         null),
-                souther.compiler.query.ClaimAnnotations.NONE, List.of(), java.util.Map.of());
+                souther.compiler.query.ClaimAnnotations.NONE, List.of(), java.util.Map.of(),
+                java.util.Map.of());
         return new AdequacyReport(AdequacyReport.SCHEMA_VERSION, "test",
                 held, WeakeningSet.none(),
                 List.of(new AdequacyReport.ModuleReport("example.wide",
                         new SourceId("wide.sou"), List.of(behavior), List.of(),
                         // Nothing this module's declarations are owed, and nothing that finding
                         // them went without: the fixture is about one behavior's own lines.
-                        new Adequacy.DeclaredBoundaries(List.of(), java.util.Map.of()))))
+                        new Adequacy.DeclaredBoundaries(List.of(), java.util.Map.of()),
+                        java.util.Map.of())))
                 .adequacy();
     }
 }
--- a/souther-cli/src/test/java/souther/cli/AReportOfOneBorder.java
+++ b/souther-cli/src/test/java/souther/cli/AReportOfOneBorder.java
@@ -278,8 +278,9 @@ final class AReportOfOneBorder {
                         new SourceId("wide.sou"), List.of(behavior), List.of(),
                         // Nothing this module's declarations are owed, and nothing that finding
                         // them went without: the fixture is about one behavior's own lines.
-                        new Adequacy.DeclaredBoundaries(List.of(), java.util.Map.of()),
-                        java.util.Map.of())))
+                        new AdequacyReport.DeclarationsShown(
+                                new Adequacy.DeclaredBoundaries(List.of(), java.util.Map.of()),
+                                java.util.Map.of()))))
                 .adequacy();
     }
 }
--- a/souther-compiler/src/main/java/souther/compiler/partition/ComparisonReadings.java
+++ b/souther-compiler/src/main/java/souther/compiler/partition/ComparisonReadings.java
@@ -15,6 +15,7 @@ import java.util.ArrayList;
 import java.util.Collections;
 import java.util.IdentityHashMap;
 import java.util.List;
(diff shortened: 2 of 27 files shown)


kawasima:
All three are fixed, and the first two came from one mistake.

**P1.** The name was minted by the act of reading a condition, and a subtree is read as a
condition more than once: the walk reads a short-circuit operator's left operand to say what stands
past it, and reads it again inside whatever encloses it — and `Condition.of` is itself recursive, so
reading `a && b && c` reads `a && b` again. Before this branch the fold was a pure function, so
re-reading produced equal values and the duplication was invisible; giving it names turned a
harmless recomputation into one condition wearing a name per reading, and `metHere` wrote every one
of them into an answer.

Reproduced first: in `if a && n > 3` the parameter came back as two different conditions on the two
accounts that carry it, and a four-term run of `&&` wrote down a triangle rather than its terms.

The register now answers for what it met. A site is the node a fold arrives at, under the names in
force there — asked after the way in has been looked through, so a truth reached through a binding
is the truth — and every reading after the first is answered with what the first came to. An arm is
met where the walk meets the fork, so a second naming of one is refused rather than answered.

Both properties are pinned, and each was watched to fail: one condition is one name however many
ways carry it, and what the reading writes down grows with the conditions rather than with the
reading.

**Two structural checks fell out of that fix, for one reason.** Making room for the register put the
recognition in a helper beside `Condition.of`, and the registers of who may read an operator name
the method that recognises. The recognition is back in the one method rather than the registers
being rewritten around it.

**P2.** Where the reading places a condition, the name and the anchor are one answer said twice, so
the constructor refuses a value whose halves name different conditions.

You were right about the check as well. It was measured on a model where the anchors also differed,
so it would have stayed green with the name taken away. What carries the claim now is a helper
called twice: one condition written once, met twice, reported at one place, and told apart by
nothing but the name — which is the case an anchor cannot answer, since it folds copies of one
construct onto one place on purpose.

**P3.** `WrittenCondition` refuses a construct no source wrote, in both arms.

**One thing I removed rather than fixed.** I had added a check that two conditions written alike are
two conditions. No mutation of the site key makes it fail — a `Core` node carries where it stands,
so every key tells them apart — so it was a green line that read like evidence and said nothing.
The reason written on the key said records make two alike comparisons equal, which is false; the key
is an identity question because a site is which node was arrived at, and the doc says that now.

On the note about #1472: understood. What is separated here is the projection, not the dependency —
when the cut lands, a caller holding a reading-owned place will need placement to be woken on its
own.

PR #1555, opened 2026-09-10
Ask a batch run where a condition is, from both places that can answer

### LINKED ISSUE #1495: A question was declared that no batch run asks, and the account of which questions those are was not touched
`Shapes.ClauseLocation` — where a clause of a declaration is written — is a
question this compiler declares and no run of the conformance corpus puts. It is
in neither of the two sets that exist to say why such a question is not put, so
`EveryQuestionThisCompilerDeclaresIsReachedOrOutsideABatchRunTest` fails on
`develop`.

The question came in with `35bd7ad04`, which separated an invariant's judgment
from where its clause is written. The capability that asks it is built where a
compilation is set up, so the question is reachable; what no batch run does is
point at a clause, which is when the capability asks.

## Why nothing saw it

That check is tagged `population`, so `mvn -o test` leaves it out and so does the
`population` job on a push, which is skipped. It is asked for by the run over
`develop` and `main` that happens on a schedule, and by naming the class with
`-Dtest=`. Every gate a change passes through on the way in stays green while it
is red.

## What has to be decided

Which of the two it is, and the sets are written so that answering that is the
work.

Either a batch run should reach it and the operations that check drives are
missing one — a report about an invariant a build could not discharge would name
the clause it is about, and a reader of that report is told where the clause is
written. If that report is one a build writes, then a run that reaches no such
report is the finding, and the question is a symptom of it.

Or nothing a batch run produces reads where a clause is written, the readers of
it being the editor and what answers a question asked about one place. Then it
belongs beside the questions written down as having no consumer in a batch run,
with the reason saying which readers those are.

Deciding it by putting it in the set because that turns the check green is the
one way out that would leave the first case unlooked at.

## Where to reproduce it

    mvn -o test -Dgroups=population -Dtest.excluded.groups=

on `develop`. Naming the class with `-Dtest=` runs it too.

## Done when

The question is reached by a run of the corpus, or written down as one nothing a
batch run reads — and either way the check that partitions them is green on
`develop`.

kawasima:
Closes #1495.

`EveryQuestionThisCompilerDeclaresIsReachedOrOutsideABatchRunTest` is red on
`develop`. The question the issue names is reached now; what is left are the
ones that place a condition on the way to a border — where it is written, the
walk that answers that for a module, and where a reading met a condition nobody
wrote a construct of.

The report reads all three: `AdequacyReport` asks for a place per reachability
gap, so a search that comes back with a row has no condition to account for and
puts none of them. Every search the corpus held certified, so none were put.
That makes this a corpus that does not exercise the two provenance forms rather
than a report that fails to read them, and putting the questions in the set for
what nothing reads would have said something untrue of the report.

So `booking` gets a module for each provenance. It already carries the questions
asked by something about to send a reader somewhere — where a clause is written,
where a declaration is written, where the declaration a line is owed at is
written — and these are the same kind of question about a condition.

One writes a string equality above the line, which is a comparison this compiler
has no cut for, so the walk declines it and the place is the writing module's
answer. The other writes a call, which is not a comparison, so the source wrote
no construct a place could be asked of by name and the reading that met it is
what places it. Both leave the line's rows composed without the condition, so
the rows do not arrive and what they were composed without is owed to a reader.

One module per provenance: the two are reached from nothing else, so a module
reaching both would say nothing about which had stopped being reached.

No production code changed, which is the check on the diagnosis. The regenerated
`booking` answers add the new modules and rewrite no existing line — the one
replaced line is the report's source index, which was empty because nothing in
that corpus pointed into a file and now names the two files it points into.

Noticed on the way and not touched here: `EverythingAnAnswerHoldsMeansSomething`
is red on `develop` for two places in an answer that mean nothing by equals, and
is red the same way with or without this change.

## What was run

The check this is for is tagged `population`, so the build on a push leaves it
out and the job for it is skipped. It is green here:

    mvn -o test -Dtest=EveryQuestionThisCompilerDeclaresIsReachedOrOutsideABatchRunTest

The whole `population` group is green too, apart from the class named above,
which is red the same way on `develop`. `mvn -o test` is green.


kawasima:
On the `String.length(value) >= 1` invariant: it is load-bearing, so it stays.

Taken off `Voucher` and `Ticket`, the corpus does not run at all — a batch over
it raises from inside the generator:

    java.lang.IllegalStateException: no class of seat/ticket holds a value it read:
    Constructed[type=conformance.waitlist.Ticket, fields={value=Text[value=w-]}]

Replacing it with one that draws no line, `Bool.not(String.isEmpty(value))`,
raises in the same place with the same value. So the shape of the invariant is
not what the corpus is resting on: with the length band the generator offers a
string it can then read back, and without one it offers `"w-"` and cannot.

That looks like something worth its own issue — a candidate this compiler
composed and then had no class for is a fact about the generator, not about a
corpus. It is not touched here.

PR #1566, opened 2026-09-10
Put the names a position wears on the classes its body draws

### LINKED ISSUE #1559: The generator composes a candidate whose position has no class that reads it
A search offers a row, and classifying the value it wrote raises from inside the
compiler:

    java.lang.IllegalStateException: no class of seat/ticket holds a value it read:
    Constructed[type=conformance.waitlist.Ticket, fields={value=Text[value=w-]}]

`InputClassifications.decided` says what that means: every class of the axis read
the value and none holds it, which is the exhaustiveness `Axis` declares of its
classes being broken. It is written as a contract failure rather than as a value
that could not be read, and that is the right reading of it — nothing here is a
measurement coming back short.

## Where the value comes from

`conformance.waitlist` in the booking corpus writes

    guard String.startsWith("w-", ticket.value) else NotOnTheList

and the value the classifier is handed is `Ticket("w-")` — the literal out of
that guard. So the composer reached for the prefix the guard names and built a
ticket out of it, and the position's classes then hold nothing that reads what
was built.

## What it depends on, and what it does not

`Ticket` carries `invariant String.length(value) >= 1`, and with it there the
corpus runs: the strings offered are ones the classes read back. Take the
invariant off and the raise is immediate. Replace it with one that draws no line
at all — `Bool.not(String.isEmpty(value))` — and the raise is the same, at the
same place, over the same value.

So this is not about the shape of the invariant, and it is not the length band
being required for a ticket to make sense. What the band does is change what the
composer offers. Without one, what it offers is a value its own classes cannot
place.

## Where to reproduce it

On `develop`, take the invariant off `Ticket` in
`souther-compiler/src/test/resources/souther/compiler/conformance/booking/waitlist.sou`
and run anything that analyses the corpus, for instance

    mvn -o test -Dtest=EveryQuestionThisCompilerDeclaresIsReachedOrOutsideABatchRunTest \
        -Dsurefire.failIfNoSpecifiedTests=false

The same holds for `conformance.vouchers` beside it.

## Done when

A value this compiler composed is one the classes of the position it wrote it at
can place, or the composer does not offer it — and either way a corpus reaching
this stops depending on a declaration's invariant to keep the composer inside
what it can read back.

### LINKED ISSUE #1565: A position measured at one number is not measured at another its rules name
A location has as many numbers as its rules name of it — what stands there and
what an operation counts of it are two, and a rule about either is a rule about
that one. Which measures a location has is asked of the location rather than of
the numbers, so a position the declarations measured at any number keeps only the
measures they gave it, and what a body says about another number of the same
place is not taken up.

## What is observed

`conformance.waitlist` in the booking corpus writes

    guard String.startsWith("w-", ticket.value) else NotOnTheList

With `Ticket` declared as a plain `String`, the position is measured at itself and
carries the two classes that guard draws:

    axis seat/ticket  derivable=true
      class ticket/String.startsWith("w-", x)
      class ticket/not String.startsWith("w-", x)

Add an invariant that says nothing about the prefix —

    data Ticket = String

— and the axis is replaced rather than joined:

    axis seat/String.length(ticket)  derivable=false
      (no classes)

The prefix classes are gone. Both numbers are there to be measured: the position
has an entry for its own value and one for its length either way, and the second
is bounded while the first is what the body divides.

## Where it happens

`Partitions.numbersMeasuring`:

    if (at.hasMeasures()) {
        List<NumericTerm.FromOnePosition> declared =
                at.axes().stream().map(Axis::term).toList();
        ...
        return numbers.stream().filter(declared::contains).toList();
    }

The predicate is about the location. With no invariant nothing has measured the
place, so the body's number is taken up and the axis exists; with one, the length
has been measured and the body's number is filtered out.

It is not silent to itself — the evidence is filed as
`ThePositionIsAlreadyMeasured`, whose own words say it is "a policy of this stage
and not a fact about the model" and that "an author is not told anything". So the
distinction the model makes is dropped, deliberately, and nobody downstream can
see that it was.

## Which contract it breaks

`Axis` and the walk that assembles measures both say a location is measured at as
many numbers as the rules name of it — which hour of a time and which minute are
two measures, and choosing one loses the other's lines. The policy caps that at
whatever the declarations happened to name first.

## Where to reproduce it

Take the invariant off `Ticket` in
`souther-compiler/src/test/resources/souther/compiler/conformance/booking/waitlist.sou`
and ask a compilation what it divides `seat` into, then put it back and ask again.
`conformance.vouchers` beside it has the same shape with an equality instead of a
call.

## Done when

Which measures a location has follows from the numbers its rules name and not
from how many the location already has: a number both sides write about is one
measure with both read into it, and a number only one of them names is a measure
of its own.

kawasima:
Closes #1559.

Reworked after review. The first attempt fixed this at `Recognitions`, by reading
a construction that carries one field called `value` as the value inside it. That
was wrong at the root, and the review was right to refuse it: whether a
declaration is a name over a value is settled by the declaration and never by the
shape of what a row wrote, and `data X = Y` and `data X = { value: Y }` are two
declarations (spec §data). The one reader that takes names off an observation
already takes them off by name — it walks the wrappers the position's type view
read and strips only where the construction's type is the name it expected — so a
record whose own field is called `value` is never mistaken for a name over one.
That patch went around that guarantee and put a second name-stripping beside it,
deciding from the runtime shape what the declarations had already decided.

## Where it actually goes wrong

Every kind of class made from a position's own declarations is published under
the names the position wears: the type view is read once, and what it says is put
on the truths, the presences, the cases and the values singled out.

The classes composed from a body's rules are not. `String.startsWith("w-", x)`
comes back as the bare set of strings, and nothing on it says a row writes one of
those under a name. So a value written under one arrived at a class drawn over
what the name holds, was read as the declaration it is, and was in none of the
sets those rules stated. Every class read it and none held it, which is what
`Axis` declares cannot happen — raised, and rightly, because it is the
partition's own exhaustiveness broken rather than a measurement coming up short.

The fix puts the names on where the measure is assembled, which is the one place
the position's type is read. The rules that composed the classes go on knowing
what they divided and nothing about what is written over it.

## The question that was asked of the arm

Fixing that turned an existing test red, and the test was right to go: it asked
whether a class is a set with an `instanceof` on the arm in hand. A class of a
position that wears a name is written down under the wearing, so the outermost
arm is the wearing and the meaning is one inside it. Asked that way, a position
divided into sets reads as one divided nowhere.

So the question moves onto the meaning, beside the two that were already asked
that way. Main code never asked the bare arm — everything about a recognition
goes through the one reader — and this was the only place that did.

One place asserts an arm and stays: its model gives the position no name, so
nothing is worn and the assertion is exact for what it tests.

## What the tests hold

A model whose body divides a string under a name, with a row on one side. It
asserts which class the row landed in rather than that some class was reached,
and that no row went unread.

Beside it, the same body over a record whose one field is called `value`. That is
where the two declarations part and why nothing downstream has to tell them apart
from what a row wrote: a name over a value is one position and the classes are of
it; a record is given up in favour of the positions under it and the classes are
of the field. Taking the names back off leaves that one green and raises on the
other, with the sentence the issue was filed about.

## What this does not fix

#1565 is beside this one: where a rule bounds some other number of the position,
these classes are dropped from the measures altogether and nothing asks which
class anything is in. That is why the booking corpus does not reach this. It is
not a prerequisite — the corpus is measured exactly as it was and the checked-in
answers are unchanged.

## What was run

`mvn -o test` is green, and the whole `population` group is green.

### DIFF
--- a/souther-compiler/src/main/java/souther/compiler/partition/PartitionClass.java
+++ b/souther-compiler/src/main/java/souther/compiler/partition/PartitionClass.java
@@ -1,8 +1,11 @@
 package souther.compiler.partition;
 
 import souther.compiler.inputs.Refinement;
+import souther.compiler.types.TypeSymbol;
 import souther.compiler.values.Value;
 import souther.compiler.values.ValueSet;
+
+import java.util.List;
 
 /**
  * One equivalence class of an input position: a set of values expected to behave the same way.
@@ -91,6 +94,22 @@ public record PartitionClass(String id, String label, Recognition recognises,
                     + " and cannot be made a class of " + number);
         }
         return new PartitionClass(id, label, recognises, representatives, denotes, selects, number);
+    }
+
+    /**
+     * The same class, asked of the value inside the names the position writes it under.
+     *
+     * <p>What the class means is about the values the position holds, and a row writes one of them
+     * under whatever names the position wears — so the two are one class and the names are how a
+     * row spells it. Said by whoever assembles the measure, which is where the position's type is
+     * read; a class made from the rules alone has no way of knowing what wears it.
+     *
+     * <p>Nothing said where the position wears nothing ({@link Recognition.Under#of}), so a class
+     * of a bare position stays the class it was rather than becoming a second spelling of it.
+     */
+    public PartitionClass under(List<TypeSymbol> worn) {
+        return new PartitionClass(id, label, Recognition.Under.of(worn, recognises),
+                representatives, denotes, selects, of);
     }
 
     /**
--- a/souther-compiler/src/main/java/souther/compiler/partition/Partitions.java
+++ b/souther-compiler/src/main/java/souther/compiler/partition/Partitions.java
@@ -678,8 +678,14 @@ public final class Partitions {
             // lies — so a position whose classes are sets has none, and a value singled out is in
             // them rather than beside them. Published as both, one distinction would be stated
             // twice in two algebras and every line would fall in no class ({@link Axis}).
+            // Under the names the position wears, which is where they are put on: the rules divided
+            // the values the position holds and knew nothing of what is written over them, and a
+            // row writes one of those values under every name the type puts on it. Left off, a
+            // class drawn over strings is asked about a construction and holds none of them —
+            // which is the classes of a position holding nothing the position holds.
             case Classing.Classed.Composed(List<PartitionClass> it) ->
-                    new Composition(it, composedFrom(dividing), false, null);
+                    new Composition(it.stream().map(each -> each.under(view.wrappers())).toList(),
+                            composedFrom(dividing), false, null);
             // No classes and nothing they were composed from, and the cuts stand: they are
             // observations of their own rather than a projection of the classes.
             case Classing.Classed.NotComposed(var why) ->
--- a/souther-compiler/src/test/java/souther/compiler/AValueWrittenUnderANameIsInTheClassesOfWhatTheNameHoldsTest.java
+++ b/souther-compiler/src/test/java/souther/compiler/AValueWrittenUnderANameIsInTheClassesOfWhatTheNameHoldsTest.java
@@ -1,0 +1,146 @@
+package souther.compiler;
+
+import org.junit.jupiter.api.Test;
+
+import souther.compiler.query.Adequacy;
+import souther.compiler.query.Compilation;
+import souther.compiler.query.PartitionEvidence;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertNotNull;
+import static org.junit.jupiter.api.Assertions.assertTrue;
+
+/**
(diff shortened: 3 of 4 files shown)


kawasima:
Both points taken.

`Recognition.saysWhichValues()` is gone. It had no reader in the compiler and the
name claimed more than it answered — a value the rules singled out, a truth and a
presence all say which values are in them. What the test wanted was narrower and
is now a recursive helper beside the assertions that use it, so no vocabulary was
added to `Recognition` for a test's sake.

The record test's own account was overclaiming and now says what it holds. The
position there is the field, so nothing hands a construction to a class at all
and a reader that read the two constructions alike would pass it as readily. What
it witnesses is the separation — a name over a value is one position, a record is
given up in favour of the positions under it — and not the absence of a reader
that would guess. What refuses the guess is that the names a class is written
under are the ones the position's type view read, and taking them off asks for
those names.

`mvn -o test` and the `population` group are green on the amended head.

PR #1567, opened 2026-09-10
Measure a location at every number its rules name, whichever side wrote about which

### LINKED ISSUE #1565: A position measured at one number is not measured at another its rules name

kawasima:
Closes #1565.

The issue's first account of the cause was wrong and has been rewritten. It read
this as classes being dropped where the position's own value carries no bound.
The classes are not dropped there — both numbers of the position exist to be
measured either way, and the declaration side is right to make no measure of a
number nothing bounds.

## Where it goes wrong

`Partitions.numbersMeasuring` asks which measures a location has of the location:

    if (at.hasMeasures()) {
        ... return numbers.stream().filter(declared::contains).toList();
    }

So a position the declarations measured at any number keeps only what they gave
it, and a body dividing a different number of the same place is filtered out. A
ticket whose length an invariant bounds and whose value a guard divides comes back
measured at its length and divided nowhere.

The evidence was filed as `ThePositionIsAlreadyMeasured`, whose own words say it
is a policy of the stage and not a fact about the model, and that an author is
told nothing. That is what let it stand: the account refuses a stage that reports
itself complete over evidence it never said what became of, and the disposition
was the sentence that satisfied it.

Which contradicts what `Axis` and the walk above it both say: a location is
measured at as many numbers as the rules name of it, and choosing one of two
loses the other's lines.

## What it is now

The question is asked about numbers. One entry per number the rules name; the
measure of each is looked up by that number; a number both a declaration and a
body speak about is one measure with both read into it, and a number only one of
them names is a measure of its own. The disposition goes with the policy —
evidence about a number is measured at that number, whichever side named it
first.

Nothing new deduplicates: the walk already keyed the measure by the number, and a
location measured twice at one number is refused where the measures are built.

## Cost

An axis more on a position whose body divides a number its declarations left
alone, and the pair space is a product over axes. That is the price of the model's
own distinctions being published rather than left out of the universe to keep the
product down.

What that costs where a model is larger than these is not measured here, and this
does not claim a budget catches it. What is known is what the corpora came to: the
pair spaces that changed were the two that had been empty, because those positions
had one axis.

## What the tests hold

A declaration bounding the length and a body dividing the value: both numbers are
measured, and the one the body divides carries what it divided into. Beside it,
the declaration bounding both numbers and the body writing twice about one: still
one measure per number, and the classes are what the body's rules come to between
them.

Putting the policy back raises rather than failing an assertion — the account
refuses a stage that dropped evidence without saying so, which is what the removed
disposition used to answer. Taking away the deduplication is refused where the
measures are built, naming the number counted twice.

## What else moved

The booking answers are regenerated: the voucher and ticket positions now carry
the axis their bodies draw, and rows are offered per class of it. The only
replaced lines are two pair spaces that were empty because those positions had one
axis.

A walk of a store now reaches the machine under a language a class of a measure
holds, at two places it could not reach before — the corpora had no such class in
an answer. Both were already written down where the declarations are walked, so
the register beside them gains the same reading.

## What was run

`mvn -o test` is green, and the whole `population` group is green.


kawasima:
The unused parameter is gone, and the call site with it. `numbersMeasuring` asks
which numbers the rules name and nothing else; with the disposition removed it has
no word to file, so it no longer takes the account. Pushed.

Worth saying what let it through here rather than only that it did. `mvn -o test`
does not run Error Prone — the gate for that is the lint build, which I do not run
locally, so CI is the first thing that reads it. The local run I reported was
green and said nothing about this, which is what a green report is worth when the
check that would have spoken was not among the ones that ran.

One correction of my own in the body. I wrote that where the pair-space cost has
to be paid it is paid by the budget that limits pair enumeration. I did not check
that, and looking for it now I do not find such a limit by that name in the
compiler. So the claim is out: what this costs on a model larger than the corpora
is not measured here, and the paragraph says only what was observed — the pair
spaces that changed are the two that were empty.

On the generator: agreed, out of scope here. `voucher /= spring` is plainly
inhabited and the class is published with no representative rather than dropped,
which is what #1565 asks for. I will raise it separately unless you would rather
fold it in.

Issue #1546, opened 2026-09-10
Decide where a declaration-reading capability may go inside a walk over what an author wrote


kawasima:
A reading of a declaration is lent: made once for a declaration under a revision and handed to whoever asks next, because the readings a question needs beyond the first are readings of the same declaration again. Where the lending is reached, a reader says where it borrows from; where it is not, the reader reads for itself and pays every reading over.

It is not reached inside a walk over what an author wrote. #1511 took it as far as the readers a store can hand it to and stopped there, because what is left is not plumbing: the walk is handed what it is reading and nothing that says where a reading comes from, and putting one there is a decision about what such a walk owns.

## Where it stops

`ValueGuarantees` reads a declaration for itself. It is reached from `RunReach` and `ReadQuantities`, and from `UniversalElementFacts` and `StepInputFacts`, which are reached from `Terms`.

`Partitions.fieldsOf` reads a declaration with a coordinate settled for itself. It is reached through `Partitions.composed`, `ValuesCarryingANumber`, `ContainersAddingUp` and `TermRealizations`.

`InvariantChecker.capabilityOf`, `ContractDischarge.of` and `PathReachability.of` build the walk itself — `Terms`, `Clauses`, `PathEngine` — and what they build has nowhere to put a lending.

## What has to be settled first

**What a borrower is handed when the reading it asks for is the one being made.** A reader under such a walk can ask about the declaration whose reading is under way, about one already made, and about one that is itself being made further up. `whileTheAnswerIsMade` holds that boundary today and holds it outside the producer only; a lending threaded down to `Terms` reaches past it from underneath. Deciding this is the whole of the issue, and threading the argument before it is decided is how the boundary gets walked around quietly.

**Whether the capability belongs to the strategy or to what invokes it.** `ValuesCarryingANumber` is a record and a `PlanComposer.Values`: a strategy for filling a plan. A lending as a component makes where a strategy borrows part of what a strategy is, and two that differ only by that compare unequal. The question is whether a capability belongs to an evaluation rather than to a strategy's identity — the record refusing to hide it is the useful part, and turning it into a class with a hidden field answers the equality and not the ownership.

**Whether `Terms` and `TermRealizations` should carry one at all.** What they want may not be somewhere to borrow from but a way to read, which is a narrower thing and would say what happens on re-entry rather than leaving it to whoever holds the lending.

## Where the population is

[...]

PR #1629, opened 2026-09-12
Carry the world a reading is made in down the walk that reads it

kawasima:
A reading of a declaration is lent. Where the lending was reached a reader said
where it borrowed from; inside a walk over what an author wrote it was not
reached, so every reader under one read each declaration whole however many
times it arrived at one.

What such a walk needed was not the lender but the world it reads in, and that
already existed: rules, a budget and where to borrow from, travelling together
so that a step substituting one of them is a step reading some other world. The
walk that builds values already carried it. The walk over the terms did not, nor
did the reading of an input, which was handed the lender and dropped it.

Both carry it now. What a value guarantees, what a step is handed and what a run
reaches read in the world their caller read in, and the way in that read without
borrowing for them is gone.

## What a borrower is handed when the reading it asks for is being made

The reading being made bounds what may be borrowed while it is made, and the
bound was put on the lender alone — so whoever put it there rebuilt the world
around it to hand it down, and the bound held only for as long as every such
reader remembered to. A walk handed the lender could reach past one nobody had
rebuilt.

The world derives it now: same rules, same budget, lender bounded. A reader
below is handed a bounded world and has no other to hand on, and the lender is
out of reach from outside the package.

## What is checked

The places this compiler starts a reading somebody else has already made were
written down as one set, and the set was saying two things at once: that a world
with nothing lent to it is a thing anybody may build, and who was building one.
Read together, the second is satisfied by the first going missing — an extraction
that stopped resolving call sites reports no such reader, which is what a
compiler with none looks like.

They are apart now. The way in is asserted to be there, and the readers to be
none. Both were put to a control: a reader reaching the way in fails the second
and not the first, and blinding the walk to how a field is named fails the first
and not the second.

## What this does not settle

The discharge analysis reads a scope of its own and is handed a world of its
own, which is what it had. A reading made in it is still one nothing lends
against, and whether the compilation should mint a source over that scope is
open.

Whether a strategy may hold a world is left as it stands. One does, and its
generated equality includes the particular world; nothing compares two of them,
so there is nothing to repair — which is not evidence that a world belongs in a
strategy's identity.

Refs #1546

### DIFF
--- a/souther-architecture-test/src/test/java/souther/architecture/NothingHereStartsAReadingSomebodyElseHasAlreadyMadeTest.java
+++ b/souther-architecture-test/src/test/java/souther/architecture/NothingHereStartsAReadingSomebodyElseHasAlreadyMadeTest.java
@@ -239,8 +239,23 @@ class NothingHereStartsAReadingSomebodyElseHasAlreadyMadeTest {
                         + " with nothing lent to it.");
     }
 
-    /** Every edge into a reading nobody else made, said in full at both ends. */
+    /**
+     * Every edge into a reading nobody else made, said in full at both ends.
+     *
+     * <p>Walked once for the class. The two questions below are two readings of one walk over every
+     * method of every compiled class, and asking each of them for its own walk reads the whole
+     * output twice to answer about the same instructions.
+     */
     private static Set<String> everyEdgeIntoAReadingOfOnesOwn() {
+        if (EVERY_EDGE == null) {
+            EVERY_EDGE = walkForEveryEdgeIntoAReadingOfOnesOwn();
+        }
+        return EVERY_EDGE;
+    }
+
+    private static Set<String> EVERY_EDGE;
+
+    private static Set<String> walkForEveryEdgeIntoAReadingOfOnesOwn() {
         Map<String, Set<MethodTypeDesc>> pairs = theShorterOfEachPair();
         Set<Named> waysIn = theWaysInThatSayTheyReadForThemselves(pairs);
         Set<String> reaching = new TreeSet<>();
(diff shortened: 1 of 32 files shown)


kawasima:
Both blockers are fixed, and the non-blocker is in too.

## The stale `@param`

The cause is that javac does not read the tags, so a signature change leaves them and the build says nothing. Fixing the two the compiler named would have left the same shape everywhere else this PR changed a signature, so the population came from the machine instead: every javadoc block in every file this PR touches, with its `@param` names checked against the parameters of the signature that follows. Those two were the only ones.

## The control stood on the wrong extraction

You are right, and the hole was wider than the one edge.

`whatItNames` has three answers, and they are three extractions: a member called, a static field read, and a member handed over to be called later among the arguments a bootstrap is given. The emptiness claim rests on the first — a reader reaches a reading of its own by calling one. The row I used as the control is named by the second. So did the mutation I ran to show the control worked, which is why it looked convincing: the control and the mutation were the same branch.

The third was uncontrolled as well. Nothing in the population is named that way today, so blinding it changed nothing visible.

Each is witnessed now, on a member known to be named that way and independent of the population: deriving a bounded world calls the lender's own, the way in reads what there is nothing to borrow from, and reading a clause on its own hands over where a name stands. Each branch was blinded in turn — blinding the call leaves the emptiness green and fails only the new control, which is the false negative you found.

One limit is written where the cases are: nothing here can ask a switch what cases it has, so a fourth way to name a member is a fourth line and is noticed by a reader rather than by a failure.

## The lender's visibility

Taken as a rule rather than left as debt, since it was cheap and the same shape as what is already here. Package-private says nothing to the walk this is about — the walk is in the package that declares the world. So the methods that reach the accessor are written down, per method rather than per class: what is entitled is a bridge to the shape that still takes the rules, the budget and the lender apart, and a class holding one such bridge is not a class whose every method may reach past the world it was given.

The rows are named rather than counted, so an extraction that stopped reading instructions comes back with nothing and fails rather than agreeing with an emptiness it never looked for. Both directions were put to a control: a walk reader rebuilding a world around the lender is a finding, and a blinded walk is one too.

Full reactor is green.

Issue #1679, opened 2026-09-15
The domains a module's behaviors take are read again for an edit to any body


kawasima:
An edit to one behavior's body reads the rules of every declaration that every behavior of the module takes.

Found while closing the same shape on the count of which declarations have no value (#1466). The count is now taken from what the module declares and asks each declaration what it settles, so an edit to a body leaves it alone. The readings that edit costs did not go away — they moved to the next reader, which is `Adequacy.Inputs`.

## What it is

`Adequacy.Inputs` is one answer per module holding the domain of every input every behavior of that module takes. It reads `Bodies.Settled` and `Bodies.SettledFn`, which carry the bodies, and its computation reads each behavior's input types and everything those reach.

So the key is the module where the fact is per behavior, and the root carries bodies where what an input domain comes to is settled by the declarations the input reaches. Either one alone would be enough to have an edit to one body reach every declaration in sight; together they are what makes the reading count of an edit say which file was touched and nothing else.

This is the cut rule the store's own documentation states: what a consumer depends on is the answer its meaning stops at, and a question asked per definition does not take a module's index as its own dependency. `Machines.OfDeclaration` and `Shapes.CardinalityPremiseOf` are the shape to follow.

## Where to measure it

`souther-bench`'s `Incremental` holds a store and edits one document; the reading count is `InvariantChecker.readingsMade()`. Rewriting a body that changes no signature is the edit to take it around, and the readings that arrive under `Adequacy.Inputs` are what this is about.

## Done when

A behavior's input domains are asked for per behavior, rooted in what the behavior declares and in the declarations its inputs reach, so that an edit to another behavior's body reads none of them. Held the way #1466's cut is held — by whether the answer was worked out again, which is what a test can see — rather than by a count that another reader can pay on its behalf.

## What this is not

It is not the whole of #1466. That issue asks for the readings an edit remakes not to grow with the declarations the edit says nothing about, and it stays open until no reader is paying that. This is the reader that pays it now.

kawasima:
Closed by PR#1681, merged into develop.

What arrives at an input is asked of the behavior whose input it is, rooted in that behavior's declaration, the implementation the body was checked against, and the declarations its inputs reach. The module's answer is which behaviors it has, and hands out the reading each of them kept.

Four ways a behavior beside it moves leave that reading alone — its body, what it states about its own answer, its being declared at all, and a rule of a declaration this input never reaches — each against a control that does move it, and each against a mutation that reddens it alone.

[...]

PR #1680, opened 2026-09-15
Count what a module declares from its declarations, and ask each what it settles

### LINKED ISSUE #1466: The canonical readings an edit remakes are chosen by its document
An edit to a declaration remakes the canonical readings its document reaches, whichever declaration it touched and whatever it did to it.

Found while measuring what an edit to a stated relation costs (#1459), against the crm corpus the benchmarks carry. The reading count is `InvariantChecker.readingsMade()`, which is there for holding a reader to when it reads; the substitution count is `AssumedContract.TAKEN_IN`, which says whether a body was checked against a contract again.

## What was measured

One document of the corpus edited in a store that already holds the answers, one shape after another against the same store, each shape alternating between two texts so that every round is an edit.

Every edit that touches a declaration makes the same readings, and that count does not change with which declaration was touched or with what the edit did to it. Renaming an `ensures` clause makes them all. Appending a behavior and its `let` at the end of the file makes them all. Restating a rule of the clause makes them all.

What the three reach after that is not the same. Restating a rule is the only one of them that reaches a substitution: a body that calls the behavior is checked against the contract again. Renaming the clause reaches none — which is the boundary #1419 drew working, since a clause's name is not part of what a caller may assume. Appending an unrelated definition reaches none either.

So the depth each edit reaches differs, and the count of readings does not. What settles the count is the document that was touched: an edit to the document nothing imports makes fewer than an edit to the one the others are built on. It is not the count of that document's declarations — a canonical reading is named by a declaration together with the source its clauses are read from and what the reading may spend, so one declaration stands behind more than one of them.

A comment that moves nothing makes none of them: written at the end of a file, or reworded to a line of the same width anywhere in it, it leaves every reading in hand. Written as a new line it makes more of them than renaming a declaration does, because every position under it moves and the answers about a module carry positions.

So the count follows neither what the edit meant nor how far it reached. Two edits arrive at it — one that changes what a declaration says, and one that changes only where the declarations are — and what comes back says which document was touched and nothing else.

The lender keeps a reading for a revision and drops what it holds when the revision moves. A revision moves whenever an input is set to a value it did not have, so an edit anywhere drops every reading in hand, and that is where some of the count comes from — left undropped, that many are borrowed rather than made. The rest are made either way.

## What this is not

It is not what an edit costs an author. The making of canonical readings was timed where it is made, nesting included, and it is a small part of a round: an edit that made none of them would leave the latency about where it is. What decides a round is #1468 for an edit that touches no declaration, and for one that does, the module-wide answers the front of the pipeline hands on — each of which genuinely comes out different when a declaration changes, so there is no equal answer there to stop at.

An earlier reading of this issue took the count for the cost and pointed at #1410, where narrowing the readings of a whole compile did move the time. That does not carry over to the edit path, and it is withdrawn.

## Where to reproduce it

`souther-bench`'s `Incremental` is the harness — it holds a store, edits one document and asks for diagnostics. Adding a shape to it and counting readings and substitutions around the round is how the above was taken. `TheRelationAnEditRestatesIsOneACallerReadsTest` shows how the substitution count is read from another module.

## Done when

The canonical readings an edit remakes do not grow with the declarations of its document that the edit says nothing about. Held against a model whose declarations are a parameter, so that what is fixed is how the count grows rather than what it is for one corpus. Or the readings are shown to be what they have to be, and that is written down where the next reader of a benchmark line will find it.

kawasima:
Refs #1466. It does not close it — see the last section.

## What was wrong

Which of a module's declarations no value satisfies was answered by `Shapes.TypesWithNoValue(module)`, which took `Bodies.Lowering(module)` as its input and ran the count over every declaration the module reaches. The lowering carries the bodies, so an edit to anything in the file asked for the count again — and asking for it reads the rules of every declaration in sight. The canonical readings an edit remade said which document was touched and nothing else.

Rewriting one behavior's body, changing no signature and no declaration, remade the reading of every data declaration beside it.

This is the cut rule the store's own documentation states, left unapplied inside a module. `Machines.OfDeclaration` is already per declaration and goes stale on that declaration's meaning; `TypesWithNoValue` was the reader still deriving per-declaration facts from a module-wide representation.

## What it does now

The count starts from what the module declares. `Front.DeclaredTypes` answers the types a module writes and the order it writes them, read off the parse — which is where the names are introduced, and where the order a group of them is reported in comes from. It is a sequence because the order is content: a group of types with no value is reported at the first of them the module declares, and its members are named in that order.

What each declaration settles before the count begins is that declaration's own answer, `Shapes.CardinalityPremiseOf`. The counts its rules ask a collection to hold decide the lattice the count rises through and have to be in hand before it starts; whether every rule of it could be read is what says a count that stopped short may have missed the rule that empties a type. Both come off one reading, and the reading is made under the declaration's own module — so it is the declaration's canonical one, and a count of a module that reaches a type of another is handed what that module's reading came to rather than reading it again under the scope of whoever reached it.

The cuts and the shortfall are then a union and an and over those answers. `TypeCardinality.solve` takes the names the count is for and asks for the premises as it walks; `reached` only ever used the names.

## Why the cardinality itself is not the cut

An answer per declaration for the cardinality would not work. The count is a least fixed point over the strongly connected components of the declaration graph, and the lattice it rises through is fixed before the rising starts — a count discovered part way gives it somewhere new to go, which is what makes it stop. `CardinalityTransfer.upperOf` depends on the neighbours already settled and on what a counterfactual grants, so it is not a fact about a declaration. It is left alone here.

## How it is held

Two tests, both reading whether an answer was worked out again rather than what it came to.

`WhatHasNoValueIsCountedFromWhatAModuleDeclaresTest` — a body written another way leaves the count where it was, and a rule stating something else takes it again, which is what says the first half is about the body rather than about an answer nothing could move.

`ADeclarationSettlesWhatACountNeedsOfItAloneTest` — a rule moving settles its own declaration again and the record whose field reaches it, and leaves every declaration outside that reach. What a declaration settles is read off its own rules and off the rules of the fields it reaches, so the reach is where the claim stops.

Both were checked by mutation: putting the module's lowering back under either answer turns the matching test red.

Two structural tests moved, each judged. `TypeSymbols.declared` gains a site where a declaration says which one it is, beside the ones already there. What makes a name a rule may write moved with the paths it is made from, out of the count and into the premise.

## Why this does not close #1466

The readings an edit remakes did not fall. They moved to `Adequacy.Inputs`, which is one answer per module holding every behavior's input domains and roots in the settled module — the same shape, one reader along. That is #1679.

Two things this turned up. The count's pre-pass was not its only reader of a declaration: the fixed point reads each one as it rises, and both were under one aggregate where the lender made them look like one reading. And the reading count of an edit is not a measure of one reader — whichever consumer reaches a declaration first pays, and the rest borrow for the revision — so a cut is held by whether the answer was worked out again, and the count is the acceptance test for #1466 once no reader is paying it.


kawasima:
Two more commits, from a walkthrough of the diff for what it costs.

`OccurrenceCounts` found the atom a question is about by projecting the reading's whole table of counted names into a new map and looking one key up in it — once per question. A count asks one question per size it wants told apart, and filling a collection from a finite element asks about every size up to how many values the element has, so the projection was built that many times per name. The reading now answers for one name beside answering for all of them.

And what a declaration settles for a count walked its spreads again to find its fields, where the source it is reading under answers for exactly that walk and the reading it is taken from is about to be given it.

Neither changes an answer, and the suite is green over both.

Not timed. Which of these is worth anything is a question about the share they take of a compile, and a call count is not that.

PR #1683, opened 2026-09-15
Say what every index a question about one definition reads is

### LINKED ISSUE #1682: A per-definition question may read a module's index, and nothing says whether it was cut
The store states three rules about what an answer may be, and two of them are checked. `EverythingAnAnswerHoldsMeansSomethingTest` says which answers hold something that is not a value; `EveryAnswerThisCompilerDeclaresIsSettledTest` says which of the two kinds a question is read under. The third — an edge is what its consumer means — is stated in `Db`'s own documentation and checked by nothing.

So a violation of it is found by someone noticing that an edit is slow. #1466 found one reader; #1679 found the one that was paying it next; the first commit of PR#1681 left two more of the same shape in the key it had just written, and the review found those. The same mechanism reproduced itself inside one pull request, which is what says the thing to close is not any one reader.

## What the rule actually is

Not "a question about one definition may not read a module's index". `Bodies.DeclaredSignature(module, behavior)` reads `DeclaredSignatures(module)`, and that is the cut, not a breach of it.

What must not happen is that the module's identity reaches a consumer whose meaning is about one definition. An index read by a question about one definition has to be cut, and there are two ways an edge can be sound:

- **A projection.** The reader's answer is one entry of the index, so an entry beside it moves the index and the projection comes out equal. `Bodies.Spec` is the shape.
- **An equal answer under a sibling edit.** The reader folds the index in, but what it comes to does not move when a definition it says nothing about is edited.

An edge that is neither is a violation. Which one an edge is cannot be read off the dependency graph — it takes the graph, what equality says about the answer, and an edit. `Bodies.CheckedBehavior` is why: it reads two of its module's indexes whole and is still left alone when a behavior is declared beside it.

## Where the edges are

Asked of the store rather than of the source. `Db.everyAnswer()` and `Db.dependenciesOf(key)` give the graph a compile actually built; an edge is a candidate where the reader's key names more than the module it is about, and the read key's only component is the module name and its answer is a collection.

Two ways that predicate misreads, both met while writing it: a key like `Shapes.FieldBindingsOf` has one component and is about one declaration, so the component has to be compared against the module name rather than counted; and `Output.Evaluated(name, arms)` has two components and is about a module, the second being a policy.

What the graph holds now, with what each edge is:

| reader | index | what it is |
|---|---|---|
| `Bodies.CalleeSigsForBody` | `Bodies.CalleeSigs` | projection |
| `Bodies.DeclaredSignature` | `Bodies.DeclaredSignatures` | projection |
| `Bodies.Stated` | `Bodies.StatedContracts` | projection |
| `Bodies.CheckedBehavior` | `Bodies.ReqSigs` | equal under a sibling declaration |
| `Bodies.CheckedBehavior` | `Bodies.RecursiveHelperConstructs` | equal under a sibling declaration |
| `Bodies.LoweredBody` | `Bodies.NamedBehaviorArity` | **taken again** |
| `Bodies.BodyForInvariantDischarge` | `Bodies.NamedBehaviorArity` | **taken again** |

The last two are live. Declaring a behavior beside a body lowers that body again and builds its discharge representation again, in a module where nothing that body reads was touched. `LoweredBody`'s own documentation says "editing another body in the same module does not expand this one again", which is true of a body and false of a declaration.

## Done when

- The census is taken from the store's graph rather than from a reading of the sources, and is what the check runs on.
- Every edge in it is one of the three, said where the check can read it — a projection, an answer equal under a sibling edit, or a violation.
- The two violations are cut, each held by a behavior declared beside the body leaving the answer alone.
- An edge that is none of the three fails the build, so the next one arrives as a red test rather than as a slow edit.

Ordering matters: the classification comes first and the check is what the classification becomes. A check written first has nothing to say but "this edge exists", and the only way to make it pass today would be to admit the two violations as legal.

Found while closing #1679 (PR#1681), which is the reader that was paying #1466. Neither of those closes this.

kawasima:
Closes #1682.

The store states three rules about what an answer may be, and the third — an
edge is what its consumer means — was stated in `Db`'s own documentation and
checked by nothing. A breach of it leaves every answer the compiler gives
unchanged, so it reached anybody only as an edit that had become slow.

## What is held

That an edit wider than what a consumer means does not move what the consumer
answers. Projection and fold are two ways that comes about; they are not the
thing held. Read as the test, a reader answering with every entry of an index
is entries of it, reads as a projection, and moves with the index whenever it
moves.

So there is one question per edge:

```
index moved?  no  -> this edit witnesses nothing
              yes -> reader moved? -> the edge is a violation
                                      otherwise -> why it held is read, and written down
```

## The census

`IndexEdges` reads the edges off the graph a compile built, through
`Db.everyAnswer` and `Db.dependenciesOf`. An edge is a candidate where the
reader's key names something its module holds and the key it read names only
the module and answers with a table.

What a key holds at a component decides that, and there is no default, because
the two halves want opposite ones. A reader holding a component nobody has read
is safer read as naming something; an index holding one is safer read as saying
which module, since reading it as naming something takes the index out of the
census and every edge into it with it. So an unread component is a third thing:
the census says which it met, and the check fails until somebody writes down
which of the two it is. What somebody wrote down beats what the component
happens to hold — a module and a definition of it may be spelled alike.

## The edits

An index that does not move leaves every reader of it alone whatever the reader
means, so every edge is held to having been met under an edit that **moved the
index it reads**, and an edge no edit moves fails as one nobody has put a
question to. Each edit is additive, each says nothing about anything already
written, and each is held to the problems the compiler is expected to have once
it is added. The register carries the edit beside the reason, so an edit that
stops moving an index arrives as a failure rather than as a register that
quietly says less.

An edge that is neither, or that nobody has said what it is, fails the build.

## What it found

`Bodies.LoweredBody` and `Bodies.BodyForInvariantDischarge` read
`Bodies.NamedBehaviorArity`, and `Bodies.CheckedBehavior` read
`Bodies.RecursiveCallSigs` and `Bodies.RecursiveHelperConstructs` — all four
whole. Declaring a behavior beside expanded every body of the module again;
declaring a recursive helper beside checked every behavior of it again. All
four are now projections of what this body's expansion reaches.

What a body reaches is two relations, because two things are asked. What is
**written into** a body stops at a recursion and leaves it out: the recursion is
lowered to a method of its own, so what its body names is named there. What
**stands in** a body stops at a recursion and answers with it: the call is what
is left, and what that recursion constructs already carries what the recursions
it calls construct. Their call half calls the same walk the helper graph is
built from rather than reading the tree a second time.

## How it was checked

Pointing any of the four readers back at its index turns the check red, naming
the reader and the index. Handing a cut the whole index back turns it red as a
reader that moved with its index. Following the written-into walk through a
recursion, or the standing-recursion walk into one, turns the relation tests
red. Leaving a component out of the register turns the unread question red
rather than changing the census silently, and letting a spelling beat the
register turns the reading itself red.

`mvn -o test` is green over the reactor. The population group has the same
failures on this branch as on `develop`, in the same tests and the same
methods.


kawasima:
A self-check of the diff turned up three things, all now in `7e01ed6`.

**The cut was wrong.** The claim it rested on — a helper may not name a behavior, so expanding one
into a body brings no behavior name the body has not got — I took from `CalleeSigsForBody`'s
documentation without checking it. Only a helper with parameters is refused. A value may name a
behavior and nothing reports it:

```
let held = twice
let through (f: (Int) -> Int, n: Int) = f(n)

behavior viaValue : (n: Int) -> Int
let viaValue (n) = through(held, n)
```

`twice` reaches `viaValue`'s tree by substitution, so its arity is wanted there. Under the first
cut it was left standing as a name where `develop` makes it a block — the emitted tree differed
for a program that compiles clean. The walk now follows every name that reaches a declaration the
table holds, and the closure of those, and the policy is part of the question because it decides
which definitions a body reaches. `TheAritiesABodyAsksForAreTheBehaviorsItsExpansionReachesTest`
pins it.

**The census was under-counting.** `anIndex` required a key with one component, so an index that
also takes a policy was invisible to it. `Bodies.CheckedBehavior -> Bodies.RecursiveCallSigs` is
one, and it is sound, but nothing was looking at it.

**The two accounts were compared as text.** `EverythingAnAnswerHoldsMeansSomethingTest` documents
why not to, in the same package, and I did it anyway. They are compared as edges now and rendered
only for the failure.

What the census does not say, now written where the word is: a reader answering with every entry
of an index is entries of it and reads as a projection. Handing the whole index back leaves the
census green — the test beside it is what refuses that.

`Db` and `Key` said things the change made false; both now name the check.

kawasima:
Both blockers are one root, and it is one I half-found in my own self-check and then failed to sweep.

**The root.** Each of the two sound forms can be satisfied without the edge being exercised. A
projection is satisfied by an answer holding nothing, because nothing is entries of every index
there is. An answer equal under a sibling edit is satisfied by an index that never moved, because
an index that stays where it was leaves every reader of it alone whatever the reader means. I put a
witness on the first and read the second off `Objects.equals(index, after)` — which, as you say,
proves the edit and not the edge. Same shape, one arm guarded.

So the fix is symmetric rather than two patches. An edge is now held to having been met under an
edit that **moved the index it reads**, and an edge no edit moves fails as one nobody has put a
question to. That took the single edit apart into seven, each additive and each about nothing
already written: a behavior declared beside, one taking a requirement, one stating a rule, a data
declaration, a row, a recursive helper, and a declaration that cannot be built. The register now
carries the edit beside the verdict, so an edit that stops moving an index arrives as a failure
rather than as a register that quietly says less.

**Your #2 was right, and it found more than `ReqSigs`.** With an edit that actually moves it,
`CheckedBehavior -> ReqSigs` comes out sound — `ReqSigs` moves, `CheckedBehavior` folds it in, and
its answer is still equal, which is the stated form. What was hiding behind the single edit was two
others:

```
Bodies.CheckedBehavior -> Bodies.RecursiveCallSigs          neither
Bodies.CheckedBehavior -> Bodies.RecursiveHelperConstructs  neither
```

A recursive helper declared anywhere in the module checked every behavior of it again. Both are cut
now, as projections of what this body's expansion reaches.

**#1 fixed as you describe.** Map against map is `entrySet().containsAll`, keys and all. A reader
answering with a single thing is still asked against the index's entries — which entry it took is
not something the census can see, and what that arm claims is only that recomputing the reader is a
lookup. Reader-collection against index-collection is no longer called a projection at all, since
nothing in the graph is that shape and `containsAll` would be dropping multiplicity for free.

**#3 taken, and it changed the default rather than the table.** The unit is now
`(Key class, component)`. The important half is which way an unjudged component falls: read as
naming something it puts its key among the readers, so its index reads are counted and have to be
judged; read as saying which module it leaves the census without a word. So unjudged now means
narrower, the register is only the components somebody had to excuse, and a new component makes the
check ask more rather than less. The register's one obligation is that no line in it is stale.

**One thing your review clears that I had already found false.** The E1818 premise — a helper may
not name a behavior, so expanding one brings in no behavior name the body has not got — is not
true for values. `let held = twice` compiles clean and carries the behavior in by substitution, and
under my first cut the emitted tree differed from `develop`'s. That was fixed in `7e01ed6` before
this review landed. What a body reaches is now one question asked once (`Bodies.ReachedByBody`), and
its call half calls the same walk the helper graph is built from rather than reading the tree a
second time — a first attempt at the recursion cut missed library operations left standing for
exactly that reason, and the suite said so.

`mvn -o test` is green over the reactor.

kawasima:
All three, with the root of each named first. #1 and #2 turned out to be the same shape from two
directions, and both were live.

## #1 — one word, two jobs, opposite safe sides

The root is not the missing `UNKNOWN`. It is that `narrows()` settled two questions whose safe
defaults are opposite, so any default it took was quietly wrong about one of them. Reader side:
unread → naming something → its index reads enter the census → stricter. Index side: unread →
naming something → the index itself leaves the census, and every edge into it with it → laxer. A
boolean cannot hold that, which is exactly your reading.

So there is no default. `WhatAComponentHolds` is now `THE_MODULE / SOMETHING_THE_MODULE_HOLDS /
HOW_TO_READ_IT / UNREAD`, the census reports every component it met that nobody has read, and the
check fails on it. The register is per `(Key class, component)`.

Reading the ones this compiler's graph actually holds was not a formality. Fifteen keys about one
declaration were being read as being about their module, and two of those answer with a table —
`Shapes.FieldBindingsOf` and `Shapes.EffectiveFieldTypesOf` — so they were being counted as
indexes. That is your failure mode, already present, hidden by the boolean:

```
an edge no edit here moves the index of ==> but was:
  [Adequacy.InputsOf -> Shapes.EffectiveFieldTypesOf,
   Adequacy.InputsOf -> Shapes.FieldBindingsOf,
   Bodies.CheckedBehavior -> Shapes.FieldBindingsOf]
```

Once every component is read, the census comes back to what it was — the classification did not
move, only the silence did.

A stale line is now also one that names a `String` component, since those are settled against the
key's own module and no word written beside one is ever asked for.

## #2 — one relation answering two questions

Correct, and the root is the word. What an expansion **writes into** a body and what a body
**reaches** are different relations: a recursion is left standing and lowered to a method of its
own, so what its body names is written into that method; what it constructs is attributed to
whoever called it, and so is what the recursions it calls construct. I had the second answering
both. `Bodies.WrittenIntoBody` stops at a recursion; `Bodies.ReachedByBody` goes through it.

Your regression as stated cannot be written — a helper may not reach a behavior at all, and E1818
is asked of the tree the expansion made, so `let carried = thrice` inside a recursion is refused
too. So it is pinned two ways instead. A helper reached only through a recursion is in
`ReachedByBody(caller)` and not in `WrittenIntoBody(caller)`; removing the stop turns that red. And
the law that makes the behavior case unobservable is now a check rather than a sentence — the day a
helper may reach a behavior, the walk through a recursion starts answering about names that are not
in that tree, and the failure arrives there.

## #3 — the branch is gone

You are right that the head and my comment disagreed; the comment described what I had decided and
not what I had written. A reader answering with a collection off a collection index is no longer
called a projection at all. What containment says about two collections depends on which collection
they are, nothing in this graph is that shape, and a word written before there is something to name
would be a word for whichever of `Set`, multiset or sequence somebody meant.

## PR body

Rewritten. The E1818 sentence you flagged was already false — a value carries a behavior in, which
my earlier self-check found and `7e01ed6` fixed — and it is out.

`mvn -o test` is green over the reactor at `1e0a68a`.

kawasima:
All three. #1 is the one that reframes the rest, and you put it better than the PR did.

## #1 — the classification was standing where the invariant belongs

The root is an inversion, not a missing case. What this holds is that an edit wider than what a
consumer means does not move what the consumer answers. Projection and fold are two ways that comes
about. I had the shape as the gate and the invariant as the consequence, so a shape that can be had
while the invariant fails went through — a reader answering with every entry is entries of the
index, reads as a projection, and moves with it.

Now there is one question per edge and it is the invariant:

```
index moved?  no  -> this edit witnesses nothing
              yes -> reader moved? -> violation
                                      otherwise -> projection ? A_PROJECTION : EQUAL_UNDER_EDIT
```

The census catches the whole-index cut itself now. Mutating `BehaviorAritiesForBody` to hand the
index back:

```
a question about one definition took its module's index whole ==>
  [Bodies.BehaviorAritiesForBody -> Bodies.NamedBehaviorArity, at BehaviorArities…]
```

which was green before.

Two things fell out of the inversion. The separate witness test is gone — `A_PROJECTION` now means
"entries of it, and some", so the vacuity guard is the classification rather than a check beside it.
And the register gained a second word on five edges, because an instance whose answer is empty held
without projecting anything: `BehaviorAritiesForBody` at a body that names no behavior is equal,
at one that names a behavior is a projection. Both are true of different instances and the register
is a set, so it says both.

## #2 — three relations, and the third has no consumer

Your reading of `SpecChecker` and `DataChecker.collectConstructs` is right, and the constructs index
already answering transitively is what settles it: one entry carries the chain, so the caller wants
the recursions its own tree calls and no more. `ReachedByBody` is replaced by
`StandingRecursionsOfBody` rather than kept beside it — the transitive relation has no consumer, and
a third word for a meaning nobody asks for is its own kind of debt.

Pinned on `caller -> spin -> deeper -> deep`:

- `WrittenIntoBody(caller)` = ∅
- `StandingRecursionsOfBody(caller)` = `{spin}` — not `deeper`
- `StandingRecursionsOfBody(spin)` = `{deeper, spin}` — its own self-call stands in its method too,
  and typing that call wants its own signature

Walking into a recursion turns the second red.

## #3 — the last default is gone

Right, and it is the same failure as the policy one with the type changed. `String` no longer means
"a name" by being `String`. The one thing settled mechanically is that a component holding the name
the key says it is about *is* that module; any other text goes to the register and is `UNREAD`
until somebody writes it down. That added fifteen lines for `#behavior` and `#fn` components, which
is the semantic classification this census was already asking for. Dropping one turns the unread
question red and takes that edge out of the census's account, visibly.

`mvn -o test` green over the reactor at `9cbae29`; CI running.

kawasima:
Fixed, plus the two non-blockers and one of the same shape I found sweeping for it.

## The blocker — an inference was overruling the judgement

The root is a precedence inversion, and it is the one the last change was supposed to remove. The
model says a role is the component's; the reading asked the text first and the register second. So
an inference about what a `String` means overruled what somebody had written down, and the two come
apart exactly where a module and a definition of it are spelled alike.

Explicit first now:

```
said?  yes -> that
       no  -> String && value == key.module() ? THE_MODULE : UNREAD
```

Your point that this cannot be caught by the unread question is why the regression is on the
reading rather than on the census: `roleAt(new Bodies.CheckedBehavior("orders", "orders"),
"behavior")` must be `SOMETHING_THE_MODULE_HOLDS`, and the module component must still be
`THE_MODULE`. Mutating the precedence back turns that red and, incidentally, empties the census —
`a census of 0 edges is not this compiler's graph` — because `orders`/`orders` is not the only
place a spelling collides once the register is ignored.

## Sweeping for the same shape found one more

A category standing in for a property is the shape you have now named twice, so I looked for the
rest of it in my own diff. `Edit` carried a boolean "the compiler is expected to speak". The
property is narrower: it says *these* problems, about the definition the edit adds. An edit that
broke something already written would satisfy the boolean, and every verdict taken under it would
be a verdict about a module whose meaning the edit had moved. It now carries the codes.

## `eitherWay`

You are right, and the honest fix was the name rather than the mechanism. `WhatItIs` is `WhyItHeld`:
what the register holds is an observation — this edit moved this index, the reader held, and here is
why it held that time — not a nature the edge has. An edge standing at a behavior that names none
and at one that names a behavior is seen both ways, and `metEitherWay` says that is what happened
rather than claiming the edge is two things.

## PR body

Rewritten against the implementation. `ReachedByBody` is gone from it, the invariant-first ordering
is what the "what is held" section now says, and the two relations are described as what stops at a
recursion and what stands in the body.

`mvn -o test` green over the reactor at `aa519b4`; CI running.

kawasima:
Done. The root is one type carrying two units, and you named both halves of it.

**The unit of what an edit did and the unit of what somebody knows are not the same.** The register
is keyed by the kind of question and the kind of index, and that reason still holds — a judgement is
about a question, and keyed by the key it would be a register of a fixture's behaviors. What was
wrong is that the same key was counting what an edit did, and there the fold is exactly the reading
this check exists to refuse: the edge that was asked standing as the witness for the edge that was
not.

`At(Key, Key)` is what a census observes; `Edge(Class, Class)` is what gets written down. `everyEdge`
and `exercised` are `At`, the difference is taken there, and the register is unchanged.

Your policy case is the one that makes it concrete, and it is prospective rather than live — the
graph today reaches `RecursiveCallSigs` only under `FULL`, so no shape currently spans two. That it
is prospective is why it wanted the type rather than an extra edit: the fixture cannot be relied on
to keep the two apart.

The assertion now reports edges, and it has teeth at that level. Making
`A_RECURSIVE_HELPER_BESIDE` add a behavior instead of a recursive helper:

```
an edge no edit here moves the index of ==>
  [Bodies.RecursiveCallSigsForBody -> Bodies.RecursiveCallSigs,
     at RecursiveCallSigsForBody[module=shop.orders, behavior=codeOf],
     reading RecursiveCallSigs[name=shop.orders, policy=FULL],
   … one line per definition it stands at]
```

I swept the rest of the diff for the same shape — a fact about one thing quantified over a fold of
it — and this was the only place left. `neither` was already per edge, `whereTheIndexMoved` folds to
the shape because why an edge held is the knowledge unit, and `unread` folds to
`(class, component)` by union, so any instance nobody has read is reported.

`mvn -o test` green over the reactor at `cbf4f0e`; CI running.

Issue #1730, opened 2026-09-16
Population contracts are red on develop for interaction evidence and uncovered findings


kawasima:
The tests tagged `population` are red on `develop`, and nothing a change waits on runs them — so the
run that reads them is the nightly one.

## What is observed

    mvn -o test -Dgroups=population -Dtest.excluded.groups=

fails on `develop` at `aec297a63`, in a clean tree with nothing else in it. Two groups of sentences
come back, and they may or may not be one fault.

**A part of a behavior's evidence does not say whether it holds a handle.**

    a part of a behavior's evidence has not said whether it holds a handle for a rule it read:
    class souther.compiler.query.InteractionEvidence

from `EveryPartOfABehaviorsEvidenceIsAskedForItsHandlesTest`, in every one of its subjects: the part
asked on its own, the whole asked for what its parts hold, and what the union would be short of
without the asking.

**A finding about an obligation joins to no entry of its account.**

    a finding about an obligation joins to it ==> expected: <[]> but was:
    <[出荷する: a pair_uncovered finding names an obligation,
      送料を出す: a interaction_uncovered finding names an obligation, ...]>

from `EveryFindingAboutAnObligationJoinsToItsAccountTest`, over findings of both `pair_uncovered`
and `interaction_uncovered`. `EverythingAnAnswerHoldsMeansSomethingTest` and
`EveryQuestionThisCompilerDeclaresIsReachedOrOutsideABatchRunTest` are red in the same run.

## Why it matters

These are not snapshots of an expected text. What
`EveryPartOfABehaviorsEvidenceIsAskedForItsHandlesTest` asks is that a kind of evidence this
compiler did not think of is refused rather than read as holding nothing, which is the check that
keeps the oracle from sharing the omission it is watching for. Red here says the contract it states
is not held, and the reading of every finding that joins to an obligation is what a reader of a
report acts on.

## Done when

The run above is green on `develop`, and each of the two sentences above is answered by what its
check is about rather than by an expectation written to match.

Whether this is one cause or two is not known. Split this if the root cause of the handle contract
and the root cause of the join turn out to be different.

PR #1742, opened 2026-09-16
Make the population contracts green again

kawasima:
Closes #1730.

The tests tagged `population` were red on `develop`, and nothing a change waits
on runs them, so the nightly is the only reader. What had collected there is
several separate faults, not one: each is a mechanism added without the check
that watches for it being asked again.

**A part of a behavior's evidence did not say whether it holds a handle.** The
evidence gained a part — which combinations of the decisions its body settles a
value by the rows were seen making — and the oracle that holds the whole to the
union of its parts refuses a shape it does not know rather than reading it as
holding nothing. So it stopped on every subject it has.

**A finding about an obligation joined to no entry of its account.** Whether a
finding names something a row is owed for is the subject's own answer, and the
document writes the key wherever it is one; the check read a list of kinds
written beside that. So the kinds added since were findings it walked past while
saying every one of them joins. The population is now the findings carrying a
key, and which kinds may carry one is asked of the kind.

**A place in an answer held something that means nothing by equals.** The
reading of a body's inputs is now asked of one behavior as well as of the
module, and the answer is one domain rather than a domain apiece — so everything
under it is met by a second name the register did not know.

**Questions were declared and reached by nothing.** Two are facts an editor asks
about a buffer, which no batch answer rests on, and each is written down with its
own reason. The other two are ordinary report paths over an ordinary model —
where an author is sent when a rule arrives and the reading stops on it, and
where a condition met on the way to a border point stands. Every source of the
corpus is written so that what it says is read, which leaves both unreached, so
the operation is run over a model stating a rule this compiler has no words for.

**No timed edit settled a choice.** The edits that add a definition leave every
declaration where it was, so no figure in the edit series answered for what
settling a choice costs. The choice is now reached the way the relation is, by
rewriting the rule that states it.

Two of these were the sentences the issue names; the rest were in the same run
and are behind the same thing, so they are here rather than split out.

`mvn -o test -Dgroups=population -Dtest.excluded.groups=` and `mvn -o test` are
both green over every module.


kawasima:
Taken. The sweep this check was written to do is the one it now states, and not
the one about kinds.

`everyFindingAboutAnObligationNamesOneEntryOfItsAccount` is now
`everyObligationIdentityThatIsPublishedJoinsToOneEntryOfItsAccount`, and the
class javadoc says outright what is not held and why: whether something is owed
has an answer per kind and an answer per subject, they are at different grains,
and a check holding a kind to publishing an identity would stand green on a
proposition already known to be false.

Measured before deciding. Restoring the universal direction goes red today on an
output case of one of the models this check compiles — `About.ACaseNoRowExpects`
is a plain `About` and the document publishes nothing under `signature.output`
for it to name. `About.AnUnansweredRow` is the same shape and the corpus has not
reached one.

That is #1743, and it names both subjects rather than either alone, with the
second classifier going away as the Done-when. The tag placement is #1744.

PR #1752, opened 2026-09-17
Ask where a clause is written where a reader is pointed at it

kawasima:
`Refs #1472`

The invariant check built its diagnostics as it ran, so checking a body resolved where every clause it judged is written. What the body means and where a reader is sent to find the rule it was judged against were one answer — and an edit that moves a rule without changing a word of it cannot leave both right: keeping the body keeps the caret where it was, moving the caret checks the body again.

What the check finds is now `InvariantFinding`, which places nothing: which type was being built, where in this body, and what was proved of each clause it is held to. `Bodies.InvariantWarnings` turns those into reports and is what asks where each clause is written — the reader that is about to point somewhere, which is what `ClauseLocations` already says of itself. It is asked beside the other warning-only questions, so the command line and the editor say the same thing.

A refusal is still said where it is met: it is what stops the check, and a body with no meaning to emit leaves no answer for a later reader to point from. That case reads a place, and what it costs is re-checking a body nothing compiles.

## What holds now

The cross-module acceptance in `AReportAboutADeclarationFollowsItWhenItMovesTest` asks all of it at once, over a workspace where an importer builds an imported type and is warned about its rule. The declaring module's two declarations are written in the other order — a comment above them would move nothing, since a place is which of the things written in a text it is.

- the body judged against the rule is not judged again
- the warning is still about the construction it was about
- the caret under the rule is on the line the rule is on now

Both caret claims were watched go red on a mutation that does not move the rule, and the first on the code before this change.

`WhereAClauseIsWrittenIsAskedByWhatPointsAtItTest` holds the boundary itself rather than the names: it walks what checking a body actually read, however far down, and asks whether an answer about where a clause is written is among it. The warning about the same construction is the comparison beside it. Both go red when the check resolves a place again.

`WhatAClauseMeansAndWhereItIsWrittenAreTwoAnswersTest` pinned that edge on the body check, which was the reader that pointed when it was written. The law it states is unchanged; the subject moved, and it now holds both halves.


kawasima:
Self-check over my own diff: four places were left where they fell rather than fixed.

**The finding listed which verdicts it is made for.** `Verdict.holds()` already answers that, and its javadoc says what is reported is settled at `report` — so the list was a second reading of the enum, and the one that fell behind would make a finding nothing can say or refuse one somebody has to. Asked of the verdict now.

**A judgment's sequence was outside its equality.** The clauses are held in the order they were declared; a report names them in it and labels their places in it. A map's equality is about entries and not about sequence, so two judgments that read out differently compared equal — which did not matter while a judgment was in no answer, and does now that one is part of what an answer about a body is kept by. The safe direction is the one taken: a sequence that differs is a report that differs.

**Which of a module's behaviors have a checked body was read in two places.** I shared half of it and left the other half — both the module check and the warnings worked out that a module whose own check stopped has no checked bodies. One reading answers it now, and both ask it.

**A claim about the warnings was untested.** The reporting question walks the bodies, so a body whose check did not come out has no findings to read. Nothing held that this is not read as "only where every body came out" — a warning quietly disappearing beside an error in the behavior under it. `AWarningIsSaidWhateverTheBodyBesideItCameToTest` holds it, names the warning rather than counting it, and goes red on the gate I had considered and rejected.

Also: a test helper hid an assertion inside the expression it was reading, and now reads the way the one beside it does.

Reactor and population green.

kawasima:
P2 is fixed. The ordering half went in after CI ran on `281412ca`, as part of a self-check (`f3972b156`) — but the defensive copy did not, and that is the other half of the cause.

## The cause

Not that `Judgment.equals` was wrong. **A record's own equality and its own constructor are right exactly when every component is a value**, and a `SequencedMap` is not one: its equality is about entries and says nothing about sequence, and its constructor keeps whatever collection the caller was writing into. Hand-writing `equals` on `Judgment` closed one symptom, left the other, and left the next component to the same trap.

So what is fixed is the component. `ClauseJudgments` holds the clauses judged at one construction in the order they were declared; `Judgment` is back to its generated `equals`/`hashCode`. The ordered comparison comes for free from a `List`, and the copy happens once, where the value is made.

What the map gave beside its order was that a clause appears once. That is a thing to refuse rather than a thing to get from a key: two answers about one clause leave no answer to which of them a report is about.

`WhatAConstructionWasJudgedAgainstIsASequenceTest` holds both directions — another order is not the same judgment, the same order is, with the same `hashCode` — and goes red on a mutation that compares as a set, which is the behaviour the map was giving. The refusal has one of its own.

## Whether anything else has the same cause

The predicate for the population is an answer whose iteration order is observable while its equality does not see it. Sweeping and measuring found `Shapes.FieldBindingsOf` and `Shapes.EffectiveFieldTypesOf`. On a declaration that spreads two others, with the spreads written the other way round:

```text
FieldBindingsOf[Pair]        order changed: true   equal: true
EffectiveFieldTypesOf[Pair]  order changed: true   equal: true
MeaningOf[Pair]                                    equal: false
```

Nothing is wrong today because of the third line: `DeclarationMeaning` holds its includes as a list, so it does see the change, and every reader of the other two also reads it. That guard is written nowhere and nothing holds it.

Not in this PR — it means changing those answers' types and the capability interfaces threaded from them, which would make this unreviewable. Filed as #1753 with the measurement. What should close it is the population rather than the two answers named there, and the Done-when says so.

Reactor and population green.

Issue #1758, opened 2026-09-17
A declaration that spreads its way back to itself ends the compile with nothing said


kawasima:
```
data Pair  = { ...Other, qty: Int }
data Other = { ...Pair,  note: String }
```

ends the compile with a `StackOverflowError` and no diagnostic. Nothing refuses a spread that reaches its own declaration — `DataChecker` has the walk that says a sum contains itself, and there is no such walk for what a product spreads.

More than one reader falls down the same hole, so the fix is not a guard at whichever one was found. `PublishedRules.governing` walks the same graph and does not stop; the expansion the field answers are projections of does stop, and stops because a spread onto a declaration already being expanded is handed back as a back edge rather than followed — which makes it finite and says nothing about what a back edge means. What is missing is the sentence: a declaration cannot be made of itself, said once where declarations are held to their rules, before anything walks the spreads.

A diamond is not this and must go on being what it is: two spreads reaching one declaration are two fields with one name, refused today as a field collision. Cutting on "reached before" rather than "on the way here" would take that refusal away.

## Done when

- a declaration reaching itself through its spreads is refused where its rules are read, with a report naming the spreads it goes round by
- what walks the spreads is total, shown on more than the one walk that was found — the population is whoever walks an include
- a diamond is still refused as a collision, and still names the two spreads that supplied the field

PR #1763, opened 2026-09-18
Refuse a data that spreads its way back to itself

kawasima:
Closes #1758.

A product that spreads its way round to itself ended the compile with a `StackOverflowError` and no diagnostic. Nothing refused it, and the walks over the spreads each decided their own termination — so what happened was whichever of them ran first ran out of stack.

## What was missing

Not a cycle detector. The sentence *a product's spread graph is well founded* existed nowhere, so every walk's `seen` or path set was an incidental stopping condition rather than a reading of the language. Making the field expansion finite moved the order the walks run in, and the next walk with no such device was the one that fell over.

So there are two contracts here, and they are separate.

**Language semantics.** A spread reaching a declaration the walk is currently inside is a declaration that reaches itself, and it is refused in one place, before any derived property of the product is read. `ProductSpreads` reads the resolved declarations and nothing else, keeps the edges as the spreads they are written as, and reports at the spread the ring is entered by, naming the spreads it goes round by (E1027). A declaration reached along two paths is not this — its fields arrive twice, which is the collision it already was.

`Shapes.WellFoundedSpreads` is the answer, and it names the declarations the walk went over rather than saying that a check ran; the count of what a declaration's values are and the elaboration of its rules both depend on it. Neither can be gated at the pass that holds a declaration to its rules, because both are asked before it.

**Traversal mechanics.** Every walk that reads a declaration's spreads recursively now cuts at the path it is on, so it comes back on a graph nobody validated. None of them reports the ring. The premise that an invalid graph cannot reach a given reader is a property of the query graph and not of the walk, and this change is what it looks like to stop writing walks that rest on it.

## What is held

A data spreading itself, a pair of them, a ring of three; two spreads meeting one declaration still refused as a field written twice; a chain written as it reads. That the readers which walk the spreads are not started at all on a ring. And walks of both lineages — the trees a world holds, and the declarations their modules publish — handed an invalid graph directly, each with a chain it does reach through so that coming back is not a walk that answered without crossing a spread.

Every claim above was watched going red under a mutation: removing either path cut, treating a declaration reached before as a ring, and dropping either gate.

## Left out

Folding this together with the sum's own cycle rule, a general graph traversal, and an API that can only be called with a validated graph. Same shape, different meaning, and the point of this change is the line between the mechanics and the semantics.


kawasima:
Fixed. The cause was one thing: the path cut was written as a condition on the iteration, so the edge it dropped left no mark on the answer being built. A cut is an arm of the walk, which is what the expansion over the same graph already had in `FieldExpansion.Include.BackEdge` — an arm has to say what it contributes, and a filter does not.

Swept by carrier rather than by the sites named. Of the walks the cut was added to, the ones whose answer carries a completeness component are `PublishedRules` and `ExpandedRules`, and both now take a spread onto the path as no clause and rules not reached. The rest — the settled clauses, `fieldType`, `hasField` and the resolved field types an editor reads — carry the clauses or a field and nothing beside them, and each says in its own doc why there is nothing to lower and where a reader asks instead.

On the order dependency: which partial answer a ring gives still turns on which of its declarations was asked first, and what keeps that from reaching a reader is that every one of them now says it is short. `InvariantChecker` already records `ClausesNotExpanded` and stops reading on that, and `whole()` refuses.

The direct-reader test now asks three lineages rather than two — what the declarations publish, what a world's expansion states, and the trees a world holds — and reads what each claims of the ring rather than only that it came back. Both completeness assertions were watched going red under the mutation that puts the old silent skip back.

The `WellFoundedSpreads` comment now says the first spread of the ring, written on the declaration it closes on, which is what the code points at.

kawasima:
Fixed, and the cause was mine to own: I removed that branch on the grounds that its population was empty, and I took the population from one reproducer I wrote — two modules importing each other — instead of from the question. The question is whether a walk started in one module can close a ring written in another, and a one-way import answers it: the ring is inside one module, but the path that reaches it is not. Reproduced the duplicate exactly as you described before fixing it.

Which module owns a ring is now asked of the ring (`ReachesItself.writtenIn`) rather than spelled at the report. Every reader that finds one has the same question and a ring has one answer to it, so there is one place to change if the language ever lets a ring span modules. A module that only reached one still goes absent, so its readers are stopped as surely as by a ring of its own — the ownership is about the sentence, not about the graph.

Both shapes are checked now. The mutual-import one stays, because it is what makes "the whole of a ring is written in one module" true rather than assumed; the new one is yours — `b` writes the ring, `a` spreads `b.Pair`, and it pins the single sentence, the position in `b`, and both queries absent. Putting the unconditional report back gives two identical `ADataSpreadsItself` reports, which is the red I watched before restoring it.

Full local run is green; CI is running.

Issue #1618, opened 2026-09-12
What a declaration publishes cannot say whether it was not declared or could not be read


kawasima:
`PublishedDeclarations.of` answers what a declaration says, or null. Two opposite facts arrive as
that one null: nothing in the compilation declares the name, and a declaration that is there whose
module could not be read. A reader that has to tell them apart cannot, and asks something else.

The reading of what rules govern a value is the reader that has to. A declaration nothing declares
carries no rule to be short of; one whose module nobody could read carries whatever its author wrote,
and a reading that reported the second as the first would say a position admits everything and say
nothing about not having looked. So it asks the world beside the publication:

```java
DeclarationMeaning said = published.of(named.key());
if (!(said instanceof DeclarationMeaning.Product product)) {
    return new PublishedRules(List.of(),
            said != null || symbols.declaredNode(named) == null);
}
```

That second question is answered out of the authored declarations, which is the thing the walk was
moved off. It is a faithful translation of what the walk over the expanded trees did — that one had
the distinction, because the lookup of a declaration's expanded clauses is total: it answers found,
unavailable, or not declared, and says which.

So the publication is the one of the two boundaries that lost a case. The reading is left holding a
world for one bit of information, and every future reader of what a declaration publishes will meet
the same absence and reach for the same second authority.

## What to do

Make the published answer total, the way the expanded one already is — found, unavailable, not
declared — and the world drops out of the walk. What the other readers of it want is the found arm,
which they can take for themselves.

## Done when

- What a declaration publishes says which of the three it is, and no reader flattens two of them into
  an absence.
- The walk over what rules govern a value reads no world, and takes what it needs from one authority.
- A check holds that a declaration whose module could not be read leaves the rules of a value short,
  and that one nothing declares does not — through the publication alone.

kawasima:
Done in #1766, merged into develop.

What a declaration publishes says which of the three it is, and no reader turns two of them into an absence: a name no module writes and a declaration whose form writes no `invariant` leave the rules of a value whole, and one whose module could not be read leaves them short. `CheckedDeclarations` refuses the two absences separately rather than under one sentence.

The walk over which rules govern a value reads no world. What a declaration spreads, what it states, and whether there is a declaration at all all come back from the publication, and a licence over the compiled walk holds every type of this compiler it names, so a second authority handed over as a narrower carrier of the same question adds a row.

[...]

PR #1766, opened 2026-09-18
Say which of the ways a declaration has nothing to publish it is

kawasima:
Closes #1618.

`PublishedDeclarations.of` answered what a declaration says, or nothing. Two opposite facts arrived as that one nothing — no module writes the name, and a declaration is there whose module could not be read — and only one of them leaves the rules of a value short. The walk over which rules govern a value has to tell them apart, so it asked the world beside the publication, which is where the declarations are written.

So `Shapes.MeaningOf` answers `PublishedDeclarationResult` and says which of the three it is, and `PublishedRules.governing` no longer takes a world. Every other reader of what a declaration publishes wants the found arm and takes it for itself.

## Telling the two kinds of nothing apart

The split needs to know whether a module writes the declaration at all, and that question is settled by the parse. Everything above the parse is a judgement about the module — the namespace it claims, its `exposing` line, its imports, whether it settles — and every judgement is an answer that can go absent. Read for the first question, each of them says the declarations that module writes are declarations nobody wrote, which is the widening the three answers exist to remove, arriving one rung down.

Two authorities were tried and both were below a judgement, so it is read off what the source was parsed as (`Front.somethingDeclares`). It stops there: a source that did not parse has no declarations to enumerate, which is the absence of something to read rather than a judgement about a module. The lookup of expanded clauses had the same split and the same defect, and is moved with it, so a reader meeting either boundary is told the same about which kind of nothing it has.

## What the three arms cost a reader

A name no module writes and a declaration whose form writes no `invariant` both leave the rules whole — there was no rule of it to be short of, and nothing it spreads to be short of either. A declaration whose module could not be read leaves them short, which is the widening a reading already has a word for. `CheckedDeclarations` refuses the two absences separately rather than under one sentence.

Two test doubles were standing in for the unavailable case with a null and now say what they mean.

## Evidence

Every arm is reached from a real workspace, measured before the checks were written. What a broken module costs is usually answered for one declaration at a time — a clause that does not type, a field naming no type, a body that does not check — so none of those reaches anything but found. What reaches the unavailable arm is answered for the module: a value defined in terms of itself, imports that form a ring, a namespace this compilation may not write.

`WhatIsPublishedTellsADeclarationNothingDeclaresFromOneItCouldNotReadTest` holds what the arms cost, through the publication, with the same declaration in a module that reads as the control. The ring case is read from a module outside the ring: both modules of one are cut and neither has a reading, but a third that imports one of them does read, and its rows were held to no rule before this.

`TheTwoBoundariesTellUnavailableFromNotDeclaredAlikeTest` holds the division the two boundaries share, over a module that does not settle, one in a ring, and one refused by its own name — the last two being the ends of the range of judgements the answer had to be lifted above. Beside them it holds what they are **not** held to: a form with no rule to write has clauses to hand over whatever became of its module, so found need not agree.

`TheWalkOverWhatDeclarationsPublishReadsNoWorldTest` is a licence over the compiled walk: every type of this compiler it names is written down. Handing it a world, or the narrower interface that answers the same question, adds a row and reddens it. What it does not hold is written down beside it.

`mvn -o test` is green.


kawasima:
A self-check over the diff found three things written from the neighbouring answer rather than from what this one does, and they are fixed in the commit above.

**The reason carried with an unavailable answer did nothing.** `Answer.of(Unavailable, mine.reports())` was copied from the expanded side, where that answer is the only carrier a report has. Here what a module was told about itself already arrives by its own path: measured with and without, the same reports come back either way. The argument is gone, and so is the sentence claiming it mattered.

**Both `Unavailable` arms claimed a case they do not cover.** The doc said a declaration whose module could not be read covers one whose imports form a ring. Measured, a ring gives `NotDeclared` — on both sides. Whether a name resolves to a declaration is read from the resolution a ring stops, so the declarations of a module in one are answered for as declarations there are none of. That sentence was already on `ExpandedClauseResult` and was copied from there, so both are corrected.

What keeps that from being a value silently held to nothing is that a compilation in that state has no reading of the importing module either. The load-bearing part is that the two boundaries agree, so `TheTwoBoundariesPutADeclarationInTheSameArmTest` holds them to dividing the cases by one question, over the workspaces that reach each arm. Answering the division from a second question reddens two of its cases in opposite directions.

**The refusal for a name nothing declares gave a reason nothing established.** Replaced with what the class already says: which shape to hand over is asked by a declarations own identity, so a name with no declaration arriving there is this compiler handing over an identity for a declaration it does not have.

`mvn -o test` is green.

kawasima:
Both findings are fixed, and the first was worse than the review could see from the head it read.

## The blocker, and its cause

You are right that `ResolvedDeclaration` is not the authority for whether a declaration exists, and the path you traced is the one the compilation takes: `Names.Declarations` returns absent for a module in a ring, above resolution, so everything below it answers that the module writes nothing.

The cause is that the three-way answer needs an *existence* authority and I took it from a query that answers whether resolution reached a declaration. Existence and reachability were being answered by one value — which is the defect this issue is about, one rung down. Making the answer total in shape did not make it total in content.

So the sweep is by that predicate rather than by the site you named. Of the readers of `ClausesExpandedFor.declarationOf`, two decide existence and feed what a value is held to: `MeaningOf` and `ClausesExpandedFor` itself, which had the identical fallback and is the boundary my new law compares against. The other four ask where a declaration is written, and for a module that could not be read, absent is the right answer — there is no caret to offer. Both of the first two are moved; the other four are left.

The existence question is now `Front.somethingDeclares`, read off what the modules were parsed as. `Front.DeclaredTypes` is above the cut, measured: in a ring it still answers `[Amount, Held]` while `Names.Declarations` and `ResolvedDeclaration` are both absent.

**It was observable, and my own note saying otherwise was wrong.** I had written that a compilation in that state has no reading of the importing module either. That was measured on the module inside the ring only. A third module that imports one of them is outside the cut and does read: before the fix its rules came back whole, after it they come back short. `WhatIsPublishedTells...` now holds that case with the third module, and putting the old discriminator back reddens it along with the ring case in the boundary law.

## What I had done to that case in the meantime

Worth saying plainly, because it is the worse half. Between your review and this reply I ran a self-check, found the ring behaviour, and wrote it up as intended — I corrected the javadoc to match the code and added `TheTwoBoundariesPutADeclarationInTheSameArmTest` asserting `NotDeclared` on both sides. That turned a defect into a contract. The doc of `Unavailable` had named the ring as one of its cases all along, and I edited the doc to fit the implementation instead of reading it as the check it was. Both are corrected: the arms say what they meant, and the law now asserts `Unavailable`.

## The architecture test

Also right, and the mutation you proposed reddens the replacement. `NameSense` carries `declares(TypeKey)` and was not among the worlds, so the old rule could be walked around by asking a narrower carrier for the same thing.

It is now written as what the walk may read rather than as what it may not: every type of this compiler the walk names is a row, read from the constant pool including member descriptors, and from the descriptors and signatures of its methods and fields. Handing it a `NameSense` adds a row and reddens it. That also closes the getter-shaped hole you mentioned — a carriers type is named in the descriptor of whatever hands it over.

## The reports argument

Already gone, in the commit before this one, for the reason you give plus a measurement: with and without it the same reports come back, because a modules own diagnostic arrives by its own path. The comment claiming otherwise went with it.

`mvn -o test` is green.

kawasima:
Right on both counts, and the counterexample is real — measured before changing anything.

```
unit    | published=Unavailable | expanded=Found
sum     | published=Unavailable | expanded=Found
product | published=Unavailable | expanded=Unavailable
```

My law passed only because every fixture I wrote happened to be a product.

## Cause

I wrote the check from the sentence I wanted to say rather than from what the code decides. What the code decides by one predicate is the split between the two kinds of nothing; whether either boundary has an answer at all is each one's own question, and `ClausesExpandedFor` answers `Found` for a form with no `invariant` to write without ever asking whether the module could be read — correctly, because such a declaration has no clauses whatever became of its module.

That is the third time in this branch I have stated a claim wider than the mechanism behind it, so it is worth naming: the two earlier ones were a check whose two arms could not disagree, and a case I called unreachable after measuring one shape of it.

## What changed

`TheTwoBoundariesPutADeclarationInTheSameArmTest` is gone. `TheTwoBoundariesTellUnavailableFromNotDeclaredAlikeTest` holds only the division they share — a module that could not be read and one in a ring are unavailable to both, a name no module writes is undeclared to both — and it carries your unit counterexample as a case of its own, asserting `Unavailable` against `Found`, so the narrower rule cannot be read as the wider one again. Putting the old discriminator back reddens the ring case there and the shortfall case in the other test.

The stale sentence about resolving is replaced wherever it appeared: the question is whether a module writes the declaration, asked of what the modules were parsed as. The arm's own doc now says what the two boundaries share and, beside it, what they are not held to.

## CI

The red on the previous head was mine and unrelated to the argument: Error Prone refused an import left unused when I rewrote the scan. It is removed. CI runs `verify`, which has Error Prone where a local `test` does not, so I swept the touched files for the same shape by hand.

`mvn -o test` is green; CI is running on `6ac1253`.

kawasima:
Confirmed and fixed. The counterexample reproduces exactly as you describe: a source writing `Amount` under a reserved namespace was answered for as a declaration nobody wrote, by both boundaries.

## Root cause

I had been moving the authority up one cut at a time instead of putting it at the rung where the question is settled. Whether a module writes a declaration is the parse's answer. Everything above the parse is a judgement *about the module* — the namespace it claims, its `exposing` line, its imports, whether it settles — and every judgement is an answer that can go absent. Read for existence, each one erases what the author wrote. Resolution was the first I hit; the exposing check was the next; there was no reason to expect it to be the last.

So `Front.somethingDeclares` now reads `Layout` for which source declares the module and `Parsed` for what that source writes, and a module off the path is answered for out of what was read back. It stops at the parse, and that stop is a different kind: a source that did not parse has no declarations to enumerate, which is the absence of something to read rather than a judgement about a module.

## The sweep

By the predicate rather than by the case you named, over every judgement between the two ends:

```
a namespace this compilation may not write   Unavailable
an exposing line naming what is not there    Found
a name written twice                         Found
imports that form a ring                     Unavailable
a module that does not settle                Unavailable
a source that does not parse                 NotDeclared
nothing wrong                                Found
```

Both boundaries agree on every row. Nothing above the parse moves existence any more; the only `NotDeclared` left is where there is nothing to read or nothing of that name.

`ClausesExpandedFor` shares the helper, so it moved with it.

The regression is in `TheTwoBoundariesTellUnavailableFromNotDeclaredAlikeTest` as a case of its own, written for what it is: a ring is a judgement about what a module imports and a refused namespace is a judgement about its own name, so they are the two ends, and an answer above both is only above them where it is read off the parse. Reading existence from the exposing check instead reddens it.

I did not widen `DeclaredTypes`, for the reason you give.

## The architecture test's limit

You are right that a second authority wrapped in a predicate over addresses adds no row, and I have overstated a check three times in this branch already, so the javadoc now says what it does not hold rather than leaving the stronger reading available.

## PR body

Rewritten to the current design. The two stale sentences you quoted are gone, along with the rest of the running commentary on how it got here.

`mvn -o test` is green; CI is running on `4a80748`.

Issue #1760, opened 2026-09-17
Nothing counts the answers whose iteration order is observed but whose equality is not


kawasima:
The third of #1753's Done conditions: whatever holds the class holds it of the population rather than of the answers that were named. What was found there was found by reading, and the next one would be found the same way.

The class is not a type. It is a pair — what a value's equality sees, and what a reader takes off it:

```
carrier equality          ×          downstream observation
```

A carrier is anything an answer is built from or carried in whose equality does not see an order: `Map`, `Set`, a record with one of those as a component, a `SequencedMap` (its equality is still about entries). An observation is a reader that turns an iteration into something positional — a list, an argument order, a descriptor, a report that names the first of several, an index, a serialization. Where the two meet on one value, an edit that moves only the order leaves the answer equal and the reader goes on saying what it said.

Order erasure ends the trail rather than clearing it. Where a value passes through `Map.copyOf`, `Set.copyOf` or a hash, the iteration order stops being anything at all — the runtime salts it, so it is not the same run to run. A reader observing an order after that is a heavier defect than this one and belongs in its own count: a compile has no justified order there to be stale about.

## Already found by reading, unfixed

`observe.FieldTypes` answers with a map and says in its own javadoc that the fields of one are an order as much as a set, and that what comes back is in the order a value is laid out. It is not a question the store answers, so a census that enumerates the store's questions will not reach it.

`check.Shape.Product` holds its fields in a `SequencedMap` and says the order is part of the answer, which is true — a reader of it lists the fields and another composes a value field by field. It is a record, so its equality is the components', and a map's equality is about entries: two products whose fields stand in different orders are one value. What reaches it is `TypeView`, and what reads it is in the partitioner.

## Done when

- the class is stated as the pair rather than as a list of types, and the count is taken by that predicate over the whole compiler
- what the count found is disposed of one way or the other: the order made part of the equality, the observation moved to a question that answers an order, or the reading shown to be commutative
- a value that has passed through an erasure is counted separately, and any reader observing an order after one is reported as the other defect

kawasima:
How the class is kept out once it has been counted, which is a question the count raises and does not answer.

Prose does not keep it out. Every carrier found so far already says in its own javadoc that its order is not part of what it answers, or that it is; the readers that disagreed with it read anyway. So the disposition of a counted instance is one of three things, and only the last of them holds:

[...]

PR #1769, opened 2026-09-18
A reading carries one work order however it was assembled

kawasima:
One finding out of #1760's census, and the walk it is about.

A reading's alternatives are asked one at a time, each block of each alternative put to a
question, and the walk stops as soon as what it holds cannot be moved. Those questions share one
allowance: a compile builds a meter once and hands it to everything asked under it, so a question
reached after the allowance has run out is answered "could not tell" rather than answered.

Which questions are reached while there is still something left was decided by the order the
alternatives and the blocks happened to be held in — and that order is the writing. The same
reading, written the other way round, came back undecided where the first one came back answered.

The algebra was never the problem and is untouched. A meet is commutative, and the walk stops
only at the element nothing can move, so what the walk computes is the same either way. What
carried the writing into the answer was beside the algebra: the allowance the questions share.
Leaving it alone and weakening what an undecided answer does to a meet would have been a fix to
the wrong thing — an undecided answer really is one this could not tell, and a join that ignored
it would say a reading stands where nothing established that it does.

So the allowance stays shared and the walk still stops early. What changes is that the order the
work is done in is read off the reading rather than off how it was written down, the way
`PlanOrder` already decides what a plan builds first. The keys are built from a block's name and
from sets and descriptions the walk is holding anyway — ordering the work never asks for the
machines the work exists to arrange — and each key is built once per walk rather than once per
comparison.

All four walks that spend one allowance go through the same order. Fixing only the one the
witness came through would have left the same defect reachable from the refusal side, against the
same meter.

## What is held

`AReadingsAnswerDoesNotFollowWhichAlternativeSpentTheSharedAllowanceFirstTest` takes the same
reading written each way round and holds both what it answers and what it spent answering, under
an allowance with room to spare and under one that runs short.

The cost half is not decoration. With the ordering removed, the short-allowance case comes back
with two different answers — and the ample-allowance case, where both spellings answer the same,
comes back having spent different amounts on different questions. That is the same defect one
step further in: the next question asked under that allowance would be the one that could not be
told. Both halves were watched failing before the fix and passing after it.

## What this does not close

#1760 is about a class of values, not this one walk, and the census behind it is recorded on the
issue. This is the first finding disposed of out of it.


kawasima:
The blocker was right, and finding it settled where the contract belongs.

`PlannedValues.refusedInEveryAlternativeAt` walked its alternatives in arrival order with the same
early return, against the same meter. Counting the walks over these carriers rather than the ones
named, there was a second: `AdmissibleValues.alternatives`, which walks alternatives while spending
from an `Allowance`. Four of six.

That is the cause rather than the slip. Nothing at a walk, at the carrier or at the allowance says
a walk spends one, so the order was right only where a reader remembered to ask. A convention
applied by hand is right where the hand went, and the two it missed are why it was the wrong shape
of fix — a walk written later would have started wrong by default, and nothing would have said so.

So the carrier holds it now. A reading's alternatives, and an alternative's blocks, are put in the
order the work over them is done where the reading is put together, and every walk gets it by
construction. The four reader-side orderings are gone; there is no `inOrder` call left outside
`PlanOrder`. Neither of the two walks the review names is fixed as such — neither can be written
wrongly any more.

`tellApart` moves to that boundary and gets the stronger form you asked for: not that subjects are
written apart, but that two unequal things a walk must do one before the other never share a key.
A tie there is the arrival order coming back at the one place that exists to stop it, so it is
refused where the order is made rather than assumed in each helper. The double key building goes
with it — the key is built once per value now, not once per walk.

## On the witnesses

The carrier-level one holds what the contract now is: the same reading, written each way round,
hands out its alternatives in one order. It discriminates — removing the canonicalization turns it
red along with both walk-level cases.

The refusal-side one I could not write, and I would rather say so than leave a green that says
nothing. Walked on its own, `refusedInEveryAlternativeAt` stops at the first alternative nothing
refuses, so what it spends is capped whichever way it went; walked after the admission on one
allowance, as a compile does, both writings came back the same in every shape I could build,
including alternatives of unequal size. I wrote it, watched it stay green under the mutation that
turns the other three red, and took it out.

That leaves the refusal walk covered by the carrier invariant and not by a witness of its own,
which is the thing the new shape is supposed to buy: the walk cannot get the order wrong because it
is not the walk's to get. It is a weaker guarantee than a red-then-green on that path, and it is
the guarantee that is actually available.

## Cost

Canonicalizing at construction rather than at each walk costs a little on the packages that
exercise these carriers — a few percent on a test-package run, which is an instrument dominated by
startup and by everything else in the suite. It is not a clean measurement and I am not claiming a
number from it. If construction turns out to be hotter than walking, the same contract can be kept
and paid for on the first walk instead; that is a change behind the carrier and would not move
anything a reader sees.

`Allowance#spent` keeps its own set of blocks "in the order they were found" and is read. Its
producer is now a canonical walk, so the order is derived rather than arbitrary — but that is a
different carrier with a different reader, and it belongs in #1760's census rather than here.

Issue #1674, opened 2026-09-14
A stepless order leaves a chooser only a run-derived candidate


kawasima:
A search for where a position may stand takes its first candidate from the run the arithmetic leaves the number, and puts it to the rest of the question afterwards. Two things can refuse such a candidate and neither of them is in the run: a set the declarations leave the position, and a place a rule holds the position away from. Where the order steps, being refused costs a step and the walk takes the next value. Where the order has no step, the one place in hand is the whole of what was offered.

## What is observed

In the vouchers corpus the denied side of a string equality is a rule of the decision that nothing composes a row for. The page says nothing could show a row can be written there and that every value tried was refused; the requirement beside it is unsettled, because nothing was composed to try.

The values that satisfy the rule are every string but one. Nothing about the model makes them hard to reach.

## What is happening

`NumericWitness` starts from `carrier.somethingInside(runs.min(), runs.max())`, which reads the ends of the run and nothing else. On an order whose values do not step, the walk outward from that place ends at once: there is no next place to step to, so the candidate offered is the only candidate.

Whichever of the two refusals then applies leaves the position with nothing. Both are reachable.

* The declarations refuse it. This is what vouchers does: the run is unbounded, so the place offered is the empty string, and the `Voucher` invariant about its length refuses it. Measured — with the length rule weakened, the same rule is composed for and the requirement settles.
* The region refuses it, because the place is a hole a disequality left. Reached where a rule bounds the run inclusively at a place another rule holds the position away from.

An earlier reading of this issue attributed the vouchers failure to the second. That is a real way to reach the same state and it is not the one this corpus takes. Both come of the same omission: the candidate was taken from the run alone.

## Direction

The crossing already exists. `Carrier.somethingOtherThan` narrows by the run, the set and the places held apart together rather than judging what came back, and builds a machine over the set once the ends of the run are exhausted — which is what `LevelRealizer` is already under.

What is missing is that the region answers which places it holds a position away from. It carries them for the proof that nothing is left and does not offer them to whoever is choosing a value, so a chooser has no way to ask before it offers.

Comes out of #1650, whose shapes are otherwise closed.

kawasima:
Closed by #1777, merged into `develop`.

The chooser now takes every candidate from the run, the values the declarations leave the position, and the places a rule holds it away from together, rather than from the run alone. In the vouchers corpus the decision rule on the denied side of the string equality moves from unsettled to required, and the finding that no row covers it appears with it.

PR #1777, opened 2026-09-18
Choose a place from everything that narrows it, and not from the run alone

kawasima:
Closes #1674.

## What was happening

`NumericWitness` took the first candidate for a position off the run the
arithmetic leaves the number, and let the rest of the question refuse it
afterwards. Two things can refuse such a candidate and neither is in the run: a
set the declarations leave the position, and a place a rule holds the position
away from.

Where the order steps, a refusal costs a step and the walk takes the next
value. Where the order has no step, `Outwards` ends `WITH_NO_STEP_TO_TAKE` with
the one place in hand — so whichever of the two refuses it leaves the position
with nothing offered, and the combination is reported as one every candidate
was refused for.

In the vouchers corpus that is the denied side of the string equality. The
candidate offered is the empty string, which the run gives up when nothing
bounds it; the `Voucher` invariant about its length refuses it; and the
decision rule beside it stays unsettled with nothing composed to try. The issue
attributed the refusal to the region's hole. The hole is a real way to reach
the same failure and this corpus does not take it — both come of the same
omission, which is the candidate having been taken from the run alone.

## What changed

The crossing is asked of the carrier, which narrows by the run, the set and the
places held apart together instead of judging what came back, and builds a
machine over the set once the ends of the run are exhausted. That is the same
arrangement the realizer is already under.

A region answers which places it holds a position away from, beside the proof
that nothing is left. The two are not one answer: one refuses a row already
standing at the place, the other is what keeps a chooser from writing one
there. It is an abstract method rather than one with an answer of its own, so
an implementation cannot leave the new half out quietly.

What the search is given beside the region is the existing `WitnessSearch`,
which pairs the sets with an allowance per crossing. Handed the sets alone, a
search would reach for a budget nothing granted it.

A position no set was worked out for is narrowed by the run and the holes and
by nothing else. What this search answers is whether the region admits an
assignment, not whether the declarations can be shown to admit the value; a
name every case of a sum spreads has no position in the reading, and deciding
here that nothing can be placed there would take the word for a row nobody
writes away from the construction that owns it.

What a set is keyed by moved to one place. A place composed for a number taken
of a position is not a value of it — a length counts a string — so the set is
asked only where the number is the position's own.

## What says so

Two fixtures at the layer that decides, one per refusal, over the same order
and the same run, with a control that the value offered is still one the run
holds. Mutating the candidate back to the run's own ends fails both; dropping
only the holes fails the one about the hole. The control stays green through
both.

The conformance answer for the vouchers decision rule moves from unsettled to
required, and the finding that the rule is uncovered appears with it — the
corpus writes no row for it.


kawasima:
Fixed, and the sweep found the same omission at a second site.

## The cause, rather than the seed

Both searches took a place from `somethingInside(run)` and handed it to `Outwards`, which walks the run. `Carrier.somethingOtherThan` is the crossing of all three narrowings, but `Outwards` had no way of being told about anything but the run — so a caller that wanted an intersected stream could only pre-filter, which is impossible for a generated stream, or post-filter, which `Carrier.admitted` documents as the thing not to do. Seeding from the crossing left the rest of the walk exactly where it was.

So the narrowings now arrive at the walk together and a caller cannot ask for one that leaves them out. A place the run holds and one of the others refuses is stepped past: not the run running out, so the walk continues, and not a place to try, so it is not yielded.

`every candidate ∈ run ∩ admitted ∖ apart` now holds, on stepping orders as well.

## What the figure counts

It bounds the places looked at rather than the values yielded, and the parameter says so. Counted the other way, a run with no end whose values are all refused is a walk nothing stops — a `Finite` set far from the seed reaches that. The ending reported is the one the arm already named, "a place the run holds was found and not taken", which stays exactly true and keeps the reader's action right: raising the figure does look further.

Where a caller passes the identity of the crossing, the places looked at and the values yielded coincide, so nothing moved for the callers that have no sets to give.

## The second site

`LevelRealizer.alongTheLine`, the pair search on a line, had the same shape — `somethingInside(together)` then `Outwards.from`, with the region and the sets in scope and unread. Its walk is over the places the anchored position stands at, so what the declarations leave that position and what the region holds it away from are what narrow it. Fixed.

The walk over a progression is not given the sets at all, so it passes the identity of that crossing and the comment says which narrowing it is without rather than leaving it to be read as an answer about the values.

## Checks

The walk's own test owns the endings, so the stepping-past behaviour is fixed there: one case per narrowing, asserting the exact places yielded and that the run still ran out. Reverting the step to the run alone fails both.

No conformance answer moved, which is what I would expect — the corpus reaches the gap only through the stepless order, and that half was already closed.

`ReadQuantities.apartFrom` said the proof that nothing is left was the whole of what a hole changes. Corrected to name both readers, matching `SearchRegion.apartFrom`.

The issue body was rewritten before this comment landed — it already attributes the vouchers refusal to the declarations and names the hole as the other way to the same state.

kawasima:
Both fixed at the cause. Head is `f360ae1`.

## One place was holding two answers

What a search does where nothing worked out a position's set is not one answer — it depends on what the search is for. A search asked whether the region admits an assignment narrows by nothing it does not know; a search that writes the value it finds into a row has nothing to compose from. Both are right, and they were being decided at each call site with an inline `instanceof` and whatever else-branch the author reached for. That is how one class came to hold both.

So each answer is named once, on `WitnessSearch`:

- `toNarrowBy(term)` — the set to narrow a search by, `ANY` where nothing worked one out, and the doc says that is the identity of the crossing rather than an answer about the position.
- `toComposeFrom(term)` — the set to compose a value out of, or null where nothing worked one out.

`valuesAt` is private now, so a call site picks one of the two names and cannot invent a third answer. `NumericWitness` takes the first; `placeMeeting` and `alongTheLine` take the second.

Fixture: a pair with one side's set unworked-out composes nothing, both ways round, with a positive control that the same pair is composed when both sets are there. Reverting `alongTheLine` to `toNarrowBy` gives `Found` where `Unknown` is expected — the exact scenario you described.

## One number was doing two jobs

Right that this is not naming. Before the filter, place inspected = candidate yielded = candidate tried, so one figure covered all three. The filter split them and I reused a figure for a meaning its own contract denies.

Split now. `Outwards` takes both figures and `Ended` has a fourth arm saying which one stopped it — `AT_THE_FIGURE_OF_CANDIDATES` and `AT_THE_FIGURE_OF_PLACES_LOOKED_AT`. Each caller maps each arm to its own budget word, so `VALUES_A_POSITION_ON_THE_WAY_IS_TRIED_AT` keeps meaning values tried and the walking has `PLACES_A_POSITION_ON_THE_WAY_IS_LOOKED_AT` beside it; the same for the pair. The progression walk is given no set and no holes, so its two figures are one number and it names one word for both arms, with a comment saying why.

The exhaustive switches made most of this a compile error rather than a choice — the word a reader meets, the word a search comes back with, and the canonical order all had to be written before it built.

Fixture: a walk whose run refuses every place but the first ends at the looking figure; the same walk with nothing refused ends at the candidate figure. The endings test is renamed for the fourth arm, since the count was in its name.

## Note

`Outwards.Ended` now has four arms and the two limits that were already there are untouched, so `WITH_NO_STEP_TO_TAKE` still travels as a population rather than a figure.

Full suite green, no conformance answer moved.

kawasima:
Right, and the rule I broke is written in that file.

## The cause

`Outwards` already states it, for the figure that was there:

> A value found and not taken, never a count that came out even. A run holding exactly this many and a run this stopped walking come back the same length, so the figure being reached says nothing on its own — what says this compiler declined to go further is a value the run holds that this did not take.

I added a second figure and read it off the count, which is that trade made for the other number. Not an off-by-one to nudge: the two checks were asking different questions, and only one of them was the question the file had already settled.

## The fix

Finding another place of the run and examining it are two things, and the allowance is for the second. So the looking figure is asked before its allowance is spent:

```java
took = true;
if (lookedAt == howManyLookedAt) {          // a place to examine, nothing to examine it with
    ended = AT_THE_FIGURE_OF_PLACES_LOOKED_AT;
    break outward;
}
lookedAt++;
if (!takenIn(next, ...)) {
    continue;
}
if (out.size() == howManyTaken) {           // a place to take, and this is not taking it
    ended = AT_THE_FIGURE_OF_CANDIDATES;
    break outward;
}
out.add(next);
```

Both figures now have the same shape, and the comment says that is why.

## The sweep

The rule is one rule and each figure is a place it has to hold, so it is pinned for both: a run exactly as long as the allowance ends `HAVING_TRIED_THEM_ALL`, a run one place longer ends at that figure — written twice, once per figure. The candidate half had the behaviour but no fixture at the boundary, so it could have drifted the same way.

Moving the increment back before the check fails the looking half with exactly your example, `expected: <HAVING_TRIED_THEM_ALL> but was: <AT_THE_FIGURE_OF_PLACES_LOOKED_AT>`.

Full suite green, no conformance answer moved. Head is `b5e74b9`.

kawasima:
Blocker reproduced, fixed at the cause, and the rule that would have caught it now exists. Head is `6844a97`.

## What the blocker was

`Realization.Unknown`'s constructor holds the walk's word and the budgets it met to each other, and says why: the word is the walk's answer and the figures travel beside it. Splitting a figure must therefore not change what its walk says — a figure split in two that takes a second word makes one walk say two things depending on which of its own numbers ran out first, and the answer carrying both cannot be assembled.

I filed `PLACES_A_PAIR_IS_LOOKED_AT` by what the figure means in the abstract instead of by what its walk answers. Filed with its twin now.

The fixture came first and fails exactly as you described, through the route an author reaches — an anchored position whose declarations leave it one value, on a line nothing bounds:

```
java.lang.IllegalArgumentException: a walk stopped by [PLACES_A_PAIR_IS_LOOKED_AT]
    does not come back with NOTHING_COMPOSED_ONE
```

## Same root, checked rather than remembered

I added two figures and got one of them right, which is the shape of a rule nothing enforces. The other new figure sits in the no-word group beside *its* twin, so both pairs are consistent — but only because I happened to.

So the pairing is data: `CompositionBudget.splitFrom()` says which figure a figure was split off, over an exhaustive switch, so a figure added has to answer. `AFigureSplitOffAnotherComesBackWithTheSameWordTest` walks whatever that states and holds each pair to one word, with a negative control that the figures are not all one word anyway and that a wordless figure still refuses. Giving the split figure its own word fails the structural check and the end-to-end fixture, at two levels.

Why the unit tests missed it: `Outwards`' own tests never assemble a `Realization`, so the walk was right and the answer built from it was unbuildable. That path is held end to end now.

## The contracts the split left behind

- `AdmittedValues` no longer states what a search does — it reports the state, and the two answers live where a search is handed what it searches with. Its third state also now says it covers a name that is no position of the reading at all, not only a reading that stopped, since that is what the shared sum name measured.
- `PLACES_A_PAIR_IS_TRIED_AT` and `HOW_MANY_PLACES_A_PAIR_IS_TRIED_AT` no longer claim that stepping past holes is what they bound.
- `Outwards.Walked` says four endings; the `howManyPlaces` reference is gone.
- The budget audit's descriptions for both `<clinit>` sites name the looking figure.

Full suite green. CI is running; I'm watching it in the background this time.

Issue #1778, opened 2026-09-18
One word stands for a position nothing read and for a position that is not this reading's


kawasima:
`AdmittedValues.Admitted.NotWorkedOut` is what a lookup comes back with for any position the sets were not built for. Two different facts reach it, and a reader that has to act differently on them cannot tell which it is holding.

## What is observed

`Partitions.witnessSearch` builds the sets off `measurements()`, keyed by each measured position's path. Anything not in that map is `NotWorkedOut`. So the word covers both:

* a reading that stopped before it reached the position, which is this compiler not having read the rules there;
* a name that is not a position of this reading at all. `r.deadline` where every case of a sum spreads `deadline` is one: the reading goes under each case, so there is no position at the shared name and there is nothing there to have been worked out.

The type's own documentation describes the first and gives an example of the second.

## Why it matters

What a reader may do differs. At the first, composing a value would be offering one at a position whose rules were never read. At the second there are no rules of that position to read, and a value placed there is refused or not by whatever writes it — which for a sum's own name is the construction, and the word it has for that is its own.

`AConditionOverASharedNameIsOneTheComposerCanPlaceTest` states that contract from the other side: the way's position is placed, and the row is refused for what is true of it.

A search that reads the sets to narrow by is not where the difference shows, because neither state narrows anything. It shows where a value is actually composed at the position, which is what the realizer does.

## What would settle it

The third state split in two, and the reading saying which. Telling them apart needs something `measurements()` does not carry today, so this widens what a search is given beside the region — which is why it is not part of #1674.

kawasima:
Closed by #1783, merged into `develop` as b88223092.

The third state is split in two and the reading says which: a path the measurement worked nothing out about, and a path it puts no position at because every case of a sum spreads the name. `NameReach.standingOf` is the one place that sorting happens; what a search narrows by takes a coarse projection of it, and what writes a value reads the cases.

Where the difference shows is where a row writes. A number read at a name the cases share is written under whichever case the row turns out to be, and the target now carries the place rather than the number's own, so a condition over such a name is placed instead of coming back as one nothing could compose a value for. A name under two sums is followed a crossing at a time until it reaches a place that was answered for.

Two things this leaves, each worth its own issue: only the first case the row can still be is tried, so a row writable under a later case is not looked for; and a condition on the way does not move the value a row composes, at an ordinary place as much as at a shared name.

PR #1783, opened 2026-09-18
Send a number read at a name the cases share to where a row writes it

kawasima:
Closes #1778.

## What the word covered

`AdmittedValues.Admitted.NotWorkedOut` was what a lookup came back with for any position the sets were not built for. Two facts reached it: a position this reading stopped short of, and a path the reading puts no position at at all, because every case of a sum spreads the name and the values stand under the cases.

They are told apart now, and the sorting is asked of the walk that saw where each name stands, once. `AdmittedValues` takes a coarse projection of it — a search narrowing by sets needs only whether there is a position here — and the side that writes a value reads the cases.

## Where the difference shows

Not in the sets. Neither state narrows anything, so a search asked whether a region admits an assignment answers alike either way.

It shows where a row is written. Where a number is read and where a row writes to move it were one location in `RealizationTarget`, and they part at a shared name: the rules name the sum's own field and a row writes one of the cases. A row asked to write at the sum's own name wrote nowhere, and a condition over such a name on the way to a point came back as one nothing here could compose a value for — which is this compiler's own word for falling short, said of a place the language never puts a value.

So the number keeps the name the rules and the report are about, and the target carries where the value goes. A name is followed one crossing at a time until it reaches a place this reading answered for, which is how the sorting answers and is what a name under two sums needs. The case at each crossing travels in what the row assumes, so every name of one sum is written under one case however many conditions name them. A name whose case the reading stopped short of routes nowhere, and neither does a place nothing worked out what it holds.

What one condition comes to is arrived at whole and then applied: where its numbers stand, where the row writes them, and what the row had to be taken as are one decision, so a condition nothing could be put under leaves the row the case it had.

## What is checked

A condition over a shared name on the way to a point is placed and the row is written under a case. Put back the way it was — the write root taken from the number's own place — and the row comes back unresolved under the old word.

Beside it, the sorting itself: a shared name, an ordinary name, and a name under two sums, each against what the walk recorded. And that a search is told the values stand under the cases rather than that the declarations leave the position everything; read as the second, the declarations would be answering for a position they were never asked about.

## What is left

A condition on the way does not move the value a row composes, at a shared name or at an ordinary position alike. Measured with a positive control before and after this change, so it is neither this routing's doing nor something the routing fixes. Worth its own issue.

And the case is taken in the order the model declares them. A row that could be written under a later case and not under the first is not looked for, which is the search this does not do.


kawasima:
The writer-side path is closed, and the invariant it would have rested on does not hold.

## What was reachable

A place with no measure of it is `NotWorkedOut` while nothing crosses there, so the sorting calls it `AtThePathItself` — and the traversal that follows a written value reaches a number at it, so a row could be written. A model that does it:

```
data L3 = { held: Int }
data L2 = { under: L3 }
data L1 = { inside: L2 }

behavior read : (o: L1, n: Int) -> Ok | No
let read (o, n) = { guard n > 3 else No; Ok }
```

Nothing divides the chain, so `o.inside.under.held` is a place the measurement worked nothing out about, `standingOf` answers `AtThePathItself`, and the write traversal answers `INT`. With a condition over it on the way, the composer put a value there.

So the finding was real rather than guarded from somewhere else, and there is a fixture for it rather than an argument.

## What decides now

The writer asks what the declarations came to at the place. A set is where the row writes; nothing worked out writes nowhere; a name that crosses is followed to where it stands.

Asked of the place rather than of the composing set, and that is the one departure from the shape the review wrote. What a search composes from carries the rule that a number taken *of* a position is not one of the position's values and is not held to its set — right for what narrows a value, and wrong for where one goes: a count read at a name the cases share would be told a set narrows nothing and sent to write at the sum's own name. Where a row writes is the place's question whichever number of it was asked for.

## The nested name

Already written, since it was a blocker when this branch was last read: a name is followed one crossing at a time until it reaches a place that was answered for, and the target holds the place rather than the one case taken. `q.s.deep` is written at `q@OA.s@IA.deep`, and the number stays the one the rules name.

Say if it should come out and go on its own — it is green and checked either way.

## What each check pins

A condition over a place nothing worked out has no value put under it. The condition is written by the check rather than read off a body, so what it establishes is what the composer does with one; the three lines before it are what say the place is reachable and writable.

A condition over a name two sums down is placed, and the row writes one of the cases all the way down. Written as a claim about the account and not about the row: a construction fills what it has to fill, so a row that is an `OA` proves nothing on its own.

Both come back refused under the change that would undo them: the write root taken from the number's own place, and the second crossing left unfollowed.

Still on the branch for their own issues: the later case is not tried, and a condition on the way does not move the value a row composes — the second at an ordinary place as much as at a name the cases share, measured with a control.

kawasima:
Done, and taken a step further than the staging the review wrote.

## The cause

A condition is placed at every position it names or at none — the places were already inside that, and the two things this branch added were outside it. What put them there was writing them as assignments into the row while walking the condition's numbers, where the existing boundary is the one line below that asks whether every number could be placed.

Swept the rest of what the walk touches per number: the numbers owed, the routes, and the two flags are all the condition's own and go out of scope with it; the region, the row's places and the account are written after the boundary or on the failing side. The two this branch added were the only ones on the wrong side.

## The shape

Local staging keeps it from leaking today and leaves the next edit free to do it again, so what a condition comes to is a value now.

```
Placed.AtAll(standing, routes, assumed)
Placed.AtNone(why)
```

The walk arrives at one of those and the caller applies all of it or none. There is no third arm holding a part of a placing beside a gap, which is the half-application this is about; a leak is not something to remember not to write, it is something that no longer compiles.

## What no check says

No fixture here can tell the two apart, and it is worth saying which rather than adding a green that says nothing.

The case at each crossing is the first the row can still be. A condition that fails after settling a sum leaves the same case the next condition would have chosen anyway, so the answers agree. The routes of a failed condition reach the map a later reader consults, but that reader only looks up numbers the row actually stands at, and a failed condition leaves none.

So it is a leak that has no witness until the later-case search exists — which is the second of the two issues still open on this branch, and the first thing that would make it one.

The whole suite is green. The line in the description that said a name crossing a second sum routes nowhere is now what the code does.

PR #1784, opened 2026-09-18
Record that division answers an exact quotient and that the affine reading keeps it

kawasima:
Two decisions, both `Proposed`, no implementation yet.

## ADR-0116: Division answers an exact Rational quotient

`Int / Int` truncates toward zero and `Decimal / Decimal` rounds to a fixed precision. Neither operand type is closed under division, so keeping the quotient in it needs a loss policy that nothing states — while `Decimal.toInt`, `Decimal.round` and `Decimal.divide` all make the model state one. The coefficient in `guard y < -1 / 2 * x + 30` comes to nought for that reason.

So `/` answers a `Rational` over `Int`, `Decimal` and `Rational` alike, and the result type is read off the operands and the operator: `4 / 2` is a `Rational` too, and constant folding decides no type. Unary minus answers a `Rational` where it is given one.

`Rational` enters arithmetic as an operand, never as a conversion. An operation with one `Rational` operand is exact arithmetic over both; an `Int` still satisfies no `Rational` binding, parameter, field or return, and `Rational.fromInt` / `fromDecimal` are where that is written. The heterogeneous signatures admit a settled pair and determine neither operand, so `let double (x) = x * 2` keeps its `Int` from the homogeneous rule rather than from whichever arm an implementation reaches first — ADR-0066 says the operator's own rule answers that, and this is it. The same holds for `<` and `==`, so `let belowHalf (x) = x < 1 / 2` takes a `Rational` and the `Int` version is written.

`Int` and `Decimal` still do not mix — `1 + 1m` has no meaning — and `1 / 1 + 1m` does: the boundary is crossed only after the source has visibly produced an exact value. Exact arithmetic starts at that subexpression and reaches nothing above it, so `a * 2 / 2` overflows where `a / 2 * 2` does not. Equality and ordering read the same exact value, so `1m == 2 / 2` is true and ADR-0009 is untouched.

What `Rational` answers is settled as a capability and not as a list of places: equality yes, ordering yes, external form no. That arm splitting off the other primitives is what refuses it at a data's field, a newtype's base and a behavior's boundary, while `List<Rational>` and `Map<Rational, V>` stay usable inside a computation. A union asks its members the same question, having answered yes unconditionally while every primitive had an external form.

Changing a value is what needs a policy — narrowing the representation is not. An exact conversion may answer a case where the carrier holds no such value; a lossy one states its rounding. `Decimal.divide` keeps its scale and mode, and `Int.divide` becomes `Int.truncatingDivide`.

A newtype inherits an operation only where it is type-closed over the wrapped type, which is not totality, so `Yen / Int` goes and `Yen + Yen` stays. `Yen / Yen` arrives: the units cancel, and the Rational is what a cancelled quotient is — the type ADR-0047 refused it for want of. The refusals left are three reasons rather than one — an unknown dimension, a surviving dimension whose value no carrier holds, and an undeclared inverse.

Amends ADR-0033, ADR-0047 and ADR-0082 — the last because a numeric fold's element is `Int` or `Decimal` "the two types `+` and `*` are defined for", and this adds a third, closed under both with identities the primitives construct. `[#invariant-discharge-terms]` is named as spec text this changes. `Rational` becomes a reserved spelling and the published-declaration boundary version moves with it.

## ADR-0117: Affine interpretation admits division by a constant and keeps its coefficients exact

Nothing is lost at the affine reading today: `BigDecimal` holds a finite decimal exactly, `AffineForms.composed` refuses `DIV` outright, and `NumericDomain` lifts the form into exact ratios before it reasons — so the third that `3 * x <= 1` puts on a position was never rounded. ADR-0116 is what changes that: `guard y < -1 / 2 * x + 30` has an affine meaning the grammar does not admit, and a coefficient the representation could not hold if it did.

A quotient is affine where its divisor reads as a form with no coefficients and a non-zero constant, so `x / 2`, `x / (1 + 1)` and `x / two` read alike — a name read through as ADR-0111 requires — while `x / y`, `1 / x` and a zero divisor make no form.

Exactness then travels to where a value is written, not to `LinearForm` alone: `AffineReading`, `QuantityKey`, `Cutting`, `LevelSpace` and its levels, `Seam.Scale`, `CutPosition`, `LevelRealizer` and what `BorderQuantity` shows carry the exact scalar, and it becomes an `Int` or a `Decimal` only where a candidate is placed on a carrier. The geometry gets by today because everything written is a finite decimal — a border at a third arrives from `3 * d <= 1` and survives as `CutPosition`'s level over its per — and `d <= 1 / 3` leaves that pair nothing to be a ratio against. A counted affine level is also not a carrier count, which `Level.ACount` already says of itself while holding a `Count(BigDecimal)`: over an `Int`, `1 / 3 * x` takes the values a third apart, so the lattice is over a ratio no decimal writes.

What a generated row carries is unchanged, Rational having no external form: a position takes a value of its own carrier, and a Decimal position whose border falls at a third gets what `3 * d <= 1` already gets rather than a fabricated `0.333…m`. `x < 1 / 3` and `3 * x < 1` draw one border.

Clearing denominators is the alternative twice and refused twice. Over the form: what an expression comes to is not a comparison against nought, so a factor taken from a comparison would reach readers making none. Over the relation entering the partition, where the factor is sound: multiplying through by a denominator that came from a Decimal's scale is the `10^scale` expansion this decision removes, arriving by another route and landing on every coefficient.

The compiler's ratio gives up the name `Rational` and keeps its contract, and `numeric` joins the packages `TheRuntimePackageIsTheBackendsToNameTest` keeps off `souther.runtime` — sharing one implementation with the backend's carrier would make an analysis depend on which backend was linked.

## How these land

ADR-0117 needs ADR-0116's decision to be worth making, and can be implemented first. ADR-0116 is one decision but more than one change, and stays `Proposed` until both halves are in, so that "`Int` exact, `Decimal` as before" is never the written rule.


kawasima:
Every claim in the review checks out against the code. Revised in 2ddab4d.

**ADR-0117's account of the present was wrong, and the `/` rule was missing.** Confirmed: `AffineForms.composed` refuses `DIV` with a comment saying a divide truncates for `Int`, and `NumericDomain.assume` lifts both the coefficients and the constant through `Rational.of`, which is exact on a finite decimal. So the reading loses nothing today and the third derived from `3 * x <= 1` was never rounded. The Context now says that, and the ADR is no longer claimed to stand without ADR-0116 — what ADR-0116 does is leave the affine grammar unable to admit an expression that has an affine meaning, and unable to hold its coefficient if it did.

The rule added is the one proposed: a quotient is affine where the divisor *reads as a form with no coefficients* and a non-zero constant, so `x / 2`, `x / (1 + 1)` and `x / two` read alike and `x / y`, `1 / x` make no form. A zero divisor makes none either — the operation answers nothing, so there is no value for a form to be about.

**ADR-0066.** Taken as stated. ADR-0066 says what an operator asks of an operand is answered where that operator's rule is stated, so this is the place. The heterogeneous signatures admit a settled pair and determine neither operand; an open operand is determined by the homogeneous signature alone. `let double (x) = x * 2` keeps its `Int`; `let scale (x) = 1 / 2 * x` gets `x : Rational` from `Rational * Rational` and not from an overload preference; `let scale (x: Int) = 1 / 2 * x` is how the heterogeneous operation is asked for.

**Capabilities.** Written as `equality yes, ordering yes, external form no`, with one correction of mechanism: `TypeOps.Requires` holds `EQUALITY` and `EXTERNAL_FORM` only — ordering is deliberately not a row there, being a witness that `Ordering` holds. The ADR now names the arm that has to split: every current primitive answers yes to both from one `case Type.Prim _`, so adding the new primitive and nothing else would hand it an external form. Collections stay homogeneous, and that is stated.

**Exact vs lossy narrowing.** Agreed and weakened. The rule is about changing a value, not about narrowing a representation, so an exact conversion may answer a case where the carrier holds no such value, and only a lossy one names a policy. Nothing here fixes the library surface.

**Alternatives.** Added to ADR-0116, with keeping `/` and adding exact division under another name first. Also added the denominator-clearing alternative to ADR-0117, refused for the reason given: a form is what an expression comes to rather than a comparison against nought, so a factor taken from a comparison would reach readers making none, and the reading and the constraint step would hold two accounts of one expression — which is what ADR-0111 removed.

**Where the naming argument lives.** Moved to ADR-0117, and it has a precise basis: `TheRuntimePackageIsTheBackendsToNameTest` refuses `souther.runtime` from the areas that reason about what a declaration is. One correction — `numeric` is in neither set that test names, so the rule does not currently reach the package the exact ratio lives in. ADR-0117 adds it to the target-neutral set, which is what makes the separation enforced rather than intended.

**Type-closed.** Stated as the static result type being the wrapped type, with a line saying it is not totality: `Int + Int` is type-closed and still overflows.

**Entering the exact domain is subexpression-local.** In ADR-0116's Consequences, with `a * 2 / 2` against `a / 2 * 2`.

kawasima:
All five verified against the code and fixed in 53c666f.

**Unary minus.** Confirmed — `Elaborator`'s `Hir.Neg` arm refuses anything but `Type.INT` and `Type.DECIMAL`, so `-r` would have failed while `0 - r` compiled. `-Rational -> Rational` is now in the Decision beside the division signatures, with the reason stated as the spelling asymmetry rather than as completeness. The library surface over a Rational stays with the library.

**Reserved spelling.** Confirmed — `TypeScope` reserves a primitive's spelling (E1502). In the Consequences now, with the migration note: nothing in this repository or the example corpus declares `data Rational`.

**Boundary version.** Confirmed — `Backend.BOUNDARY_VERSION` is what `ModuleReadback` checks before it reads a published header. A line in the Consequences says it moves when Rational becomes writable in a helper signature, for the reason ADR-0076 moved it.

**The resource hazard is older than the semantic one.** Right, and the two were run together. `Rational.of(BigDecimal)` does build `BigInteger.TEN.pow(scale)`. ADR-0117's Context now separates them: the semantic need follows ADR-0116, the eager power of ten is already there, and the representation rule would be worth having even if the affine reading never changed. The rule's own section names it as what goes.

**PR body.** Rewritten — both sentences you quoted were left over from the first draft and said the opposite of the current text.

One thing worth flagging, unrelated to the review: this working directory had `git pull origin develop` run against it between reviews, which rebased the branch and rewrote the two pushed commits. The content was identical, so I force-pushed the rebased history with a lease on the old tip after checking that `docs/adr` on the remote was already contained in it. That is also what the `mergeable: false` was — the branch was still based on the develop from before #1781, and the README table row conflicted. It reads `MERGEABLE` now.

kawasima:
All three closed in 3ada2f4, plus the union note as an implementation requirement.

**`N / N`.** Taken. Confirmed the current shape: `ArithmeticCheck` reaches `ln != null && rn != null` and refuses `DIV` outright, the `lt.equals(rt)` distinction existing only for `ADD`/`SUB`. The old reason — the quotient is a value in neither newtype — was true while nothing could hold a dimensionless exact number, and ADR-0116 is what supplies one. So `N / N -> Rational` where both operands are the same newtype: not the newtype unwrapping, but the units cancelling, with the Rational being what a cancelled quotient is.

The ADR now separates the three refusals that stay, which had been one reason between them:

- `Amount / Quantity` — the dimension is unknown rather than cancelled
- `Yen / Int` — the dimension survives and the value is a Rational no `Yen` holds
- `Int / Yen` — an inverse dimension the model has not declared

The middle one is where this meets ADR-0047: scalar division was inherited *because* it preserved the dimension, and it goes for not being type-closed, whether or not the dimension survives.

**ADR-0082.** Confirmed the wording — "The element MUST be `Int` or `Decimal`", justified as "the two types `+` and `*` are defined for", which is the sentence this decision falsifies. Rational is admitted as an element of both. It is closed under `+` and `*`, and the identities are exact nought and one, which the primitives construct — no seed for an author to write and no literal rule involved. The empty-list rule and the newtype-element refusal are untouched, so `let total: Rational = List.sum([])` is answered the way the other two already are. ADR-0116's status line amends ADR-0082 now, since leaving it would have two accepted ADRs disagreeing about what a numeric type is.

**Comparison and equality inference.** Right that the section stated the rule only over the arithmetic signatures above it. It now says it covers every heterogeneous numeric operator the decision defines, with `let belowHalf (x) = x < 1 / 2` and `let isHalf (x) = x == 1 / 2` both answering `x : Rational`, and the `Int` version written.

**Union external form.** Added as a requirement in the capability section rather than left to the implementation. `TypeOps.answers` does answer `EXTERNAL_FORM` unconditionally for `Type.Union`, which held while every primitive had one; the arm descends into the members the way the list, set, option and map arms already do. Otherwise `hasExternalForm` says yes about a type the boundary refuses, and the capability the ADR states is not one `TypeOps` holds.

kawasima:
Right, and the chain is as described. Closed in abe9d8b.

Verified every step: `AffineReading(LinearForm<NumericTerm> form, BigDecimal cut, …)`, `QuantityKey(Map<NumericTerm, BigDecimal> direction)` with its `per`, `Cutting.per`, `Seam.Scale(BigDecimal per, Carrier onto)`, `CutPosition(Level written, BigDecimal per)`, and the coefficient maps `LevelRealizer` and `BorderQuantity` work in. Changing `LinearForm` alone would have cut it one step in.

The part worth writing down is the one you identified: the geometry gets by today because everything written in a model is a finite decimal, so a border at a third arrives only from something like `3 * d <= 1` and survives as `CutPosition`'s pair — a level over a per, exact because it is a ratio of two written numbers. `d <= 1 / 3` makes the written number itself a third, and there is no second number for it to be a ratio against. That is the sentence the ADR was missing, and it is now in the Decision:

> An exact affine scalar stays exact through the partition geometry, and becomes an Int or a Decimal only where a candidate is placed on a carrier.

I took the option you recommend rather than clearing denominators at the border. The border-level version is added to Alternatives and distinguished from the one already refused — that one rewrote `LinearForm`, where a factor from a comparison has no business; this one rewrites only the relation, where the factor is sound. It is refused for cost: a coefficient that came from a Decimal of a large scale has a denominator the size of that scale, so multiplying through is the `10^scale` expansion this decision removes, arriving by another route and landing on every coefficient of the relation instead of on one embedding. It would also leave the geometry holding a scaled relation while the reading holds the form, which is the two-accounts problem again.

What reaches a row is stated as unchanged, with the reason being the capability rather than a rule about generation: Rational answers no external form, so a row is written with `FixtureTemplate.integer` / `decimal` on the position's own carrier. The four models are written into the ADR as what generation has to keep answering, plus the pair you suggested — `x < 1 / 3` and `3 * x < 1` drawing one border, which is the witness that `QuantityKey`'s direction-not-size reading survived exact arithmetic. The Decimal one carries the explicit obligation not to fabricate `0.333…m` as a point on the line.

kawasima:
Added in 1a52773. Right that the chain has a layer between the cut and the reading, and that it is the easiest place to leave a `BigDecimal` in the middle.

Verified: `BorderQuantity.levels()` builds a `LevelSpace`, whose `steppingBy` and `overFiniteDecimals` both take a `BigDecimal` generator, and a counted level is `Level.ACount(Count at)` over `Count(BigDecimal at)`.

The sentence that makes this a decision rather than a list is the one you gave: **a counted affine level is not a carrier count.** `Level.ACount`'s own javadoc already says it — "a number the quantity itself counts to, which is on no coordinate's carrier" — while `Count` is defined as the number a value counts to *on* a carrier's order. The two shared a representation only while every level a form could take was a finite decimal, and `guard 1 / 3 * x < 2` over an `Int` ends that: the quantity takes the values a third apart, so the lattice itself is over a ratio no decimal writes. The ADR states that `LevelSpace`, its intervals, generators and comparisons carry the exact scalar and that a counted level does too, and leaves the shape — `ACount` holding the ratio, or a level that separates the carrier case from the exact one — to the implementation. What it forbids is a `Count` of a `BigDecimal` there.

`AdditiveImage` is noted as the reason nothing mathematical is open: it takes `Map<A, Rational>`, computes its generator with `Rational.gcd`, and answers `OverWholeNumbers` / `OverFiniteDecimals` in exact ratios. Only the representation handed to it was behind.

`guard 1 / 3 * x < 2` is in the generation fixtures, with the distinction spelled out: the first two put a fractional cut on a whole-numbered lattice, this one makes the lattice fractional, and a geometry carrying only its cut exactly would pass those and fail this.

PR body rewritten to the current text — it was missing the ADR-0082 amendment, the union arm, `Yen / Yen`, and the generation geometry. Merging once the build is green.

Issue #1785, opened 2026-09-18
The compiler's exact ratio holds the name the language is about to take


kawasima:
ADR-0116 introduces a language type named `Rational`. `souther.compiler.numeric.Rational` already holds that name, and its contract is the opposite one — its own javadoc says it is "not a number a model writes", and it is deliberately kept apart from `Place` so that a third never reaches a row.

ADR-0117 decides that the compiler's concept keeps its meaning and gives up the name.

## What to do

Rename `souther.compiler.numeric.Rational` to `ExactRatio`, along with `Rational.of`, `Rational.gcd` and the other entry points, in `NumericDomain`, `AdditiveImage`, `RationalCut`, `AffineConstraint`, `CanonicalForm` and everything else in `numeric` that names it. Nothing about its behaviour changes.

`RationalCut` is worth a look while renaming: it is named after the type, not after the language's Rational, so whether it follows the rename is a judgement about what reads best rather than a rule.

## Why the two are not one implementation

They hold the same mathematical domain, so sharing is the obvious question. `TheRuntimePackageIsTheBackendsToNameTest` is the answer: it refuses any mention of `souther.runtime` from the areas that reason about what a declaration is, because what a declaration is and what one backend calls it are two things. The runtime carrier of a language `Rational` is the JVM backend's physical representation; an analysis reasoning in it would be one whose soundness depended on which backend was linked.

## What the test does not currently cover

That test names its target-neutral areas explicitly:

```java
private static final Set<String> TARGET_NEUTRAL =
        Set.of("types", "check", "stdlib", "semantics", "partition", "inputs", "core", "flow");
```

`numeric` is in neither that set nor `MAY_NAME_IT`, so the rule does not reach the package this exact ratio lives in. Add it to the target-neutral set, which is what turns ADR-0117's reasoning into something a run can fail on.

Prerequisite for the rest of ADR-0116 and ADR-0117. It changes no behaviour, so it stands alone.

ADR-0116, ADR-0117.

PR #1790, opened 2026-09-18
Give the compiler's exact ratio a name the language is not taking

kawasima:
Closes #1785.

`souther.compiler.numeric.Rational` holds the exact arithmetic the constraint algebra reasons in, and its own account says it is not a number a model writes. ADR-0116 gives that spelling to a language type that is exactly that, and ADR-0117 decides the compiler's concept keeps its meaning and gives up the name.

So `Rational` becomes `ExactRatio`, everywhere in `numeric` and at the three places outside it that name one — `inputs.ReadQuantities`, `partition.CandidateDomain`, `partition.LevelRealizer`, `partition.LevelSpace`. Nothing about the behaviour moves.

`RationalCut` follows it as `ExactCut`. It is named after the type rather than after anything a model writes, and the new name says what its javadoc already contrasts it with: an `Endpoint` stops on a carrier's order and is spelled in the decimals that carrier counts in; this one stops while the reasoning is still exact.

## The rule that says the two ratios are not one implementation

`TheRuntimePackageIsTheBackendsToNameTest` refuses any mention of `souther.runtime` from the areas that reason about what a declaration is — which is why the language's `Rational` and this one cannot share a carrier, the runtime one being the JVM backend's physical representation. `numeric` was in neither that set nor `MAY_NAME_IT`, so the rule did not reach the package the exact ratio lives in. It is in the target-neutral set now.

Adding it changes no run today: `numeric` names the runtime package nowhere. Putting a `"souther.runtime"` literal in `ExactRatio` turns the test red and names the file, so the area is checked and not merely listed.

`mvn -o test` is green.


kawasima:
The finding holds, and I measured it before changing anything: with `import souther.runtime.DecimalMath;` and a use of it added to `numeric/ExactRatio.java`, the rule stayed green. `souther-runtime` is a `provided` dependency of this module, so that mutation compiles — nothing else refuses it either.

The check now reads each source with its comments taken out and looks for the package in what is left. That keeps the three forms that are a dependency — an import, a fully qualified mention, and the string a backend's own table is written with — and drops the one that is an account of why there is none, which `types.TypeSymbol` and `check.ScaleRange` each carry in a javadoc. Broadening to a plain `contains` would have made those two the finding, as you say.

The reading is the one thing the rule is a predicate over, so it has its own fixture: a source carrying the package in a javadoc, a line comment, an import, a fully qualified field type, and two strings — one of them a string whose content looks like a comment.

Measured in `numeric`, each on the real tree:

- `import souther.runtime.DecimalMath;` — red, naming `numeric/ExactRatio.java`
- `souther.runtime.DecimalMath.class` with no import — red
- `"souther.runtime"` as a string — red, which is what the old check caught
- `{@code souther.runtime}` in a javadoc — green

`mvn -o test` is green.

PR #1798, opened 2026-09-19
Hold a form's ordered answers to the order its terms decide

kawasima:
The order a form is walked in is settled in one place, and the readers whose answers have an order
in them ask for it there. That the one order exists is held where it is written; that these readers
go through it was said nowhere, which is the half a contract on the order cannot state.

So the readers answering with a sequence or a text — what a form comes to when it is ordered, where
a reading files what it found, what a quantity's terms are, and how a report spells a direction —
are put to one form twice over.

Written both ways round, which tells a reader taking the terms as they come from one that asks. A
control beside it says the two writings really do hand their terms over in different orders, without
which each answer would hold while saying nothing.

And written as two terms a name cannot tell apart, which is the half the first witness does not
reach: a location written as one name and the same location reached a field at a time are spelled
alike and are two terms, and the one order refuses the walk rather than handing one coefficient over
twice. A reader sorting the form by what its terms are called answers alike for both writings and
does exactly that, so without this pair the rule would pass a reader that had gone back to keeping
its own order.

Each of the four is asked on its own rather than the first of them standing for the rest, and each
goes red on its own for both witnesses.

This is a behavioural regression guard at the choke points a reader's ordered answer goes through,
and not a detector for everything the reading behind the order-blind carriers found at an affine
form. A rule over the raw coefficients would name every consumer that holds the mapping and would
ask for a reading of each — which is that reading again, written as something to keep. The
coefficients are held as they were handed in, so what settles the order is asked for by each reader;
what is held here is that the readers there are go on asking.


kawasima:
## Where the weakness came from

The order these readers ask for is two things, and the witness reached one of them. Walking by the
names is the first; the second is that a name is not what tells two terms apart, and that one is
what a reader keeping its own order does not have. Both writings of a form over two terms spelled
differently are told apart by any sort by name, so the fixture separated a reader taking the terms
as they came from everything else, and separated nothing else from anything.

Which is the half that matters most when it goes: a walk by the names over two terms spelled alike
hands one coefficient over twice and the other never, and the form comes out a term short with the
arithmetic looking fine.

Both witnesses are in now. Each reader is asked on its own — the first of them answering says
nothing about the rest — and each goes red on its own: raw iteration fails the two writings, a sort
by what the terms are called fails the pair a name cannot tell apart, and only asking for the one
order passes both. Checked by putting each reader back to a sort of its own, one at a time.

## Whether the same weakness is anywhere else

Five places here settle an order a value does not hold. The hole needs both halves of what happened
at a form: the order has a refusal in it that a plain sort lacks, and each reader asks for it rather
than the value having been put in the order on the way in.

Where a plan's positions are ordered, the order is asked for as the value is built, so no reader can
be written that does not have it — a reader-side rule there would be about a path nothing takes.
Where a canonical form is walked, the walk is the only way to the coefficients and it refuses what it
cannot tell apart, and a rule already holds the readers to going through it. Where a value is shown
in one order, what is settled is a rendering and there is no refusal to miss. And what answers with
the one thing a collection holds refuses a collection that is not of one, which is nothing to do with
how anything is spelled.

So a form is the one place where both halves hold, which is what the coefficients being kept as they
were handed in comes to: the order is asked for by each reader, and a reader that forgets is a
reader this rule is here to find.

The subjects were the other half of the mistake — they came off a list I had rather than off what
answers with an order. Counted again: these four are the ordered answers a caller reaches, one more
is a helper inside the search that no caller sees, and three write a refusal's message from the same
call on a path already failing. That is in the test's own words now, along with what it does not
cover: a reader added tomorrow is not one it says anything about.

Issue #1852, opened 2026-09-21
A scope threaded through Terms, TotalityChecker and Elaborator copies its whole state at every nested binder


kawasima:
Found while fixing #1843: `BlockReaches`'s walk built a fresh copy of its accumulated bound-name set at every `let`, `match` arm and nested block it entered, so a chain of N nested scopes summed to O(N²) — the same shape as #1814's exponential blow-up, one nesting form short of it.

The same anti-pattern is not unique to that new query. At least three existing recursive walks over `Core`/`Hir` thread a `Map`/`Set` of what is bound so far and copy the whole thing at every nested scope instead of extending it in O(1):

- `souther.compiler.check.Terms`: `naming`'s `Core.LetIn`, `Core.Block`, `Core.Match` and `Core.IfConstructed` cases (via the now-removed `binding()` helper and inline `new HashMap<>(bound)`); `isWritten`'s `Core.LetIn` case; `writtenSyntaxOf`'s `Core.LetIn` case (via `withGiven`).
- `souther.compiler.check.TotalityChecker`: `walk`'s `Hir.Match`, `Hir.LetIn` and the container-combinator step body case, via the private `with(Map, BindingId, Set)` helper.
- `souther.compiler.check.Elaborator`: `collectApplications`'s `Hir.Block`, `Hir.LetIn`, `Hir.IfConstructed` and `Hir.Match` cases, via the private `with(Set, List<Hir.Binder>)` helper.

### Done when

* each of the three walks extends its threaded scope state in O(1) per binder rather than copying everything already bound, so the total cost of walking a chain of N nested scopes is O(N) instead of O(N²);
* `TotalityChecker`'s fix does not break the one place this differs from the other two: `RecCall` retains the `lt`/`eq` relation in scope at the moment a recursive call was recorded, read back long after `walk` has moved past that point (`buildScgs`) — a plain mutate-then-restore on a shared mutable map would silently corrupt an already-recorded `RecCall` the next time the walk backtracks past it, so this one needs a relation that is safe to keep past the call that extended it (an immutable environment layered over its parent works: extending it is one allocation, and nothing beneath a layer is ever mutated once made);
[...]

Issue #1843, opened 2026-09-21
What a block reaches outside itself is worked out again by each backend


kawasima:
An escaping block has to carry what its body reaches outside the block. That is one fact about the block, but each output walks the `Core` tree for itself to find it.

The JVM emitter has `BodyGen.freeVars` and `collectFree`. They find both bindings from the enclosing body and injected behaviors the block calls, because the generated lambda class has to carry both. `souther-wasm-compiler` has `FreeReads`, which walks the same `Core` structure to find the bindings outside the block, which become the closure's captured values.

The two walkers differ because their physical closures differ, but the question they start from does not:

> what does this block reach that the block itself does not bind?

Scope, nested blocks, `let`, match-arm binders and attempted constructions are properties of the checked tree. An output should decide how to carry those reaches, not rediscover which reaches exist.

There should be one compiler query over a checked block that answers its external reaches, keeping the distinct kinds distinct — at least enclosing bindings and externally supplied behaviors — in an order an output can assign its own fields or slots from.

Most of this exists already, as `BodyGen`'s private `Reaches`: it holds the reads and the behaviors separately, in first-seen order, and hands each out as a list. Promoting it is largely a move. The order is part of what a reader is promised, so it stays a list on the way out; a set whose type does not say it is ordered would leave the promise where nothing holds it.

### Done when

* the compiler has one canonical query answering what a `Core.Block` reaches outside itself;
* the answer distinguishes binding captures from declaration and requirement reaches, rather than making an output tell them apart afterwards;
* the order the answer is read in is carried by the type of the answer;
* nested binders and blocks are handled by that query once;
* `BodyGen.freeVars` and `collectFree` no longer walk `Core` to discover the capture set;
* an output remains free to choose how each reach is represented physically.

Deleting the second walk is the consumer half: souther-lang/souther-wasm-compiler#7.

The query is about reachability, not closure layout. A JVM field, a Wasm capture object and a native environment are deliberately outside its result.

kawasima:
Merged in #1850 (PR merged to develop at 151ca693c).

PR #1854, opened 2026-09-21
Stop copying a scope's whole bound state at every nested binder

kawasima:
## Summary

Found while fixing #1843: `BlockReaches`'s walk copied its whole accumulated bound-name set at every nested `let`/`match`/block instead of extending it in O(1), costing O(N²) for a chain of N nested scopes. The same anti-pattern was already present in three existing walks — this PR fixes those, closing #1852.

- `Terms.naming` (LetIn/Block/Match/IfConstructed), `Terms.isWritten` (LetIn), `Terms.writtenSyntaxOf` (LetIn): none of these retain their threaded `Map`/`Set` past the call that received it, so each now mutates one shared, owned map/set and restores it on the way back out of a scope (add before recursing, remove after) — the same fix already applied to `BlockReaches` in #1850. The now-dead `binding()`/`withGiven()` copy helpers are removed.
- `Elaborator.collectApplications` (Block/LetIn/IfConstructed/Match): same fix, same shape.
- `TotalityChecker.walk`: differs, and needed a different fix. `RecCall` retains the `lt`/`eq` relation in scope at the moment a recursive call was recorded, read back long after `walk` has moved past that point (`buildScgs`) — mutate-then-restore on a shared map would corrupt an already-recorded `RecCall` the next time the walk backtracks past it. Replaced the copied `Map<BindingId, Set<BindingId>>` with `Layered`, an immutable relation over a parent: extending it is one allocation, and nothing beneath a layer is ever mutated once made, so a `RecCall` can hold one indefinitely.

## Test plan

- [x] `mvn -o test` — full reactor, 176/176 passing, run in an isolated worktree off `develop` with only these three files changed
- [x] Targeted runs before the full pass: `CompileTotalityTest`, `CompileInvariantQuantifierTest`, `CompileRecursiveHelperTest`, `TheRecursiveHelpersAreAnsweredInDeclarationOrderTest` (TotalityChecker); `CompileBlockTest`, `CompileFunctionBindingTest` (Elaborator); the full `souther-compiler` suite (Terms)

Fixes #1852


kawasima:
Both points addressed in \`d7b6425\`.

1. \`Layered\` made extension O(1) but left \`get\` O(depth) on a miss — and a bare reference to the function's own parameter is a guaranteed miss at every layer, since parameters are never keys in \`lt\`/\`eq\`. Confirmed this stayed O(N²) for a chain of N nested matches/lets each referencing the outer parameter directly. Fixed per the suggested design: \`RecCall\` now carries a precomputed \`ArgRelation\` (the \`strict\`/\`roots\` sets) per argument, settled at the moment the call is found rather than the environment itself. This removes the retention requirement entirely, so \`lt\`/\`eq\` are back to one shared, mutated \`HashMap\` apiece (O(1) average get/put), giving the genuine O(N) total the issue asks for. \`Layered\` is deleted.
2. Rebased onto current \`develop\` (\`235efaaa7\`) — clean, no conflicts. \`mergeStateStatus\` is now \`CLEAN\`.

Also dropped the duplicated comment line in \`Terms.writtenSyntaxOf\`'s \`LetIn\` case you flagged.

Full reactor (\`mvn -o test\`) 176/176 passing after both the fix and the rebase.

